PR #6, opened 2019-06-05
chore: changes to repo to fix failing tests and netlify errors

jbadan:
When initializing travis/automated release, I accidentally added some incorrect entries to the changelog. This should clean it up for future releases.

Additional changes added to fix failing travis tests and netlify errors.


greg-a-smith:
@jbadan Should probably use `chore:` prefix.

package.json
```diff
@@ -91,5 +90,6 @@
       "lint:pre-commit"
     ],
     "silent": false
-  }
+  },
+  "browserslist": ["last 2 versions"]
```
jbadan: Netlify was throwing a ton of warnings that were overloading and causing the build to fail by having this in the postcss config. Works the same as having a browserslist - this is now the accepted way of doing it.

PR #14, opened 2019-06-05
fix breadcrumb bottom  negative margin

JYSW380:
Closes SAP/fundamental-styles#12

set bottom margin to 0 at breadcrumb component  to avoid overlapping of components.


package-lock.json
```diff
@@ -10787,4 +10787,4 @@
       "dev": true
     }
   }
```
bcullman: unrelated change. please remove from commit.

PR #30, opened 2019-06-06
feat: toggle updates (#1466)

saad-mo:
Bring in Toggles/Switch updates from [fundamentals repo](https://github.com/SAP/fundamental/pull/1466)

* Update toggle component to match design specs


.stylelintrc.yml
```diff
@@ -27,7 +27,7 @@ rules:
   selector-pseudo-class-no-unknown: true
 
   ## General / Sheet
-  no-duplicate-selectors: true
+  # no-duplicate-selectors: true
```
jbadan: Why are we needing to disable a lint rule?

saad-mo: I was marking this line below as a duplicate selector. is there a way to add exceptions to this linting rule to get past this? the toggle/switch scss is a bit convoluted. 

`&+.#{$block}__switch{`

https://github.com/SAP/fundamental-styles/blob/a245d359b1b9afae28261323aa9aae2e5b781a2e/scss/components/toggle.scss#L58

MattL75: Is there no way to clean it up or organize it in a different way which wouldn't require an exception to the lint rule?

saad-mo: i am open to ideas if you have any

jbadan: If we can't come up with another solution, we can add `/* stylelint-disable-line no-duplicate-selectors */` on the line in the scss file so that we don't need to disable the entire rule everywhere else :)

saad-mo: this is false positive. we are not really creating a duplicate selector, rather reusing the existing block element that was defined.

saad-mo: turns out the linter was not complaining about `#{$block}`, but there was indeed a duplicate rule defined which I corrected.

Issue #34, opened 2019-06-07
Bug: button text alignment on localizationEditor


jbadan:
![Screen Shot 2019-06-07 at 9 25 30 AM](https://user-images.githubusercontent.com/29607818/59118907-40b62a00-8906-11e9-8e6a-1db299d45a6e.png)

"EN" Text on button is not center aligned like it was in fundamental < 1.5.3.

rengare:
Issue is already fixed 
![image](https://user-images.githubusercontent.com/10849982/70638016-754dca00-1c38-11ea-84e9-40982d2030e1.png)

salarenko:
Issue has been fixed. (Cross browser check)

PR #35, opened 2019-06-07
fix: remove 100% height from localizationEditor button

jbadan:
Closes SAP/fundamental-styles#34

Before: 
Screen Shot 2019-06-07 at 9 35 17 AM

After:
Screen Shot 2019-06-07 at 9 34 48 AM

We might want to look into making our tests more sensitive - it should have caught this error when I introduced it in https://github.com/SAP/fundamental/pull/1401 and it should have caught these changes in this pr in the tests. 

I can't find any side effects of removing this, but since our tests aren't picking up these changes I'm nervous about just deleting it. Thoughts?

### DIFF
--- a/scss/components/localization-editor.scss
+++ b/scss/components/localization-editor.scss
@@ -29,11 +29,6 @@ $block: #{$fd-namespace}-localization-editor;
         }
     }
 
-    &__button {
-        height: 100% !important;
-        max-height: 100% !important;
-    }
-
     .fd-popover__body {
         width: 100%;
     }


bcullman:
are we missing a lint rule here? `!important` almost always a bad idea...

jbadan:
@bcullman We should definitely add that in the future, but there are `!important`'s everywhere that we would need to remove/rewrite. I think that's more of a story for the backlog.

jbadan:
I need to look at this further - removing the height has now affected the hover states.

jbadan:
@bcullman @droshev so it looks like the hover issue was actually always present, this 100% height on localization editor just covered it up. It's present on all buttons inside input groups. I think that is part of a larger fix than this. 
https://github.com/SAP/fundamental-styles/issues/36

Thoughts?

bcullman:
we should track down the bigger issue

scss/components/localization-editor.scss
```diff
@@ -29,11 +29,6 @@ $block: #{$fd-namespace}-localization-editor;
         }
     }
 
-    &__button {
-        height: 100% !important;
-        max-height: 100% !important;
-    }
-
```
greg-a-smith: So, I agree this seems odd to have here and also with the `!important` declarations, however, simply removing this causes issues when hovering or focusing on the button.  This _seems_ like something that should be fixed within input groups rather than here.  I am seeing the same weirdness on hover and focus on buttons within an input group.

jbadan: Yeah my plan was to leave this hanging until we have time to resolve the input group hover/focus problems. I remember having this issue in nui-widgets, especially with the border hovers on timepicker and whatnot, but the code is so different the solution we came up with there didn't seem to apply.

jbadan:
closing in favor of addressing #36

PR #53, opened 2019-06-09
docs: add 2 separate scrollbars for left nav and main

droshev:
Closes SAP/fundamental-styles#52

### DIFF
--- a/docs/_includes/side-nav.html
+++ b/docs/_includes/side-nav.html
@@ -5,70 +5,82 @@
 
 
 <ul class="side-nav__links-container">
-{% for entry in sidebar %}
+    {% for entry in sidebar %}
     {% for folder in entry.folders %}
-        {% if folder.output contains "web" %}
+    {% if folder.output contains "web" %}
 
-        {% assign folder-name = folder.title | downcase %}
+    {% assign folder-name = folder.title | downcase %}
 
-            {% if page.url contains folder-name %}
-            <li class="side-nav__headers active">
+    {% if page.url contains folder-name %}
+    <li class="side-nav__headers active">
+        {% else %}
+    <li class="side-nav__headers">
+        {% endif %}
+        <a class="side-nav__headers-link"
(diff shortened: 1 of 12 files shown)


MattL75:
Doesn't this feel weird? Mostly talking about the sticky footer.

image

There's no difference in color so it just looks like the components blend in with the footer, makes it look weird

MattL75:
The scrollbar on the sidebar also feels weird

image

jbadan:
@MattL75 I think the scrollbar looks weird because of the background color, we do have visible scrollbars on the other fundamental-* libraries. Maybe setting the background color on it? Although this won't work in Firefox or IE/Edge. Or we could swap the background color of the sidenav and the background for the component container so that it blends better in all systems.

MattL75:
@jbadan Maybe we defer the scrollbar look to a later conversation? Possibly involving designers? 😅 Honestly not too sure of the best direction to go with this. It's fine for now I guess, so the remaining issue would be the white on white with the footer.

jbadan:
@MattL75 Sounds good to me! 

I think options for the footer: not sticky or change the container backgrounds and wait for designer input.

droshev:
I don't want to modify the colors/design more as we will have something news but I am open for tiny changes.
What do you think we should do in this ticket?

jbadan:
@droshev Sounds good to me - can we just fix the footer issue? Have it behave like the one in fundamentals?

droshev:
this doesn't look good:
Screen Shot 2019-06-14 at 2 52 14 PM

MattL75:
@droshev Hard to say without live example, not even sure what you changed there.

droshev:
- updated the header on the landing page not to be sticky
- I removed the footer when there is a left nav since it looks odd.

droshev:
@jbadan @MattL75

jbadan:
@droshev There's still one small problem with the expand collapse.

droshev:
@jbadan

Issue #37, opened 2019-06-07
Bug: search on components page not working


DorinR:
> What version of the Fundamental Styles are you using?
v0.1.0

**To Reproduce**
1. Go to https://sap.github.io/fundamental-styles/components/index.html
2. Enter a search query in the search box in the top-left corner of the page

**Expected behavior**
1. Auto-complete suggestions appearing while typing in search query.

**Screenshots**
![Screen Shot 2019-06-07 at 4 38 33 PM](https://user-images.githubusercontent.com/22306864/59132498-5b5ac400-8943-11e9-96cd-85873e7215a9.png)

**Desktop:**
 - OS: MacOS 10.14.5 (Mojave)
 - Chrome
 - Version 74.0.3729.169

bcullman:
(I think this came over broken)

jbadan:
See: https://github.com/SAP/fundamental/issues/1476#issuecomment-499642792

droshev:
Yes, it came like that from fundamental

PR #54, opened 2019-06-09
docs: fixed the seach box in the docsite

droshev:
Closes SAP/fundamental-styles#37


droshev:
it doesn't work in netlify :(

droshev:
The issue was coming from invalid json file. It is fixed now!

PR #60, opened 2019-06-10
fix: Self-contained styles for button

greg-a-smith:
This PR adds a new `components` top-level directory that will be part of the distribution package and within it are the `*.scss` files that can be used for self-contained styles in the functional libraries.

The first component to be created is `button.scss`.

There are a number of visual tests references updated, but it appears it can be linked to some of the global resets that changed.

#### Old
old_visual_regression_fundamental_button_0_example-containerfd-button_0_desktop
#### New
visual_regression_fundamental_button_0_example-containerfd-button_0_desktop

#### Old
old_visual_regression_fundamental_button_1_example-containerfd-button__n1_0_desktop
#### New
visual_regression_fundamental_button_1_example-containerfd-button__n1_0_desktop

#### Old
old_visual_regression_fundamental_button_14_example-containerfd-button__n14_0_desktop
#### New
visual_regression_fundamental_button_14_example-containerfd-button__n14_0_desktop


components/button.scss
```diff
@@ -0,0 +1,6 @@
+$fd-icons-path: "../scss/icons/";
+$fd-fonts-path: "../scss/fonts/";
```
greg-a-smith: Variables added here so they are not forced upon the consuming library.

scss/_settings.scss
```diff
@@ -51,14 +51,14 @@ $fd-type--base: 14 !default;
...
-  "5": #{36/$fd-type--base}rem 40/36 20px, //36/40 20
-  "6": #{48/$fd-type--base}rem 52/48 24px, //48/52 24
+  "-1": 12px 16/12 8px, //12/16 8 btm
+  "0": 14px 20/14 12px, //14/20 12
+  "1": 16px 20/16 12px, //16/20 12
+  "2": 18px 24/18 12px, //18/24 12
+  "3": 20px 28/20 12px, //20/28 12
+  "4": 24px 32/24 16px, //23/32 16
+  "5": 36px 40/36 20px, //36/40 20
+  "6": 48px 52/48 24px, //48/52 24
```
greg-a-smith: These were doing calculations to figure out the `rem` value, but were using the `px` value to do the math.  Without a specified default font size on the `<body>`, we need to specify all this directly on the components.

scss/icons/_settings.scss
```diff
@@ -4,10 +4,10 @@
...
 $fd-icons-sizes: (
     "default": $fd-font-size, //14px
-    "s": #{12/$fd-type--base}rem, //12px
-    "m": #{16/$fd-type--base}rem, //16px
-    "l": #{20/$fd-type--base}rem, //20px
-    "xl": #{24/$fd-type--base}rem //24px
+    "s": 12px,
+    "m": 16px,
+    "l": 20px,
+    "xl": 24px
```
greg-a-smith: These were doing calculations to figure out the `rem` value, but were using the `px` value to do the math.  Without a specified default font size on the `<body>`, we need to specify all this directly on the components.

scss/components/button.scss
```diff
@@ -90,20 +88,20 @@ a.#{$block} {
       @include fd-hover() {
           --fd-button-background-color: var(--fd-color-action-hover);
           --fd-button-color: var(--fd-color-action-2);
-          @include fd-var-color("color", fd-color("action",2));
-          @include fd-var-color("background-color", fd-color-state("hover","action"));
+          @include fd-var-color("color", fd-color("action",2), --fd-button-color);
+          @include fd-var-color("background-color", fd-color-state("hover","action"), --fd-button-background-color);
```
greg-a-smith: The majority of the changes in this file are adding the additional argument for the mixin so all CSS variable references also include a default value.

scss/mixins/_mixins.scss
```diff
@@ -5,6 +5,12 @@
     line-height: $fd-line-height;
     @include fd-var-color("color", $fd-color, --fd-color-text-1);
     font-family: $fd-font-family;
+    -webkit-font-smoothing: antialiased;
+    box-sizing: border-box;
+    &::before,
+    &::after {
+        box-sizing: inherit;
+    }
```
greg-a-smith: These were previously being included on the `<html>` or `<body>` elements, which we can no longer depend on.

scss/mixins/_mixins.scss
```diff
@@ -140,7 +146,6 @@
   display: inline-block;
   margin: 0;
   padding: 0;
-  font-smoothing: antialiased; /* stylelint-disable-line property-no-unknown */
```
greg-a-smith: Moved to a more global reset mixin.

test/templates/layout.njk
```diff
@@ -7,23 +7,38 @@
...
-        {% elif selfContained %}
-            {% if id === 'icon' %}
-                {% set sass = 'icons.scss' %}
-            {% elif id === 'radio' or id === 'checkbox' or id === 'input' %}
-                {% set sass = 'components/form.scss' %}
-            {% elif id === 'panel' or id === 'panel-grid' or id === 'layout' %}
-                {% set sass = 'layout.scss' %}
+        {% if selfContained %}
+            {% set sass = id + '.scss'  %}
+            {{ sass | component_css }}
```
greg-a-smith: All the self-contained component pages will now be missing styles except for button, but I think that is what we want so we can validate them as more and more components are added to the self-contained styles.

bcullman:
Do the visual diffs actually generate different visuals?

greg-a-smith:
@bcullman They are slightly different, but the differences are very minor.  They seemed to be related to font display.

droshev:
How about we include the semantic theming parameter colors as well?

greg-a-smith:
@droshev I think that's a larger story.  Although there is a separate file for consumers to use for self-contained styles, it imports the same file that is used in the "all" CSS file.  Because of that, I think we need to take a more macro approach to the use of semantic theming variables.

bcullman:
another followup story here might be trim this file - it's exceptionally large. some candidates to remove would be named classes that serve as alternates for pseudo selector classes. Both should not be needed.

Issue #67, opened 2019-06-11
Bug: Text of disabled & pressed & hovered Button in Button Group not legible


SaiShan:
Fiddle: https://jsfiddle.net/christiankienle/3gjhzyxL/2/

Hover over the disabled button in the button group and you will notice that the text is not legible.

![text-disabled](https://user-images.githubusercontent.com/25252963/59301010-979d6580-8c5e-11e9-80f1-f4f2a8342264.png)

PR #75, opened 2019-06-12
fix: Fixed button text and border when disabled button is hover or pressed

SaiShan:
Closes SAP/fundamental-styles#67.

Fixed button text when disabled and border when disabled button is pressed as per issue #67. 
See screenshot. 

Screen Shot 2019-06-13 at 10 52 50 AM


SaiShan:
> @SaiShan Can you include a snippet of the change you made in the nunjucks files to make a button disabled in a button group? Or better yet, should we add an example with a disabled button in a button group?

Thank you for your advise, I added an example in the playground. The changes can be seen in the newest version of the changes.

scss/components/button-group.scss
```diff
@@ -51,3 +51,13 @@ $block-alt: #{$fd-namespace}-segmented-button;
     }
   }
 }
+.fd-button-group .fd-button--compact:hover[aria-disabled=true],
+.fd-button-group .fd-button:hover[aria-disabled=true] {
```
greg-a-smith: I'm not sure if `aria-disabled` is the best selector to use here.  Looking through other examples, I see buttons are using an `is-disabled` class and in another library we use, we are using the `disabled` HTML attribute.  I think the `disabled` HTML attribute is probably going to be the best option going forward since there will be other a11y benefits by using the HTML spec.

The consuming library may also add the `aria-disabled` attribute, but we shouldn't base styling upon it.

SaiShan: I totally agree. Thank you for pointing it out.

scss/components/button-group.scss
```diff
@@ -51,3 +51,17 @@ $block-alt: #{$fd-namespace}-segmented-button;
...
+.fd-button-group .fd-button:hover {
+  &.is-disabled {
+    --fd-button-background-color: white;
+    @include fd-var-color("background-color", white);
+  }
+}
+
+.fd-button-group .fd-button--compact:focus,
+.fd-button-group .fd-button:focus {
+  &.is-disabled {
```
greg-a-smith: I don't think disabled should be implemented as a CSS class (like `is-disabled`).  Although it can be used for styling, it provides no a11y support.  When going through the playground page with a screen reader, the disabled buttons sound just like the enabled buttons.  A disabled button should actually remove itself from the tab order.

I know this isn't unique to buttons or button groups, but the elements should be getting the `disabled` attribute (for elements that support it) and then we can use _that_ in our styling selectors.

scss/components/button.scss
```diff
@@ -82,7 +82,12 @@ a.#{$block} {
           margin-left: 0;
         }
       }
-
+      &:focus {
+        &.is-disabled,
+        &[aria-disabled="true"] {
```
greg-a-smith: Shouldn't be using `aria-*` attributes for styling selectors.

scss/components/button-group.scss
bcullman: I would go further and suggest we *strip* `&.is-disabled` styles from the CSS. 

Using it will consistently lead to bad a11y practices.

InnaAtanasova: @droshev

droshev:
@SaiShan the thread is quite long. Here is what i think we should do in that PR:

- make the PR to start with `feat:` instead of `fix:`
- get rid of the class `is-disabled` and use `disabled` as an HTML attribute as css selector
- do not use `aria-*` as css selector
- update the title of the PR describing that we get rid of class `is-disabled`
- keep the scope for `button-group` (and raise separate issue(s) for the other component(s)

scss/components/button-group.scss
```diff
@@ -51,3 +51,10 @@ $block-alt: #{$fd-namespace}-segmented-button;
     }
   }
 }
+.fd-button-group .fd-button--compact:hover,
+.fd-button-group .fd-button:hover {
+  &:disabled {
+    --fd-button-background-color: white;
+    @include fd-var-color("background-color", white);
```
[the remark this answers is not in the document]
greg-a-smith: You know what?  As I look at this closer, I don't think any of these changes should be in this file.  Button groups are a method for putting a group of buttons together.  The state handling of a button (_e.g._ disabled) should be handled for all buttons -- not just buttons in button groups.

test/templates/button-group/index.njk
```diff
@@ -52,8 +52,65 @@
...
+        aria={ })
+      }}
+    {{-  button(
+        { label: 'Second' },
+          modifier={ block: [size] },
+          aria={ })
+        }}
+    {{-  button(
+          { label: 'Third' },
+            modifier={ block: [size] })
```
greg-a-smith: Example wise, there should be examples that show different buttons as the disabled version (not always the first one).

scss/components/button-group.scss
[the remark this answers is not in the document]
SaiShan: @greg-a-smith the `disabled` is handled in all other buttons. Moving this part over to button.scss make the button-group not behave properly.

scss/components/button.scss
```diff
@@ -607,4 +608,4 @@ a.#{$block} {
       }
     }
 }
-/* stylelint-enable */
\ No newline at end of file
+/* stylelint-enable */
```
greg-a-smith: Changes in this file should be reverted.

scss/components/button-group.scss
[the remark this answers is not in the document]
greg-a-smith: Let's connect on this one.  There is likely something else getting in the way of that.

scss/components/button.scss
SaiShan: The changes were reverted. I do not see them anymore in the files changed.

PR #88, opened 2019-06-13
fix: fixed misalignment issue with icon

### LINKED ISSUE #64: List item elements vertical alignment
Elements inside sidenav list item are not vertically centered. 

See screenshot:
Screen Shot 2019-06-11 at 3 17 05 PM

JYSW380:
Closes #64
fixed list item misalignment issue in side nav scss file

Before screenshoot:
Screen Shot 2019-06-13 at 2 56 57 PM

After screenshoot:
Screen Shot 2019-06-13 at 2 57 32 PM


scss/components/side-nav.scss
```diff
@@ -100,6 +100,8 @@ $block: #{$fd-namespace}-side-nav;
     }
 
     &__link {
+        display: flex;
+        align-items: center;
```
bcullman: should this go higher? 

`&__link,&__sublink`, line 75

JYSW380: I tried to put higher, please see screenshot:
Screen Shot 2019-06-14 at 9 17 36 AM

and it showed result as:
Screen Shot 2019-06-14 at 9 17 43 AM

so i think we should keep as before:
Screen Shot 2019-06-14 at 9 26 32 AM

bcullman: did you leave the `display: block;` on line 79?

JYSW380: yes, i did not change it.

bcullman: um, you would have to remove that. thats how the cascade works.

JYSW380: ok, i fixed it, and now the code is at higher position and After screenshot:
Screen Shot 2019-06-14 at 3 52 54 PM

please let me know if that is ok now? thanks.

droshev:
@jeannewalters @LeoT7508 could you comment @greg-a-smith 's question as it is more a design question?
thanks

jeannewalters:
@greg-a-smith I would prefer the first line and the icon aligned from a scannability perspective. @LeoT7508 @saad-mo Is there a reason why it has to stay center-aligned with all the text rather aligned with the first line only?

LeoT7508:
@jeannewalters @saad-mo @greg-a-smith 

The positioning of the icon should always be anchored to the top position for consistency, if we move them when there are 2 lines vs 1 the icons will jump around and the list will look broken. This is also a case of common sense, we should never let a label get longer than 2 lines, in this case it's time to brainstorm and think of a short label, these shouldn't be sentences.

greg-a-smith:
@LeoT7508 Completely agree that they shouldn't be sentences.  There are times when things will wrap to a second line (especially with localized text) so thanks for the feedback on the desired alignment in those cases.

scss/components/side-nav.scss
```diff
@@ -105,7 +105,7 @@ $block: #{$fd-namespace}-side-nav;
         &.has-child {
             background-image: url(#{$fd-down-arrow-icon});
             background-repeat: no-repeat;
-            background-position: calc(100% - #{$fd-forms-select-width--background-image}) center;
+            background-position: left calc(100% - 10px) top 15px;
```
greg-a-smith: I see you are trying to make a similar change to the child down arrow icon.  It is better, however, the side nav item itself needs some more right padding as text can bleed over onto the icon (see screenshot).  It should also probably have similar spacing on the right of the icon as the list item has on the left side.
Screen Shot 2019-06-19 at 10 21 17 AM

scss/components/side-nav.scss
```diff
@@ -75,7 +75,7 @@ $block: #{$fd-namespace}-side-nav;
     &__link,
     &__sublink {
         display: flex;
-        align-items: center;
+        align-items: flex-start;
```
greg-a-smith: It seems like there's a balance that can be achieved here.  This keeps the icon anchored to the top, however, for single line text, the icon doesn't align to it very well (it's almost the same as the original state, just off on the bottom side).  I think the desired effect is to have the icon appear to be center-aligned to the first row of text regardless of whether there is one or two lines of text.
Screen Shot 2019-06-19 at 10 23 29 AM

LeoT7508:
@JYSW380 

There should be 20px padding on the left side of that icon so the text doesn't run into it like in the screenshots above.

JYSW380:
> @JYSW380
> 
> There should be 20px padding on the left side of that icon so the text doesn't run into it like in the screenshots above.

sorry for the confusion, i edited above screenshot, the first one is with icon & the second one is without icon, they both have 20px padding on the left

docs/pages/components/side-navigation.md
```diff
@@ -20,32 +20,32 @@ The left navigation can always display or expand/collapse using the menu icon wi
...
         <li class="fd-side-nav__item">
             <a class="fd-side-nav__link" href="#">
-                Link Item
+                <span class="fd-side-nav__link-icon--l">Link Item</span>
             </a>
         </li>
         <li class="fd-side-nav__item">
             <a class="fd-side-nav__link" href="#">
-                Link Item
+                <span class="fd-side-nav__link-icon--l">Link Item</span>
```
greg-a-smith: Not sure what all these changes are for.  Many of these do not have icons on them.

scss/components/side-nav.scss
```diff
@@ -25,7 +25,7 @@ $block: #{$fd-namespace}-side-nav;
     $fd-side-nav-hover-background-color: fd-color("neutral", 1);
     $fd-side-nav-link-color: fd-color("text", 2);
     $fd-side-nav-max-width: 249px;
-    $fd-side-nav-link-padding: fd-space(2.5) fd-space(5);
+    $fd-side-nav-link-padding: fd-space(2.5) fd-space(7.5) fd-space(2.5) fd-space(5);
```
greg-a-smith: Not sure the right padding is large enough yet.  See screenshot:
Screen Shot 2019-06-24 at 12 20 31 PM

scss/components/side-nav.scss
```diff
@@ -74,12 +74,34 @@ $block: #{$fd-namespace}-side-nav;
...
+
+            &--l {
+                line-height: 24px;
+            }
+
+            &--xl {
+                line-height: 28px;
+            }
+        }
+
```
greg-a-smith: _Seems_ excessive to make another class just for this.  Can't we leverage the existing classes that already identify icons and/or the arrow to expand/collapse children links?

docs/pages/components/side-navigation.md
JYSW380: those classes are for in case people want to add an icon before, and we need these classes to make sure that if others want to add an icon, then the icon is at centered aligned with the first row of text whether one or two text.

scss/components/side-nav.scss
```diff
@@ -100,12 +101,11 @@ $block: #{$fd-namespace}-side-nav;
     }
 
     &__link {
-
         //STATES *******************************************
         &.has-child {
             background-image: url(#{$fd-down-arrow-icon});
             background-repeat: no-repeat;
-            background-position: calc(100% - #{$fd-forms-select-width--background-image}) center;
+            background-position: left calc(100% - 14px) top 15px;
```
greg-a-smith: I think this can be simplified to avoid using the `calc` function.  If believe the following line achieves the same results:
```scss
background-position: right 14px top 15px;
```

scss/components/side-nav.scss
```diff
@@ -124,12 +124,31 @@ $block: #{$fd-namespace}-side-nav;
...
+        margin-top: -1px;
+
+        &.sap-icon--s {
+            margin-top: 0;
+        }
+
+        &.sap-icon--m {
+            margin-top: 0;
+        }
+
```
greg-a-smith: I don't think you need small or medium blocks since the value is just 0.  I don't think icons have a top margin by default.

scss/components/side-nav.scss
```diff
@@ -124,12 +124,31 @@ $block: #{$fd-namespace}-side-nav;
...
+        }
     }
 
     &--icons {
-        min-width: auto;
+        min-width: 75px;
         .#{$block}__icon {
             padding-right: 0;
+            position: relative;
+            left: calc(50% - 4px);
```
greg-a-smith: When is this class used?

scss/components/side-nav.scss
JYSW380: yeah, it will give the same position. thanks for the comments.

scss/components/side-nav.scss
JYSW380: there is one more case that is sap-icon w/o any size indicated, and it is  margin-top: -1px, so if i don't give small or medium blocks, it will take margin-top: -1px.

scss/components/side-nav.scss
JYSW380: this is for Side navigation collapsed with icons, if there is no that two lines code,
please see screenshot:
Screen Shot 2019-06-27 at 4 50 48 PM

scss/components/side-nav.scss
bcullman: `calc()` is a perf sink. It should only ever be used specific use cases. (see https://codepen.io/nicolasjengler/post/don-t-misuse-the-calc-css-function)

we should actually take a pass where it can refactored out.

scss/components/side-nav.scss
greg-a-smith: This entire `&--icons` block is sort of messed up.  It's basically like band-aids and top of band-aids, but never really addresses the primary issue.  I think that entire block can be replaced with this:
```scss
    &--icons {
        min-width: auto;

        .#{$block} {
            &__link,
            &__sublink {
                padding: fd-space(2.5) fd-space(5);
                justify-content: center;
            }

            &__icon {
                padding-right: 0;
            }
        }
    }
```

Rationale:
* Keep the `min-width` set to `auto` so it sizes to the width of its content.
* The padding on links and sublinks needs to be reset so it's the same on the left and right therefore content can be centered without `calc` statements.
* Change the flex box `justify-content` for links to force icons to be centered.
* Keep the removal of `padding-right` for icons.

Issue #71, opened 2019-06-11
Popover in action bar is not aligned with other buttons


JYSW380:
Is this issue related to a specific component?

Action Bar

What version of the Fiori Fundamentals are you using?

v1.4.3

What offering/product do you work on? Any pressing ship or release dates we should be aware of?

Kyma / CDS / ModelT

What front-end framework are you implementing in, e.g., Angular, React, etc.?

Angular

To Reproduce
Create an action bar that has action buttons and popover on the right. Buttons and popover are not vertically aligned.

Expected behavior
Buttons and popover are vertically aligned. .fd-popover {top: 0;} should help to resolve the issue.

Screenshots
<img width="1338" alt="Screen Shot 2019-06-11 at 3 44 53 PM" src="https://user-images.githubusercontent.com/41170611/59301567-e4ce0700-8c5f-11e9-81b1-f5776ee28626.png">

PR #96, opened 2019-06-17
fix: popover button alignment in action bar

JYSW380:
Closes SAP/fundamental-styles#71

fixed popover misalignment issue with different buttons.

Before screenshot -- left side is regular button and right side is a popover
Screen Shot 2019-06-17 at 11 03 45 AM

After screenshot -- left side is regular button and right side is a popover
Screen Shot 2019-06-17 at 11 03 53 AM


JYSW380:
> I'm not sure this is the right solution - should we be hardcoding a top property? Why are the buttons different sizes?

yeah, the problem is in the action bar component, for button component there should not have --l 
see screenshot:
Screen Shot 2019-06-17 at 12 09 21 PM

After screenshot:
Screen Shot 2019-06-17 at 11 53 25 AM

scss/components/action-bar.scss
```diff
@@ -122,7 +122,7 @@ $block: #{$fd-namespace}-action-bar;
           }
         }
         .fd-popover {
-            top: - fd-space(1);
+            top: 0;
```
greg-a-smith: Is this just an issue when popovers wrap components (buttons, in this case) in an action bar?  I suppose so simply because of the selector at work.  Is there a reason to keep the declaration with `top: 0` or should this block just be removed entirely?

JYSW380: yeah, i tried removed  .fd-popover {top: 0}, and it still works. thanks

PR #115, opened 2019-06-19
fix: self-contained styles for label component

### LINKED ISSUE #102: Self-contained styles for Label
Similar as #60

DorinR:
Closes #102 

This PR contains the self-contained styling for the label component. 
This PR is part of the larger task of making all components self-contained.

### DIFF
--- a/components/label.scss
+++ b/components/label.scss
@@ -1,0 +1,6 @@
+$fd-icons-path: "../scss/icons/";
+$fd-fonts-path: "../scss/fonts/";
+
+@import "../scss/components/label";
+@import "../scss/icons/icon";
+@import "../scss/fonts/72";
--- a/components/label.scss
+++ b/components/label.scss
@@ -2,5 +2,5 @@ $fd-icons-path: "../scss/icons/";
 $fd-fonts-path: "../scss/fonts/";
 
 @import "../scss/components/label";
-@import "../scss/icons/icon";
+@import "../scss/core/elements";
 @import "../scss/fonts/72";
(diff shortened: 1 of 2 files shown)


components/label.scss
```diff
@@ -0,0 +1,6 @@
+$fd-icons-path: "../scss/icons/";
+$fd-fonts-path: "../scss/fonts/";
+
+@import "../scss/components/label";
+@import "../scss/icons/icon";
```
greg-a-smith: Do we need icons passed along with labels?

test/resources/theme-ugly.css
```diff
@@ -1,5 +1,6 @@
 :root {
     --fd-color-action-1: magenta;
+    --fd-color-action: orangered;
```
greg-a-smith: Do we really have CSS variables names `--fd-color-action` and then `--fd-color-action-1`?

DorinR: It depends on the component. For button they use the variables you defined previously (the ones with a number appended), but for more simple components such as the label, it is `--fd-color-action` (without a number appended) that is used.

components/label.scss
DorinR: Not with this component. 👍

test/resources/theme-ugly.css
droshev: We need to tackle the colors, but it can be done separately.
We can just import the sass(and css variable) file from the theme designer and use only those variables.

DorinR:
Currently working on another change for this PR, please do not review just yet!

PR #118, opened 2019-06-20
fix: added self-contained style for link component

### LINKED ISSUE #101: Self-contained styles for Link
Similar as #60

DorinR:
Closes #101 

This PR contains the self-contained styling for the link component. 

### DIFF
--- a/components/link.scss
+++ b/components/link.scss
@@ -2,6 +2,5 @@ $fd-icons-path: "../scss/icons/";
 $fd-fonts-path: "../scss/fonts/";
 
 @import "../scss/components/link";
-@import "../scss/core/elements";
 @import "../scss/icons/icon";
 @import "../scss/fonts/72";
--- a/components/link.scss
+++ b/components/link.scss
@@ -1,0 +1,7 @@
+$fd-icons-path: "../scss/icons/";
+$fd-fonts-path: "../scss/fonts/";
+
+@import "../scss/components/link";
+@import "../scss/core/elements";
+@import "../scss/icons/icon";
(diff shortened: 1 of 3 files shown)


components/link.scss
```diff
@@ -0,0 +1,7 @@
+$fd-icons-path: "../scss/icons/";
+$fd-fonts-path: "../scss/fonts/";
+
+@import "../scss/components/link";
+@import "../scss/core/elements";
```
greg-a-smith: Importing the entire core elements file is not what we want to do here.  We need to find out what from this file is necessary to render links properly and bring those into the link class definitions -- either via common mixins or directly.  The whole point of self-contained styles is not only bring along what is needed and not do the changes using elements.

greg-a-smith:
Additionally, I have created #142 to handle changing the way the links are "underlined".  They need to use the `text-decoration` style property and not use a bottom border.  Not part of the this PR, but needs to be changed.

scss/components/link.scss
```diff
@@ -10,3 +10,10 @@ $block: #{$fd-namespace}-link;
 .#{$block} {
   @include fd-link();
 }
+
+a.#{$block} {
+  font-size: $fd-font-size;
+  font-family: $fd-font-family;
+  line-height: $fd-line-height;
+  -webkit-font-smoothing: antialiased;
+}
```
greg-a-smith: These styles are all in the `fd-reset` mixin.  You could remove this block and the change the block above to this:
```scss
.#{$block} {
    @include fd-reset();
    @include fd-link();
}
```

PR #120, opened 2019-06-21
fix: added self-contained style for breadcrumb

### LINKED ISSUE #100: Self-contained styles for Breadcrumb
Similar as #60

DorinR:
Closes #100 

This PR contains the self-contained styling for the breadcrumb component. 


components/breadcrumb.scss
```diff
@@ -0,0 +1,7 @@
+$fd-icons-path: "../scss/icons/";
+$fd-fonts-path: "../scss/fonts/";
+
+@import "../scss/components/breadcrumb";
+@import "../scss/core/elements";
```
greg-a-smith: Importing the entire core elements file is not what we want to do here. We need to find out what from this file is necessary to render links properly and bring those into the link class definitions -- either via common mixins or directly. The whole point of self-contained styles is not only bring along what is needed and not do the changes using elements.

components/breadcrumb.scss
```diff
@@ -0,0 +1,7 @@
...
+@import "../scss/icons/icon";
```
greg-a-smith: Would icons come with this component?

greg-a-smith:
PR should have a `fix:` prefix in its title.

DorinR:
Thanks @greg-a-smith , my most recent commit to this branch addresses your comments.

scss/components/breadcrumb.scss
```diff
@@ -67,3 +67,26 @@ $block: #{$fd-namespace}-breadcrumb;
...
+    &:focus {
+      outline-style: dotted;
+      outline-width: 1px;
+      @include fd-var-color("outline-color", fd-color-state("hover", "action"), --fd-color-action-focus);
+    }
+    @include fd-disabled() {
+      @include fd-var-color("outline-color", fd-color-state("disabled", "action"), --fd-color-action-disabled);
+      cursor: not-allowed;
+    }
+}
```
greg-a-smith: Styles should not be done on plain elements (`<a>` in this case) otherwise these styles can bleed into other components or elements on a page.

scss/components/breadcrumb.scss
```diff
@@ -66,4 +66,27 @@ $block: #{$fd-namespace}-breadcrumb;
...
+        &:focus {
+          outline-style: dotted;
+          outline-width: 1px;
+          @include fd-var-color("outline-color", fd-color-state("hover", "action"), --fd-color-action-focus);
+        }
+        @include fd-disabled() {
+          @include fd-var-color("outline-color", fd-color-state("disabled", "action"), --fd-color-action-disabled);
+          cursor: not-allowed;
+        }
+    }
```
greg-a-smith: This block needs to be changed because it's tied directly to an element (`<a>` in this case).  Although it's a child of the entire block class, styles can bleed in or out of this without a class declaration.  Additionally, are all these really needed?  Shouldn't the links just be using the `.fd-link` class anyway and then offer minor adjustments to it?

scss/components/breadcrumb.scss
greg-a-smith: I think that block could be switched over to be the `.fd-breadcrumb__link` class.

greg-a-smith:
@DorinR After merging the latest from `master`, which had some fixes for the un-normalize toggle, I added two calls to the `fd-reset` mixin to fix these up.

PR #121, opened 2019-06-21
fix: added self-contained styles for the badge component

### LINKED ISSUE #99: Self-contained styles for Badge
similar as #60

DorinR:
Closes #99 

This PR contains the self-contained styling for the badge component. 


components/badge.scss
```diff
@@ -0,0 +1,7 @@
+$fd-icons-path: "../scss/icons/";
+$fd-fonts-path: "../scss/fonts/";
+
+@import "../scss/components/badge";
+@import "../scss/core/elements";
```
greg-a-smith: Importing the entire core elements file is not what we want to do here. We need to find out what from this file is necessary to render links properly and bring those into the link class definitions -- either via common mixins or directly. The whole point of self-contained styles is not only bring along what is needed and not do the changes using elements.

components/badge.scss
```diff
@@ -0,0 +1,7 @@
...
+@import "../scss/icons/icon";
```
greg-a-smith: Do icons need to come with this component?

test/resources/theme-ugly.css
```diff
@@ -1,5 +1,6 @@
 :root {
     --fd-color-action-1: magenta;
+    --fd-color-action: orangered;
```
greg-a-smith: Looks like there are multiple other CSS variables you could put into the ugly theme.  Check the `scss/components/badge.scss` file for other uses.

test/resources/theme-ugly.css
```diff
@@ -14,4 +14,17 @@
...
+    --fd-color-text-5: gold;
+    --fd-color-action: rgb(3%, 96%, 28%);
+    --fd-color-positive: rgb(87%, 33%, 21%);
+    --fd-color-alert: rgb(49%, 35%, 38%);
+    --fd-color-negative: rgb(9%, 75%, 29%);
+    --fd-color-background-3: rgb(47%, 36%, 91%);
+    --fd-color-background-4: rgb(51%, 6%, 2%);
+    --fd-color-alert: rgb(84%, 13%, 91%);
+    --fd-color-background-5: rgb(13%, 71%, 90%);
+    --fd-color-negative: rgb(31%, 59%, 38%);
```
greg-a-smith: `--fd-color-alert` and `--fd-color-negative` are in this list twice.

PR #122, opened 2019-06-21
fix: added self-contained style for the status-label component

### LINKED ISSUE #98: Self-contained styles for Status Label
similar as #60

DorinR:
Closes #98 

This PR contains the self-contained styling for the status-label component. 


components/status-label.scss
```diff
@@ -0,0 +1,8 @@
+$fd-icons-path: "../scss/icons/";
+$fd-fonts-path: "../scss/fonts/";
+
+@import "../scss/components/status-label";
+@import "../scss/core/elements";
+@import "../scss/core/root";
```
greg-a-smith: Importing the entire core elements and root files is not what we want to do here. We need to find out what from these files is necessary to render status labels properly and bring those into the link class definitions -- either via common mixins or directly. The whole point of self-contained styles is not only bring along what is needed and not do the changes using elements.

components/status-label.scss
```diff
@@ -0,0 +1,8 @@
...
+@import "../scss/icons/icon";
```
greg-a-smith: Do icons need to come with this component?

test/resources/theme-ugly.css
```diff
@@ -1,5 +1,6 @@
 :root {
     --fd-color-action-1: magenta;
+    --fd-color-action: orangered;
```
greg-a-smith: Looks like there are multiple other CSS variables you could put into the ugly theme.  Check the `scss/components/status-label.scss` file for other uses.

greg-a-smith:
PR should have a `fix:` prefix in its title.

scss/components/status-label.scss
```diff
@@ -180,3 +180,11 @@ $fd-status-label-icon-size: fd-space(4) !default;
...
     }
 }
+
+:root {
+    --fd-color-text-1: #32363a;
+    --fd-color-text-2: #51555a;
+    --fd-color-positive: #0a7e3e;
+    --fd-color-alert: #e9730c;
+    --fd-color-negative: #bb0000;
+}
```
greg-a-smith: The definition of these root-level variables shouldn't occur within a component's scss file.  The component scss file can use these variables, but it shouldn't define their values.

scss/components/status-label.scss
```diff
@@ -45,10 +45,8 @@ $fd-status-label-icon-size: fd-space(4) !default;
 
 
 .#{$block} {
-
-
-  --fd-status-label-color: var(--fd-color-text-2);
-  --fd-status-label-icon-background-color: var(--fd-color-text-2);
+  @include fd-var-size("--fd-status-label-color", $fd-color-text-2, --fd-color-text-2);
+  @include fd-var-size("--fd-status-label-icon-background-color", $fd-color-text-2, --fd-color-text-2);
```
greg-a-smith: I don't think this is the proper use of the this mixin.  The first parameter should be the CSS property (_e.g._ `color`, `background-color`, `height`, etc.). Same for the additional uses below.

PR #123, opened 2019-06-21
fix: added self-contained styles for alert component

### LINKED ISSUE #97: Self-contained styles for Alert
similar as https://github.com/SAP/fundamental-styles/pull/60

SaiShan:
Closes #97 

This PR contains the self-contained styling for the alert component.
This PR is part of the larger task of making all components self-contained #136


components/alert.scss
```diff
@@ -0,0 +1,9 @@
+$fd-icons-path: "../scss/icons/";
+$fd-fonts-path: "../scss/fonts/";
+
+@import "../scss/components/alert";
+@import "../scss/components/link";
+@import "../scss/core/elements";
+@import "../scss/core/root";
```
greg-a-smith: Importing the entire core elements and root files is not what we want to do here. We need to find out what from these files is necessary to render alerts properly and bring those into the link class definitions -- either via common mixins or directly. The whole point of self-contained styles is not only bring along what is needed and not do the changes using elements.

scss/components/alert.scss
```diff
@@ -161,3 +161,8 @@ $block: #{$fd-namespace}-alert;
       }
     }
 }
+
+p {
+  margin-bottom: 0;
+  margin-top: 0;
+}
```
greg-a-smith: Styles cannot be applied directly to elements (`<p>` in this case) as they can bleed into other components on the page.

scss/mixins/_mixins.scss
```diff
@@ -239,10 +242,18 @@
...
   @include fd-disabled() {
     @include fd-var-color("color", map-get($fd-colors-action-states, "disabled"), --fd-color-action-disabled);
     @include fd-var-color("border-bottom-color", map-get($fd-colors-action-states, "disabled"), --fd-color-action-disabled);
+    @include fd-var-color("outline-color", fd-color-state("disabled", "action"), --fd-color-action-disabled);
+    cursor: not-allowed;
+  }
+  &:focus {
+    outline-style: dotted;
+    outline-width: 1px;
+    @include fd-var-color("outline-color", fd-color-state("hover", "action"), --fd-color-action-focus);
```
greg-a-smith: If all the changes in this file are needed, I would expect these to be part of the link component's PR.  Once they are there, this PR can take advantage of those same styles.

scss/components/alert.scss
SaiShan: @greg-a-smith Thank you for the review, will create a class for it to prevent the bleeding.

SaiShan:
Do not review this PR just yet, might have another fix.

PR #130, opened 2019-06-26
fix: added self-contained styles for spinner component

JYSW380:
Closes SAP/fundamental-styles#127

This PR contains the self-contained styling for the spinner component.


components/spinner.scss
```diff
@@ -0,0 +1,4 @@
+$fd-fonts-path: "../scss/fonts/";
```
jbadan: I don't think this component needs fonts to work

JYSW380: no, we dont need fonts, thanks for point out.

test/templates/spinner/index.njk
```diff
@@ -18,13 +18,12 @@
...
-<h2>[aria-busy="true"]</h2>
-<p>The spinnner is most often used within another component to indicate the loading state. Any container (even <code>ui-, app-, and page-level</code>) can be "activated" by setting <code>aria-busy="true"</code>.
+<h2>is-busy</h2>
+<p>The spinnner is most often used within another component to indicate the loading state. Any container (even <code>ui-, app-, and page-level</code>) can be "activated" by setting <code>is-busy</code>.
 </p>
-<p>The spinner should be included inside the container. Its visibility can be toggled in relation to the <code>aria-busy</code> attribute. They should always be opposites, i.e, if currently loading, <code>section[aria-busy="true"]</code>, <code>.fd-spinner[aria-hidden="false"]</code>, once the content is loaded, toggle to false and true respectively.</p>
 
     {% set example %}
-<div class="fd-panel" aria-busy="true">
+<div class="fd-panel is-busy">
```
greg-a-smith: The "hidden" example should be redone as it no longer is triggered by `aria-hidden`.  The text should be changed and the example itself should implement the `is-hidden` class.

test/resources/theme-ugly.css
```diff
@@ -1,33 +1,33 @@
...
-
-    --fd-color-text-1: rgb(93%, 2%, 45%);
-    --fd-color-background-6: powderblue;
-    --fd-color-text-5: gold;
-    --fd-color-positive: rgb(87%, 33%, 21%);
-    --fd-color-negative: rgb(9%, 75%, 29%);
-    --fd-color-background-3: rgb(47%, 36%, 91%);
-    --fd-color-background-4: rgb(51%, 6%, 2%);
-    --fd-color-alert: rgb(84%, 13%, 91%);
-    --fd-color-background-5: rgb(13%, 71%, 90%);
```
greg-a-smith: Really just reorganized/reordered these.  The only addition was `--fd-color-background-1`.

saad-mo:
this component (`fd-spinner`) is deprecated in favor of `fd-loading-spinner` and `fd-loading-dots` - https://sap.github.io/fundamental/components/loading-spinner.html

this should be removed from the styles lib and we should work on the new ones instead.

@greg-a-smith @jbadan @droshev  thought?

greg-a-smith:
@saad-mo If this component is deprecated, I agree we should begin work on its replacements.  That said, until those replacements exist, this needs to stay so we should push this PR through and create a GitHub issue to have it removed once the replacement components are available.

Issue #131, opened 2019-06-27
Self-contained styles for side-nav component


DorinR:
Similar to #60

Issue #136, opened 2019-06-27
Self-contained styles for all components


SaiShan:
Making all components self-contained. 

Legend: 
`In Progress` - the component name has a link to the relevant GitHub issue 
`Ready for Review` - the component is ready to be reviewed 
`Completed` - the checkbox in front of the component name is checked.

- [x] Action Bar #211 `Completed`
- [x] Alert #97 
- [x] Badge #99 
- [x] Breadcrumb #100 
- [x] Button #60 #193 
- [x] Button-group #145 `Completed`
- [x] Button-split `Completed`
- [x] Calendar `Completed`
- [x] Combobox Input `Completed`
- [x] Contextual Menu `REMOVED`
- [x] Counter #134 `Complete`
- [x] Date Picker `REMOVED`
- [ ] Dropdown
- [x] Form #167 `in Progress` JENNA
- [x] Icon #147`Completed` 
- [X] Identifier #128 `Completed` 
- [x] Image #124 `Completed` 
- [x] Inline Help #146 `Completed`
- [x] Input Group #165 `In Progress` GREG
- [x] Label #102 `Completed`
- [x] Link #101 `Completed`
- [x] List Group #212  `Completed`
- [X] Loading Spinner #129  (to be removed once the replacement is ready)
- [ ] Localization Editor
- [x] Menu #158 `Completed`
- [ ] Modal #256 `Ready for Review` MARK
- [x] Multi Input `Completed`
- [x] Pagination #220 `Complete`
- [x] Panel #222 `Complete`
- [x] Panel-grid `Complete`
- [x] Popover #148 `Completed`
- [x] ~~Product-menu #266~~ `DONE`
- [x] ~~Product-switcher #265~~ `Ready for Review` DENO
- [x] Product-tile #160 `Completed`
- [ ] Search Input
- [ ] Shellbar
- [x] Side Navigation #131 `Completed`
- [x] Status Label #98 `Completed`
- [x] Table #278 `Completed`
- [X] Tabs #166 `Completed` 
- [x] Tile #149 `Completed`
- [x] Tile-grid `Spec Change for Layout Grid`
- [x] Time `Completed` MARK
- [x] Time Picker `REMOVED`
- [x] Token #132 `Completed` 
- [x] Toggle #213 `Completed`
- [x] Tree `Ready for Review` JENNA
- [X] ~~User Menu #247~~ `Removed`
- [ ] Overlay #291  `Ready for Review` GREG
- [x] Section #271 `DONE`
[...]

PR #137, opened 2019-06-27
fix: added self-contained styles for side-nav component

JYSW380:
Closes SAP/fundamental-styles#131

This PR contains the self-contained styling for the side nav component.


components/side-nav.scss
```diff
@@ -0,0 +1,5 @@
+$fd-icons-path: "../scss/icons/";
+
+@import "../scss/components/side-nav";
+@import "../scss/icons/icon";
```
greg-a-smith: I'm pretty sure this will need fonts as well.

scss/components/side-nav.scss
```diff
@@ -133,3 +136,7 @@ $block: #{$fd-namespace}-side-nav;
         }
     }
 }
+ul, ol {
+    margin-top: 0;
+    margin-bottom: 0;
+}
```
greg-a-smith: Styles should not be tied to element selectors (`<ul>` and `<ol>` in this case) as styles can bleed in or out to other components on a page.

scss/components/side-nav.scss
```diff
@@ -54,6 +54,8 @@ $block: #{$fd-namespace}-side-nav;
         text-transform: uppercase;
         color: $fd-side-nav-link-disabled-color;
         background: $fd-side-nav-title-background-color;
+        font-size: $fd-font-size;
+        font-family: $fd-font-family;
```
JYSW380: give this 2 lines code because @import "../scss/fonts/72"; didn't change the font style that shown in the playground

greg-a-smith: This component should just include the mixin `fd-reset`, which will do this and a few other things that we can no longer depend on from the top-level DOM nodes.

greg-a-smith:
Somewhere along the line, there was a merge conflict that wasn't resolved properly and the alignment of side nav icons was off.  I have fixed that with the latest commit.

Issue #65, opened 2019-06-11
Bug: Input-group button overflow


SaiShan:
Input-group button overflows when you click on it, seems to be on both active and focus.
Components affected by this issue:
* Combobox Input
* Date Picker
* Input Group
* Localization Editor
* Multi Input
* Search Input
* Time Picker

![button-overflow](https://user-images.githubusercontent.com/25252963/59300650-b3ecd280-8c5d-11e9-85bf-05ae18b5c4fb.png)

Issue #140, opened 2019-06-28
Update light button 


SaiShan:
Update the light button component to match the newest design sheet.

Issue #138, opened 2019-06-28
Better selector definitions for block-level classes


greg-a-smith:
There are some components that use the following syntax to define base styles for its component:
```scss
$block: #{$fd-namespace}-identifier;

@at-root {
    [class*="#{$block}"] {
        // style definitions go here
    }
}
```

Selectors like this can be considered a performance issue.  See https://github.com/csslint/csslint/wiki/disallow-selectors-that-look-like-regular-expressions for more information.

Instead, these base styles should be tied to a base class selector.  Using the example above, it should look like this:
```scss
$block: #{$fd-namespace}-identifier;

.#{$block} {
    // style definitions go here
}
```

This will require the markup specs to change.  Currently, elements for these components don't specify a base class -- they only specify the extensions of it and then rely on the attribute regex selector to apply the base styles.

droshev:
it is only for columns for now - let's close and see whether there will be new components in the future

PR #77, opened 2019-06-12
fix: Fixed datepicker button overflow

SaiShan:
Closes SAP/fundamental-styles#65

Fixed datepicker button overflow as per issue #65. 
See screenshot. 

Screen Shot 2019-06-13 at 10 39 28 AM

PR #141, opened 2019-06-28
fix: fixed Datepicker button overflow

SaiShan:
Closes SAP/fundamental-styles#65

Note: On-hover, the blue border does not appear due to styling issues with the light button. An issue for it was already created for it #140 .
Screen Shot 2019-06-28 at 11 42 24 AM


scss/components/button.scss
```diff
@@ -221,6 +224,8 @@ a.#{$block} {
   }
   //.fd-button--light
   &--light {
+    --fd-forms-height: 34px;
+    --fd-forms-height-compact: 26px;
```
greg-a-smith: It doesn't _seem_ like this is the right place for these variables.  It seems like height-related values should be for all buttons, not just the `--light` variation.

scss/components/button.scss
```diff
@@ -76,7 +76,10 @@ a.#{$block} {
       }
       &:empty {
         padding: 0;
-        &::before,
+        &::before {
+          margin-right: 0;
+          margin-left: 0;
+        }
```
greg-a-smith: I _think_ there's a much easier way of doing this.  Instead of resetting margins back to 0, we should just _only_ add the margins when it's a button with an icon _AND_ text (not:empty).  That said, that will probably come as a future refactoring effort because it will require base class selectors (_e.g._ `.sap-icon`).  This will probably become part of #138.

For this PR, you could simply combine the `before` and `after` pseudo elements into a better sass syntax:
```scss
&:empty {
    padding: 0;

    &::before,
    &::after {
        margin-right: 0;
        margin-left: 0;
    }
}
```

scss/components/input-group.scss
```diff
@@ -217,4 +223,9 @@ $block: #{$fd-namespace}-input-group;
             z-index: map-get($fd-z-index-levels, "second");
         }
     }
+    .fd-input {
+        &:focus {
+          box-shadow: none;
```
greg-a-smith: So, is this the new focus styling for inputs in input groups (or just inputs in general)?

SaiShan: @greg-a-smith This fixes the difference in styling pointed out in this comment https://github.com/SAP/fundamental-styles/pull/77#pullrequestreview-250640425. Also, represent the latest design. 

As for its application, I reached out to @LeoT7508 to have more details. For the mean time, I can use a more specific class to avoid bleeding while getting a response from the design team.

SaiShan:
> I think this PR needs to be refocused to be a change to input groups as other input groups using buttons (Combobox Input, Time Picker, etc.) also have similar issues. The behavior of the inputs and the buttons with regards to focus and hover states need to be consistent. Also, these are not working correctly when viewing in RTL. Fixing generic input groups should, in turn, fix the other implementations of the input groups.

I agree, focusing on input-group will resolve issues of component using the input-group with button. For the consistency, the design were changed for inputs and light button. Issues were created to fix this; #153 and  #140 for inputs and light button respectively.

SaiShan:
> These changes still seem focused on just DatePicker. I think this PR should be renamed to include all input groups and then make changes that fix all implementations of it.

Ok. Will reopen the PR when the changes for all components will be addressed.

Issue #147, opened 2019-07-03
Self-contained styles for Icon


SaiShan:
Similar to #60.

SaiShan:
Not required.

PR #155, opened 2019-07-04
fix: Added self-contained style for icon component

SaiShan:
Closes SAP/fundamental-styles#147

This PR contains the self-contained styling for the icon component.


components/icon.scss
```diff
@@ -0,0 +1,3 @@
+$fd-icons-path: "../scss/icons/";
+
+@import "../scss/icons/icon";
```
greg-a-smith: There is additional work needed in the `scss/icons/icon.scss`.  If you look at this in the playground and turn the "un-normalize" toggle on, the unsized (default) icon is showing up as 24px (because of the default font size on the body).  If the default is supposed to be 14px, then the icon base needs to define that.

On a separate note, I am not a big fan of this selector:
```scss
[class*="#{$block}"] {
  @include fd-icon-base();
}
```
But, I know these exist all over the place and it's probably a larger scale story to change.

SaiShan: @greg-a-smith Proper changes have been addressed.

Issue #69, opened 2019-06-11
Bug: is-disabled class


SaiShan:
The is-disabled class does not render the disabled state for a `<span>`.

See the screenshot.

<img width="928" alt="broken" src="https://user-images.githubusercontent.com/25252963/59301177-05499180-8c5f-11e9-932a-9138b0492144.png">

Related issue: #68.

salarenko:
Issue has been fixed
![image](https://user-images.githubusercontent.com/17496353/73940119-2a27fe80-48eb-11ea-9bd6-043892d4b30e.png)

PR #156, opened 2019-07-04
fix: Fixed is-disabled class on span element in time picker component

SaiShan:
Closes SAP/fundamental-styles#69

Fixed is-disabled class on span element in time picker component. See screenshot below. 

Screen Shot 2019-07-04 at 11 39 27 AM

### DIFF
--- a/scss/components/input-group.scss
+++ b/scss/components/input-group.scss
@@ -163,6 +163,11 @@ $block: #{$fd-namespace}-input-group;
                 width: 0;
             }
         }
+        &.is-disabled {
+            border-color: var(--fd-color-neutral-2);
+            opacity: 0.4;
+            cursor: not-allowed;
+        }
     }
     &--before {
         #{$fd-elements-inputs--text} {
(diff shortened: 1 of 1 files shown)


scss/components/input-group.scss
```diff
@@ -174,6 +179,9 @@ $block: #{$fd-namespace}-input-group;
         #{$fd-elements-inputs--text} {
             border-top-right-radius: 0;
             border-bottom-right-radius: 0;
+            &:disabled {
+                opacity: 0.4;
+            }
```
greg-a-smith: Both of these changes look like they should be coming from a mixin.  Also seems like this should apply to the `--before` modifier too.

SaiShan: > Although these changes are in the `input-group.scss` file, I think this PR should be more focused on input groups as a whole and then time picker will just inherit the fix. I believe there is already a set of changes queued up to address input groups so maybe this PR gets closed in favor of those?

Ok will do.

Issue #128, opened 2019-06-25
Self-contained styles for Identifier


DorinR:
Similar to #60

PR #157, opened 2019-07-04
feat: added self-contained style for identifier component

JYSW380:
Closes SAP/fundamental-styles#128

This PR contains the self-contained styling for the identifier component.


scss/components/identifier.scss
```diff
@@ -111,9 +112,53 @@ $block: #{$fd-namespace}-identifier;
...
+.fd-has-background-color-accent-8 {
+  background-color: var(--fd-color-accent-8, map-get($fd-colors-accent, 8));
+}
+
+.fd-has-background-color-accent-9 {
+  background-color: var(--fd-color-accent-9, map-get($fd-colors-accent, 9));
+}
+
+.fd-has-background-color-accent-12 {
+  background-color: var(--fd-color-accent-12, map-get($fd-colors-accent, 12));
```
greg-a-smith: Where is accent 12 coming from?  I only see 9 in the playground and docs.  And if there is a "12", what about "10" and "11"?

test/resources/theme-ugly.css
```diff
@@ -30,4 +31,15 @@
...
+    --fd-color-accent-1: #bc3f8c;
+    --fd-color-accent-2: #6a68a8;
+    --fd-color-accent-3: #4f9344;
+    --fd-color-accent-4: #32c71c;
+    --fd-color-accent-5: #0cd06b;
+    --fd-color-accent-6: #253470;
+    --fd-color-accent-7: #a8fac7;
+    --fd-color-accent-8: #945ef7;
+    --fd-color-accent-9: #e3ba20;
+    --fd-color-accent-12: #41cc50;
```
greg-a-smith: Same comment about accent 12.

scss/components/identifier.scss
JYSW380: accent 12 is the color for this at Spec section:
Screen Shot 2019-07-11 at 9 23 12 AM

scss/components/identifier.scss
```diff
@@ -101,19 +102,58 @@ $block: #{$fd-namespace}-identifier;
...
-      --fd-identifier-color: var(--fd-color-text-1);
+    &.#{$block}--transparent {
       --fd-identifier-background-color: transparent;
-      @include fd-var-color("color", $fd-color);
+      @include fd-var-color("color", map-get($fd-colors-text, 1), --fd-color-text-1);
       @include fd-var-color("background-color", transparent);
     }
 }
+
+.fd-has-background-color-accent-1 {
```
greg-a-smith: These classes seem to be independent of identifier so they should either be renamed as part of identifier or be moved to a separate location to be shared.

Also, the `$fd-colors-accent` map has 15 colors defined -- we should include classes for each of them.

test/resources/theme-ugly.css
```diff
@@ -40,5 +40,10 @@
     --fd-color-accent-7: #a8fac7;
     --fd-color-accent-8: #945ef7;
     --fd-color-accent-9: #e3ba20;
+    --fd-color-accent-10: #5c819d;
+    --fd-color-accent-11: #d17f15;
     --fd-color-accent-12: #41cc50;
+    --fd-color-accent-13: #2b78c5;
+    --fd-color-accent-14: #6367de;
+    --fd-color-accent-15: #984bd2;
```
greg-a-smith: Since we have 15 accent backgrounds now, seems like we should show all 15 on the playground page (instead of just 9).

greg-a-smith:
The changes now require the use of the `fd-identifier` base class so this is a breaking change.  I am changing the PR prefix to `feat:`.

Issue #149, opened 2019-07-03
Self-contained style for tile component


JYSW380:
Similar to #60 

Self-contained styles for all components #136

PR #159, opened 2019-07-05
fix: added self-contained style for tile component

JYSW380:
Closes SAP/fundamental-styles#149

This PR contains the self-contained styling for the tile component.


components/tile.scss
```diff
@@ -0,0 +1,7 @@
+$fd-icons-path: "../scss/icons/";
+
+@import "../scss/components/tile";
+@import "../scss/icons/icon";
+@import "../scss/components/image";
+@import "../scss/components/button";
+@import "../scss/components/identifier";
```
greg-a-smith: This is another component that is a little strange to apply self-contained styles to because the majority of the style content loaded here is for components other than the tile.  I know these were added because of the playground examples, but honestly the playground examples should be reworked to _not_ include buttons, identifiers or images as those are completely self-contained themselves.  The playground examples can show a `<button>` element, but it shouldn't have any `fd-button` classes on it.  Similarly, images and identifiers can show an `` or a `<div>` element, but those should not have any `fd-image` or `fd-identifier` classes.

On the main docs site, we can keep the buttons, images and identifiers in place as those are more use case-related.

scss/components/button.scss
```diff
@@ -554,8 +554,9 @@ a.#{$block} {
     &--secondary {
       --fd-button-border-color: transparent;
       @include fd-hover() {
-        --fd-button-background-color: var(--fd-color-neutral-1);
-        --fd-button-color: var(--fd-color-action-1);
+        @include fd-var-color("color", map-get($fd-colors-action, 1), --fd-button-color);
+        @include fd-var-color("background-color", map-get($fd-colors-neutral, 1),--fd-button-background-color);
+
```
greg-a-smith: These changes shouldn't be part of self-contained styles for tiles.

scss/components/identifier.scss
```diff
@@ -111,9 +111,8 @@ $block: #{$fd-namespace}-identifier;
     }
     //background
     &--transparent {
-      --fd-identifier-color: var(--fd-color-text-1);
       --fd-identifier-background-color: transparent;
-      @include fd-var-color("color", $fd-color);
+      @include fd-var-color("color", map-get($fd-colors-text, 1), --fd-color-text-1);
```
greg-a-smith: These changes shouldn't be part of self-contained styles for tiles.

scss/components/tile.scss
```diff
@@ -75,9 +75,13 @@ $block: #{$fd-namespace}-tile;
...
-    margin-bottom: 0;
+    @include fd-reset-spacing();
   }
   &__actions {
     padding: fd-space(3) $fd-tile-content-padding-x fd-space(3) fd-space(3);
   }
+
+  .fd-has-background-color-accent-3 {
+    background-color: var(--fd-color-accent-3, map-get($fd-colors-accent, 3));
+  }
```
greg-a-smith: This class has nothing to do with the tile component.  It should either be renamed to be part of tile or moved to a shared location.  See #157 for a similar comment.

test/templates/tile/index.njk
```diff
@@ -99,13 +78,8 @@
...
     {%- endcall %}
     {%- call tile_actions() -%}
-    {{  button(
-        {
-            icon: 'overflow'
-        },
-        modifier={ block: ['light'] },
-        aria={ label: 'More' })
-    }}
+    <button class="sap-icon--overflow" aria-label="More"></button>
```
droshev: maybe we can avoid any `fundamental-styles` classes. Instead you can put a custom class and style it like mode menu

scss/components/tile.scss
```diff
@@ -2,15 +2,15 @@
 @import "./../mixins";
```
jbadan: We should remove use of `&[aria-disabled="true"]` on line 50, we should also update the documentation site with these changes: https://github.com/SAP/fundamental-styles/blob/master/docs/pages/components/tile.md#disabled-state

We missed updating the docs site for button and the others that we removed the aria-* selectors in the scss.

test/templates/tile/index.njk
JYSW380: so you wanna me to remove the icon?

components/tile.scss
```diff
@@ -0,0 +1,4 @@
+$fd-icons-path: "../scss/icons/";
+
+@import "../scss/components/tile";
+@import "../scss/icons/icon";
```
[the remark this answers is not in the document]
JYSW380: we need icons for the overflow icon inside button, and with or without fonts didn't change anything in this component, so i didn't put it.

docs/pages/components/tile.md
```diff
@@ -147,18 +147,18 @@ Add `role=button` to rendering a tile as a button
 <br>
 
 ## Disabled state
-Add class `is-disabled` and/or `aria-disabled="true"` attribute
+Add class `is-disabled` attribute !!!!!
```
jbadan: why so many !!!!! 😆

components/tile.scss
jbadan: Does the Tile component need icons?

docs/pages/components/tile.md
JYSW380: oops, forgot to remove :sweat_smile:

components/tile.scss
JYSW380: please see the screenshot:
Screen Shot 2019-08-02 at 4 08 57 PM
the overflow icon inside button

test/templates/tile/index.njk
droshev: you can have an icon but without using directly fundamental classes

JYSW380: checked at Greg did the button self contained component, icons can be used directly, and it's used @import "../scss/icons/icon" to achieve it
Screen Shot 2019-08-05 at 9 17 25 AM

docs/pages/components/tile.md
```diff
@@ -147,18 +147,18 @@ Add `role=button` to rendering a tile as a button
 <br>
 
 ## Disabled state
-Add class `is-disabled` and/or `aria-disabled="true"` attribute
+Add class `is-disabled` attribute
```
greg-a-smith: This should just say "Add class `is-disabled`" (remove the word "attribute").

docs/pages/components/tile.md
```diff
@@ -147,18 +147,18 @@ Add `role=button` to rendering a tile as a button
...
 
 {% capture tile %}
-<div class="fd-tile" aria-disabled="true">
+<div class="fd-tile" is-disabled>
```
greg-a-smith: The `is-disabled` needs to be within the quotes for the `class` attribute.

docs/pages/components/tile.md
```diff
@@ -147,18 +147,18 @@ Add `role=button` to rendering a tile as a button
...
     <div class="fd-tile__content">
         <h3 class="fd-tile__title">Tile Title</h3>
     </div>
 </div>
 
 <br>
 
-<div class="fd-tile" aria-disabled="true">
+<div class="fd-tile" is-disabled>
```
greg-a-smith: The `is-disabled` needs to be within the quotes for the `class` attribute.

docs/pages/components/tile.md
```diff
@@ -169,7 +169,7 @@ Add class `is-disabled` and/or `aria-disabled="true"` attribute
 
 <br>
 
-<div class="fd-product-tile" aria-disabled="true">
+<div class="fd-product-tile" is-disabled>
```
greg-a-smith: The `is-disabled` needs to be within the quotes for the `class` attribute.

test/templates/tile/component.njk
```diff
@@ -6,7 +6,7 @@ tile:
     aria={}
 -->
 {% macro tile(properties={}, modifier={}, state={}, aria={}, classes=[]) -%}
-<div class="fd-tile{{ modifier.block | modifier('tile') }}{{ state | state }}{{ classes | classes }}"{{ aria | aria }}>
+<div class="fd-tile{{ modifier.block | modifier('tile') }}{{ state | state }}{{ classes | classes }}"{{ aria | aria }} is-disabled>
```
greg-a-smith: The `is-disabled` needs to be adjusted here since it's not an attribute.

components/tile.scss
```diff
@@ -0,0 +1,2 @@
+
+@import "../scss/components/tile";
```
greg-a-smith: This component will need the font imported.

test/templates/tile/index.njk
```diff
@@ -153,14 +127,14 @@
 
     <br><br>
     <h2>disabled</h2>
-    <p>Add <code>aria-disabled=true</code></p>
+    <p>Add <code>is-disabled</code></p>
     {% set example %}
-{%- call tile(aria={disabled: true}) -%}
+{%- call tile() -%}
```
greg-a-smith: These examples no longer appear disabled.  I think a parameter will need to be passed, but just not an aria attribute.

docs/pages/components/tile.md
jbadan: I think we should keep the `aria-disabled='true'` in the examples, but just not on line 150 where we say what selector to use.

Issue #148, opened 2019-07-03
Self-contained style for popover component


DorinR:
Similar to #60

PR #162, opened 2019-07-05
fix: Added self-contained style for Popover component

SaiShan:
Closes SAP/fundamental-styles#148

This PR contains the self-contained styling for the Popover component.


components/popover.scss
```diff
@@ -0,0 +1,8 @@
+$fd-icons-path: "../scss/icons/";
+$fd-fonts-path: "../scss/fonts/";
+
+@import "../scss/components/popover";
+@import "../scss/components/menu";
+@import "../scss/components/image";
```
greg-a-smith: This is another component that is a little strange to apply self-contained styles to because the majority of the style content loaded here is for components other than the popover.  I know these were added because of the playground examples, but honestly the playground examples should be reworked to not include any specific fundamental-related content.  Images and menus shouldn't be used as those are completely self-contained themselves. The playground examples can show a plain element to anchor the popover to. Similarly, the menu content in the popover should be taken out.  It can use a plain `<ul>` with `<li>` elements, but should not have any fd-menu classes.

On the main docs site, we can keep the images and menus in place as those are more use case-related.

droshev:
in the playground there is an example missing with default placement(without specifying explicitly):
Screen Shot 2019-08-01 at 8 04 24 PM
maybe it would be good to have one more example where the popover is:
`<div class="fd-popover">...`

droshev:
I find using `fd-popover__control` in 2 nested elements a bit weird + in this situation we don't gain anything. On top of that neither of implementation libraries are applying it on the child level. Can we remove it?

SaiShan:
@droshev Changed requested were addressed in the latest version!

scss/components/popover.scss
```diff
@@ -270,3 +270,8 @@ $block: #{$fd-namespace}-popover;
         }
     }
 }
+
```
jbadan: We should remove the `[aria-hidden="true"]` selector on line 151.

Is `.popover-body__pgd` just for the playground? If so, it should live in https://github.com/SAP/fundamental-styles/blob/master/test/resources/layout.css

docs/pages/components/date-picker.md
```diff
@@ -24,7 +24,7 @@ This component mostly relies on the CSS of other components and has very little
 			<div class="fd-input-group fd-input-group--after">
 				<input type="text" value="" placeholder="Pick a date">
 				<span class="fd-input-group__addon fd-input-group__addon--after fd-input-group__addon--button">
-				<button class="fd-popover__control fd-button--light sap-icon--calendar" aria-controls="Itcgq828" aria-expanded="false" aria-haspopup="true" aria-hidden="true"></button>
+				<button class="fd-button--light sap-icon--calendar" aria-controls="Itcgq828" aria-expanded="false" aria-haspopup="true" aria-hidden="true"></button>
```
jbadan: Can we remove this? I think this is needed for the implementation of popper.js in the fundamental-* libraries.

docs/pages/components/popover.md
```diff
@@ -49,7 +49,7 @@ There are four placement options:
 {% capture default-popoverwmenu %}
 <div class="fd-popover">
     <div class="fd-popover__control">
-        <button class="fd-button fd-popover__control" aria-label="Image label" aria-controls="jhqD0558a" aria-expanded="false" aria-haspopup="true">Left Aligned (default)</button>
+        <button class="fd-button" aria-label="Image label" aria-controls="jhqD0558a" aria-expanded="false" aria-haspopup="true">Left Aligned (default)</button>
```
jbadan: This one can be removed ✅ 😃

docs/pages/components/time-picker.md
```diff
@@ -27,7 +27,7 @@ A basic example of a time picker.
             <div class="fd-input-group fd-input-group--after">
                 <input type="time" class="fd-input " id="" placeholder="hh:mm am/pm">
                 <span class="fd-input-group__addon fd-input-group__addon--after fd-input-group__addon--button">
-                                  <button class=" fd-button--light sap-icon--history fd-popover__control" aria-controls="rthHR811" aria-expanded="false" aria-haspopup="true"></button>
+                                  <button class=" fd-button--light sap-icon--history" aria-controls="rthHR811" aria-expanded="false" aria-haspopup="true"></button>
```
jbadan: This needs `fd-popover__control` for the implementation libraries.

scss/components/popover.scss
SaiShan: @jbadan thank you, will do. Wasn't quite sure in which file it should be.

jbadan:
After chatting with @SaiShan about removing the duplicate `fd-popover__control`, I would prefer leave it as is in this pr and do it separately because
1) that's a breaking change and the fundamental-* libraries will need to be updated
2) It _might_ be necessary for the implementation of popper.js. We need to play around with this. See https://github.com/SAP/fundamental-react/blob/master/src/Popover/Popover.js#L69

Is that ok with you @droshev?

components/popover.scss
```diff
@@ -0,0 +1,6 @@
+$fd-icons-path: "../scss/icons/";
+$fd-fonts-path: "../scss/fonts/";
+
+@import "../scss/components/popover";
+@import "../scss/icons/icon";
+@import "../scss/fonts/72";
```
greg-a-smith: I don't believe popovers should require either fonts or icons.  The popover is literally just the container that holds overlay content.  Whatever content goes into the popover will bring with it the font and/or icons that might be needed.

test/templates/popover/index.njk
```diff
@@ -1,13 +1,22 @@
...
-{% from "./component.njk" import popover, popover_popperjs %}
+{% from "./component.njk" import popover, popover_popperjs, list %}
 {% from "./../menu/component.njk" import menu %}
 
 {% block content %}
 
 <style media="screen">
     .fd-popover{ margin-left: 100px; margin-right: 100px;}
+
+    .popover-body__pgd {
```
greg-a-smith: I think this should should be renamed to _not_ include "body" in the class name.  There is already a `fd-popover__body` class that defines the body container of the popover, but this class is really specific to the content that goes in that body.  It is also local to this playground file.  Maybe something like `.sample-content`?

test/templates/popover/index.njk
```diff
@@ -1,13 +1,22 @@
...
+    font-family: '72', -apple-system, BlinkMacSystemFont, Segoe UI, Roboto, Helvetica Neue, Arial, sans-serif, Apple Color Emoji, Segoe UI Emoji, Segoe UI Symbol;
```
greg-a-smith: I would avoid using the font 72 here.  This is a playground page and is only intended to highlight the popover styles -- not the popover content styles.

test/templates/popover/component.njk
```diff
@@ -37,3 +36,7 @@ popover:
   <div class="fd-popover__arrow"></div>
 </div>
 {%- endmacro -%}
+
+{%- macro list(properties={}, modifier={}, state={}, aria={ hidden: true }) -%}
```
greg-a-smith: Maybe rename this macro (_e.g._ `popover-content`).

greg-a-smith:
@SAP/fundamental-styles-core-team It shows the ❌ , but that's because Mergeable is waiting on approvals -- somewhat misleading.  This is ready for review.

components/popover.scss
bcullman: (does it use an "X" for closing? - that might be an icon)

jbadan:
@greg-a-smith Can we remove https://github.com/SAP/fundamental-styles/blob/master/scss/components/popover.scss#L151

greg-a-smith:
@jbadan I saw that and looked into it, however, that is currently woven into the fabric of the library and the docs pages.  Eventually, this should be removed, but I think there's a larger story and how to "wire up" examples like this that must occur first.  I just didn't want this PR to go out of control.

Issue #145, opened 2019-07-03
Self-Contained style for button-group component


DorinR:
Similar to #60

PR #163, opened 2019-07-05
fix: added self-contained style for button-group component

JYSW380:
Closes SAP/fundamental-styles#145 

This PR contains the self-contained styling for the button-group component.


components/button-group.scss
```diff
@@ -0,0 +1,5 @@
+$fd-icons-path: "../scss/icons/";
+
+@import "../scss/components/button";
+@import "../scss/components/button-group";
+@import "../scss/icons/icon";
```
greg-a-smith: This is another component that is a little strange to apply self-contained styles to because the majority of the style content loaded here is for the button component, which has its own self-contained styles. I know these were added because of the playground examples, but honestly the playground examples should be reworked to not include any specific fundamental-related content.  Although this is the button group component, the examples shouldn't use the button component for the children.  The essence of a button group is a wrapper around multiple elements (typically buttons although the CSS is written such that it wouldn't have to be) that have rounded corners and borders to make sure they format correctly when in a group.  The playground examples can show a `<button>` element, but shouldn't use any of the `fd-button` classes.  They could be styled with some custom CSS on the playground page itself using a temp class variable name (_e.g._ `btn-playground`).

On the main docs site, we can keep the buttons in place as the children as those are more use case-related.

JYSW380: @greg-a-smith If I understand correctly, I should remove the `fd-button` class and add some custom class called `btn-playground` which has the same style as the fd-button? Where would I put this new class? Is there a file where we put custom sass in the docs? Or maybe just in `<style>` tags?

scss/components/button-group.scss
```diff
@@ -2,6 +2,13 @@
 @import "./../mixins";
 @import "./../functions";
 
+
+.btn-playground {
```
droshev: I think this shouldn't be in that file

droshev:
it would be great to add some values in the `ugly` theme for the btn group.
Also I don't see any constrains of having different than 3 buttons in a group, so it would be good to have examples with different set of buttons

components/button-group.scss
```diff
@@ -0,0 +1,4 @@
+$fd-icons-path: "../scss/icons/";
+
+@import "../scss/components/button-group";
+@import "../scss/icons/icon";
```
greg-a-smith: Icons shouldn't be needed here.

test/templates/button-group/index.njk
```diff
@@ -1,129 +1,74 @@
...
-          aria={ label: 'Chart', selected: true })
-        }}
-    {{-  button(
-          { icon: 'pool' },
-            modifier={ block: [size] },
-            aria={ label: 'Pool' })
-          }}
+    <button class="btn-playground sap-icon--survey" aria-label="Survey"></button>
+    <button class="btn-playground sap-icon--pie-chart" aria-label="Chart"></button>
+    <button class="btn-playground sap-icon--pool" aria-label="Pool"></button>
```
greg-a-smith: These buttons should _not_ be using icon classes from the style library.  These "plain" buttons do not need to look like SAP buttons or look like previous playground examples.  These should literally be very simple-styled buttons that exist only to verify button group styles (_e.g._  border radius on or off, borders not doubling, etc.)

test/templates/button-group/index.njk
```diff
@@ -1,129 +1,74 @@
...
-          aria={ label: 'Chart' })
-        }}
-    {{-  button(
-          { icon: 'pool' },
-            modifier={ block: [""] },
-            aria={ label: 'Pool' })
-          }}
+    <button class="btn-playground sap-icon--survey" aria-label="Survey" disabled></button>
+    <button class="btn-playground sap-icon--pie-chart" aria-label="Chart"></button>
+    <button class="btn-playground sap-icon--pool" aria-label="Pool"></button>
```
greg-a-smith: Same comment as above about icons.

test/templates/button/component-anchor.njk
```diff
@@ -5,6 +5,6 @@
 {%- macro anchor_button(properties={}, modifier={}, state={}, aria={}, classes="", size="") -%}
 {%- set is_default = modifier.block and modifier.block.length and not (modifier.block | join("")) %}
 
-<a class="{{ 'fd-button' if not modifier.block or modifier.block == "" or is_default }}{{ modifier.block | modifier('button') }}{{ ' fd-button--'+size if size }}{{ ' sap-icon--'+properties.icon if properties.icon }}{{ state | state }}{{ classes | classes }}"{{ aria | aria }}>{%- if properties.label %}{{properties.label}}{%- endif %}</a>
+<a class="{{ 'button' if not modifier.block or modifier.block == "" or is_default }}{{ modifier.block | modifier('button') }}{{ ' button--'+size if size }}{{ ' sap-icon--'+properties.icon if properties.icon }}{{ state | state }}{{ classes | classes }}"{{ aria | aria }}>{%- if properties.label %}{{properties.label}}{%- endif %}</a>
```
greg-a-smith: What's this change for?  Seems unrelated to this PR.

JYSW380: you are right, will go back to the original code, thanks.

test/templates/button-group/index.njk
```diff
@@ -1,129 +1,44 @@
...
-        }}
-    {{-  button(
-          { label: 'Third' },
-            modifier={ block: [size] },
-            aria={ })
-          }}
-  {%- endcall %}
 
+<style>
+  .btn-playground {
```
jbadan: @greg-a-smith Can we move this to layout.css or some other shared css file. I think we have a lot of components coming up that need a placeholder button.

greg-a-smith:
@jbadan @hertweckhr1 The `layout.css` actually isn't loaded on the self-contained pages.  That said, the point is still valid so I'll create a new CSS file that _IS_ loaded on self-contained pages that just contains some resusable styles for playground purposes.

test/resources/layout-shared.css
```diff
@@ -0,0 +1,12 @@
+body {
+    background-color: rgb(243, 244, 245);
+    transition: background-color .25s;
```
greg-a-smith: Added this part to shared just to make the self-contained frame easier to see.

PR #168, opened 2019-07-05
fix: possible fix for dimension issue with links

### LINKED ISSUE #142: Link component needs to change underline styling
Currently, the link component (`.fd-link`) is not using `text-decoration` for the underline, but rather a bottom border.  The biggest problem with that is this will change the dimensions of links, which makes them difficult to put inline in a block of text or in other components.

bcullman:
Closes #142 

using bottom-border as a stand-in for underline has some stylistic advantages (it will underline text and icons paired together for example), however it can cause the bounding box to have a different dimension than text without a bottom-border.

This solution uses a `:after` absolutely positioned element to get the same underline effect, but without affecting bounding-box dimension changes.

worth considering, but may not be merge worthy

### DIFF
--- a/scss/mixins/_mixins.scss
+++ b/scss/mixins/_mixins.scss
@@ -225,28 +225,47 @@
 
 @mixin fd-link($color: $fd-color--action) {
   @include fd-var-color("color", $fd-color--action, --fd-color-action-1);
-  text-decoration: none;
+  //text-decoration: underline;
   display: inline-block;
   transition: all $fd-animation--speed ease-in;
-  border-bottom-style: solid;
-  border-bottom-width: 1px;
-  @include fd-var-color("border-bottom-color", $fd-color--action, --fd-color-action-1);
+  //border-bottom-style: solid;
+  //border-bottom-width: 1px;
+  text-decoration: none;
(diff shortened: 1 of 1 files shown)


jbadan:
This looks like a reasonable solution - the only diffs in the tests look like they are due to fixing the problematic bounding-box dimension changes.

scss/mixins/_mixins.scss
```diff
@@ -225,28 +225,44 @@
...
-  border-bottom-width: 1px;
-  @include fd-var-color("border-bottom-color", $fd-color--action, --fd-color-action-1);
+  text-decoration: none;
+  position: relative;
+  &::after {
+    content: '';
+    width: 100%;
+    position: absolute;
+    left: 0;
+    bottom: 1px;
```
greg-a-smith: Looking at the `fd-link` instances from `master`, I think this value should be `bottom: 0;`.

bcullman:
given the long term goal of this library is to align with the other big Firoi3 project, UI5 web components and that ui5 web components uses `text-decoration:underline` for their underline style, the need for this "clever" solutions will become unnecessary. 

closing

PR #169, opened 2019-07-09
fix: Added padding, margin and border to fd-reset mixin

greg-a-smith:
After checking pages with the "Un-normalize" playground toggle, it was found that a number of components were allowing styles to bleed in.  This adds padding, margin and border resets to the `fd-reset` mixin.

Once that was in place, there were a few components that were adjusted to prevent the bleed in of external styling.  They were:

* Alert
* Link (no additional changes were necessary after the `fd-reset` mixin was adjusted)
* Side Nav
* Status Label (changed the modifier classes to be sibling classes rather than standalone)

### DIFF
--- a/scss/components/status-label.scss
+++ b/scss/components/status-label.scss
@@ -48,133 +48,119 @@ $fd-status-label-icon-size: fd-space(4) !default;
   --fd-status-label-color: var(--fd-color-text-2);
   --fd-status-label-icon-background-color: var(--fd-color-text-2);
 
-    //LOCAL VARS (set all themeable properties, always include !default)
+  //LOCAL VARS (set all themeable properties, always include !default)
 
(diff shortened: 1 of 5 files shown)


scss/components/status-label.scss
```diff
@@ -48,133 +48,119 @@ $fd-status-label-icon-size: fd-space(4) !default;
   --fd-status-label-color: var(--fd-color-text-2);
   --fd-status-label-icon-background-color: var(--fd-color-text-2);
 
-    //LOCAL VARS (set all themeable properties, always include !default)
```
greg-a-smith: Unfortunately, this file shows a bunch of changes -- mostly because of fixing the indentation level for many lines.  The essence of the change was to remove the two mixins that weren't being used and then changing the seven modifier classes to be sibling selectors rather than standalone classes.

test/resources/unnormalize.css
```diff
@@ -1,14 +1,18 @@
 body {
     font-size: 24px;
     font-family: Courier, monospace;
```
bcullman: comic sans might be better here. (this is a real comment)

greg-a-smith: Updated.

PR #176, opened 2019-07-11
fix: add SAP Theme designer Quartz colors

droshev:
temporarily include the color mapping from Theme Designer. We can use:
- css variables
- sass variables


droshev:
@greg-a-smith - i just wanted to include the files as we will remove them once the npm package is ready. 
I wanted to use them in the components we are refactoring right now

greg-a-smith:
@droshev One more thing ... since we are on version 0, this should actually be labeled with `fix:` since `feat:` for version 0 is technically a breaking change.

PR #178, opened 2019-07-11
fix: updated light button styling

SaiShan:
Closes SAP/fundamental-styles#140

Updated light button style. A border now appears when you hover. See screenshot. 

Screen Shot 2019-07-11 at 4 11 55 PM


scss/components/button.scss
```diff
@@ -236,6 +236,7 @@ a.#{$block} {
         @include fd-var-color("color", fd-color("action", 1), --fd-button-color);
         @include fd-var-color("background-color", transparent, --fd-button-background-color);
       }
+      box-shadow: 0 0 0 1px var(--fd-color-action-hover, map-get($fd-colors-action-states, "hover"));
```
greg-a-smith: I think there is more to this change as a disabled light button still shows an outline on hover.

PR #181, opened 2019-07-15
feat: added tag component

ChristianKienle:
This PR adds a *tag*-component.

- I have absolutely no strong feelings about the naming.
- The *tag*-component tries to be just like the already existing *token*-component with one small difference: The *close*-button is it's own element.

Other than that those two components should be identical.

Why is that important?

1. `fd-tag` can be used without the *close*-button: Just don't add `<span class="fd-tag__close"></span>` inside the `fd-tag`-span.
2. You can detect clicks on the close-button/element and differentiate clicks on the close button from clicks on the tag-title.

I have adjusted the docs and include two examples: Default tags and RTL-tags.

I would also like to add some kind of `selected`-state – but let's take one step at a time. I have a feeling this PR may require some talking. :D 

Screenshot 2019-07-15 at 15 12 21

### DIFF
--- a/docs/pages/components/tag.md
+++ b/docs/pages/components/tag.md
@@ -1,0 +1,73 @@
+---
+title: Tag
+id: tag
+keywords: tag, token
+sidebar: left-navigation-sidebar
+toc: false
+permalink: components/tag.html
+folder: components
+summary:
+---
+
+Tags are used to represent contextualizing information. They can be useful to show applied filters, selected values for form field or object metadata.
+{: .docs-intro}
+
+<br>
+
+{% capture default-tag %}
+<span class="fd-tag" role="button">
+  Bibendum
+  <span class="fd-tag__close"></span>
+</span>
+
+<span class="fd-tag" role="button">
+  Lorem
+  <span class="fd-tag__close"></span>
+</span>
+
+<span class="fd-tag" role="button">
+  Dolor
+  <span class="fd-tag__close"></span>
+</span>
+
+<span class="fd-tag" role="button">
+  Filter
+  <span class="fd-tag__close"></span>
+</span>
+
+<span class="fd-tag" role="button">
+  Filter
+  <span class="fd-tag__close"></span>
+</span>
+{% endcapture %}
+{% include display-component.html component=default-tag %}
+
+{% capture rtl-tag %}
+<div dir="rtl">
+  <span class="fd-tag" role="button">
+    Bibendum
+    <span class="fd-tag__close"></span>
+  </span>
+
+  <span class="fd-tag" role="button">
+    Lorem
+    <span class="fd-tag__close"></span>
+  </span>
+
+  <span class="fd-tag" role="button">
+    Dolor
+    <span class="fd-tag__close"></span>
+  </span>
+
+  <span class="fd-tag" role="button">
+    Filter
+    <span class="fd-tag__close"></span>
+  </span>
+
+  <span class="fd-tag" role="button">
+    Filter
+    <span class="fd-tag__close"></span>
+  </span>
+</div>
+{% endcapture %}
+{% include display-component.html component=rtl-tag %}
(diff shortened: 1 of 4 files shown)


MattL75:
What do you think about making the `<span class="fd-tag__close">` an actual button? 

Also what is the point of the token component if we add this? I feel like we could just make Tag the new token. I really hate how the Token has the x button rendered as a pseudo-element.

@greg-a-smith

ChristianKienle:
> token

I have thought about making the close-thing an actual button. I have not done this just because the other ui-component libraries/css frameworks I looked at (for reference) also are not doing it. I assume that there might be a good reason as to why but I am not sure what their reason is. Maybe this might cause problems down the road once you want to embed the tags inside an input? Not sure…

The token component could be deprecated. I have not just modified token because of backwards compatibility.

saad-mo:
@ChristianKienle `fd-tag` has been deprecated a while ago in favor of `fd-token` to align with Fiori 3 naming convention - https://sap.github.io/fundamental/components/tag.html 

so token is here to stay not tag :)

MattL75:
@saad-mo Sure, that's fine. We can just replace the token code with the tag code Christian wrote up. The important thing is the optional close button.

@ChristianKienle I would much prefer the button personally, leave it up to everyone though.

greg-a-smith:
@MattL75 @ChristianKienle The "X" element _should_ be an actual `<button>` as it provides us multiple a11y features right out of the box.  That said, the tag/token itself shouldn't be clickable as you can't have a clickable element inside another clickable element.  I _suppose_ the tag/token could be clickable if no "X" button was present, but then what would the use case be for making that clickable?

ChristianKienle:
>  The "X" element should be an actual <button>
Agreed. Will make the necessary change. I was aware about the button debate :) and the only thing that kept me from using a button was that I have never seen it done that way. I looked at several ui component libraries but we certainly do not have to repeat the mistakes of others.

> That said, the tag/token itself shouldn't be clickable as you can't have a clickable element inside another clickable element

We have a private token component that has an x and is selectable. I was trying to use Fundamental Styles for that (`fd-token`) but failed due to multiple reasons – some of which would be solved by this PR.

Our tokens can be quickly deleted or selected.

I am also are about the "no nested clickable elements"-rule. This limitation is kind of sad because it makes complex but useful controls like this:

ezgif com-video-to-gif (1)

impossible. But this is a debate for another time. :)

So – assuming we have a real close-`<button>` then the tag/token itself can't be clickable – right? This means that you won't be able to implement drag and drop (of tokens) and you won't be able to select tokens. I am now trying to think how this could be used by an app. In an App I could either display a bunch of tags/tokens without a close button (and have them selectable/drag and droppable) or display them with the close button but then without the possibility to select/dnd them. Right? I just want to understand the limitations of our potential choices. :)

ChristianKienle:
> @ChristianKienle `fd-tag` has been deprecated a while ago in favor of `fd-token` to align with Fiori 3 naming convention - https://sap.github.io/fundamental/components/tag.html
> 
> so token is here to stay not tag :)

As stated earlier – "I have absolutely no strong feelings about the naming.".

Token was already taken and this change is incompatible – at least potentially. The current `fd-token`-implementation may be aligned with Fiori from a naming perspective but isn't naming one of the least important aspects of a component?

saad-mo:
> Token was already taken and this change is incompatible – at least potentially. The current `fd-token`-implementation may be aligned with Fiori from a naming perspective but isn't naming one of the least important aspects of a component?

We need to have consistency in the naming of components across design files, component libraries and implementation in the project. it does not matter what exactly is called, but we need to maintain consistency. We renamed this component once already to be consistent and does not make sense to go back to the old name.

greg-a-smith:
@ChristianKienle I am all for creating innovative and useful controls, but there are a11y concerns we have to consider when doing this.  As a starting point, ask yourself how would you interact with your control using only a keyboard.  For the example you showed, although each item _looks_ like a token, maybe its semantic meaning is different.  Maybe it's possible to create something like that, but using token as a catch-all component may not be the best route.

MattL75:
My idea of the token is essentially what we currently have, but have an optional close button element which is actually a `button` tag.

droshev:
@ChristianKienle do you think you can follow up on @greg-a-smith 's comments?

droshev:
@ChristianKienle

droshev:
@ChristianKienle we will close the PR for now. You can always open it again

PR #182, opened 2019-07-15
fix:Update toggle design

saad-mo:
Bring in updated toggle design from - https://github.com/SAP/fundamental/pull/1466

#### Before and After Screenshots
This PR is similar to https://github.com/SAP/fundamental/pull/1466
#### Test

* http://localhost:3030/toggle


greg-a-smith:
First thing is this PR (and subsequent merge commit message) should be prefixed with `fix:` as `feat:` is technically a breaking change for version 0.

Issue #166, opened 2019-07-05
Self-contained styles for tabs component


DorinR:
Similar to #60

PR #184, opened 2019-07-15
fix: added self-contained style for Tabs component

SaiShan:
Closes SAP/fundamental-styles#166

This PR contains the self-contained styling for the Tabs component.

### DIFF
--- a/components/tabs.scss
+++ b/components/tabs.scss
@@ -1,0 +1,1 @@
+@import "../scss/components/tabs";
--- a/components/tabs.scss
+++ b/components/tabs.scss
@@ -1,2 +1,4 @@
+$fd-fonts-path: "../scss/fonts/";
+
 @import "../scss/components/tabs";
 @import "../scss/fonts/72";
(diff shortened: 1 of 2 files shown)


SaiShan:
> There are some styles that bleed in and/or out of the tabs (for both self-contained and not). On the playground pages, check the "Un-Normalize" toggle.

It seems like the box-sizing inherits from the body. However, the two html's do not have the same padding and margin which creates the two fd-tabs element to not have the same box-sizing. However, the issue is not located in the component itself.

greg-a-smith:
@SaiShan It's actually not related to box sizing.  The `fd-tabs__link` class needs to reset padding, margin and border as they are allowing global `a` element styles in (from `unnormalize.css`) to bleed in.  This isn't a self-contained styles problem -- it happens in both the left and right frames.

components/tabs.scss
```diff
@@ -0,0 +1 @@
+@import "../scss/components/tabs";
```
greg-a-smith: I think this needs to load the font 72 -- I think that's the minor difference being seen in the widths of the tabs.

PR #188, opened 2019-07-15
fix: Alert updates

saad-mo:
Design tweaks and `--no-icon` modifier based on this PR - https://github.com/SAP/fundamental/pull/1493

#### Test

* http://localhost:3030/alert
* http://127.0.0.1:4000/components/alert.html

#### Changelog

**New**

* no icon modifier

**Changed**

* minor design tweaks, sizing of icon, close button 

**Removed**

nothing


docs/pages/components/alert.md
```diff
@@ -76,20 +84,72 @@ Add the modifier class `fd-alert--error` for Error alert.
...
 </div>
 {% endcapture %}
 
 {% include display-component.html component=default-alert %}
+
+<style>
+  .docs-component__no-icon .fd-alert--no-icon{
+    margin-bottom: 20px;
+  }
+</style>
```
greg-a-smith: Maybe remove this and just put a `<br />` in between each type (like all the other docs components do?

test/templates/alert/index.njk
```diff
@@ -71,6 +71,71 @@
 
 <br><br>
 
+<h1>alert --no-icon</h1>
+
+<!-- output the component example and the code snippet -->
+{% set example %}
+{{  alert(
+        properties={
+            text: '<p class="fd-alert__text">Integer turpis augue, tincidunt dignissim mauris id, rhoncus dapibus purus. Maecenas et enim odio. Nullam massa metus, varius quis vehicula sed, pharetra mollis erat. Sed bibendum, sapien at posuere interdum, libero est sollicitudin magna, ac gravida tellus purus eu ipsum. Proin ut quam arcu. <a href="#" class="fd-link">Learn More</a></p>',
```
greg-a-smith: The `<a href="#" class="fd-link">Learn More</a>` links should be removed from here (multiple places on this page) as the playground is serving as a style reference specific to alerts and the self-contained styles playground will not handle these as those styles come packaged with the link component.  The main docs can still use links as those are more use case-based examples.

docs/pages/components/alert.md
saad-mo: done

test/templates/alert/index.njk
saad-mo: done

Issue #187, opened 2019-07-15
Self-contained style for nav component


JYSW380:
Similar to #60

PR #189, opened 2019-07-15
fix: added self-contained style for nav component

JYSW380:
Closes SAP/fundamental-styles#187

This PR contains the self-contained styling for the nav component.


greg-a-smith:
@JYSW380 @saad-mo Just curious ... is this component still active?  I don't see a page for it on the main docs site -- just on the playground.

saad-mo:
> @JYSW380 @saad-mo Just curious ... is this component still active? I don't see a page for it on the main docs site -- just on the playground.

In the early stage of the library, we had the idea of having a generic nav component that can be used as a side nav(vertical) or sub-navigation(horizontal) in the header or the footer of the application. The shell specs have moved away from this (due to Fiori 3) so I would say that this component can be deprecated and even removed.

PR #190, opened 2019-07-16
fix: update XL identifier size

saad-mo:
Update identifier size from 88px to 80px 

similar to - https://github.com/SAP/fundamental/pull/1495

#### Test

* http://localhost:3030/identifier

#### Changelog

**New**

 nothing 

**Changed**

* XL identifier size from 88px to 80px

**Removed**

nothing


saad-mo:
> I think this will be fine, but let's let #157 get merged first and then merge the latest from `master` into this branch.

done.

Issue #191, opened 2019-07-16
Self-contained style for button component -- additional changes


JYSW380:
open this issue refer to 
fix: added self-contained style for tile component #159

PR #193, opened 2019-07-16
fix: fixed secondary button part

JYSW380:
Closes SAP/fundamental-styles#191

This PR just fixed the secondary button part.
This PR contains the self-contained styling for the button component. 


scss/components/button.scss
```diff
@@ -553,9 +553,12 @@ a.#{$block} {
     //DEPRECATE --secondary
     &--secondary {
       --fd-button-border-color: transparent;
+
       @include fd-hover() {
         --fd-button-background-color: var(--fd-color-neutral-1);
         --fd-button-color: var(--fd-color-action-1);
+        @include fd-var-color("background-color", map-get($fd-colors-neutral, 1),--fd-button-background-color);
+        @include fd-var-color("color", map-get($fd-colors-action, 1), --fd-button-color);
```
greg-a-smith: I am trying to figure out _what_ this is changing/fixing.  I am not seeing a difference between `master` and this branch.

Also, the secondary button is marked as deprecated so is this change even necessary?

JYSW380: w/o code at line 560 and 561, when hovered on the secondary button, shown as below screenshots:
self-contained style (left side of comparison playground): blue color
Screen Shot 2019-07-17 at 9 33 06 AM


left side of comparison playground: light grey color
Screen Shot 2019-07-17 at 9 33 16 AM

greg-a-smith: @saad-mo Wanted to get your take on this.  The playground page shows "Secondary Button" as being deprecated.  Is this true?  Also, is there a difference between "Secondary" and "Default Light"?  It kind of looks like "Default Light" is meant to replace "Secondary".

saad-mo: > @saad-mo Wanted to get your take on this. The playground page shows "Secondary Button" as being deprecated. Is this true? Also, is there a difference between "Secondary" and "Default Light"? It kind of looks like "Default Light" is meant to replace "Secondary".

`fd-button--secondary` is indeed deprecated. Fiori 3 guidelines called for a new naming convention which is `emphasized` (blue filled), `regular` (outline also same as secondary), and `light` (no borders). Along with semantic flavors of these three (standard, positive, medium, negative).

We left the deprecated version in the code to maintain backward compatibility and make non-breaking changes when introducing the new naming convention. I don't think we need to maintain this in fd-styles anymore since we are free to break things :)

PR #195, opened 2019-07-17
fix: table updates

saad-mo:
Closes sap/fundamental#1500

adds semantic highlighting support for table rows 

Screen Shot 2019-06-21 at 10 28 03 AM

#### Test

* http://127.0.0.1:4000/components/table.html
* http://localhost:3030/table

#### Changelog

**New**

* `fd-table__row--valid`, `fd-table__row--warning`, `fd-table__row--error` and `fd-table__row--information` modifiers for semantic row highlighting 

**Changed**

nothing

**Removed**

nothing

## Updated tests 

visual_regression_fundamental_table_6_example-containerfd-table__n6_0_desktop
visual_regression_fundamental_table_7_example-containerfd-table__n7_0_desktop
visual_regression_fundamental_table_12_example-containerfd-table__n12_0_desktop
visual_regression_fundamental_table_11_example-containerfd-table__n11_0_desktop
visual_regression_fundamental_table_3_example-containerfd-table__n3_0_desktop
visual_regression_fundamental_table_5_example-containerfd-table__n5_0_desktop
visual_regression_fundamental_table_4_example-containerfd-table__n4_0_desktop

New screenshots for updated tests (Jenna):
diff is due to adding new padding for table header cells - I'm still not clear why the font and padding is less for the header
Screen Shot 2019-09-09 at 8 38 53 AM
Screen Shot 2019-09-09 at 8 38 58 AM


droshev:
@saad-mo can you bring the description from fundamental-ui?

saad-mo:
@jbadan This is ready for review again.

scss/components/table.scss
```diff
@@ -240,4 +254,78 @@ $block: #{$fd-namespace}-table;
             }
         }
     }
+
+    &__row--valid,
+    &__row--warning,
+    &__row--error,
+    &__row--information {
+        & td:first-child {
```
jbadan: We cannot use element selectors.

docs/pages/components/table.md
```diff
@@ -19,11 +19,11 @@ A table is a set of tabular data. Line items can support data, images and action
 <table class="fd-table">
     <thead class="fd-table__header">
         <tr class="fd-table__row">
-            <th class="fd-table__cell" scope="col"></th>
-            <th class="fd-table__cell" scope="col">Column Header</th>
-            <th class="fd-table__cell" scope="col">Column Header</th>
-            <th class="fd-table__cell" scope="col">Column Header</th>
-            <th class="fd-table__cell" scope="col">Column Header</th>
+            <th class="fd-table__cell fd-table__cell--header" scope="col"></th>
```
jbadan: Why do we need this class?

jbadan:
Just those 2 issues, otherwise this is looking great 💯

scss/components/table.scss
saad-mo: fixed

docs/pages/components/table.md
saad-mo: `th` has different padding than `td` and since we are using the class `fd-table_cell` for td, i created a modifier `--header` that applies this padding for `th`

hertweckhr1: @saad-mo instead of creating an extra class `fd-table_cell--header`, can you just add the `padding-top` and `padding-bottom` in the location of under cell styles in the header section? (between lines 78-88 in table.scss).

jbadan:
@saad-mo I made a couple changes to keep this more in line with the class names already present.

scss/components/table.scss
```diff
@@ -79,6 +142,8 @@ $block: #{$fd-namespace}-table;
             border: none;
             @include fd-type("-1");
             @include fd-weight();
+            padding-top: $fd-table-header-cell-padding;
+            padding-bottom: $fd-table-header-cell-padding;
```
jbadan: This is the part I'm not sure about - did we want to reduce the padding here?

saad-mo: Yes, this is based on Fiori 3 designs. Below is the screenshot from their master sketch file. 

Screen Shot 2019-09-11 at 2 12 16 PM

jbadan: Sounds good!

test/templates/table/index.njk
```diff
@@ -68,7 +68,7 @@
             <th class="fd-table__cell" scope="col">Column Header</th>
             <th class="fd-table__cell" scope="col">Column Header</th>
             <th class="fd-table__cell" scope="col">Column Header</th>
-            <th class="fd-table__cell" scope="col">Column Header</th>
+            <th class="fd-table__cell " scope="col">Column Header</th>
```
bcullman: unneeded whitespace

scss/components/table.scss
```diff
@@ -59,6 +60,68 @@ $block: #{$fd-namespace}-table;
               @include fd-var-color("background-color", fd-color-state("selected-hover"), --fd-color-background-selected-hover);
             }
         }
+
+        &--valid,
+        &--warning,
+        &--error,
+        &--information {
+            border: none;
+            & .#{$block}__cell:first-child {
```
bcullman: interesting technique. does it work in IE?

jbadan: @bcullman sure does!

Screen Shot 2019-09-17 at 10 09 21 AM

jbadan:
@markpalfreeman no stress! Linting was important to knock out 💯

PR #197, opened 2019-07-18
fix: add asset upload component

saad-mo:
bring in new asset upload component from - https://github.com/SAP/fundamental/pull/1509

#### Test

* http://localhost:3030/asset-upload
* http://127.0.0.1:4000/components/asset-upload.html

#### Changelog
detailed changlog - https://github.com/SAP/fundamental/pull/1509


docs/_data/sidebars/left-navigation-sidebar.yml
```diff
@@ -72,6 +72,10 @@ entries:
       url: /components/alert.html
       output: web
 
+    - title: Asset Upload
+      url: /components/asset-upload.html
+      output: web
+
```
greg-a-smith: When on the `http://localhost:4000/components/index.html` page, I am not seeing **Asset Upload** in the left nav, however, if I navigate to a component page (_e.g._ **Alert**), then I do see it.

docs/pages/components/asset-upload.md
```diff
@@ -0,0 +1,52 @@
...
+toc: false
+permalink: components/asset-upload.html
+folder: components
+summary:
+---
+
+Asset Upload provides a stylized component for file upload button.
+{: .docs-intro}
+
+## Default Asset Upload
```
greg-a-smith: It seems odd to call this the "default" when the version with no message simply omits the `fd-asset-upload__message` element.  It seems like the default should have required information (_e.g. "text") so that it would have an icon and text, but no message.  Then there should be an example with additional message and then the last option with no icon.  I'm just thinking of how this would be implemented in a functional library.  There would a "no icon" option (which deviates from the default) and then it would seem strange to have a "no message" option when all you would really need to do is just not pass the message text.

docs/_data/sidebars/left-navigation-sidebar.yml
saad-mo: I don't see this problem. maybe the docs site was not generated with the latest? 

Screen Shot 2019-07-22 at 12 07 36 PM

docs/pages/components/asset-upload.md
saad-mo: I changed the default test to "Asset upload with Icon and Message" there are no clear guidelines on what the default should be from the design side. 

Screen Shot 2019-07-22 at 12 12 31 PM

Issue #199, opened 2019-07-22
remove UI Kit from the website


droshev:
As we are considering harmonizing our components with other SAP components' libraries our UI Kit might become outdated for time being. Let's not publish it until we figure the next step out.
Revisit the content of https://sap.github.io/fundamental-styles/resources/ui-kit.html and https://sap.github.io/fundamental-styles/
https://sap.github.io/fundamental-styles/resources/index.html

Issue #201, opened 2019-07-22
remove the color page


droshev:
As the library will be migrating over the theming parameter we have to remove the fundamental-styles color system so we need to remove 
https://sap.github.io/fundamental-styles/foundation/colors.html

PR #210, opened 2019-07-25
docs: Remove Colors and UI Kit pages

saad-mo:
Closes SAP/fundamental-styles#199, #201

Removes UI Kit and Colors pages 

#### Test

* http://127.0.0.1:4000/index.html

#### Changelog

**New**

nothing

**Changed**

nothing 

**Removed**

* UI Kit page
* Colors page


droshev:
@greg-a-smith I was wondering whether we should even remove the build and files like `colors.yml`. the next step is to completely remove that feature from the doc before we discontinue the fundamentals' colors.
What do you think?

greg-a-smith:
@droshev Not quite sure.  I'm not quite sure where all the color touch points are within the build.  I was just checking the visible docs pages.

Issue #134, opened 2019-06-27
Self-contained styles for Counter component


SaiShan:
Similar to #60

PR #219, opened 2019-08-01
fix: Added self-contained style for Counter component

SaiShan:
Closes SAP/fundamental-styles#134

This PR contains the self-contained styling for the Counter component.

Screen Shot 2019-08-14 at 8 23 37 AM


scss/components/counter.scss
```diff
@@ -11,7 +11,7 @@ $block: #{$fd-namespace}-counter;
   $fd-counter-radius: $fd-border-radius !default;
   $fd-counter-background: fd-color("status", 4) !default;
   $fd-counter-color: fd-color("text", 5) !default;
-  $fd-counter-font-size: 0.785714285714286em !default;
+  $fd-counter-font-size: 0.785714285714286rem !default;
```
greg-a-smith: We need to get rid of the `rem` value as it's possible to make the self-contained and the "all" styles appear in different sizes.  I know we have `rem` values coming from theme designer so that is going to need to be addressed more globally going forward.

test/templates/counter/index.njk
```diff
@@ -1,70 +1,35 @@
 {% extends "layout.njk" %}
-{% from "./component.njk" import counter %}
```
greg-a-smith: If we're no longer importing the counter component, then we should remove that file.  Or, use the component file and redo the examples to call that.

droshev: @greg-a-smith that's what I did, didn't I?

scss/components/counter.scss
droshev: @greg-a-smith I am not sure how this change had happened. Reverted.

components/counter.scss
```diff
@@ -0,0 +1,6 @@
+$fd-icons-path: "../scss/icons/";
+$fd-fonts-path: "../scss/fonts/";
+
+@import "../scss/components/counter";
+@import "../scss/icons/icon";
```
greg-a-smith: This shouldn't need icons.

scss/components/counter.scss
```diff
@@ -20,13 +20,16 @@ $block: #{$fd-namespace}-counter;
   @include fd-var-color("color", $fd-counter-color, --fd-counter-color);
   @include fd-var-color("background-color", $fd-counter-background, --fd-counter-background-color);
   font-size: $fd-counter-font-size;
+  font-family: $fd-font-family;
+  -webkit-font-smoothing: antialiased;
```
greg-a-smith: These should be pulled in from the `fd-reset` mixin instead.

scss/components/counter.scss
```diff
@@ -20,13 +20,16 @@ $block: #{$fd-namespace}-counter;
...
   line-height: 1;
   display: inline-block;
   text-align: center;
   padding: fd-space(1);
   &--notification {
+    font-size: $fd-counter-font-size;
```
greg-a-smith: Why is there a font size declaration in here?  This is just a modifier class used in conjunction with the base class.

test/templates/counter/index.njk
greg-a-smith: You are not importing it, but the file still exists.  To match all the other playground examples, we should actually keep the file, import it and pass the necessary parameters to it.

test/templates/counter/index.njk
```diff
@@ -1,70 +1,35 @@
...
+    {% set example %}
+    <p>Read<span class="fd-counter" arial-label="Unread count">1</span></p>
+    <br/>
+    <button aria-label="Notifications">Open<span class="fd-counter fd-counter--notification" arial-label="Unread count">24</span></button>
+    <br/><br/>
+    <button aria-label="Notifications">Check<span class="fd-counter" arial-label="Unread count">300</span></button>
+    <br/><br/>
+    <a aria-label="Notifications">close<span class="fd-counter" arial-label="Unread count">6000</span></a>
+    <br/><br/>
+    <span class="sap-icon--bell"><span class="fd-counter fd-counter--notification" arial-label="Unread count">24</span></span>
```
greg-a-smith: Some of the notification examples are missing the `fd-counter--notification` class.

droshev: @greg-a-smith that is true as we have 2 designs for the counters:
 - default
 - notification

test/templates/counter/index.njk
```diff
@@ -1,70 +1,41 @@
...
-{{ counter(count=0, modifier="") }}
-{{ counter(count=5, modifier="") }}
-{{ counter(count=25, modifier="") }}
-{{ counter(count=101, modifier="") }}
-{{ counter(count=4001, modifier="") }}
+    <span class="fd-counter" aria-label="Unread count">5</span>
+    <span class="fd-counter" aria-label="Unread count">25</span>
+    <span class="fd-counter" aria-label="Unread count">101</span>
+    <span class="fd-counter" aria-label="Unread count">4001</span>
+    <span class="fd-counter" aria-label="Unread count">9999+</span>
```
greg-a-smith: I think these should all be changed back to use the counter imported from the `component.njk` file.  I know they are very simple, but this keeps it consistent with all the other playground page constructs.

scss/components/counter.scss
```diff
@@ -16,6 +16,7 @@ $block: #{$fd-namespace}-counter;
   --fd-counter-color: var(--fd-color-text-5);
   --fd-counter-background-color: var(--fd-color-status-4);
 
+  @include fd-reset();
```
greg-a-smith: The line above `$fd-counter-font-size: 0.785714285714286em !default;` should be changed to use `px` otherwise you can get different size text between self-contained and "all" styels.  There will be a larger discussion/story about the use of `px` vs `rem` or `em` in the future, but right now, the self-contained text is larger than the "all" styles.

scss/components/counter.scss
```diff
@@ -25,8 +26,9 @@ $block: #{$fd-namespace}-counter;
   text-align: center;
   padding: fd-space(1);
   &--notification {
+    font-size: $fd-counter-font-size;
```
greg-a-smith: Why is there a font size declaration in here? This is just a modifier class used in conjunction with the base class.

scss/components/tabs.scss
```diff
@@ -39,7 +39,7 @@ $block: #{$fd-namespace}-tabs;
   @include fd-var-color("border-bottom-color", $fd-tabs-background-color, --fd-tabs-background-color);
   &__item {
     @include fd-reset();
-      padding: 0 $fd-tabs-link-padding-x;
+    padding: 0 $fd-tabs-link-padding-x;
```
greg-a-smith: Although a "good" fix, it's unrelated to this PR and should be reverted.

scss/components/counter.scss
droshev: I think the counter font-size was initial in `em` as dependant to the parent element. Let's just keep it as it was and then we can tackle it later on

test/templates/counter/index.njk
droshev: reverted it

Issue #153, opened 2019-07-04
Update input and input-group styling


SaiShan:
Update input and input-group styling to match newest design sheet. The design sheet can be found here https://app.zeplin.io/project/5a8c458b6e231fc65bddbdfb/screen/5c0054fc8b9cac0193d42be9

PR #221, opened 2019-08-01
fix: Updated input-group style

SaiShan:
Closes SAP/fundamental-styles#153
Closes SAP/fundamental-styles#69
Closes SAP/fundamental-styles#65

This PR refactors the styling of the input group component. Previously, the input had a thick border when on-focus. Now, it has a thin border. See screenshot. 

Before
Screen Shot 2019-08-01 at 1 54 25 PM

After 
Screen Shot 2019-08-01 at 1 53 03 PM

This PR also fixes the input-group button overflow seen in many component. See screenshot. 

Before
Screen Shot 2019-08-01 at 1 58 18 PM

After
Screen Shot 2019-08-01 at 1 57 54 PM

Lastly, it also fixes is-disabled class on span element in components using input group. See screenshot for Time-picker example.

before

after


greg-a-smith:
I fear some of these changes won't play well with self-contained styles.  Closing this PR in lieu of the self-contained work for input group which will aim to address some of these fixes.

PR #228, opened 2019-08-05
UI5 Web Components shared styles POC

jbadan:
This pr edits our build pipeline to combine styles from ui5-webcomponents and our self-contained styles to create a new, processed css file per component. 
> NOTE: This removes the old pipeline and should not be merged as is. 

## To Do List
🚫: blocked by webcomponents
💅 : blocked until all component self contained styles in fundamental-styles are complete. 

[ ]  postcss-generateClassNames plugin: pass filename to plugin in order to replace `fd-button` with `fd-component`
https://jira.concur.com/browse/CORUI-6961

[ ] 💅 rework playground to serve styles from `components` directory only, removing the `scss` directory reliance 
https://jira.concur.com/browse/CORUI-6953

[ ] 💅 rework testing suite to only use `components` directory (self-contained styles)
[ ] update readme and other (documentation)[https://github.com/SAP/fundamental-styles/wiki]
https://jira.concur.com/browse/CORUI-6953

 [ ] 💅 edit build pipeline to bundle all self contained styles together - this will be needed for the [docs site](https://github.com/SAP/fundamental-styles/blob/master/docs/css/docs-all.scss#L5) https://jira.concur.com/browse/CORUI-6953

[ ] rewrite scss to remove this pattern:  .#{$block}, [class*="#{$block}--"], a[class*="#{$block}--"], the [class*="#{$block}--"] selectors are getting picked up by the attribute selector block. We can either edit the plugin to account for this or fix this pattern in general. 
https://jira.concur.com/browse/CORUI-6962

[ ] rewrite scss to use ui5-webcomponents variables/theme designer variables - this is for the `postcssCustomProperties` and `postcssAddFallback` plugins. `parameters-bundle` _is_ currently published. We would just like this to live in styles repo. This is not blocked. We can also pass an array with our own variables.

[ ] 🚫 replace styles with those from ui5-webcomponents

## ui5-webcomponents asks
[ ] publish custom css - `main/dist/css` folder. - currently publishing only css.js 
[ ] publish [`postcssAddFallback` plugin](https://github.com/SAP/ui5-webcomponents/blob/master/packages/main/lib/postcss-add-fallback/index.js), convert to take array
[x] publish [`sap_fiori_3/parameters-bundle.css`](https://github.com/SAP/ui5-webcomponents/blob/master/packages/main/src/themes/sap_fiori_3/parameters-bundle.css)


config/postcss-generate-classes/index.js
```diff
@@ -0,0 +1,54 @@
+const postcss = require('postcss');
+const parser = require('postcss-selector-parser');
+
+//TO DO: 
+//* figure out how to pass file name- to replace fd-button with component file name
+//* handle .#{$block}, [class*="#{$block}--"], a[class*="#{$block}--"], the [class*="#{$block}--"] selectors are getting picked up by the attribute selector block
+
```
jbadan: Additional to do: handle `[disabled]` attribute selector from fundamental-styles scss. Do not convert this to class.

PR #238, opened 2019-08-07
chore: add mergeable.yml

droshev:
add initial version of `mergeable.yml` file

### DIFF
--- a/.github/mergeable.yml
+++ b/.github/mergeable.yml
@@ -1,0 +1,32 @@
+
+version: 2
+mergeable:
+  - when: pull_request.*, pull_request_review.*
+    validate:
+      - do: title
+        must_exclude:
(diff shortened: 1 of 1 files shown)


.github/mergeable.yml
```diff
@@ -0,0 +1,32 @@
...
+        must_include:
+          regex: ^(feat|docs|chore|fix|test)(\(\w+\))?:.+$
+          message: Semantic release conventions must be followed.
+      - do: description
+        must_exclude:
+          regex: \[ \]
+          message: There are incomplete TODO task(s) unchecked.
+      - do: approvals
+        min:
+          count: 2
```
jbadan: Can we change this to 1? The `or` doesn't seem to be picking up Inna's approval.

PR #239, opened 2019-08-08
fix: add dependency to sap theme designer package and remove the local files

droshev:
remove the local copies of sap theme designer files and add a dependency to their package


jbadan:
Are we going to import these into a variables file so we can use it with our scss?

droshev:
@jbadan we can start importing them in the self-containted files and in the global one

PR #244, opened 2019-08-09
feat: alert refactoring

InnaAtanasova:
- adds theme designer variables for colors 
- updates paddings and spacing according to Fiori3 guidelines
- removes extra code


scss/components/alert.scss
```diff
@@ -1,45 +1,66 @@
...
 @import "./../mixins";
-@import "./../functions";
 @import "./../icons/mixins";
-/*!
-.fd-alert+(--information, --success, --warning, --error, --no-icon)
-    .fd-alert__close
-*/
+
+@import "./../../theme-designer/Base/baseLib/sap_fiori_3/sass_variables.scss";
+@import "./../../theme-designer/Base/baseLib/sap_fiori_3/css_variables";
```
jbadan: Should wait for https://github.com/SAP/fundamental-styles/pull/239 to be merge and then import from `@sap-theming/theming-base-content`?

droshev: this is done :)

components/alert.scss
```diff
@@ -1,6 +1,9 @@
 $fd-icons-path: "../scss/icons/";
 $fd-fonts-path: "../scss/fonts/";
 
+@import "./../node_modules/@sap-theming/theming-base-content/content/Base/baseLib/sap_fiori_3/sass_variables.scss";
+@import "./../node_modules/@sap-theming/theming-base-content/content/Base/baseLib/sap_fiori_3/css_variables";
+
```
jbadan: Do we need the `scss` reference here if it's already defined in settings that are being imported into alert.scss?

InnaAtanasova: We don't need it, I removed it. Thanks for noticing it

InnaAtanasova: We don't need it, I removed it. Thanks for noticing it

jbadan:
@InnaAtanasova can you merge with master to bring in those latest changes for easier review (I know, easier said than done 😆 ). Thank you!

PR #246, opened 2019-08-11
fix: added self-contained styles for action bar

### LINKED ISSUE #211: Self-contained style for action-bar
Refactoring to make the action-bar component support self-contained style.

droshev:
fixes #211


droshev:
Maybe we can create separate test files when we have combined components - e.g. action bar with popover as we may want to have those combinations in test handy for future testing
@jbadan @greg-a-smith @hertweckhr1 @InnaAtanasova what do you think?

test/templates/action-bar/index.njk
```diff
@@ -12,87 +10,32 @@
     .fd-action-bar {
         margin: 0 20px;
     }
```
jbadan: Can we use the placeholder buttons already in layout-shared: https://github.com/SAP/fundamental-styles/blob/master/test/resources/layout-shared.css#L6

droshev: done!

components/action-bar.scss
```diff
@@ -0,0 +1,6 @@
+$fd-icons-path: "../scss/icons/";
+$fd-fonts-path: "../scss/fonts/";
+
+@import "../scss/components/action-bar";
+@import "../scss/icons/icon";
```
greg-a-smith: This component shouldn't need icons.

test/resources/layout-shared.css
```diff
@@ -10,3 +10,28 @@ body {
...
+    padding: 0;
+    border: none;
+    border-radius: 4px;
+    background-color: white;
+    min-height: 28px;
+    width: 26px;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+}
```
greg-a-smith: Can we replace both `.test-btn-placeholder` and `.test-btn-back` with `.btn-playground`?

test/templates/action-bar/component.njk
```diff
@@ -1,32 +1,27 @@
...
-    <div class="fd-action-bar__back">
-        <button class="fd-button--light sap-icon--nav-back"></button>
-    </div>
-    {%- endif %}
+{% set headlines = ['h2', 'h3', 'h4', 'h5', 'h6'] %}
+
+{%- macro action_bar(properties={}, aria={}, title_size="h3") %}
+<div class="fd-action-bar">
+{%- if properties.back_button %}
+    <div class="fd-action-bar__back"><div class="test-btn-back sap-icon--nav-back"></div></div>
```
greg-a-smith: Can we replace `test-btn-back` with `btn-playground`?  Also, we should not be using the class `sap-icon--nav-back` in the playground examples.

test/templates/action-bar/index.njk
```diff
@@ -1,98 +1,16 @@
...
-                { "label": "Option 2" },
-                { "label": "Option 3" },
-                { "label": "Option 4" }
-                ]
-            })
-        }, modifier={ block: ['left']})
-    }}
-{%- endset %}
+{%- macro render_txt_button() %}<div class="test-btn-placeholder">Button</div>{%- endmacro %}
+{%- macro render_icon_button() %}<div class="test-btn-placeholder sap-icon--overflow"></div>{%- endmacro %}
```
greg-a-smith: Can we replace `test-btn-placeholder` with `btn-playground`?  Also, we shouldn't be using the class `sap-icon--overflow` in the playground.

greg-a-smith:
I pushed changes for the last two items noted.

PR #250, opened 2019-08-12
fix: added self contained styles to panel component

### LINKED ISSUE #222: Self-contained style for panel component
Refactoring to make the panel component support self-contained style.
Similar to #60

hertweckhr1:
Closes #222

This PR contains the self-contained styling for the panel component.

All tests passing with origin master:
image


jbadan:
Should we just delete https://github.com/SAP/fundamental-styles/blob/master/scss/components/panel.scss while we're doing this?

jbadan:
@hertweckhr1 Can you include visual regression test update screenshots please :)

scss/layout/_panel.scss
```diff
@@ -75,28 +80,33 @@ $block: #{$fd-namespace}-panel;
...
     &__body {
+        @include fd-reset();
         padding: $fd-panel-padding;
         &--full-bleed,
         &--bleed {
           padding: 0;
         }
+        p {
+            @include fd-reset-block();
+        }
```
greg-a-smith: The more I look at this one, the more I don't like it.  A panel, like other container components, should not really have any rules about _what_ content can appear in its body.  Because of that, we shouldn't be handling any child element styling here.  This example if for a `<p>`, but what I the consumer put a `<ul>` (or any other element) in the body?

scss/mixins/_mixins.scss
```diff
@@ -19,6 +19,15 @@
...
 }
+
+@mixin fd-reset-block {
+  margin-top: 0;
+  margin-bottom: $fd-margin-bottom;
+  &:last-child {
+      margin-bottom: 0;
+  }
+}
+
```
greg-a-smith: This block probably also goes away because of the noted comment above.

scss/layout/_panel.scss
greg-a-smith: Actually, how about we remove this `p` block (and the `fd-reset-block` mixin) and replace it with this (like it's being done for modals):
```scss
    > :first-child {
      margin-top: 0;
    }

    > :last-child {
      margin-bottom: 0;
    }
```
This should accomplish the same thing, but not lock us down to individual elements.

jbadan:
The font sizes aren't right with description:

jbadan:
@hertweckhr1 Can you please add screenshots when you update the tests :)

hertweckhr1:
image

^^Update on tests

PR #251, opened 2019-08-12
fix: self contained styles for calendar component

jbadan:
https://github.com/SAP/fundamental-styles/issues/136

Self contained styles for calendar component.

**This may look weird due to 3 factors:**
1) removed hard coded white background from testing suite 
2) docs site is also hard coding background: https://github.com/SAP/fundamental-styles/blob/master/docs/_sass/_docs-display-component.scss#L362
> replaced both of these with white background inside `__content`

3) These styles need to be updated in general to match newest design: https://wiki.wdf.sap.corp/wiki/display/visualcore/Calendar+%28Fiori+3%29+-+Date+Picker

Tests were updated due to using placeholder buttons and removing hard-coded white background for testing suite. I also removed the background color for the chart showing different item states so we can actually see what's happening. Examples now match docs site, see before and after: 

Screen Shot 2019-08-15 at 1 16 04 PM

Screen Shot 2019-08-15 at 1 16 13 PM


components/calendar.scss
```diff
@@ -0,0 +1,7 @@
+$fd-icons-path: "../scss/icons/";
+$fd-fonts-path: "../scss/fonts/";
+
+@import "../scss/components/calendar";
+@import "../scss/icons/icon";
+@import "../scss/fonts/72";
```
jbadan: Icons are included because of this line: https://github.com/SAP/fundamental-styles/blob/master/scss/components/calendar.scss#L155

Although, it might be fine without them. Thoughts @greg-a-smith ?

greg-a-smith: @jbadan Definitely some interesting CSS. 😄 Although that method of flipping the arrow icons is used (creative), if you check the rendered markup, there is actually a button component (class `fd-button`) that the calendar component will use internally.  Since button already comes with icons, they should be left out of calendar.

test/templates/calendar/index.njk
```diff
@@ -1,14 +1,8 @@
...
 {% from "./../format.njk" import format %}
-{% from "../button/component.njk" import button %}
 {% from "./component.njk" import calendar %}
 
 {% block content %}
-<style>
-main {background-color: white;}
-
-</style>
-
```
greg-a-smith: There is similar styling on the docs pages:
```css
.docs-component .fd-calendar__content {
    background: white;
}
```
It _seems_ like this should just be part of the calendar styling.  Without it, there is some bleed through from a non-white background because of the rounded corners.

jbadan: So I thought this at first too, but that isn't what matches the spec on the wiki page. This component needs to be updated in general, but you should be able to see the rounded corners. I left the hardcoded value in the docs site to prevent confusion until we update the styles: https://wiki.wdf.sap.corp/wiki/display/visualcore/Calendar+%28Fiori+3%29+-+Date+Picker

jbadan: Now that I look at the specs again, I don't know anymore. I think it is a white background, but the numbers shouldn't be white?

jbadan: After discussing offline with @greg-a-smith, we are going to add the white background to the scss file for now. This component will need to be updated with the newest design spec in the future.

scss/components/calendar.scss
```diff
@@ -51,36 +50,30 @@ $fd-calendar-table-width: $fd-calendar-table-item-width * 7 !default;
...
-  @include fd-var-color("color", $fd-calendar-item-color--current);
+  @include fd-var-color("color", $fd-calendar-item-color--current, --fd-color-action-selected);
   &::after {
     content: "";
     position: absolute;
     top: 0;
     right: 0;
     bottom: 0;
     left: 0;
-    height: $fd-calendar-table-item-width;
```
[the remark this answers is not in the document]
jbadan: It was not

test/templates/calendar/index.njk
```diff
@@ -1,21 +1,14 @@
...
 
 <p>The <code>.fd-calendar__item</code> container takes the majority of styling. The rest of the markup provides the structure and layout.
 </p>
 <p>Classes available for <code>.fd-calendar__item</code>. All are available to dates, months, and years.</p>
 <style media="screen">
-    .calander-info{ background: #ccc;}
-    .calander-info td{ background: #fff;}
+    .calander-info{ background: #f3f4f5;}
 </style>
```
jbadan: I added this so that the `.fd-calendar` white background didn't get added here. This way we can see the calendar items.

scss/components/calendar.scss
```diff
@@ -152,6 +142,9 @@ $fd-calendar-table-width: $fd-calendar-table-item-width * 7 !default;
     align-items: center;
   }
   &__action {
+    & > * {
+      @include fd-reset-child-spacing();
+    }
```
greg-a-smith: I added this mixin to use in action-type containers that will keep its children from blowing out with crazy margins.  I want to reuse this in `action-bar` as well.

jbadan:
@greg-a-smith checked out your changes - 👍

PR #254, opened 2019-08-13
fix: self contained styles for panel grid

hertweckhr1:
Self contained styles for panel-grid.  

Part of larger story #136 to self contain all components. 

image

Screen Shot 2019-08-14 at 7 46 28 AM


test/templates/panel-grid/index.njk
```diff
@@ -1,23 +1,17 @@
...
 {% from "./component.njk" import panel_grid %}
-{% from "./../panel/component.njk" import panel, panel_header, panel_actions, panel_filters, panel_body, panel_footer %}
 
 {% set panel_item %}
-{% call panel() -%}
-    {% call panel_body() -%}
-      <p>.fd-panel</p>
-    {%- endcall %}
-{%- endcall %}
+    <div class="panel-playground">.fd-panel</div>
```
jbadan: Can we change `.fd-panel` here to `panel-playground` or `placeholder` so that we know this is a placeholder element?

jbadan:
After fixing that small nit I'm super happy with is :)

scss/layout/_panel-grid.scss
```diff
@@ -7,7 +7,7 @@
 $block: #{$fd-namespace}-panel-grid;
 .#{$block} {
   //VARS
-
+  @include fd-reset();
```
markpalfreeman: nit: this is slightly awkward underneath `// VARS` :)

it might be better above the properties on line 14?

hertweckhr1: 😅 ha! yes

test/resources/layout-shared.css
```diff
@@ -10,3 +10,12 @@ body {
     background: beige;
     font-size: 14px;
 }
+
+.panel-playground, .panel-body-playground {
```
greg-a-smith: I think you can remove `.panel-body-playground`.

test/resources/layout-shared.css
```diff
@@ -10,3 +10,12 @@ body {
...
+    border: 1px solid navy;
+    border-radius: 4px;
+    padding: 10px;
+    background: gold;
+    font-size: 14px;
+    box-shadow: 0 5px 20px 0;
```
greg-a-smith: Minor nit: box shadows should be defined with a color.  Without it, the playground box shadows are actually appearing themeable.

PR #256, opened 2019-08-14
fix: add self-contained styles for Modal component

markpalfreeman:
Self-contained styles for Modal.

### Before and After Screenshots
**Modal-only styles (Left) / All Fundamental styles (Right)**
Screen Shot 2019-08-14 at 9 18 54 AM

**Visual tests**
Screen Shot 2019-08-14 at 9 13 36 AM

### Test

* Updated visual tests for playground Modal (and some stale ones from Menu)

### Changelog

**Changed**
* Moved heading base styles to a mixin to share elsewhere (e.g. `.fd-modal__title`)

**Removed**
* Custom buttons within Modal playground examples (will come from Button directly)


scss/components/modal.scss
```diff
@@ -30,25 +30,31 @@ $block: #{$fd-namespace}-modal;
...
+
   &__header {
     position: relative;
     border-bottom-style: solid;
     border-bottom-width: 1px;
     @include fd-var-color("border-bottom-color", $fd-modal-border-color, --fd-modal-border-color);
   }
+
   &__title {
+    @include fd-heading();
```
markpalfreeman: To replicate underlying `h*` classes from global fundamental CSS.

scss/components/modal.scss
```diff
@@ -70,13 +77,20 @@ $block: #{$fd-namespace}-modal;
     height: fd-space(9);
     @include fd-focus();
   }
+
   &__body {
     max-height: calc(100vh - #{$fd-modal-header-height} - #{$fd-modal-footer-height});
     overflow-y: auto;
-    > p {
-      padding: 0;
```
markpalfreeman: Most default paragraphs don't have default padding—this wasn't doing anything on our playground examples.

scss/components/modal.scss
```diff
@@ -70,13 +77,20 @@ $block: #{$fd-namespace}-modal;
...
+
+    > :first-child {
```
markpalfreeman: Added these just for top and bottom margin; everything else should be styled at the (incoming content) child level.

test/templates/modal/component.njk
```diff
@@ -1,45 +1,26 @@
...
-                  },
-                  classes=['modal__close'])
-          }}
-      </header>
-      <div class="fd-modal__body">
+    <div class="fd-modal__content" role="document">
+        <header class="fd-modal__header">
+            {% set title_size = properties.header_size if properties.header_size else "h3" %}
+            <{{title_size}} class="fd-modal__title">{{ properties.header }}</{{title_size}}>
+            <button class="fd-modal__close"></button>
```
markpalfreeman: Primary changes are using basic `<button>`s (this one has a class for the icon).

test/templates/modal/index.njk
```diff
@@ -1,17 +1,14 @@
...
 {% from "./component.njk" import modal %}
-{% from "../icon/component.njk" import icon %}
 
 {% block content %}
 
-
 <p>
-NOTE: The <code>modal</code> itself does not provide the outer container. Use the <code>.fd-overlay</code> layout component in conjunction with the <code>.fd-ui__overlay</code> template container to hide and show the modal. See <a href="http://localhost:3030/pages/overlay">the overlay example</a>.
-
+  NOTE: The <code>modal</code> itself does not provide the outer container. Use the <code>.fd-overlay</code> layout component in conjunction with the <code>.fd-ui__overlay</code> template container to hide and show the modal. See <a href="http://localhost:3030/pages/overlay">the overlay example</a>.
```
greg-a-smith: What the heck does this sentence even mean?  Seems like we could get rid of that in the playground.

scss/components/modal.scss
jbadan: Do we need something like: 
```
> * {
      @include fd-reset();
    }
``` 

here as well so the un-normalize sheet can't affect the interior of the `fd-modal__body`? Or should we allow that?

greg-a-smith: That _seems_ like a bit much (at least at this time) until we know more about how self-contained styles will affect overall page and application styling.

test/templates/modal/component.njk
markpalfreeman: There's a Slack thread with Greg debating whether `.fd-modal__close` (the modal CSS) should bring in the icons for the close "X" icon, or if we can remove that and leave it with the button CSS, trusting the component libraries to apply the right icon class to the button.

I could go either way. @jbadan do you have a leaning one way or the other?

jbadan: No real preference, I'd rather keep these as simple as possible so I'm more leaning towards removing it and trusting the implementation libraries to know what to do.

greg-a-smith: I think there is enough chatter about this and no real detractors so let's go ahead and make this change.  We are version 0, so let's do this. 😄  This will allow to remove icons from modal self-contained styles.  We will keep the `fd-modal__close` class, but it will just handle the positioning of its child (button).  The implementation libraries will then be using their respective button component within the modal and will supply the proper icon to it.

components/modal.scss
```diff
@@ -0,0 +1,6 @@
+$fd-icons-path: "../scss/icons/";
+$fd-fonts-path: "../scss/fonts/";
+
+@import "../scss/icons/icon";
```
greg-a-smith: Per discussion, let's get rid of icons in the modal self-contained styles.

test/templates/modal/component.njk
jbadan: Excellent - this also just came up in the `tree` component I'm working on. I'll follow this pattern ✅

PR #257, opened 2019-08-14
fix: Self contained combobox-input and multi-input

jbadan:
#136 
BREAKING CHANGE

>NOTE: These components were only combined because it was the same fix for both.

Screen Shot 2019-08-14 at 9 24 18 AM

Multi-input and combobox input had styling specific to fundamental-styles for popover. This was only required for the docs site as the implementation libraries will need to handle child classes in popper differently due to it's location in the dom. 

As the combobox-input component had nothing else, I removed it from the `scss` directory. I've left it on the docs site so that we have examples to go by, just no specific css in necessary.

Changes:
1. added styles needed for docs site display to docs site scss
1. removed scss/components/combobox-input + testing page 

Test suite changes:
1. removed combobox-input tests
1. edited multi-input to use placeholders as the only scss left in that file is for adding a margin between items --> therefore large diff. 
1. removed multi-input compact test - this was only relevant in the context of buttons


scss/components/popover.scss
```diff
@@ -170,6 +170,10 @@ $block: #{$fd-namespace}-popover;
         box-shadow: 0 5px 20px 0 var(--fd-color-neutral-3), 0 2px 8px 0 var(--fd-color-neutral-2);
         opacity: 1;
 
+        &--full-width {
+          width: 100%;
+      }
+
```
greg-a-smith: I don't think there is a need for this here.  Because popper.js-based popovers are placed in a different place in the DOM (child of body), setting the width to 100% will just span the entire body rather than matching its size to its reference element.  Sizing of the popover in relation to the reference element is going to be the responsibility of the implementation library.

test/templates/popover/index.njk
```diff
@@ -140,5 +148,22 @@ is simply a styling example.</p>
...
+{{  popover_popperjs(properties={
+    id: _id7,
+    control: sample_control_input(properties={id: _id7}),
+    body: popover_content()
+  }, modifier={ block: ['no-arrow', 'full-width']})
+}}
+
+{% endset %}
+{{ format(example) }}
+
```
greg-a-smith: I'm not sure we need this example because popper.js-positioned components won't be able to have their width set with just CSS since they will exist in a different DOM node (child of body).

scss/components/popover.scss
jbadan: Ooof I thought about this wrong! Good call

PR #258, opened 2019-08-14
fix: add self contained styles to button-split component

jbadan:
Removed many tests that were testing the different combos of the button - out of scope for self contained styles. 

**Updated with post pr feedback screenshots**
2 tests updated - switch from `fd-button` to `btn-playground` and `btn-playground--emphasized` placeholder style. 
The rest deleted as they were repeats of the same test but with different fd-buttons.

Removed styles added to `popover.scss` that were specific for the fundamental-styles implementation on the docs site.

Screen Shot 2019-08-15 at 1 49 22 PM

docs site (to see emphasized button implementation): 
Screen Shot 2019-08-14 at 2 25 19 PM


scss/components/button-split.scss
```diff
@@ -14,35 +14,33 @@ $block: #{$fd-namespace}-button-split;
   display: inline-flex;
   vertical-align: middle;
   position: relative;
-  // & > * {
-    margin: 0;
-    [class*="fd-button"]:first-child {
+  margin: 0;
+  :first-child {
```
jbadan: Is this a good way of doing this or a really bad idea? 😆  @greg-a-smith

test/templates/button-split/index.njk
```diff
@@ -1,90 +1,15 @@
...
-              ]
-            })
-        },
-        modifier={ block: ['no-arrow', 'right']},
-        classes="contextual-menu"
-        ) | indent(4)
-    }}
-
-    {%- endcall %}
+{% call button_split() -%}{%- endcall %}
```
greg-a-smith: We should probably make an example that injects children with a `blah---emphasized` class to make sure the 1-pixel margin is getting put there.

scss/components/popover.scss
```diff
@@ -132,20 +132,6 @@ $block: #{$fd-namespace}-popover;
...
-              right: 0;
-            }
-          }
-
-          &.fd-popover__body--right {
-            right: 0;
-            @include fd-rtl() {
-              left: 0;
-            }
-          }
```
greg-a-smith: This block is interesting.  It seems like an override within the same file.  Up a few lines from this are `&--left` and `&--right` blocks that are also doing this, but with slightly different values.  Hmmm. 🤷‍♂ The blocks above are using a `-3px` value which makes no sense.  If we are removing these from popover, we should look into removing the other blocks as well.

scss/components/button-split.scss
```diff
@@ -14,35 +14,33 @@ $block: #{$fd-namespace}-button-split;
...
       border-bottom-right-radius: 0;
+      border-left-width: 1px;
     }
-    [class*="fd-button"]:nth-child(2) {
+    :first-child {
       border-top-left-radius: 0;
       border-bottom-left-radius: 0;
+      border-top-right-radius: $fd-border-radius;
+      border-bottom-right-radius: $fd-border-radius;
       border-left-width: 0;
```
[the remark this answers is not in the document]
jbadan: Good catch! The spacing threw me off and I didn't notice the border

scss/components/popover.scss
jbadan: Yeah I've been slowly trying to identify needed for popover styles and "needed for fundamental-styles popover styles and moving them as I encounter them. I believe this had to do with a problem in the original pr where the popover was slightly misaligned on the emphasized button due to adding that margin for the white gap.

scss/components/button-split.scss
```diff
@@ -14,35 +14,33 @@ $block: #{$fd-namespace}-button-split;
...
   position: relative;
-  // & > * {
-    margin: 0;
-    [class*="fd-button"]:first-child {
+  margin: 0;
+  :first-child {
+    border-top-right-radius: 0;
+    border-bottom-right-radius: 0;
+  }
+  :nth-child(2) {
```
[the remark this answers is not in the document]
jbadan: So this was another "fundamental-styles" specific case - due to using an example with popover:
```

<div class="fd-button-split fd-has-margin-right-small" role="group" aria-label="button-split">
  <button class="fd-button sap-icon--cart" aria-label="button">Button with text</button>
  <button class="fd-button sap-icon--slim-arrow-down" aria-controls="t4c0o273" aria-haspopup="true" 
  aria-expanded="false" aria-label="More"></button>
<Popover here />
</div>
```

PR #227, opened 2019-08-05
fix: Added self-contained style for Toggle component 

SaiShan:
Closes SAP/fundamental-styles#213

This PR contains the self-contained styling for the Toggle component.
This PR is part of the larger task of making all components self-contained. #136

Screenshots for test updates incoming --- only one test was actually modified - see screenshot below. The rest were deleted examples.

Screen Shot 2019-08-13 at 10 55 33 AM

Screen Shot 2019-08-13 at 10 57 38 AM

PR #262, opened 2019-08-14
docs: toggle on docs site missing fd-toggle__input class

jbadan:
In https://github.com/SAP/fundamental-styles/pull/227 I changed the selector from `input` element to class `fd-toggle__input` and forgot to update the docs site where we are using them in our display containers.

Also removed some unnecessary css overrides.

Before:
screen_shot_2019-08-14_at_3 00 29_pm

After:


greg-a-smith:
@jbadan One other thing ... maybe change the prefix to `docs:`?

Issue #266, opened 2019-08-16
add self-contained styles for Product Menu component


droshev:
Part of larger #136 for all components.

PR #267, opened 2019-08-16
feat: remove product menu as a component

droshev:
fixes: #266 
BREAKING CHANGE
Removing product menu as a component. It doesn't make sense to keep this as a separate component. It is completely specific to `shellbar`


scss/components/shellbar.scss
```diff
@@ -257,4 +257,44 @@ $block-user-menu: #{$fd-namespace}-user-menu;
...
+    &::after {
+      @include triangle(10px 5px, fd-color("shell", 2), down, --fd-color-shell-2);
+      content: "";
+      position: absolute;
+      top: calc(50% - 2.5px);
+      margin-left: 8px;
+      margin-right: 8px;
+    }
+  }
+  .fd-popover__body--right {
```
jbadan: This block is most likely fundamental-styles specific as popovers will live separately in the dom in the implementation libraries. See: https://github.com/SAP/fundamental-styles/blob/master/docs/_sass/_docs-display-component.scss#L43

test/templates/index.njk
```diff
@@ -1,118 +1,266 @@
...
+        <li>
+          <a href="pages/grid">Grid</a>
+        </li>
       </ul>
 
     </div>
-    {% endif %}
-</div>
+  {% endif %}
+</div>
```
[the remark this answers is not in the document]
droshev: that is my IDE settings :)

test/templates/product-menu/index.njk
```diff
@@ -1,26 +0,0 @@
-{% extends "layout.njk" %}
-{% from "./../format.njk" import format %}
-{% from "../button/component.njk" import button %}
-{% from "./component.njk" import product_menu %}
```
greg-a-smith: If we're removing this, shouldn't we also remove the `data.json` and `component.njk` files in the `product-menu` directory?

droshev: it is used in shellbar testing

PR #268, opened 2019-08-16
fix: self contained styles for tree component

jbadan:
## Related Issue
SAP/fundamental-styles#136

## Description

I originally intended to remove the icons from tree and just use fd-button for the `__control` (like was done with `__actions`. This created a lot of follow up problems, especially with differing hover, active, focus states than any of the `fd-button`s.  I'm open to opinions on this. It's unfortunate to include the icon library for one icon, although our roadmap includes replacing this icon setup so the problem might solve itself.

## Screenshots

tests **before** making any scss changes, only updating the test files to use placeholder buttons and badges:
Screen Shot 2019-08-16 at 9 55 38 AM

**after** scss changes:
Screen Shot 2019-08-16 at 10 06 12 AM


scss/components/tree.scss
```diff
@@ -160,11 +163,9 @@ $block: #{$fd-namespace}-tree;
         @include fd-var-color("color", $fd-link-color, --fd-color-action-1);
         @include fd-focus();
         @include fd-rtl() {
-          margin-left: fd-space(3);
```
jbadan: this margin-left was unnecessary

jbadan:
@greg-a-smith good catch! looks like that margin-left was necessary 😆

jbadan:
Updates:
Screen Shot 2019-08-19 at 2 42 55 PM
Screen Shot 2019-08-19 at 2 44 48 PM

hertweckhr1:
Spacing still seems a little off - or maybe its just a quirk in the test?

jbadan:
@hertweckhr1 good catch! It's not so much spacing as the font size for the headings is too big on my branch when compared to master.

Update incoming :)

jbadan:
@greg-a-smith @hertweckhr1 https://wiki.wdf.sap.corp/wiki/pages/viewpage.action?pageId=2060038188
Looks like the fonts were supposed to be 14px all along, we had them arbitrarily set to 12px. Updated with tests.

Screen Shot 2019-08-20 at 8 32 06 AM

PR #269, opened 2019-08-16
feat: add self-contained styles for Time component

markpalfreeman:
## Related Issue
#136

## Description
Self-contained styles for Time.

~**NOTE / HELP WANTED**:~
~I removed the selector at this level that shrinks its child `input` padding for these small inputs.
I need guidance on where this should go instead—I don't think we want that for all `.fd-form__control`s, it's specific to this context 🤷‍♂~ _addressed_ 

### Before:
Screen Shot 2019-08-16 at 10 13 42 AM

### After:
~❗️ notice the padding issue now that it's been reverted the "default"~ _addressed_
Screen Shot 2019-08-16 at 10 13 57 AM

Screen Shot 2019-08-16 at 10 22 14 AM


scss/components/time.scss
```diff
@@ -11,17 +11,13 @@ $block: #{$fd-namespace}-time;
...
       margin-right: fd-space(1);
+
       &:last-child {
         margin-right: 0;
       }
     }
 
     &__input {
-        margin: fd-space(2) 0;
-        input {
```
markpalfreeman: Most primary change—the removal of this padding messes things up, but I'm not sure where to put it instead.

test/resources/layout-shared.css
```diff
@@ -25,7 +25,8 @@ body {
 }
 
 .input-playground {
-  width: 250px;
+  width: 100%;
```
markpalfreeman: This is the default input style, which I think makes sense to have here too; without it, playground inputs with "limited" width extend way beyond their containers.

test/templates/time/component.njk
```diff
@@ -1,48 +1,31 @@
...
+{%- macro time(properties={}, modifier=[], aria=[]) -%}
+<div class="fd-time{{ modifier | modifier('time') }}"{{ aria | aria }}>
   {%- for item in properties.items %}
     {{ time_item(item) }}
   {%- endfor %}
 </div>
 {%- endmacro %}
 
-{%- macro time_item(properties={}, state={}, modifier=[], classes=[], aria=[]) -%}
+{%- macro time_item(properties={}) -%}
```
markpalfreeman: Cleaning up properties that aren't needed with just a single example.

scss/components/time.scss
greg-a-smith: 🤔  You are right that these styles are specific to this context.  We have n-number of "items", each with an input and two buttons.  Let's style it that way.  Let's keep the `fd-time__item` styles.  Then, let's remove (🔥) the current `fd-time__input` wrapper `<div>` and put that class directly on the `<input>` and make the styling:
```css
&__input {
    margin: fd-space(2) 0;
    padding-left: fd-space(2);
    padding-right: fd-space(2);
    text-align: center;
}
```
I would also put the `fd-time__input` as a child selector to `fd-time__item` to make sure our specificity wins on the `<input>` (since that will have styles already coming from `fd-form__control`).

This will all be a breaking change, but hey -- we're on version 0. 😄

markpalfreeman: This makes a lot of sense, good suggestion on just repurposing the class (maybe that's what Jenna was hinting at earlier and I didn't fully grasp it).

markpalfreeman:
Fixes addressed, the docs page has been restored: 
Screen Shot 2019-08-16 at 3 17 54 PM

components/time.scss
```diff
@@ -0,0 +1,4 @@
+$fd-fonts-path: "../scss/fonts/";
+
+@import "../scss/fonts/72";
```
greg-a-smith: Not sure why I didn't see this before, but the font isn't actually needed for this component.  The form input will bring the font with it.

scss/components/time.scss
```diff
@@ -11,17 +11,17 @@ $block: #{$fd-namespace}-time;
...
-          padding-left: fd-space(2);
-          padding-right: fd-space(2);
-          text-align: center;
-        }
+    // Spacing and narrower side padding for child inputs
+    &__item > &__input {
+      margin: fd-space(2) 0;
+      padding-left: fd-space(2);
+      padding-right: fd-space(2);
+      text-align: center;
```
greg-a-smith: You can leverage the power of sass and write this like:
```scss
&__item {
    {{ existing styles }}

    .#{$block}__input {
        margin: fd-space(2) 0;
        padding-left: fd-space(2);
        padding-right: fd-space(2);
        text-align: center;
    }
}
```
I think it's safer to write `fd-time__input` as a descendant class to `fd-time__item`, but not necessarily a direct descendant (using the `>` notation).

markpalfreeman: Yeah I went back and forth on both of those points :)

Changes made.

scss/components/time.scss
```diff
@@ -11,17 +11,17 @@ $block: #{$fd-namespace}-time;
       max-width: fd-space(10);
       text-align: center;
       margin-right: fd-space(1);
+
```
greg-a-smith: I think I missed this before (probably because it didn't show up directly in the file diff), but what exactly is this `width: 184px` for?  The phrase "quick fix" worries me.  And _what_ overflow issue?  This _seems_ like it might be related to the use of `fd-time` within `fd-time-picker`, but this strikes me as a problem. In fact, see the attached screenshot.

Screen Shot 2019-08-19 at 9 49 30 AM

The 4th `__item` is wrapping due to this arbitrary width.  I know it's the playground with crazy margin values being applied, but there is still no reason why any of this should wrap to a new line.

Maybe the `fd-time` class should provide a flex box wrapper around the `fd-time__item` blocks to prevent them from ever "wrapping".  Thoughts?

markpalfreeman: I agree that it should be changed (any fixed width seems like a hack), but I wasn't sure this was the right time to do it...

greg-a-smith: You know what, maybe it isn't the time to do it.  I've created #280 to track this work.

markpalfreeman:
@greg-a-smith thanks, I agree. I'm wrapping up some TimePicker changes I was going to bundle here, but for the sake of reviews I'll just open another small PR

PR #270, opened 2019-08-17
fix: self contained styles for tile grid component

hertweckhr1:
## Description
Self contained styles for tile grid component. Part of #136.

image
image


test/resources/layout-shared.css
```diff
@@ -26,3 +26,35 @@ body {
...
+} 
+
+.fd-has-grid-column-span-2 {
+  grid-column: span 2;
+}
+
+.fd-has-background-color-accent-7 {
+  background-color: var(--fd-color-accent-7, #0cd7ed) !important;
+}
+
```
hertweckhr1: ^^ Not sure if this is the best place for the fd-has-gride-row-span-2, fd-has-grid-column-span-2, and fd-has-background-color-accent-7 helper functions.

scss/components/tile-grid.scss
```diff
@@ -17,6 +18,7 @@ $block: #{$fd-namespace}-tile-grid;
   position: relative;
 
   & > * {
+    @include fd-reset();
```
greg-a-smith: This _seems_ wrong.

test/templates/tile-grid/index.njk
```diff
@@ -1,50 +1,37 @@
 {% extends "layout.njk" %}
 {% from "./../format.njk" import format %}
-{% from "../tile/component.njk" import tile, tile_content, tile_media, tile_actions %}
-{% from "../product-tile/component.njk" import product_tile, product_tile_content, product_tile_media %}
-{% from "../button/component.njk" import button %}
 {% from "./component.njk" import tile_grid %}
 {% from "../identifier/component.njk" import identifier %}
```
greg-a-smith: Identifier shouldn't be imported or used here.

jbadan:
@greg-a-smith I'd be happier with less repetitive code. I'm 💯 on board with replacing with a generic style. I think now is the time to do the replacements we want and group it with all the other large breaking changes.

hertweckhr1:
reworking for a layout-grid component.

PR #272, opened 2019-08-17
fix: add new section playground; section self-contained

### LINKED ISSUE #271: add self-contained styles for Section

droshev:
fixes #271 
- create new playground test for `section`
- add self-contained `section.scss`


test/templates/section/index.njk
```diff
@@ -0,0 +1,23 @@
+{% extends "layout.njk" %}
+{% from "./../format.njk" import format %}
+{% from "./component.njk" import section,
+section_headlines %}
```
greg-a-smith: Should be on the line above it.

scss/layout/_section.scss
```diff
@@ -13,55 +15,59 @@
...
+    flex: 1;
+    margin-bottom: 0;
+  }
+  &__actions {
+    margin-left: auto;
+  }
+  &__footer {
+    display: flex;
+    justify-content: center;
+  }
```
greg-a-smith: It's tough to tell if anything actually changed in this file with the reformatting.  Were there any changes?

scss/layout/_section.scss
```diff
@@ -1,6 +1,8 @@
 @import "./../settings";
 @import "./../functions";
 @import "./../mixins";
+@import "./../core/root";
```
greg-a-smith: Why is this being imported?

droshev: there are some css variables like `--fd-padding-ui` being defined there

scss/layout/_section.scss
droshev: I reverted the reformatted code

greg-a-smith:
I created #282 as a follow-up for sections and panels, but it is outside the scope of this PR.

PR #274, opened 2019-08-17
fix: add page self-contained; add test in the playground

### LINKED ISSUE #273: add self-contained styles for Page

droshev:
fixes #273


scss/layout/_page.scss
```diff
@@ -1,6 +1,7 @@
 @import "./../settings";
 @import "./../functions";
 @import "./../mixins";
+@import "./../core/root";
```
greg-a-smith: Why is this being imported?

test/templates/page/index.njk
```diff
@@ -0,0 +1,14 @@
+{% extends "layout.njk" %}
+{% from "./../format.njk" import format %}
+
+{% block content %}
+    {% set example %}
+    <div style="border: 1px solid">
+        <article class="fd-page">
```
greg-a-smith: I know this has been spec'd this way, but `<article>` seems like the wrong tag to use for a "page".  Almost by definition, shouldn't this class be applied to a `<main>` tag?

From https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main:
> The **HTML `<main>` element** represents the dominant content of the `<body>` of a document. The main content area consists of content that is directly related to or expands upon the central topic of a document, or the central functionality of an application.

components/page.scss
```diff
@@ -0,0 +1,4 @@
+$fd-fonts-path: "../scss/fonts/";
+
+@import "../scss/layout/page";
+@import "../scss/icons/icon";
```
greg-a-smith: Icons are being imported with no variable set and no font is being imported, but a variable is being set.  Something seems amiss. 😄 Seems like fonts are needed, but not icons.

scss/layout/_page.scss
droshev: there are some css variables like `--fd-padding-ui` that are defined there

greg-a-smith:
Created #283 as a follow-up story.

Issue #275, opened 2019-08-19
Localization Editor should be redesigned


greg-a-smith:
The Localization Editor component should be redesigned and use cases adjusted.

* The popover should only show when the button is clicked (not on focus of the input)
* The use of input groups for both the default language and all subsequent languages is confusing because the default one is a button and all others are just addon text
* If the other languages are going to use input groups, there needs to be some aria attributes added to help describe the inputs

droshev:
we will remove it

PR #276, opened 2019-08-19
feat: Self-contained styling for localization editor

greg-a-smith:
## Related Issue
n/a

## Description
Self-contained styles for localization editor.  This is a breaking change.  This component should be redesigned to make it easier to use, but that will be tracked in a separate story.  See #275 

## Screenshots
Screen Shot 2019-08-19 at 1 52 07 PM


docs/pages/components/localization-editor.md
```diff
@@ -22,31 +22,31 @@ summary:
             <input type="text" class="" id="" aria-controls="NJvVh542c" aria-expanded="false"
             aria-haspopup="true" placeholder="Enter Label">
             <span class="fd-input-group__addon fd-input-group__addon--after fd-input-group__addon--button">
-            <button class="fd-button--light fd-localization-editor__button"
+            <button class="fd-button--light"
```
greg-a-smith: This class should not be needed as input groups should already handle this.

docs/pages/components/localization-editor.md
```diff
@@ -22,31 +22,31 @@ summary:
...
             aria-controls="test" aria-expanded="false" aria-haspopup="true">EN*</button>
             </span>
          </div>
       </div>
       <div class="fd-popover__body fd-popover__body--no-arrow"  aria-hidden="true" id="NJvVh542c">
          <nav class="fd-menu">
-            <ul class="fd-menu__list fd-localization-editor__list">
```
greg-a-smith: This class is functionally being replaced by `fd-localization-editor__language` which will be placed directly on the `<li>` elements to adjust spacing.

scss/components/localization-editor.scss
```diff
@@ -2,48 +2,11 @@
...
-        width: 100%;
-    }
-
-    .fd-menu {
-        display: block;
-    }
-    //STATES *******************************************
-    &.is-disabled,
-    &[disabled] {
-        opacity: 0.5;
```
greg-a-smith: Changes:

* The `__list` class has been replaced by `__language`
* The `__button` class is removed because input groups should already handle this
* Popover styling will be handled differently in an implementation library because popover will be placed in a different DOM node
* Menu styling does not belong here
* The localization editor `<div>` is not what gets the disabled class/attribute

This may eventually all go away when it gets redesigned.

test/templates/localization-editor/component.njk
```diff
@@ -1,85 +1,20 @@
...
-<div class="fd-localization-editor">
-    {{  popover(properties={
-        id: _id,
-        control: localization_editor_main_input_item(properties, aria={controls: _id},size=size),
-        body: localization_editor_input_body(properties)
-      }, modifier={ block: ['no-arrow'] })
-    }}
 </div>
-{%- endmacro %}
+{% endmacro %}
```
greg-a-smith: For the playground, all we need to show is the list of languages to validate the spacing between them.

PR #278, opened 2019-08-19
fix: self contained styles for table

hertweckhr1:
## Related Issue
#136

## Description
Self-contained styles for table.  
* got rid of playground examples with images and checkboxes since table is not responsible for their styles 
* for playground purposes, for contextual menu examples, the menu is empty since popover and menu are the ones handling those styles. 

## Screenshots
3 tests failing and updated. 
* 2 tests are for the tests I ommitted (with images and checkboxes) 
* last test looks funny/off? 
image

image


components/table.scss
```diff
@@ -0,0 +1 @@
+@import "../scss/components/table";
```
jbadan: The table probably needs fonts

hertweckhr1: Yup I think you're right, forgot to add it in.

scss/components/table.scss
```diff
@@ -82,20 +83,19 @@ $block: #{$fd-namespace}-table;
             }
         }
     }
-    tbody {
-        tr {
+    &__body {
+        &-row {
```
markpalfreeman: Are these going to win specificity wars with some generic element selector like `table tbody tr` etc?

#ihatetables

test/templates/table/component.njk
```diff
@@ -41,36 +41,36 @@ table:
 <div class="fd-table--fixed">
 {% endif %}
 <table  class="fd-table{{ modifier.block | modifier('table') }}{{ state | state }}{{ classes | classes }}"{{ aria | aria }}>
-    <thead>
-        <tr>
+    <thead class="fd-table__header">
+        <tr class="fd-table__header-row fd-table__row">
```
jbadan: Can we simplify these class names at all? So it only requires 1? Like since it's the child of `fd-table__header`, can we just have `fd-table__row` class and handle that in the scss?

test/templates/table/component.njk
```diff
@@ -41,36 +41,36 @@ table:
...
-            <td>{{ item }}</td>
+            <td class="fd-table__col fd-table__body--col">{{ item }}</td>
             {%- endif %}
             {%- endfor %}
         </tr>
         {%- endfor %}
     </tbody>
 </table>
-{%- if properties.fixed_col_table %}
+{%- if properties.ç %}
```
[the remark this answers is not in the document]
hertweckhr1: new keyboard is getting the best of me 🤣

scss/components/table.scss
hertweckhr1: I think the goal is to not have these styles bleed over into other components, hence the making of these classes instead of using the element selectors

scss/components/table.scss
```diff
@@ -82,20 +83,19 @@ $block: #{$fd-namespace}-table;
...
             border-right: none;
         }
-        td {
+        &-col {
             word-break: break-word;
         }
     }
-    td,
-    th {
+    &__col {
```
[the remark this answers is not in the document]
hertweckhr1: fd-table__col will cover both td and th 👍

hertweckhr1:
Updated tests (although they don't seem to be rendering the same examples): 
image

scss/components/table.scss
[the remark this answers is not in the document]
markpalfreeman: I know, I'm speaking more about the name. The selector `td, th` was previously "any cell in a table" whereas `fd-table__col` implies it's just for columns...

I'm just always trying to wear the maintainability/are-we-going-to-understand-this-in-3-months hat 🎩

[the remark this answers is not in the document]
jbadan: Maybe `fd-table__cell` might be a better fit here?

test/templates/table/index.njk
```diff
@@ -281,11 +243,7 @@
                 label: "Header Column",
                 id: "col1",
                 context_menu: true,
-                menu: [
-                    { label: "Ascending" },
-                    { label: "Decensing" },
-                    { label: "Fix Column" }
-                ],
+                menu: [],
```
jbadan: We can probably just remove the menu option from the table macro because it's handled by popper.

scss/components/table.scss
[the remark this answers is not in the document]
hertweckhr1: @greg-a-smith thoughts? I believe it might be best then to have this split up into `fd-table__header-cell` and `fd-table__col` instead of just `fd-table__cell` for ease on labeling purposes? Otherwise I will need to label all <td> tags with `fd-table__col` and `fd-table-cell`. 

With them split up I can label all `<th>` with `fd-table__header-cell` and keep all <td> as `fd-table__col`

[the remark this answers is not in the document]
greg-a-smith: @hertweckhr1 I like @markpalfreeman's suggestion.  He's right that `fd-table__cell` is a better replacement for `td` and `th` than `fd-table__col`.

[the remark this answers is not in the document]
hertweckhr1: coolio changing it to fd-table__cell on Mark and Jenna's good thinkings 👍

test/templates/table/component.njk
```diff
@@ -41,29 +41,29 @@ table:
 <div class="fd-table--fixed">
```
jbadan: I know you didn't write this, but can we do this cleaner? I guess it could be a followup but needing 2 extra divs when we have a fixed table seems weird to me.

test/templates/table/component.njk
```diff
@@ -41,29 +41,29 @@ table:
...
 {% endif %}
 <table  class="fd-table{{ modifier.block | modifier('table') }}{{ state | state }}{{ classes | classes }}"{{ aria | aria }}>
-    <thead>
-        <tr>
+    <thead class="fd-table__header">
+        <tr class="fd-table__row">
             {%- for header in properties.headers %}
             {%- if header.context_menu -%}
-            <th class="fd-table__context-menu {{header.classes | classes}}" style="left:{{0 if header.fixed_col else header.offset}}; width:{{header.width}}" {{header.aria | aria}}  scope="col" >
+            <th class="fd-table__context-menu {{header.classes | classes}} fd-table__cell " style="left:{{0 if header.fixed_col else header.offset}}; width:{{header.width}}" {{header.aria | aria}}  scope="col" >
```
jbadan: Should this have inline styles? I think that defeats the purpose of testing the styles.

test/templates/table/component.njk
```diff
@@ -41,29 +41,29 @@ table:
...
     </thead>
-    <tbody>
+    <tbody class="fd-table__body" >
         {%- for row in properties.rows %}
-        <tr{% if properties.selectedRow === loop.index %} aria-selected="true"{% endif %}>
+        <tr{% if properties.selectedRow === loop.index %} aria-selected="true"{% endif %}  class="fd-table__row">
             {%- for item in row %}
             {%- if item.label %}
-            <td class="{{item.class}}" style="left:{{0 if item.fixed_col else item.offset }}; width:{{item.width}}">{{ item.label }}</td>
+            <td class="{{item.class}} fd-table__cell" style="left:{{0 if item.fixed_col else item.offset }}; width:{{item.width}}">{{ item.label }}</td>
```
jbadan: same for the inline styles here - seems weird

jbadan:
Can we delete all of these no longer used variables too: https://github.com/SAP/fundamental-styles/blob/5e30025255115b7443e7bfca3511a8d66618fba3/test/templates/table/index.njk#L32

scss/components/table.scss
```diff
@@ -82,20 +85,31 @@ $block: #{$fd-namespace}-table;
             }
         }
     }
-    tbody {
```
jbadan: I think `__fixed-col` a couple lines above should also be changed to `__fixed-cell`, but I could be getting confused on everything moving around.

scss/components/table.scss
```diff
@@ -176,9 +190,6 @@ $block: #{$fd-namespace}-table;
     &__fixed-col {
```
jbadan: same here - should this be `__fixed-cell`?

scss/components/table.scss
```diff
@@ -192,21 +203,40 @@ $block: #{$fd-namespace}-table;
...
 
-        td {
+        .#{$block}__cell {
             min-width: 200px;
         }
     }
 
     &--no-borders {
         border: none;
-        & tbody tr {
```
jbadan: I doesn't look like the block is working as expected, look at the `--no-borders` example, there shouldn't be any borders between the rows:
Screen Shot 2019-08-21 at 11 55 27 AM

jbadan:
One last thing - the header background color isn't picking up `fd-color("neutral", 1)` in theme ugly - it should be turning dark reddish  ` --fd-color-neutral-1: rgb(117, 7, 7);`

scss/components/table.scss
hertweckhr1: __fixed_col refers to an entire column that is fixed, so in this case it seems to be described correctly?

scss/components/table.scss
hertweckhr1: same reply as above.

scss/components/table.scss
markpalfreeman: Yeah I'm fine leaving this if it was already thought through as a "column" style.

scss/components/table.scss
```diff
@@ -82,20 +85,31 @@ $block: #{$fd-namespace}-table;
...
+        font-size: $fd-font-size;
+        -webkit-font-smoothing: antialiased;
+        box-sizing: border-box;
+        padding: 0;
+        margin: 0;
+        border: 0;
+        &::before,
+        &::after {
+            box-sizing: inherit;
+        }
```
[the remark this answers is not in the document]
hertweckhr1: almost, but not quite - I had to leave out the font color being set in fd-reset and since its only being used once, it didn't seem necessary to add an additional mixin

scss/components/table.scss
hertweckhr1: I ended up switching it to cell - it seems like its styling the individual cells rather than a whole, so thought better to change to keep consistent with our previous change to fd-table__cell

hertweckhr1:
Replaced downward arrow since that was using popper as a placeholder.

docs/pages/components/table.md
```diff
@@ -163,40 +163,40 @@ A table is a set of tabular data. Line items can support data, images and action
...
 <table class="fd-table">
-    <thead>
-        <tr>
-            <th scope="col"><input type="checkbox"></th>
-            <th scope="col">Column Header</th>
-            <th scope="col">Column Header</th>
-            <th scope="col">Column Header</th>
-            <th scope="col">Column Header</th>
+    <thead class="fd-table__header">
+        <tr class="fd-table__row>
```
jbadan: Just a small typo here - missing the `"` and then I think we're ready to 🚢 🎉

greg-a-smith: I pushed the changes to this quick as I was reviewing it and I agree with @jbadan -- this looks good.

scss/components/table.scss
```diff
@@ -166,19 +181,18 @@ $block: #{$fd-namespace}-table;
         }
     }
 
-    th.fd-table__context-menu {
+    .#{$block}__header .#{$block}__cell.fd-table__context-menu {
```
greg-a-smith: Couple things:

First, for better readability and searching, let's change this to `.#{$block}__contextual-menu`.

Second, can we write this in better sass syntax with nesting for descendant and sibling classes?

scss/components/table.scss
```diff
@@ -166,19 +181,18 @@ $block: #{$fd-namespace}-table;
...
         &:hover {
             @include fd-var-color("background-color", $fd-table-sort-column-header-background-color, --fd-color-neutral-2);
             cursor: pointer;
         }
     }
 
-    &__fixed-col {
+    &__fixed-cell {
+        left: 0;
+        max-width: 200px;
```
greg-a-smith: When using a fixed column, why does the column have to be 200 pixels wide?  I see lots of hard-coded references to 200px and although it works, it seems very limiting.  That seems like a detail for the implementation library to handle.

A little out of scope for this PR, but maybe we come back to this in a separate issue?

scss/components/table.scss
```diff
@@ -188,25 +202,46 @@ $block: #{$fd-namespace}-table;
...
             min-width: 200px;
         }
     }
 
     &--no-borders {
+        @include fd-reset();
         border: none;
-        & tbody tr {
+
+        .#{$block}__body .#{$block}__row {
```
greg-a-smith: Better sass syntax (nesting)?

scss/components/table.scss
```diff
@@ -188,25 +202,46 @@ $block: #{$fd-namespace}-table;
...
+            text-align: right !important;
+        }
+    }
+    @each $i in 2,3,4,5,6,7,8,9 {
+        &.#{$fd-namespace}-has-nth-child-#{$i}-text-align-right {
+            .#{$block}__cell:nth-child(#{$i}) {
+            text-align: right !important;
+            }
+        }
+    }
```
greg-a-smith: Do these class names make anyone else cringe? 🤢 Don't need to change them now, but there _has_ to be a better way.

scss/components/table.scss
```diff
@@ -188,25 +202,46 @@ $block: #{$fd-namespace}-table;
...
-        & tbody tr {
+
+        .#{$block}__body .#{$block}__row {
             border: none;
         }
     }
 
     &--striped {
-        tbody tr:nth-child(even) {
+        .#{$block}__body .#{$block}__row:nth-child(even) {
```
greg-a-smith: Better sass syntax (nesting)?

scss/components/table.scss
```diff
@@ -60,42 +61,56 @@ $block: #{$fd-namespace}-table;
...
+        .#{$block}__row {
             &:hover {
                 background-color: inherit;
             }
         }
 
-        td,
-        th {
+        .#{$block}__cell,
+        .#{$block}__row {
```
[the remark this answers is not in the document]
hertweckhr1: This actually impacts the styling a bit, but can rework row separately to match.

scss/components/table.scss
hertweckhr1: I can remove this, and add back the inline styling in example.  With docs instruction, it looks like there is direction for the consumer to set width here.

hertweckhr1: Separate story opened: https://github.com/SAP/fundamental-styles/issues/292

PR #284, opened 2019-08-20
fix: revert multi input test/new shared layout input class

jbadan:
Add new full-width input class for shared components to prevent clashes/fix failing test on master.

+ Revert multi-input test accidentally updated here: https://github.com/SAP/fundamental-styles/pull/269


test/resources/layout-shared.css
```diff
@@ -30,14 +30,18 @@ body {
...
-  width: 100%;
-  box-sizing: border-box;
+.input-playground, .input-playground-full-width {
+  width: 250px;
   background: beige;
   padding: 10px;
   font-size: 14px;
 }
 
+.input-playground-full-width {
```
markpalfreeman: This is so non-BEM of you 😉

jbadan: BEM schmem ;)

PR #286, opened 2019-08-20
feat: remove product switcher component

### LINKED ISSUE #265: add self-contained styles for Product Switcher component
Self-contained styles for Product Switcher.

droshev:
fixes #265 

Breaking Changes:
Remove Product Switcher as a component; remove the playground test


scss/components.scss
```diff
@@ -40,6 +40,5 @@
 @import "components/localization-editor";
 @import "components/search-input";
 @import "components/multi-input";
-@import "components/product-switcher";
```
markpalfreeman: Was this sass file deleted previously?

scss/components/shellbar.scss
```diff
@@ -289,4 +289,62 @@ $block-product-menu: #{$fd-namespace}-product-menu;
...
+$fd-product-switcher-product-width: 140px;
+$fd-product-switcher-product-height: 40px;
+
+/*!
+.fd-product-switcher
+*/
+$block-product-switcher: #{$fd-namespace}-product-switcher;
+
+
+.#{$block-product-switcher} {
```
markpalfreeman: Was all this just ported over directly from somewhere?

scss/components/shellbar.scss
```diff
@@ -289,4 +289,62 @@ $block-product-menu: #{$fd-namespace}-product-menu;
...
+  position: relative;
+  display: inline-block;
+
+  &__body {
+    padding: 20px 10px 0 10px;
+    color: initial;
+    > nav {
```
markpalfreeman: I thought we were trying to avoid using direct element selectors for all of these components 🤔

test/templates/shellbar/component.njk
```diff
@@ -236,7 +235,7 @@ shellbar:
         {%- endif %}
         {%- if switcher %}
             <div class="fd-shellbar__action fd-shellbar__action--collapsible">
-              {{ product_switcher() }}
+              <button class=" fd-button--shell sap-icon--grid"></button>
```
markpalfreeman: nit: ✂️ space between `class=" fd-button`

scss/components/shellbar.scss
jbadan: This will be fixed when we handle shellbar - I think this is just the stop gap so the docs site examples don't break. We did the same with product-menu.

droshev:
done

PR #291, opened 2019-08-23
fix: Self-contained styles for overlay

greg-a-smith:
## Description
Self-contained styles for `fd-overlay`. Created a new playground page to demonstrate this that would work with self-contained styles.  Technically, overlay isn't its own component, but it gets used in conjunction with other components.


test/templates/index.njk
```diff
@@ -55,10 +56,10 @@ li {
           <li><a href="side-nav">.fd-side-nav</a></li>
           <li><a href="spinner">.fd-spinner</a></li>
           <li><a href="table">.fd-table</a></li>
+          <li><a href="tabs">.fd-tabs</a></li>
           <li><a href="tile">.fd-tile</a> | <a href="product-tile">.fd-product-tile</a></li>
           <li><a href="tile-grid">.fd-tile-grid</a></li>
           <li><a href="time">.fd-time</a>
-          <li><a href="tabs">.fd-tabs</a></li>
```
greg-a-smith: Couldn't stand seeing tabs out of alphabetical order any longer. 😜

test/resources/layout-shared.css
```diff
@@ -45,3 +50,21 @@ body {
...
 }
+
+.overlay-playground {
+    width: 400px;
+    padding: 50px 0;
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    align-items: center;
+}
```
markpalfreeman: What do you think about nesting the "modifiers" here?

I could be convinced either way, but I'm getting nervous we'll confuse our future selves mixing "real" and "fake" classes together when scanning down the classes in this file.

```
.overlay-playground {
  ...

  .fd-overlay--modal & {
    background: white;
  }
}
```

test/templates/overlay/index.njk
```diff
@@ -0,0 +1,27 @@
+{% extends "layout.njk" %}
+
+{% block content %}
+    <script>
+    function toggle(type, show) {
+        var overlay = document.querySelector('.fd-overlay--' + type);
+        if (overlay) {
+            overlay.setAttribute('aria-hidden', (show ? "false" : "true"));
```
markpalfreeman: Are these still styled based on the `aria-hidden` attribute?

markpalfreeman:
I _think_ I'm leaning towards importing the overlay styles into `modal.scss` and `alert.scss` directly, because they are required styles that those "components" need.

It makes more sense for me for our `components` library folder match 1:1 with the docs site components.

test/resources/layout-shared.css
greg-a-smith: I can't do nesting in this file because it's a CSS file.  We _could_ refactor it into sass, but that seemed like overkill.

test/templates/overlay/index.njk
greg-a-smith: They are, yes, but in the source sass file, there was no class-based alternative.  I can look into adding an `is-hidden` class to it.

greg-a-smith:
@markpalfreeman So, I _think_ it is actually optional for alerts.  I think alerts are styled in such a way that they can either be dropped in as an overlay (as seen in these playground examples) or they can be used as inline content similar to the NUI Widgets MessageBar component.  Either way, I'm still on the fence about it.

markpalfreeman:
@greg-a-smith me too (on the fence). I'm fine if it's a separate file that can be utilized only when needed, but I think it's the mismatch of the `components/` folder with actual documented components that's bothering me most.

test/templates/overlay/index.njk
markpalfreeman: I think if this is something doing for all other "self-contained" components, this is the time to do it for `overlay` too.

hertweckhr1: ^Agree with Mark for consistency

Issue #293, opened 2019-08-26
Input placeholder text is not italicized


KevinOkamoto:
**Describe the bug**

The placeholder text of the input element and related elements (e.g. combobox, search input) is not italicized.

fundamental-styles 0.1.0
fundamental-ngx 0.9.0

**To Reproduce**

Using the fundamental-ngx  `fd-combobox` component, thusly:

<fd-combobox glyph="search" placeholder="Search…" [dropdownValues]="dropdownValues"></fd-combobox>

I get the following result:
![image](https://user-images.githubusercontent.com/48103491/63704539-1fb3f580-c7e0-11e9-9c05-44f47668e2f6.png)

**Expected behavior**

The placeholder text "Search…" should be italicized:
![image](https://user-images.githubusercontent.com/48103491/63704583-36f2e300-c7e0-11e9-8dea-c377ef04256d.png)

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Mac OS
 - Browser: Chrome, Safari, Firefox
 - Version: 0.1.0

**Smartphone (please complete the following information):**

**Additional context**

PR #301, opened 2019-08-28
fix: Making input placeholder text italic

KevinOkamoto:
## Related Issue
Closes SAP/fundamental-styles#293

## Description
Applied `font-style: italic` to input placeholder text.

### Before:
Screen Shot 2019-08-28 at 8 57 45 AM

### After:
Screen Shot 2019-08-28 at 9 03 43 AM

### DIFF
--- a/scss/components/input.scss
+++ b/scss/components/input.scss
@@ -12,6 +12,7 @@ $block: #{$fd-namespace}-input;
     &::placeholder {
         @include fd-reset();
         @include fd-var-color("color", $fd-forms-color--placeholder, --fd-color-neutral-4);
+        font-style: italic;
     }
     &--compact {
         @include fd-var-size("height", $fd-forms-height--compact, --fd-forms-height-compact);


jbadan:
@KevinOkamoto I was thinking about this pr a little more and I think we might run into some problems. `fd-input` is also part of the building blocks for datepicker and timepicker, which do not have italicized placeholders (according to the wiki I briefly glanced at).

@droshev any thoughts on that?

bcullman:
@jbadan, @droshev: 
I *think* that placeholder text is meant to be italic, while mask text (like what you might see initially in a date-picker) is not italic.

jbadan:
@bcullman @KevinOkamoto that sounds reasonable. I think once the tests are updated this is ready to 🚢

jbadan:
@KevinOkamoto are you able to update this pr with conflicts + unit tests?

jbadan:
@KevinOkamoto Please reopen this pr when you have time to update your tests and conflicts.

PR #305, opened 2019-08-30
feat: shellbar refactor and self contained styles

jbadan:
## Related Issue
SAP/fundamental-styles#136

[BREAKING CHANGE]

## Description
* self contained styles for Shellbar component. 
* Restructured to use input group, button, identifier, etc. simplified markup - removed some extraneous divs
* Created sub components shellbar directory to store product_menu, product_switcher and search_input (specific to shellbar)
* regular search input component turned out to be more of a pattern (like combobox input)
       - created new docs section for patterns to show how to put these together

**search input does not match docs site because:**
Screen Shot 2019-08-30 at 8 30 26 AM

## Screenshots

Updated screenshot post pr feedback:

Screen Shot 2019-09-03 at 1 59 40 PM


scss/components/shellbar.scss
```diff
@@ -65,15 +74,65 @@ $block: #{$fd-namespace}-shellbar;
             }
         }
     }
+
+    &__button {
```
bcullman: should button, title/subtitle and logo be split out as sub files?

jbadan: yeah that would make sense - then I can reuse the styles for title in product-menu as well

test/resources/layout-shared.css
```diff
@@ -12,6 +12,12 @@ body {
     padding: 20px;
 }
 
+.input-group-button-playground, .input-group-playground {
+  font-size: 14px;
+  padding: 0;
+  margin: 0;
+  border: none;
+}
```
greg-a-smith: I know these PRs crossed in the night, but hopefully the input-group branch will help reduce the need for this.

jbadan:
> From a self-contained point of view, I think this looks pretty good (just a few comments). Some other random feedback I have, but feel free to queue all this up in a separate story as it really has nothing to do with self-contained styling.
> 
> 1. There's a `fd-shellbar__title` and then a `fd-product-menu` and a `fd-product-menu__title` depending on whether the "Title" is a menu or not.  This seems like it should be just one thing.
> 2. Why is `fd-shellbar__group--start` a flex layout with items aligned center, but neither `fd-shellbar__group--middle` nor `fd-shellbar__group--end` are.  Seems like they could all benefit from this nice optimaization and keep shellbar layouts similar.
> 3. The modifiers for `fd-shellbar__group` (`--start`, `--middle` and `--end`) seem very specific to their placement.  Maybe rename more to their content (_e.g._ `--product`, `--help` and `--actions` or something along those lines)?  You can always use the flex box `order` to move them around visually.
> 4. Aren't most (if not all) `fd-shellbar__action` elements going to contain a button child?  Could this and `fd-shellbar__button` be combined into one or the other?  I do see search input within one and that makes me wonder a bit about the name. 🤷‍♂
> 5. The class `fd-shellbar__logo--image-replaced` seems like an odd name for the SAP logo. 😄

1. replaced both with `fd-shellbar__title`
2. I have no idea-  they now are all updated to a sane flexbox implementations. 
3. Renamed `fd-shellbar__group--(product, copilot, actions)
4. eh while they probably _should_, there's a case for using a `fd-identifier`, I've left the actions in place for now. Also replaced the `--collapse` and `--collapsible` with `--mobile` and `--desktop` as I had a hard time telling the difference between the two with the previous name.
5. I have no idea what the purpose of this class is. Was it meant to be a fallback if the logo wasn't found? Was it just for the docs site? I have no idea so I left it alone.

scss/components/shellbar.scss
jbadan: I decided not to do that for now - the naming convention would be thrown off unless we made them mixins/they make sense in the shellbar base file. Thoughts?

docs/pages/components/shellbar.md
```diff
@@ -34,7 +34,8 @@ The shellbar handles layout and has some built-in elements but relies on standal
 * `.fd-shellbar__subtitle` (optional) displays an application context. _This should be used rarely._
 * `.fd-product-menu` (optional) for navigating to applications within the product.
 * `.fd-shellbar__action` (required) container for each product action and link.
-* `.fd-shellbar__collapse` (optional) for product actions on mobile screens.
+* `..fd-shellbar__action--mobile` (optional) for product actions only visible on mobile screens.
```
greg-a-smith: Minor nit: Remove the double dots (`..`)

scss/shellbar/_search-input.scss
```diff
@@ -9,43 +9,40 @@ $block-search-input: #{$fd-namespace}-search-input;
...
+        transition: border-color .125s;
+        padding-left: 12px;
+        padding-right: 12px;
+        &:focus {
+            box-shadow: none;
+        }
+        &[aria-hidden="true"],
+        &.is-hidden {
+        opacity: 0;
+        visibility: hidden;
```
greg-a-smith: Not a big fan of this right now, but I have a feeling we will be iterating on shellbar a few times.  I think there are some further optimizations we can achieve, but this is probably okay for now.

Issue #165, opened 2019-07-05
Self-contained style for input-group component


DorinR:
Similar to #60

PR #307, opened 2019-08-31
feat: Self-contained styles for input group

greg-a-smith:
## Related Issue
Closes SAP/fundamental-styles#165

## Description
Refactored `input-group.scss` to be much simpler and self-contained.  This will be a **breaking change** for consumers.  Changes outlined below:

### Things Removed
* `type="number"` _(new modifier in `input.scss`)_
* `type="search"` _(will be part of search input component or put on consumers to handle)_
* `fd-input-group--no-border` _(no examples shown or used ... and why???)_
* `fd-input-group--compact` _(will be handled directly at sub component level by all "compact"-able components such as inputs, textareas and buttons)_
* `fd-input-group__button` _(all button element styling comes from button)_
* `fd-input-group__button--step-up` _(button already handles icons and will be responsibility of consumers)_
* `fd-input-group__button--step-down` _(button already handles icons and will be responsibility of consumers)_
* `fd-input-group__button--clear` _(button already handles icons and will be responsibility of consumers)_
* `fd-input-group--before` _(pseudo selectors replace the need)_
* `fd-input-group--after` _(pseudo selectors replace the need)_
* `fd-input-group__addon--before` _(pseudo selectors replace the need)_
* `fd-input-group__addon--after` _(pseudo selectors replace the need)_

## Screenshots
Due to the overhaul of the changes and the rewriting of the playground examples, showing before and after diffs would be almost useless.


scss/components/button.scss
```diff
@@ -116,6 +116,19 @@ a.#{$block} {
...
+    @include fd-var-size("max-height", $fd-forms-height/2, --fd-forms-height-half);
+    line-height: 1;
+
+    &.#{$block}--compact {
+      --fd-forms-height-half: #{$fd-forms-height--compact/2};
+      @include fd-var-size("height", $fd-forms-height--compact/2, --fd-forms-height-half);
+      @include fd-var-size("max-height", $fd-forms-height--compact/2, --fd-forms-height-half);
+      line-height: 1;
+    }
+  }
```
greg-a-smith: Added to allow buttons to support the stacked up/down on the number "widget" with input group.

scss/components/button.scss
```diff
@@ -235,16 +248,16 @@ a.#{$block} {
...
       @include fd-var-color("background-color", fd-color("neutral", 1), --fd-button-background-color);
       @include fd-disabled() {
         --fd-button-background-color: transparent;
         --fd-button-color: var(--fd-color-action-1);
         @include fd-var-color("color", fd-color("action", 1), --fd-button-color);
         @include fd-var-color("background-color", transparent, --fd-button-background-color);
-        box-shadow: none;
       }
-      box-shadow: 0 0 0 1px var(--fd-color-action-hover, map-get($fd-colors-action-states, "hover"));
+      @include fd-var-color("border-color", fd-color("action", 1), --fd-button-border-color);
```
greg-a-smith: Why this button used box shadow for hover when default and emphasized didn't, beats me.  🤷‍♂  There are other buttons that do it too, but the entire button styling needs to be overhauled, simplified and brought up to Fiori 3 specs.

test/templates/input-group/index.njk
```diff
@@ -5,275 +5,133 @@
...
 {% block content %}
-<style media="screen">
-    .fd-input-group {
-        margin-bottom: 20px;
-    }
-</style>
 
-<h1>w/ text addon</h1>
-{% set example %}
+{% set types = ["text", "textarea"] %}
```
greg-a-smith: Changed most of this file to use for loops to generate multiple examples rather than all being hard-coded.

greg-a-smith:
@bcullman To your points:

1. I'm assuming you are referring to the blue focus ring on the playground pages.  If so, that's a browser thing and nothing we are concerned about.  On the docs pages, when you assemble the input groups with `fd-input`, `fd-textarea` and `fd-button` components, they have their own focus ring.  It's still not to Fiori 3 specs, but that's a whole nothing issue.
1. I'll work on the z-index adjustments for focus rings.
1. Regarding border radius values being different, I'm assuming this was from the playground page again.  This was actually intentional to show and prove that input-group's primary job is to remove corners from touching boundaries and let sub components handle their own.

greg-a-smith:
@jbadan @bcullman I think I've addressed all PR feedback.

1. I decided to leave the number input in the docs and playground as there was some input logic to get rid of rounded corners on stackable buttons.  Ultimately, I do think this should be a separate component in an implementation library, but for styling purposes, it can go here.  I did, however, move the suppression of the browser's type=number helper to a modifier class on fd-input rather than being specific to an input group.
1. Did some work to clean up focus rings on the docs site with real subcomponent classes.  The playground will (intentionally) not deal with those.
1. Standardized the border-radius values on the playground inputs.

Issue #289, opened 2019-08-22
New component: Notifications


InnaAtanasova:
Implement Notifications component

![Screen Shot 2019-08-22 at 3 26 34 PM](https://user-images.githubusercontent.com/39598672/63543629-3fea6880-c4f1-11e9-8306-d84b81cbd89e.png)

saad-mo:
can this component be used inline on the page? I need to a replacement component for what we have in the cloud portal as shown below. 

<img width="1105" alt="Screen Shot 2019-09-11 at 2 15 33 PM" src="https://user-images.githubusercontent.com/15215400/64727577-c665ea80-d49e-11e9-87fe-4ffa20d8ea30.png">

Issue #296, opened 2019-08-27
Edit playground node-sass to consume same postcss config as build pipeline


jbadan:
https://github.com/SAP/fundamental-styles/blob/master/test/app.js#L32

The testing playground is importing `scss` files directly from `scss` folder and transforming them into css using `node-sass`. The files are not being run through our postcss config found here: https://github.com/SAP/fundamental-styles/blob/master/config/postcss.config.js.

This causes a huge pain point when developing for IE - there is no way to rely on the autoprefixer. 

app.js should be edited to also use postcss config.

PR #314, opened 2019-09-05
feat: notifications

InnaAtanasova:
## Related Issue
Closes SAP/fundamental-styles#289

## Description
This is a new component. 
Notifications have 3 sizes: default, medium and small
The types of the notifications are: default, success, error, warning, and information.
There is a vertical stripe at the left hand side of the notification representing the Priority Indicator. An optional semantic icon can be added in the Title area.


docs/_data/versions.yml
```diff
@@ -190,6 +190,10 @@ multi-input:
   since: 1.1.0
   updated: 1.2.1
 
+notification:
+  since: 1.1.0
+  updated:
+
```
jbadan: I think we got rid of this file

jbadan:
@InnaAtanasova Can you merge master into your branch so it's updated with latest. Only other thing to do with a new component is to add an import to this file: https://github.com/SAP/fundamental-styles/blob/master/src/fundamental-styles.scss

docs/_data/versions.yml
droshev: done

scss/components/notification.scss
```diff
@@ -0,0 +1,311 @@
...
+
+
+    // ------------------------ HEADER ------------------------
+    &__header {
+        @include fd-reset();
+        display: flex;
+        box-shadow: inset 0 -0.0625rem var(--sapPageHeader_BorderColor, $sapPageHeader_BorderColor);
+    }
+
+    &__indicator {
```
jbadan: I think we need some sizing resets here:
Screen Shot 2019-09-11 at 3 14 23 PM

test/templates/notification/index.njk
```diff
@@ -0,0 +1,443 @@
...
+<div class="fd-notification fd-notification--success">
+    <div class="fd-notification__header">
+        <div class="fd-notification__indicator--success"></div>
+        <div class="fd-notification__title">This is the notifications title Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.</div>
+        <button class="btn-playground fd-notification__close"></button>
+    </div>
+    <div class="fd-notification__body">
+        <div class="fd-notification__content">
+            <div class="fd-notification__avatar">
+                <span>WW</span>
```
jbadan: If you put a placeholder class on this -- like one of the ones on `layout.css`, it won't throw you off on the unnormalize checkbox.

jbadan:
Other than those two minor things I'm on board! Looks great 🎉

test/templates/notification/index.njk
droshev: done

scss/components/notification.scss
droshev: done

docs/_data/components.yml
```diff
@@ -45,6 +45,12 @@
     - title: Modal
       meta: modal
 
+    - title: Multi Input
```
jbadan: Should multi input be here? I thought it moved ot patterns.

scss/_settings.scss
```diff
@@ -1,6 +1,8 @@
 //library namespace
 $fd-namespace: fd !default;
 
+@import "./../node_modules/@sap-theming/theming-base-content/content/Base/baseLib/sap_fiori_3/sass_variables.scss";
```
jbadan: are we taking the sass variables or the css variables? Should this also be passed to the postcss config? I think there's some infrastructure concerns here. 

I'm ok with this as long as someone has checked that the postcss color backups still work as expected.

docs/_data/components.yml
jbadan: Nevermind it should be!

scss/_settings.scss
droshev: i thought we agreed to pass both(css and scss variables), if this is not the case we can clean it up

jbadan: No that's fine, but did we check this works as expected with postcss?

You'll need to check the docs site as the playground isn't wired up to have the scss run through postcss, see: https://github.com/SAP/fundamental-styles/issues/296

scss/components/notification.scss
```diff
@@ -0,0 +1,312 @@
...
+    $fd-notification-indicator-color-success: var(--sapSuccessColor, $sapSuccessColor) !default;
+    $fd-notification-indicator-color-information: var(--sapInformationColor, $sapInformationColor) !default;
+
+    $fd-notification-indicator-border-color-error: var(--sapErrorBorderColor, $sapErrorBorderColor) !default;
+    $fd-notification-indicator-border-color-warning: var(--sapWarningBorderColor, $sapWarningBorderColor) !default;
+    $fd-notification-indicator-border-color-success: var(--sapSuccessBorderColor, $sapSuccessBorderColor) !default;
+    $fd-notification-indicator-border-color-information: var(--sapInformationBorderColor, $sapInformationBorderColor) !default;
+
+    $fd-notification-title-color: var(--sapGroup_TitleTextColor, $sapGroup_TitleTextColor) !default;
+    $fd-notification-content-text-color: var(--sapContent_LabelColor, $sapContent_LabelColor) !default;
```
jbadan: so I think if we setup postcss correctly, none of this is necessary. You would just need `var(--sapContent_LabelColor)` and that's it.

PR #329, opened 2019-09-16
chore: lint src and scss directories

markpalfreeman:
## Description
We need stylistic consistency across the fundamental repo, especially for new contributors, so all the code in the repository looks the same to assist readability and understanding.

I started down the path of modifying lint rules, but ultimately scrapped the existing config and just used `stylelint-config-standard` which bases its style lint rules on the most agreed-upon practices from popular libraries. There are only a couple "manual" overrides/changes which I'll describe inline in the PR.

I'm aware not all of these are "ideal" changes, but IMO making small sacrifices to align on _something_ is better than inconsistency.

### Notes
- This PR does **not** yet lint the `/docs` folder, although I think it should as a next round.
- I included a `.prettierrc.json` file—using a Prettier extension in your editor with auto-save is pretty glorious; it can acknowledge the `.stylelintrc` and auto-format the files for you ✨


scss/components/alert.scss
```diff
@@ -1,183 +1,225 @@
...
-      @include fd-var-color("border-color", $fd-alert-border-color--information, --fd-alert-border-color);
-      @include fd-icon("message-information", "m");
-      @include fd-rtl() {
-        @include fd-icon("message-information", "m", "after");
-      }
+
+    @include fd-var-color("background-color", $fd-alert-background-color--error, --fd-alert-background-color);
+    @include fd-var-color("border-color", $fd-alert-border-color--error, --fd-alert-border-color);
+    @include fd-icon("message-error", "m");
+
```
hertweckhr1: why is this space created?

markpalfreeman: I'm pretty sure it's because the next declaration has an open block `{ }`. I'm guessing it's a different _style_ of include, so it helps to visually separate it.
The equivalent in JS would be wanting a line break between these two:

```
const thing = 1;

function () {
  console.log(thing);
}
```

.travis.yml
```diff
@@ -12,7 +12,7 @@ script:
 jobs:
   include:
   - stage: Lint and Test
-    script: npm run lint && npm run docs:lint && npm run test:ci-config && npm run test
```
markpalfreeman: Removing this for now since the linting is more strict. I'll create a follow-up to lint the `docs/` folder too, but didn't want to increase the scope.

package.json
```diff
@@ -11,11 +11,9 @@
     "docs:config": "./scripts/create-library-config.js",
     "docs:prod": "npm run docs:compile && cd docs && bundle install && bundle exec jekyll build --config _config.yml,_config-library.yml,_config-prod.yml && cd ..",
     "docs": "npm run docs:compile && cd docs && bundle install && bundle exec jekyll build --config _config.yml,_config-library.yml && cd ..",
-    "docs:lint": "stylelint './docs/_sass/*.scss'",
```
markpalfreeman: Just deleting these entirely because I imagine the future will have a single lint script for the entire repo.

PR #331, opened 2019-09-17
fix: Menu item name not truncating

### LINKED ISSUE #317: Menu item name not truncating when name is too long
**Describe the bug**
The item name should truncate when it is too long in accordance with the spec.
Actual:
image

Menu component

> What version of the Fundamental Styles are you using?
0.1.0

Angular 7

**To Reproduce**
1. Open https://sap.github.io/fundamental-ngx/menu
2. Inspect element on any menu item name and change it to a long text
3. Observe that truncation does not happen.

**Expected behavior**
(spec is same as for title truncation in action bar):
image

**Additional context**
A bug had initially been raised in ngx repo as well https://github.com/SAP/fundamental-ngx/issues/1175

stefanoScalzo:
fixes #317 
fixes https://github.com/SAP/fundamental-ngx/issues/1175

## Description
This modification allows the text to truncate if the title is too long for the textbox.

### Before:
truncation_not_applied

### After:
Screen Shot 2019-09-17 at 11 45 26 AM


docs/_sass/_docs-display-component.scss
```diff
@@ -14,6 +14,7 @@
     transition: background-color .25s;
 
     > .fd-tile__content {
+        width: 100%;
```
jbadan: I think this will only fix the problem on the docs site, no? Shouldn't we fix it in the actual scss?

stefanoScalzo: Fixed!

jbadan:
@stefanoScalzo Great! You will also need to update the test you edited. Directions can be found here - https://github.com/SAP/fundamental-styles/wiki/Visual-Testing#automated-visual-testing 

TLDR: `npm run test`, take a screenshot of failed test + diff, `npm run test:update`, make sure you only commit changes that you meant to make to tests (ie ones that failed). Add, commit, push and you should pass the travis step.

docs/_sass/_docs-display-component.scss
```diff
@@ -14,6 +14,7 @@
     transition: background-color .25s;
 
     > .fd-tile__content {
+        // width: 100%;
```
hertweckhr1: ^ Don't forget to delete if it's no longer needed 👍

stefanoScalzo: removed thanks :)

Issue #336, opened 2019-09-18
Add dropdown modifier to Button component


InnaAtanasova:
> Is this issue related to a specific component?
Yes, Button Component

**Describe the solution you'd like**
<img width="1802" alt="Screen Shot 2019-09-18 at 4 47 37 PM" src="https://user-images.githubusercontent.com/39598672/65184741-2545d980-da34-11e9-9303-e54c588fa10e.png">
<img width="1775" alt="Screen Shot 2019-09-18 at 4 47 47 PM" src="https://user-images.githubusercontent.com/39598672/65184742-2545d980-da34-11e9-9b87-d91ad9771d6e.png">

In Fundamental-ngx, the dropdown component is an opinionated composition of the “fd-popover” and “fd-menu” components with the use of a styled button - `fd-dropdown__control`. We need the Dropdown component from Fundamental-styles just to style the button. Our Select component is also using this styled button. We can add the styling for the dropdown control button to the Button component and this way avoid importing the Dropdown component

PR #337, opened 2019-09-18
fix: adds fd-button--menu styling

InnaAtanasova:
## Related Issue
Closes SAP/fundamental-styles#336

## Description
added `--menu` modifier styling

## Screenshots
Screen Shot 2019-09-19 at 4 34 34 PM


scss/components/button.scss
```diff
@@ -447,4 +448,102 @@ a.#{$block} {
...
+        }
+      }
+      &:active,
+      &.is-active,
+      &[aria-expanded="true"],
+      &:focus {
+        box-shadow: 0 0 0 1px var(--fd-color-status-3, map-get($fd-colors-status, 3));
+      }
+    }
+  }
```
greg-a-smith: This seems like a lot of unnecessary styling to just add a down arrow icon for `fd-button--dropdown`.  This is just a modifier and _should_ be able to co-exist with other base button styling and other elements and modifiers.

test/templates/button/index.njk
```diff
@@ -94,4 +94,19 @@
...
+
+{% set example %}
+
+<button class="fd-button fd-button--dropdown">Dropdown</button>
+<br>
+<button class="fd-button fd-button--compact fd-button--dropdown">Dropdown</button>
+<br>
+<button class="fd-button fd-button--dropdown full-width">Dropdown</button>
+<br>
+<button class="fd-button fd-button--dropdown full-width error">Dropdown</button>
```
greg-a-smith: Don't "error" buttons just use the `fd-button--negative` class?

docs/pages/components/button.md
```diff
@@ -204,3 +204,35 @@ Button with multiple actions
...
+  <br>
+   <button class="fd-button fd-button--compact fd-button--dropdown">Dropdown</button>
+  <br>
+  <button class="fd-button fd-button--dropdown" disabled>Dropdown</button>
+  <br>
+  <button class="fd-button fd-button--dropdown full-width">Dropdown</button>
+  <br>
+  <button class="fd-button fd-button--dropdown full-width error">Dropdown</button>
+  <br>
+  <button class="fd-button fd-button--dropdown full-width error" disabled>Dropdown</button>
```
greg-a-smith: Don't "error" buttons just use the `fd-button--negative` class?

test/templates/button/index.njk
InnaAtanasova: Screen Shot 2019-09-19 at 1 53 14 PM
Screen Shot 2019-09-19 at 1 53 28 PM
Agreed, that was my original idea, but  `fd-button--negative` is a button with red text and red border, on hover it has red background and white color. In our case we only need a red border, so the amount of styling is less than overwriting the default behavior of `fd-button--negative`

PR #343, opened 2019-09-23
feat: remove icons and fonts from component styles

jbadan:
## Description
[ BREAKING CHANGE ]
- removed scss directory

This pr separates fonts and icons from the other components, the implementation libraries should now handle this logic.

Files moved using `git -mv` so history should be intact (besides with the stylelint pr as the merge got dicey)

Reasons for large test diffs:
1. buttons - there was something funky going on before with icons in buttons, see gif below: 

1. identifier: removed example testing `fd-has-background-color`, this should be in a helper test, it has nothing specifically to do with identifier.
1. others: components that didn't have fonts available to them before now do, therefore there were some changes in the reference images due to using font72 on the "playground placeholder" text


src/root.scss
```diff
@@ -1 +1 @@
-@import "../scss/core/root";
+@import "./core/root";
```
greg-a-smith: Should we just move the `/src/core/_root.scss` file to `/src` and remove the underscore?

test/templates/layout.njk
```diff
@@ -7,6 +7,8 @@
     <title>{{ id }}</title>
     <link rel="stylesheet" href="static/layout.css">
     <style>
+        {{ 'fonts.scss' | component_css }}
```
jbadan: @greg-a-smith should I load the fonts and icons here like you did in the example you showed me with the variables instead?

greg-a-smith: I think it will need to be like this for now because these are `*.scss` files and need to be converted to CSS first.  The other hard-coded `<link>` tags are just loading files that are already CSS.  If we ever refactor the playground to use generated CSS files, then it could just be dropped in as a `<link>` tag.

src/shellbar.scss
```diff
@@ -1,2 +1,234 @@
-@import "../scss/components/shellbar";
-@import "../scss/fonts/72";
+@import "./settings";
+@import "./mixins";
+@import "./functions";
+@import "./shellbar/product-menu";
+@import "./shellbar/product-switcher";
+@import "./shellbar/search-input";
+
```
bcullman: should shellbar's subcomponents just be promoted to top level? the datepicker and timepicker are (or should be) is just a wrappers around other subcomponents as well.

src/layout.scss
```diff
@@ -1,5 +1,5 @@
-@import "../scss/layout/container";
-@import "../scss/layout/col";
-@import "../scss/layout/ui";
-@import "../scss/layout/shell";
-@import "../scss/layout/app";
+@import "./layout/container";
+@import "./layout/col";
+@import "./layout/ui";
+@import "./layout/shell";
+@import "./layout/app";
```
bcullman: should these layout ones be grouped into a scss/css that can deliver them?

jbadan: That is the purpose of this file (if I understand the question)

src/shellbar.scss
jbadan: We can, either way is fine with me 🤷‍♀ 

Datepicker and timepicker don't even have files since there's no styles associated with them

PR #347, opened 2019-09-26
fix: Added Select and List components

### LINKED ISSUE #277: Dropdown should be refactored to Select
The current `Dropdown` component (`fd-dropdown`) is described as being more flexible than a normal `<select>`.  This component is intended to be a non-native select component, but it has been written using buttons and is very tightly coupled with button styling.  This should be refactored into a new `Select` component and styled as such.

See UI5 Web Components: https://sap.github.io/ui5-webcomponents/playground/components/Select/
See Global Design Wiki: https://wiki.wdf.sap.corp/wiki/pages/viewpage.action?pageId=2051306683

greg-a-smith:
## Related Issue
Closes #277 

## Description
The existing **Dropdown** component was not implemented or styled correctly.  It was essentially trying to be both a **Select** and a **MenuButton**, but wasn't doing either of them correctly.

To better align with SAP Global Design, I created a new **Select** component and a new **List** component.  The **List** component can be used as a standalone component, but also serves as the popover content for the **Select** component.

### Notes/Disclaimers
1. Not all **List** features were implemented.  The remaining features can be added as future enhancements.  Things like "Growing List", "Selection", "Standard List with Byline" and "Wrapping Behavior" were not yet implemented.  See https://wiki.wdf.sap.corp/wiki/pages/viewpage.action?pageId=2036843636 for more details.
1. The **List** component uses _ONLY_ SAP Fiori 3 CSS variables. 💯 
1. The **Select** component is still using a number of fundamental-based mixins, functions and styles so it doesn't match the design specs 100%.
1. The PostCSS config has been altered to now include the Fiori 3 CSS variables for fallbacks and custom properties.

### Questions
1. Should we completely remove the **Dropdown** (`fd-dropdown`) component and make this a breaking change?  Or do we leave it and mark it as deprecated?

   > I think **Dropdown** should stay for now until such time as the **MenuButton** component is added.

1. Should we completely remove the **List Group** (`fd-list-group`) component and make this a breaking change?  Or do we leave it and mark it as deprecated?

   > I'm not really sure _what_ **List Group** is supposed to be.  Is it just a mis-named **List**?  Is it something else entirely?  There is nothing in the global design docs called "List Group".

## Screenshots
None of the existing tests were modified.  Only new test references were created for the new **Select** and **List** components.


config/postcss.config.js
```diff
@@ -24,7 +24,8 @@ module.exports = {
     sourcesContent: true,
     plugins: [
         postcssImport(),
-        postcssAddFallback({importFrom: 'dist/root.css'}), //replace with theme-designer variables
+        postcssAddFallback({importFrom: 'dist/root.css'}),
+        postcssAddFallback({importFrom: 'node_modules/@sap-theming/theming-base-content/content/Base/baseLib/sap_fiori_3/css_variables.css'}),
```
greg-a-smith: The `importFrom` does not accept an array of files so we just needed to call this plugin twice.

config/postcss.config.js
```diff
@@ -33,8 +34,7 @@ module.exports = {
             level: 1
         }),
         postcssCustomProperties({
-            preserve: true,
-            importFrom: 'dist/root.css' //replace with theme-designer variables
```
greg-a-smith: There is no need to import the file again.  The fallback values for the CSS variables (from the `postcssAddFallback` plugin) will be used to generate the custom properties.

docs/_sass/_docs-general.scss
```diff
@@ -201,7 +201,7 @@ pre {
 }
 
 .docs-intro {
-    font-size: 1.42857rem;
+    font-size: 1.1rem;
```
greg-a-smith: The intro text was just too big in relation to everything else.

test/templates/layout.njk
```diff
@@ -6,6 +6,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <title>{{ id }}</title>
     <link rel="stylesheet" href="static/layout.css">
+    <link rel="stylesheet" href="sap-theming/css_variables.css">
```
greg-a-smith: Until we have PostCSS working in the playground, this will load the Fiori 3 CSS variables.

src/list.scss
```diff
@@ -0,0 +1,159 @@
...
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
+
+  &__title {
+    flex: 3 3 10%;
+    font-size: var(--sapFontLargeSize);
+  }
+
+  &__secondary {
```
markpalfreeman: Question: do we want to call this `status` to align with global design? `fd-list__secondary` makes me think of a sub-list or something... but if you think it needs to be generic to handle other things, no problem.

https://wiki.wdf.sap.corp/wiki/display/visualcore/List+%28Fiori+3%29+-+Standard+List#List(Fiori3)-StandardList-Statusincluded

src/select.scss
```diff
@@ -0,0 +1,81 @@
...
+
+    @include fd-icon-glyph("slim-arrow-down","after") {
+      transition: transform $fd-animation--speed ease-in;
+    }
+
+    position: relative;
+    padding-right: $fd-forms-height + $fd-forms-padding - 2;
+
+    &::after {
+      @include fd-var-color("color", $fd-color--action, --fd-color-action-1);
```
jbadan: We don't need to use the `fd-var-color` mixins anymore - you can put `var(--fd-color-action-1)` and the postcss plugins will take care of the rest :)

markpalfreeman:
Re #2 this is already pretty well encapsulated, so I'm all for deleting `List Group` but maybe as a second PR?

src/list.scss
greg-a-smith: I thought about that as well, however, I wanted to try and keep this more generic (if possible) as there are other examples of this second column of data.

See https://wiki.wdf.sap.corp/wiki/display/visualcore/List+%28Fiori+3%29+-+Item+Counter

greg-a-smith:
@markpalfreeman I think I'd rather get a read on _what_ **List Group** is (or was supposed to be) before we 🔥 it.

greg-a-smith:
I removed the use of `fd-var-*` from the `select.scss` and then had to bring in `root.scss` to the playground so those variables were present.  This caused a few test failures, however, that was a _GOOD_ thing since the test references were wrong.  See screenshot:

Screen Shot 2019-09-26 at 10 49 56 AM

greg-a-smith:
I also added a deprecation message on the **Dropdown** and **List Group** pages.

PR #359, opened 2019-10-02
[NOT-FOR-MERGE] localization-editor, input-group

JKMarkowski:
## Description
I added some fixes, due to latest self contained styles adapt. there was still some visual bugs/weird behaviours.

### Before:

Localization Editor:
image

### After:

Localization Editor:
image


src/input-group.scss
```diff
@@ -49,8 +49,6 @@ $block: #{$fd-namespace}-input-group;
   .#{$block}__input {
     @include fd-reset-child-spacing();
 
-    flex: 1 0 auto;
```
jbadan: This was necessary for the implementation libraries (definitely fundamental-react), can you link this change and show it doesn't affect anything?

src/localization-editor.scss
```diff
@@ -10,4 +10,11 @@ $block: #{$fd-namespace}-localization-editor;
 
     padding: fd-space(2);
   }
+
+  // Elements
+  &__list {
+    li {
```
jbadan: Due to the nature of self contained styles, we can't use element selectors to prevent bleeding. This either needs a different selector or new class.

src/input-group.scss
greg-a-smith: If you remove this declaration, what are the `flex-grow`, `flex-shrink` and `flex-basis` values?

src/localization-editor.scss
JKMarkowski: I will add some class then.

JKMarkowski: It's done by other PR.

src/input-group.scss
JKMarkowski: As I could notice setting these properties makes the `__input` to take whole width of his parent, not matter if there is any sibling. (it's changed there by `width:auto`, but it's still not correct)
image
I dont know why you need it there, because by default, without `flex:...` it's gonna take whole width of parent minus the size of addon.

bcullman:
unsure about a lot of the regression image updates. can someone provide a reason why these were altered here?

InnaAtanasova:
Please refer to this 
[PR](https://github.com/SAP/fundamental-styles/pull/368)

PR #361, opened 2019-10-03
fix: refactor alert component to use sap-theming and align with global design

InnaAtanasova:
- adds theme designer variables for colors 
- updates paddings and spacing according to Fiori3 guidelines
- removes extra code


src/_settings.scss
```diff
@@ -4,6 +4,8 @@
 // library namespace
 $fd-namespace: fd !default;
 
+@import "./../node_modules/@sap-theming/theming-base-content/content/Base/baseLib/sap_fiori_3/sass_variables.scss";
```
jbadan: We don't need the scss variables. The foundation should be in place to simply use the css variables.

src/alert.scss
```diff
@@ -10,41 +10,64 @@
...
+  @mixin fd-alert-close-btn-container {
+    position: absolute;
+    width: 2rem;
+    height: 1.625rem;
+    top: 0.375rem;
+    right: 0.125rem;
+  }
+
+  // SUCCESS
+  $fd-alert-background-color--success: var(--sapSuccessBackground, $sapSuccessBackground) !default;
```
jbadan: These can just be `var(--sapSuccessBackground)`, postcss will do the rest

src/_settings.scss
InnaAtanasova: Removed the import

src/alert.scss
InnaAtanasova: Removed the sass variables

jbadan:
@InnaAtanasova I know it's actually a feature, but for versioning we should label it a fix since it's not breaking.

Issue #363, opened 2019-10-04
Fix menu to have icon within anchor


stefanoScalzo:
Before opening:
- [Search for duplicate or closed issues](https://github.com/SAP/fundamental-styles/issues?utf8=✓&q=is%3Aissue)
- Read the [contributing guidelines](https://github.com/SAP/fundamental-styles/wiki/Contribution-Guidelines)

Feature requests must include:
Currently the hover will go on either the icon or on the anchor item but not both!

**Is your feature request related to a problem? Please describe.**
The hover does not change the whole background color but only the anchor OR icon
![Screen Shot 2019-10-04 at 11 50 24 AM](https://user-images.githubusercontent.com/50607147/66221525-2b2af400-e69d-11e9-820a-71be0ed17706.png)

> Is this issue related to a specific component?
Menu component
> What browser are you working in?
Google chrome, safari
> What version of Fundamental Styles are you using?
0.3.0-rc.6"
> What offering/product do you work on? Any pressing ship or release dates we should be aware of?
Fundamental-Styles

> What front-end framework are you implementing in, e.g., Angular, React, etc.?
Angular

**Describe the solution you'd like**
Allow both a focus and hover to be implemented on the whole list item

PR #364, opened 2019-10-04
feat: menu hover

stefanoScalzo:
## Related Issue
Closes SAP/fundamental-styles#363

## Description
Added the span icon element within the anchor tag
Screen Shot 2019-10-04 at 11 52 34 AM

### After:

Screen Shot 2019-10-04 at 11 53 35 AM


scss/components/action-bar.scss
```diff
@@ -0,0 +1,130 @@
+@import "./../settings";
```
jbadan: These files now live in `src` directory

stefanoScalzo: done

droshev:
@jbadan we discussed this with @stefanoScalzo as this is a breaking change.
@stefanoScalzo can you update the description of the PR stating that this is breaking change?
Also we should open issue in each library referencing this PR as a source.

PR #368, opened 2019-10-08
fix: localization-editor, input-group

InnaAtanasova:
This PR fixes the same issue addressed by @JKMarkowski in his
[PR](https://github.com/SAP/fundamental-styles/pull/359)
For some reason in his branch a couple of tests are failing and I wasn't able to fix it. 
I created a new branch from the latest master, applied the changes and all tests now pass.

Before the fix:
Screen Shot 2019-10-08 at 11 35 01 AM
Screen Shot 2019-10-08 at 11 34 54 AM

After:
Screen Shot 2019-10-08 at 11 35 14 AM
Screen Shot 2019-10-08 at 11 35 09 AM

Locally running the tests:
Screen Shot 2019-10-08 at 12 10 26 PM

### DIFF
--- a/src/input-group.scss
+++ b/src/input-group.scss
@@ -49,8 +49,7 @@ $block: #{$fd-namespace}-input-group;
   .#{$block}__input {
     @include fd-reset-child-spacing();
 
-    flex: 1 0 auto;
-    width: auto;
+    width: 100%;
     z-index: map-get($fd-z-index-levels, "first");
   }
 


bcullman:
Shouldn't there be some sort of visual tests that we affected by this? Certainly not *all* of them, but a few.

InnaAtanasova:
> Shouldn't there be some sort of visual tests that we affected by this? Certainly not _all_ of them, but a few.

There's only one png file for localization editor test and it doesn't "test" the component when the menu is open. The current change can't be detected in this case and that's why the tests are not affected. 
Screen Shot 2019-10-08 at 1 52 24 PM

Issue #207, opened 2019-07-24
Bug: Side Navigation styling inconsistency with Fiori 3


InnaAtanasova:
**Describe the bug**

> Is this issue related to a specific component?
Side Navigation

**Expected behavior**
- Fundamental-styles doesn't have Utility Section (bottom aligned)
- The collapsed state of the elements with two levels has the wrong type of arrow. 

![Screen Shot 2019-07-24 at 12 11 32 PM](https://user-images.githubusercontent.com/39598672/61809920-385d7400-ae0c-11e9-8e11-e62e8559226c.png)

PR #380, opened 2019-10-21
feat: refactor side navigation

InnaAtanasova:
## Related Issue
Closes SAP/fundamental-styles#207

## Description
- added a new (INTERNAL for the lib) reusable component called Nested list which will allow multiple levels of nesting (up to 10)
- use the Nested List component in Side Navigation component 
- refactor the Side Nav component to match the latest specs
- The Side nav now has 2 sections - Main Section and Utility Section. Each of these sections has a Nested List for displaying the navigation items. 

## BREAKING CHANGES
- Removed classes: fd-side-nav__list, fd-side-nav__item, fd-side-nav__link, fd-side-nav__group, fd-side-nav__title, fd-side-nav__sublist, fd-side-nav__subitem, fd-side-nav__sublink, fd-side-nav__icon, fd-side-nav--icons


bcullman:
Should this even be needed?

With proper CSS, *any* `li` that is a direct child of a `li ul` can be set to indented - no need for an `indented` class or a collection of classes to name what level deep your are

bcullman:
note that plainoldhtml indents children

```
<html>
<body>
    <ul>
        <li>1</li>
        <li>2</li>
        <li>3
            <ul>
                <li>a</li>
                <li>b</li>
                <li>c
                    <ul>
                        <li>i</li>
                        <li>ii</li>
                        <li>iii</li>
                    </ul>
                </li>
            </ul>
        </li>
    </ul>
</body>
</html>
```

src/nested-list.scss
```diff
@@ -0,0 +1,295 @@
...
+  font-style: normal;
+  font-weight: normal;
+  color: var(--sapList_TextColor);
+  font-size: var(--sapFontHeader4Size);
+  text-decoration: inherit;
+  text-transform: none;
+  speak: none;
+}
+
+@mixin fd-flex-center {
```
markpalfreeman: This would probably better live in a shared utilities folder, as it's not particular to this file.

src/nested-list.scss
```diff
@@ -0,0 +1,295 @@
...
+  list-style: none;
+  width: 100%;
+
+  @for $i from 2 through 10 {
+    .level-#{$i} {
+      $paddingLeft: ($i - 2) + 2.75rem;
+
+      .#{$block}__link,
+      .#{$block}__content {
+        padding-left: $paddingLeft;
```
markpalfreeman: As @bcullman mentioned previously, I don't think we want to be determining the "level" of nesting and calculating padding for each... wouldn't it be preferred to instead put padding on each "parent" `ul` (with some modifier like `.fd-list--expandable`) and let the children indent themselves (recursively)? It would eliminate the need for this custom math.

As a visible example from UI5 wiki:
Screen Shot 2019-10-22 at 9 31 03 AM

https://wiki.wdf.sap.corp/wiki/pages/viewpage.action?pageId=2075534845

InnaAtanasova: Yes, in most of the cases this would be the approach, but not in this one. If you add padding to the ul element then the `selected` styling of the subitems (children) wouldn't be correct. The background color and the underline border should go to the ends of the Navigation list.
Here is how it looks now:
Screen Shot 2019-10-22 at 1 25 50 PM

Here is the case when we add padding to the `ul` element and don't do the calculations
Screen Shot 2019-10-22 at 1 29 20 PM

InnaAtanasova:
> Should this even be needed?
> 
> With proper CSS, _any_ `li` that is a direct child of a `li ul` can be set to indented - no need for an `indented` class or a collection of classes to name what level deep your are

@bcullman , applying padding to the `li` adds indentation but doesn't have the correct styling for `hover`, `is-selected` and `focus` state. 
Screen Shot 2019-10-22 at 1 36 39 PM

droshev:
> to only create fundamental-styles "components" that match to UI5 controls? If so, I'm not sure we need a `NestedList` as a separate one unless we're actuall

It is not only needed for the side navigation and we are not first creating that

InnaAtanasova:
> @greg-a-smith how does this jive with the work you did on standard `List`?
> 
> Are we trying to only create fundamental-styles "components" that match to UI5 controls? If so, I'm not sure we need a `NestedList` as a separate one unless we're actually planning to reuse it somewhere else besides `SideNavigation`.

@markpalfreeman , my approach at the beginning was to refactor the List component and achieve the nesting there. But it required too many changes. For starters, the current implementation uses `flex` for situating the icons and text. This causes problems when you add a `ul` inside the `fd-list__item`. 
Second, you can "mix" items with and without icons, and the items without icons should have the same text alignment as the items with icons. 
Third, the "after" icon (the arrow) should be dynamic - it should have different states when the child list is expanded and not-expanded. This can't be achieve with CSS and the current implementation of the List
Screen Shot 2019-10-22 at 2 04 15 PM

src/nested-list.scss
markpalfreeman: Makes sense, I didn't see the hover/active states previously.

markpalfreeman: I still don't know that we want the consumers of this to have to keep track of the "level" of indentation in their markup to supply the correct class.

Seems like using parent/child relationship selectors allows consumers to just apply classes more generically, and the CSS is smart enough to style itself, like:
```css
.list
    .list__item
    .list__item
        .list__item
        .list
            .list__item
            .list__item
        .list__item
    .list__item
```

droshev:
> If we go forward with `Nested List`, _I think_ tit'd be the first time this library would be exporting a non-Fiori3 component and referencing it in public docs/navigation.
> 
> If this is something that needs to be shared in multiple places, my preference would be to use it as a sass mixin that gets brought into the components that need it, but the class names stay aligned to Fiori3 components so we have a 1:1 match.
> 
> Probably worth deciding as a team, but to me it gets confusing if we start creating "non-standard" components here—how do we keep track of which ones are Fiori3 and which ones are "helpers"?

it is a fiori 3 component

src/nested-list.scss
InnaAtanasova: Done

src/_nested-list.scss
```diff
@@ -0,0 +1,307 @@
+@import "./settings";
+@import "./mixins";
+@import "./functions";
+
+$block: #{$fd-namespace}-nested-list;
+
+@mixin fd-side-nav-arrow {
```
markpalfreeman: nit: since this is `nested-list.scss`, this is probably better named `fd-nested-list-arrow` to not couple to the "side-nav" concept

src/_nested-list.scss
```diff
@@ -0,0 +1,307 @@
...
+          @include fd-selected() {
+            border-bottom: var(--sapList_BorderWidth) solid;
+            border-bottom-color: var(--sapList_SelectionBorderColor);
+          }
+        }
+      }
+    }
+  }
+
+  &[aria-hidden="true"] {
```
markpalfreeman: I think we want an additional `.is-hidden` class to not fully rely on `aria` attributes for styles?

src/_nested-list.scss
```diff
@@ -0,0 +1,307 @@
...
+  speak: none;
+}
+
+.#{$block} {
+  @include fd-reset();
+
+  list-style: none;
+  width: 100%;
+
+  @for $i from 2 through 10 {
```
markpalfreeman: I'm not sure if my previous comment got lost or if I forgot to post it, but do we want consuming libraries to have to calculate `.level-#` for classes?

I love that this is a mixin, but it seems like a preferred hierarchy would just be something like this, and the styles would be ready to support all tiers of parent/child relationship (and padding size) up to a certain level.

Thoughts from the rest of the team? (pseudo-code)

```
.list
    .list__item
    .list__item
        .list
            .list__item
            .list__item
       ...
```

markpalfreeman: Brad did rightly mention, though, that the tradeoff for this is making really deep parent-child selectors in the CSS 😕 

`.list > .list__item > list > .list__item {}` etc

I'm not sure which is best

src/_nested-list.scss
jbadan: so I think we never fully decided how we want to handle all the aria-selectors. We've got it implemented many ways across components with no real standard. We should figure this out in general, but maybe not here.

src/_nested-list.scss
droshev: @markpalfreeman what would you suggest?

markpalfreeman: We can probably just leave it like this; both ways have tradeoffs.

InnaAtanasova:
> Appears that a few things aren't self-contained for styling yet (see screenshot).
> Screen Shot 2019-10-25 at 2 41 41 PM

Hi @greg-a-smith ,
I added a reset for the TEST list items. 

Screen Shot 2019-10-28 at 2 01 09 PM
Screen Shot 2019-10-28 at 2 01 02 PM

test/resources/layout.css
```diff
@@ -174,3 +174,28 @@ body {
...
+}
+
+.test-sidenav-list__item{
+  box-sizing: border-box;
+  font-size: 14px;
+  padding: 0;
+  margin: 0;
+  border: 0;
+  list-style: none;
+}
```
greg-a-smith: These shouldn't be necessary, should they?  The playground is supposed to demonstrate the component and its styles.  The "test" classes are for extraneous components included within another component (_e.g._ buttons in an action bar or modal, icons in a panel, etc.)

InnaAtanasova: The Side Nav component is:
- setting the width of the component in expanded state and in condensed state
- adding 2 sections - Main and Utility 

What is inside the 2 sections is the Nested List which is "external" to Side Nav and that's why I use 
`<ul class="test-sidenav-list">
        <li class="test-sidenav-list__item">Item 1</li>
        <li class="test-sidenav-list__item">Item 1</li>
        <li class="test-sidenav-list__item">Item 1</li>
        <li class="test-sidenav-list__item">Item 1</li>
 </ul>`
just because I need "content" to test the Side Nav.

I am struggling to understand your comment above ("Appears that a few things aren't self-contained for styling yet (see screenshot).")

test/templates/side-nav/index.njk
```diff
@@ -1,386 +1,783 @@
...
-                  { "label": "item link" },
-                  { "label": "item link" }
-              ]
-            }
-          ]
-        }
-    )}}
-    {% endset %}
-    {{ format(example) }}
+<h2>Nested List Without Icons</h2>
```
greg-a-smith: All these nested list examples should be in their own page:  `test/templates/nested-list/index.njk`.  This file should only have the examples that show the `fd-side-nav*` classes.

Issue #385, opened 2019-10-23
self contained button component


stefanoScalzo:
Feature requests must include:
- As much detail as possible for what we should add and why it's important
- Relevant links to prior art, screenshots, or live demos whenever possible
-->

**Is your feature request related to a problem? Please describe.**

> Is this issue related to a specific component?
Must change all fd-colors to self contained button styles. The Medium element no longer exists, fd-button--standard is black and white, fd-button looks like fd-button--standard, fd-button--emphasized  fd-button--negative/positive have no hover colers and base background
> What browser are you working in?
Google chrome
> What version of Fundamental Styles are you using?
0.3.0

**Describe the solution you'd like**
Make styles self contained

PR #386, opened 2019-10-23
feat: add visual tests

stefanoScalzo:
## Related Issue
Closes SAP/fundamental-styles#385

## Description
-Refactored all of the button to remove anything using maps and fd-colors and replaced to corresponding colors. 
-The Medium was removed, and the regular fd-button now has the same styles as fd-button--standard. There is no grey option. 
-For fd-button--emphasized, no hover or background color given by designer for negative and positive, thus applying only bold for text and keeping the same background colors.
##Breaking changes: 
Medium class is removed

### Before:
Screen Shot 2019-10-23 at 10 21 58 AM
### After:
Screen Shot 2019-10-23 at 10 22 23 AM
### Before:
Screen Shot 2019-10-23 at 10 07 19 AM
### After:
Screen Shot 2019-10-23 at 10 03 51 AM
### Before:
Screen Shot 2019-10-23 at 10 05 06 AM
### After:
Screen Shot 2019-10-23 at 10 22 18 AM
### Before:
Screen Shot 2019-10-23 at 10 22 12 AM
### After:
Screen Shot 2019-10-23 at 10 04 07 AM
### Before:
Screen Shot 2019-10-23 at 10 23 23 AM
### After:
Screen Shot 2019-10-23 at 10 04 43 AM


bcullman:
@stefanoScalzo: i assume we are seeing a shift on standard button from grey to blue because we are aligning with the colors provided by the theme vars - can you confirm?

this one looks wrong:
image

stefanoScalzo:
> @stefanoScalzo: i assume we are seeing a shift on standard button from grey to blue because we are aligning with the colors provided by the theme vars - can you confirm?
> 
> this one looks wrong:
> image

Confirmed for the colors. Actually that one shows fd-button--light fd-button--negative for disabled but ill fix the screenshot

bcullman:
@stefanoScalzo: the focus state appears to be missing - tabbing though the buttons they should get some sort of visual feedback.

in ui5 webcomponents, this appears to be some sort of inset dashed or dotted line.

https://sap.github.io/ui5-webcomponents/playground/components/Button/

image

the UI5 fiori spec spec itself seems to be lacking in describing focus style.

droshev:
The focus parameters are listed here: https://github.com/SAP/theming-base-content/blob/1bcf23f0e592b71e26372f33af6c45c77cef063b/content/Base/baseLib/sap_fiori_3/css_variables.css#L93

bcullman:
@stefanoScalzo this is much better. 

However, I would suggest that specifically for the `emphasized` style, an alternate color should be used for the focus line, it's almost impossible to see the line on the dark blue button background.

suggest setting to white.

stefanoScalzo:
> @stefanoScalzo this is much better.
> 
> However, I would suggest that specifically for the `emphasized` style, an alternate color should be used for the focus line, it's almost impossible to see the line on the dark blue button background.
> 
> suggest setting to white.

fixed

bcullman:
@stefanoScalzo its happening now onpress, but still not while tabbing onto the button.

bcullman:
@stefanoScalz - I was looking at:
https://deploy-preview-386--fundamental-styles.netlify.com/components/button.html
which yesterday did not show this, but checking today, it looks good.

🚢

droshev:
@stefanoScalzo can you list all the classes that are being removed

PR #387, opened 2019-10-23
fix: tile theme parameter refactoring

stefanoScalzo:
Description
Refactored all of the tile css file to remove any fd-colors and apply the appropriate theming
No changes done to actual visuals
No classes removed

### DIFF
--- a/src/mixins/_mixins.scss
+++ b/src/mixins/_mixins.scss
@@ -296,7 +296,7 @@
   &:focus {
     outline-style: dotted;
     outline-width: 1px;
-    outline-color: var(--sapContent_FocusColor);
+    outline-color: var(--sapLinkColor);
   }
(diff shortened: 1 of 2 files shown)


src/mixins/_mixins.scss
```diff
@@ -267,59 +267,36 @@
...
   &:visited {
-    @include fd-var-color("color", map-get($fd-colors-action-states, "visited"), --fd-color-action-visited);
-    @include fd-var-color(
-      "border-bottom-color",
-      map-get($fd-colors-action-states, "visited"),
-      --fd-color-action-visited
-    );
-  }
-
-  @include fd-active-pressed-selected() {
```
jbadan: should we be getting rid of aria-selected?

PR #398, opened 2019-10-28
fix: migrate breadcrumbs to use theming parameters

droshev:
fixes: #397

PR #401, opened 2019-10-28
fix: migrate breadcrums to use theming parameters

droshev:
fixes #398


droshev:
I lost the previous PR somehow...

Issue #404, opened 2019-10-30
Input placeholder text not vertically centered in Firefox browser


KevinOkamoto:
**Describe the bug**

In a Firefox browser (v69.0.3), the input field placeholder text is not vertically centered.

> Is this issue related to a specific component?

This bug affects text input elements.

> What version of the Fundamental Styles are you using?

Fundamental Styles v0.3.0

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

I'm working on Fundamental NGX - Platform.

> What front-end framework are you implementing in, e.g., Angular, React, etc.?

Angular.

**To Reproduce**
1. In a Firefox browser
2. Go to 'https://sap.github.io/fundamental-styles/components/multi-input.html'
3. You'll see that the place holder text is not centered vertically

**Expected behavior**
The placeholder text should be vertically centered.

**Screenshots**

<img width="1006" alt="Screen Shot 2019-10-30 at 10 52 59 AM" src="https://user-images.githubusercontent.com/48103491/67884666-77821c00-fb03-11e9-990d-a0e462ff8b85.png">

**Desktop (please complete the following information):**
 - OS: macOS Mojave
 - Browser: Firefox 
 - Version: 69.0.3

PR #406, opened 2019-10-30
fix(input): Fixing vertical alignment of placeholder text in input field

KevinOkamoto:
Fixing `height` and `line-height` values for input so that the placeholder text is vertically centered for Firefox browsers.

## Related Issue
Closes SAP/fundamental-styles#404

## Description
Added CSS rules to set `height` and `line-height` to the same values for input field for both "default" and "compact" modes.

Updated CSS rules to use the SAP theming variables.

Added missing SAP theming variables to test/playground "ugly" theme.

## Screenshots

Using Firefox (69.0.3):

### Before:
Screen Shot 2019-10-30 at 3 14 44 PM
Screen Shot 2019-10-30 at 3 14 54 PM

### After:
Screen Shot 2019-10-30 at 3 14 04 PM
Screen Shot 2019-10-30 at 3 14 13 PM

### DIFF
--- a/src/input.scss
+++ b/src/input.scss
@@ -15,11 +15,17 @@ $block: #{$fd-namespace}-input;
   width: 100%;
 
   &::placeholder {
+    @include fd-reset();
+    @include fd-var-size("line-height", $fd-forms-height, --fd-forms-height);
     @include fd-var-color("color", $fd-forms-color--placeholder, --fd-color-neutral-4);
   }
 
   &--compact {
     @include fd-var-size("height", $fd-forms-height--compact, --fd-forms-height-compact);
+    @include fd-var-size("line-height", $fd-forms-height--compact, --fd-forms-height-compact);
+  }
+
+  &--compact::placeholder {
     @include fd-var-size("line-height", $fd-forms-height--compact, --fd-forms-height-compact);
   }
 
--- a/src/input.scss
+++ b/src/input.scss
@@ -10,16 +10,17 @@ $block: #{$fd-namespace}-input;
 .#{$block} {
   @include fd-form-text();
   @include fd-var-size("height", $fd-forms-height, --fd-forms-height);
+  @include fd-var-size("line-height", $fd-forms-height, --fd-forms-height);
 
   width: 100%;
 
   &::placeholder {
-    @include fd-reset();
     @include fd-var-color("color", $fd-forms-color--placeholder, --fd-color-neutral-4);
   }
 
   &--compact {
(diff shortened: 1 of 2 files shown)


src/input.scss
```diff
@@ -10,16 +10,17 @@ $block: #{$fd-namespace}-input;
 .#{$block} {
   @include fd-form-text();
   @include fd-var-size("height", $fd-forms-height, --fd-forms-height);
+  @include fd-var-size("line-height", $fd-forms-height, --fd-forms-height);
 
   width: 100%;
 
   &::placeholder {
-    @include fd-reset();
```
jbadan: This should probably not be removed, just adding what you needed to override.

src/input.scss
```diff
@@ -10,16 +10,17 @@ $block: #{$fd-namespace}-input;
 .#{$block} {
   @include fd-form-text();
   @include fd-var-size("height", $fd-forms-height, --fd-forms-height);
+  @include fd-var-size("line-height", $fd-forms-height, --fd-forms-height);
```
droshev: Since we are touching the component maybe we should migrate it to use the parameters from the theming package

src/input.scss
KevinOkamoto: OK. I've reimplemented the `fd-reset()` and have added overrides to `line-height` in subsequent CSS rules.

KevinOkamoto:
> @KevinOkamoto do you think you can help migrating the component to use the theming parameters instead of the variables from fundamental

@droshev I'm unfamiliar with the term "migrating the component". Is this a new plan to depricate the use of `$fd-*` variables in fundamental-styles?

KevinOkamoto:
@droshev I've change the CSS variables the the SAP theming variables.

src/input.scss
```diff
@@ -9,17 +9,24 @@ $block: #{$fd-namespace}-input;
 
 .#{$block} {
   @include fd-form-text();
-  @include fd-var-size("height", $fd-forms-height, --fd-forms-height);
+  @include fd-var-size("height", $fd-forms-height, --sapElement_Height);
```
droshev: once we consume the theming parameters, we don't have to call the mixin. The code would look:
```
height: var(--sapElement_Height);
line-height: var(--sapElement_Height);
```

jbadan:
@KevinOkamoto Can you update tests so we can merge this? Thanks!

InnaAtanasova:
This is already fixed in the Forms refactoring PR
Screen Shot 2019-11-14 at 10 46 56 AM

PR #412, opened 2019-11-03
fix: implement busy indicator

### LINKED ISSUE #411: add busy indicator
implement busy indicator

droshev:
## Related Issue
closes #411 

Introducing the busy indicator:
- 3 sizes (default, medium, large)
- normal and contract modes as part of Quartz Light

Playground:
- introduce an animation toggle for the automated tests

Screen Shot 2019-11-03 at 4 16 04 PM
Screen Shot 2019-11-03 at 4 16 16 PM


droshev:
I see issues with the reference images as the busy indicator is changing it state all the time.

jbadan:
@droshev if these tests cause problems with the animations, you can always ignore them here like Spinner: https://github.com/SAP/fundamental-styles/blob/master/test/visual-regression-tests/config/backstopConfig.js#L29

droshev:
@jbadan i thought that it would be good to test even the current state without animation.
That's why I added a check box to toggle the animation - so the automated test case check this initial state
and we can still see the animation when we manually test

src/busy-indicator.scss
```diff
@@ -0,0 +1,105 @@
...
+  position: relative;
+  width: $fd-dots-size;
+  height: $fd-dots-size;
+  margin: $fd-dots-margin;
+  display: inline-block;
+  border-radius: 50%;
+  background-color: currentColor;
+}
+
+.#{$block}--circle-0 {
```
jbadan: Is there a way to do this with all sharing the same class name?

droshev: @jbadan  the animation delay is different and it is good to align

Issue #415, opened 2019-11-05
Contextual menu styling is broken


InnaAtanasova:
**Describe the bug**

> Is this issue related to a specific component?
Contextual Menu
> What version of the Fundamental Styles are you using?
Current

**Screenshots**
<img width="543" alt="Screen Shot 2019-11-05 at 10 20 06 AM" src="https://user-images.githubusercontent.com/39598672/68220687-2d42e400-ffb6-11e9-95cc-a562b97f779c.png">
<img width="316" alt="Screen Shot 2019-11-05 at 10 19 55 AM" src="https://user-images.githubusercontent.com/39598672/68220688-2d42e400-ffb6-11e9-8566-1f7fb5d79021.png">

PR #416, opened 2019-11-05
fix: menu width issue

InnaAtanasova:
## Related Issue
Closes SAP/fundamental-styles#415

## Description
The popover body is not taking the correct width. It's currently taking the width of the popover control, not the width of the component inside (Menu).
See the screenshots.

The cause of the bug:
 - Menu component is using display: table and table-layout: fixed

Solution: 
- Use display: inline-block; and `fd-ellipsis` mixin to handle very long text

### Before:
Screen Shot 2019-11-05 at 1 40 04 PM
Screen Shot 2019-11-05 at 1 39 47 PM
Screen Shot 2019-11-05 at 1 39 40 PM

### After:
Screen Shot 2019-11-05 at 1 49 16 PM
Screen Shot 2019-11-05 at 1 49 00 PM


src/menu.scss
```diff
@@ -33,14 +33,13 @@ $block: #{$fd-namespace}-menu;
...
   @include fd-reset();
+  @include fd-ellipsis();
 
   position: relative;
-  white-space: nowrap;
   padding-top: fd-space(base);
   padding-bottom: fd-space(base);
   width: 100%;
-  display: table;
-  table-layout: fixed;
```
bcullman: table-layout: fixed assumes you have pre-computed all rendered scenarios. It it almost *never* a solution that should be used in production.

InnaAtanasova:
> Can you update on why the tests are failing/what changed?

These is from running the tests locally

Screen Shot 2019-11-05 at 3 54 07 PM
Screen Shot 2019-11-05 at 3 53 54 PM

PR #419, opened 2019-11-07
feat: forms refactoring

InnaAtanasova:
## Description
Refactored:
- `fd-form-label` - added `--required` modifier for styling the `*`
- added `fd-form-item--horizontal` modifier class to display form items inline (horizontally)
- added `fd-form-label--inline-help` for form items with inline help
- refactored interactive states colors to use the theme designer
- refactored the form-messages to reflect the specs - now we have information, valid, invalid, warning
- removed `fd-form-message--help` - `BREAKING CHANGE`
- refactored `fd-textarea` - added interaction states, compact mode
- refactored `fd-radio` - added interaction states, cosy and compact mode
- refactored `fd-checkbox` - added interaction states, compact mode
- added `fd-form-label--radio`, `fd-form-label--checkbox` and `fd-form-label--toggle` modifier classes to label

### Before:
B5
B4
B3
B2
B1

### After:
A7
A6
A5
A4
A3
A8
A2
A1

About tests:
Running locally the following tests failed. They have been updated
Screen Shot 2019-11-07 at 4 51 25 PM


docs/_sass/_docs-display-component.scss
```diff
@@ -15,7 +15,7 @@
 
     > .fd-tile__content {
```
droshev: I think those stylings should be even removed. They create false expectations

src/input.scss
```diff
@@ -9,17 +9,29 @@ $block: #{$fd-namespace}-input;
...
+  padding: 0 0.625rem;
+  margin-bottom: 0;
+
+  &:last-child {
+    margin-bottom: 0.25rem;
+  }
 
   &::placeholder {
     @include fd-reset();
     @include fd-var-color("color", $fd-forms-color--placeholder, --fd-color-neutral-4);
```
droshev: can we replace the old values with theming parameters?

src/mixins/_mixins.scss
```diff
@@ -6,7 +6,7 @@
 
   @include fd-var-color("color", $fd-color, --fd-color-text-1);
```
droshev: can we replace with theming parameters?

src/input.scss
InnaAtanasova: done

src/mixins/_mixins.scss
InnaAtanasova: done

docs/_sass/_docs-display-component.scss
InnaAtanasova: This rule is applied only when the fd-tile is child of `.docs-component`.
The padding is just for the content to look nice and the width is to show the cases when we have very long text and truncation.
We want to be able to see this:
Screen Shot 2019-11-08 at 10 52 21 AM

Without the width we see this:
Screen Shot 2019-11-08 at 10 53 57 AM

ChristianKienle:
@InnaAtanasova JFYI: the forms documentation still refers a lot to "form groups". But the form groups have been removed.

src/form-item.scss
```diff
@@ -11,25 +11,25 @@ $block: #{$fd-namespace}-form-item;
...
   @include fd-reset();
 
-  margin-bottom: fd-space("small");
+  display: flex;
+  flex-direction: column;
 
-  &--inline {
-    float: left;
-    margin-right: fd-space(4);
+  &--horizontal {
```
jbadan: What are the different use cases for `--horizontal` versus `--inline`?

InnaAtanasova: `--inline` is what we had before for displaying the checkboxes and radio buttons inline.
Screen Shot 2019-11-13 at 2 30 45 PM
This was only the elements go to one line, the label is on another line. 

But then we needed to add inline items - label and input inline. This was achieved by using flexbox. 
Screen Shot 2019-11-13 at 2 31 06 PM

Stefano is working currently on Forms Layout - there's going to be vertical and horizontal layout. 

So, to minimize the breaking changes, we decided to keep the `--inline` modifier for displaying the checkboxes and radio btns inline, and add `--horizontal` for displaying the items inline

InnaAtanasova:
> @InnaAtanasova JFYI: the forms documentation still refers a lot to "form groups". But the form groups have been removed.

Thanks for pointing this out. You are absolutely right. We will def improve the documentation.

PR #420, opened 2019-11-09
fix: migrate buttons radius over theming parameters

droshev:


src/button-split.scss
```diff
@@ -32,13 +32,13 @@ $block: #{$fd-namespace}-button-split;
...
     }
 
     :nth-child(2) {
-      border-top-left-radius: $fd-border-radius;
-      border-bottom-left-radius: $fd-border-radius;
+      border-top-left-radius: var(--sapButton_BorderCornerRadius);
+      border-bottom-left-radius: var(--sapButton_BorderCornerRadius);
       border-top-right-radius: 0;
       border-bottom-right-radius: 0;
       border-left-width: 1px;
```
InnaAtanasova: This could also be a variable. Technically should be the same as border-width (--sapButton_BorderWidth)

droshev: done

Issue #426, opened 2019-11-14
fd-form-item + horizonal mode + form messages


ChristianKienle:
It seems like the combination mentioned above it not supported – or is this a bug?

<img width="687" alt="Screenshot 2019-11-14 at 17 24 36" src="https://user-images.githubusercontent.com/153225/68875815-ba4b1300-0703-11ea-9486-d57da9fad087.png">

Issue #425, opened 2019-11-14
fd-form-item in horizontal mode not aligned


ChristianKienle:
<img width="273" alt="Screenshot 2019-11-14 at 17 23 32" src="https://user-images.githubusercontent.com/153225/68875713-91c31900-0703-11ea-844b-53821cbdefb4.png">

(Can also be seen in the official docs)

ChristianKienle:
I talked to @InnaAtanasova in Slack. While doing so it became apparent that this bug report is ambiguous. It is not clear which alignment this bug report talks about.

The image below is the same as above but now has an alignment guide added to it which shows what I mean:

![dd](https://user-images.githubusercontent.com/153225/68924149-57e42800-0780-11ea-8090-311e9b4276c5.png)

The other alignment that is a bit off is this one:

![dd](https://user-images.githubusercontent.com/153225/68924241-9974d300-0780-11ea-8e2a-2469ee4ce6ae.png)

So the alignment is off in both directions – but originally I was only talking about the first alignment issue.

droshev:
is this still valid?

PR #428, opened 2019-11-15
Feat: form layout input

stefanoScalzo:
## Related Issue
Closes SAP/fundamental-styles#425
 Closes SAP/fundamental-styles#426

## Description
Adding layout grid to fix alignments

### Before:
No grid available

### After:
Screen Shot 2019-11-15 at 1 59 30 PM
Screen Shot 2019-11-15 at 1 59 23 PM


jbadan:
Is this a fiori3 component?

stefanoScalzo:
> Is this a fiori3 component?

yes

droshev:
@jbadan the changes are reflecting Fiori 3 specs

src/form-layout.scss
```diff
@@ -0,0 +1,139 @@
...
+.fd-form-horizontal
+*/
+
+$block: #{$fd-namespace}-form-layout;
+
+.#{$block} {
+  @include fd-reset();
+
+  background-color: transparent;
+  background-color: var(--sapGroup_ContentBackground);
```
InnaAtanasova: In the specs, the background is `transparent`. Why are you setting it first to `transparent` and then to `var(--sapGroup_ContentBackground)`?

src/form-layout.scss
```diff
@@ -0,0 +1,139 @@
...
+  padding: 0.5rem 1rem 1rem;
+  grid-gap: var(0.5rem, 0.5rem);
+  width: 100%;
+
+  &_horizontal--4-7-1 {
```
InnaAtanasova: I think we can remove the 4-7-1 from the class name.
could be just `fd-form-layout--horizontal` which will be 4-7-1 layout, and `fd-form-layout--vertical` which will be 12

@droshev , what do you think?

src/form-layout.scss
```diff
@@ -0,0 +1,139 @@
...
+      min-width: 0;
+
+      .fd-input {
+        margin-top: 0.25rem;
+        margin-bottom: 0.25rem;
+        height: 2.25rem;
+
+        &--compact {
+          margin-top: 0.1875rem;
+          margin-bottom: 0.1875rem;
```
InnaAtanasova: These values are repeated a lot. Maybe a good idea is to make them variables

src/form-layout.scss
```diff
@@ -0,0 +1,139 @@
...
+
+      .fd-form-label {
+        align-self: center;
+        justify-self: right;
+        padding-left: 0.75rem;
+        margin-right: 0.75rem;
+
+        &--compact {
+          margin-top: 0.1875rem;
+          margin-bottom: 0.1875rem;
```
InnaAtanasova: These values are repeated a lot. Maybe a good idea is to make them variables

InnaAtanasova:
Screen Shot 2019-11-21 at 4 34 37 PM
Looks like when you have two inputs inline, the padding inside the inout is missing.

The second issue from the screenshot is the alignment of the label text. The `Password:*` is not really in the center

InnaAtanasova:
There's a `</br>` tag at the bottom of the doc page. Also, I think we need some description for the Grif Layouts (some short explanation under the 4-7-1 and 12 layouts)

stefanoScalzo:
Will be implemented in new PR as inline forms have changed completely

Issue #444, opened 2019-11-20
Style(core): Button on active/selected state button focus is not as per Fiori 3 standards


sKudum:
Feature requests must include:
Focus should not exceed button border as per fiori3 standards image attached for the reference.

**Describe the bug**
On select/active state button focus is exceeding button border as shown in image.

> Is this issue related to a specific component?
Button, input

> What version of the Fundamental Styles are you using?
3.0

> What front-end framework are you implementing in, e.g., Angular, React, etc.?
Angular 8

**To Reproduce**
1. Go to fundamental-styles/ngx pages
2. Click on button
3. Observe the focus is exceeding the button border.

**Expected behavior**
Button focus should not exceed the button border.

**Screenshots**
In fundamentals

<img width="467" alt="Screen Shot 2019-11-20 at 1 03 38 PM" src="https://user-images.githubusercontent.com/53487468/69218638-265fc800-0b97-11ea-9cd6-3e32c99fba8e.png">

In Fiori 3
<img width="465" alt="Screen Shot 2019-11-20 at 1 04 34 PM" src="https://user-images.githubusercontent.com/53487468/69218650-2fe93000-0b97-11ea-81a7-b72e5ab36a0f.png">

**Desktop (please complete the following information):**
 - OS: all
 - Browser :all
 - Version : all

**Smartphone (please complete the following information):**
 - Device: all
 - OS: all
 - Browser :all
 - Version :all

**Additional context**
Having below code in .fd-button.is-focus can help to resolve this issue.

   outline: -webkit-focus-ring-color auto 0px;
    border: 1px solid #085caf;
    box-shadow: inset 0px 0px 0px 0px 085caf;
    box-sizing: border-box;

PR #445, opened 2019-11-20
fix: button when is-selected

stefanoScalzo:
## Related Issue
Closes SAP/fundamental-styles#444

## Description
Removes the extra border for the new specifications given by fiori 3.
Also removed fd-button--emphasized fd-button--positive and  fd-button--emphasized fd-button--negative since these classes no longer exist
### Before:
image
### After:
Screen Shot 2019-11-20 at 10 08 19 AM


bcullman:
why so many diffs?

InnaAtanasova:
> why so many visual regression changes?

Repeat Brad's question. SideNav, Table, Image, etc don't have buttons. Their tests shouldn't be updated.

Also, in the description you mention you "Also removed fd-button--emphasized fd-button--positive and fd-button--emphasized fd-button--negative since these classes no longer exist" but I still see `positive` and `negative` in the documentation.
Screen Shot 2019-11-20 at 3 48 53 PM


Last, removing classes is Breaking Changes and should be documented.

bcullman:
nested list still in visual regression tests. - are they needed?

Issue #62, opened 2019-06-11
Redefine form groups


greg-a-smith:
Tasks:

1. Change all examples that have a group of radio buttons or checkboxes to have a wrapping `<div>` with the `fd-form__group` class
1.  Create new class `fd-form__group--inline`
1.  Remove `fd-form__item--inline` class
1.  Update examples of inline radios or checkboxes to use `fd-form__group--inline` on wrapping `<div>`

PR #451, opened 2019-11-22
feat: input group for checkbox/radio inline

stefanoScalzo:
## Related Issue
Closes SAP/fundamental-styles#62

## Description
Remove form-item--inline and add form-group--inline class

Nothing has visually changed but because the visual tests for inline elements have moved to the form-group level they are now in the form-group-visual regression tests

## Breaking Change
* Remove `fd-form-item--inline` class
* Add `fd-form-group--inline` class, which is parent of multiple `fd-form-item`s


src/form-item.scss
```diff
@@ -27,9 +27,9 @@ $block: #{$fd-namespace}-form-item;
     }
   }
 
-  &--inline {
-    display: block;
-    float: left;
-    margin-right: fd-space(4);
-  }
+  // &--inline {
```
jbadan: Can we just delete this?

test/templates/form-group/index.njk
```diff
@@ -27,4 +27,56 @@
   {% endset %}
   {{ format(example) }}
 
+  <h2>Inline radio</h2>
+
+  {% set example %}
+    <fieldset class="fd-fieldset">
+      <legend class="fd-fieldset__legend">Inline Radio buttons</legend>
+      <div class="fd-form-group--inline">
+          <div class="fd-form-item fd-form-item">
```
jbadan: this example has double `fd-form-item` on each div

InnaAtanasova:
`test/templates/form-group/index.njk` still has double `fd-form-item` on each div

docs/pages/components/form.md
```diff
@@ -92,25 +92,25 @@ Do not use the input field if:
...
     <br />
-    <div class="fd-form-item fd-form-item--horizontal">
+    <div class="fd-form-item--horizontal">
         <label class="fd-form-label" for="input-1a">Default Input:</label>
         <input class="fd-input" type="text" id="input-1a" placeholder="Field placeholder text">
     </div>
     <br />
     <br />
-    <div class="fd-form-item fd-form-item--horizontal">
+    <div class="fd-form-item--horizontal">
```
JKMarkowski: Is it good usage of BEM things? I see `fd-form-item` gives `fd-reset()`

src/form-group.scss
```diff
@@ -17,3 +17,12 @@ $block: #{$fd-namespace}-form-group;
     margin-bottom: 0;
   }
 }
+
+.#{$block}--inline {
+  display: flex;
+
+  .fd-form-item {
+    float: left;
```
JKMarkowski: Shouldn't we avoid floats generally ?

docs/pages/components/form.md
JKMarkowski: Is it good usage of BEM things? I see `fd-form-item` gives `fd-reset()` Normally it should be `fd-form-item fd-form-item--horizontal`

test/templates/form-group/index.njk
```diff
@@ -27,4 +27,56 @@
...
+{% endset %}
+{{ format(example) }}
+<br>
+
+  <h2>Checkbox inline </h2>
+
+  {% set example %}
+    <fieldset class="fd-fieldset">
+        <legend class="fd-fieldset__legend">Checkboxes inline</legend>
+        <div class="fd-form-group--inline">
```
InnaAtanasova: Should be `fd-form-group fd-form-group--inline`. `fd-form-group` brings the reset

src/form-group.scss
```diff
@@ -17,3 +17,12 @@ $block: #{$fd-namespace}-form-group;
...
   }
 }
+
+.#{$block}--inline {
+  display: flex;
+  justify-content: flex-start;
+
+  > :not(:last-child) {
+    margin-right: fd-space(4);
+  }
```
InnaAtanasova: This is scss, you can use nesting to specify that an element is a child of another element or `&`

PR #473, opened 2019-11-28
feat:popover refactoring

InnaAtanasova:
## Description
Update the popover component to match Fiori3 specs:
- border radius
- box-shadow
- using theme designer parameters for colors
- arrow width and height

Added Header, Subheader and Footer to the popover body

## Breaking changes 
New Classes in popover component
* Header (without a subheader): `fd-popover__body-header`
* Header with subheader: `fd-popover__body-header--with-subheader` should be applied if the header is followed by a subheader
* Header in compact mode: `--compact` modifier class on the `fd-popover__body-header`
* Subheader: `fd-popover__body-subheader`
* Subheader in compact mode: `--compact` modifier class on the `fd-popover__body-subheader`
* Footer: `fd-popover__body-footer`
* Footer in compact mode: `--compact` modifier class on the `fd-popover__body-footer`

### Before:
Screen Shot 2019-11-28 at 3 30 35 PM

### After:
Screen Shot 2019-11-28 at 3 30 53 PM
Screen Shot 2019-11-28 at 3 30 45 PM


JKMarkowski:
@InnaAtanasova  I see on specs, arrow has same border as rest of popover. On your examples there is no border in arrow.

stefanoScalzo:
The popover control should not include the label for the one with input

InnaAtanasova:
> @InnaAtanasova I see on specs, arrow has same border as rest of popover. On your examples there is no border in arrow.

True! The "border" of the popover body is achieved by box-shadow. For the "arrow" we can not use box-shadow like it's specified in the design specs. They are using a square that is rotated and half of it is hidden by the popover body. We are using borders to create the "arrow". We could change the implementation of the arrow (to make a square and rotate it and apply box-shadow), but the popper.js is using a similar to our implementation  and in order to overwrite their styles we need a border color. Since there is no color specified in the specs, we contacted the design team and they suggested for the arrow to use --sapGroup_ContentBorderColor. Until they provide a color for the arrow (same as the popover body) we will use this one.

src/popover.scss
```diff
@@ -6,24 +6,44 @@ $block: #{$fd-namespace}-popover;
 
 .#{$block} {
 
-  $fd-popover-top-position: 120% !default;
+  $fd-popover-top-position: calc(100% + 0.5rem) !default;
```
stefanoScalzo: you actually don't need calc when using sass it does it for you

InnaAtanasova: Hmm not really because we are using 100%.

Here is a good explanation from Stackoverflow:
"SASS height: 100px / 2; will compile to height: 50px; <- and this is what the browser will see always no matter what.

CSS width: calc(100% - 20px) will always be this even on the browser, and at that point the browser will do the calculations depending on what that 100% looks like."

https://stackoverflow.com/questions/52869421/sass-calc-vs-normal-calculation

stefanoScalzo:
The popover should not open when clicking on the label! Only on the input field

stefanoScalzo:
There should be a max width on the popover according to the viewport which will then need to add overflows for the inner contents

stefanoScalzo:
There should be a limit on the size of the button or it will overflow
Screen Shot 2019-12-18 at 2 02 24 PM

InnaAtanasova:
> There should be a limit on the size of the button or it will overflow
> Screen Shot 2019-12-18 at 2 02 24 PM

This is not in the scope of this PR

InnaAtanasova:
> There should be a max width on the popover according to the viewport which will then need to add overflows for the inner contents
> Screen Shot 2019-12-18 at 1 59 30 PM


When the popover has a specified width (max-width), the Header, the Subheader and the footer will have ... if the content is too long

Screen Shot 2020-01-08 at 3 08 21 PM

stefanoScalzo:
Is it normal that both the image and icon are not tab-able?
Also, where did you get the info for the subheader as ui5 only has a header and a footer?
Shouldn't the header have var(--sapContent_ShadowColor) and not var(--sapContent_HeaderShadow)

InnaAtanasova:
`Is it normal that both the image and icon are not tab-able?` - not in the scope of this PR
`Also, where did you get the info for the subheader as ui5 only has a header and a footer?` - in Header specs
`Shouldn't the header have var(--sapContent_ShadowColor) and not var(--sapContent_HeaderShadow)` - not according to the specs

JKMarkowski:
LGTM :fireworks:

PR #475, opened 2019-11-29
fix: Match link and breadcrumb to fiori3

JKMarkowski:
## Related Issue
Closes SAP/fundamental-styles#474

## Description
There is link and breadcrumb fixed, now the focus/hover works in safari/IE/ME. Also other border is removed and the link follows latest fiori3 specs.

## Breaking changes 
* No breaking changes

### Before:
Links:
image

BreadCrumb
image

### After:
Links:
image

BreadCrumb:
image


src/breadcrumb.scss
```diff
@@ -83,10 +83,8 @@ $block: #{$fd-namespace}-breadcrumb;
 
     @include fd-link();
```
stefanoScalzo: See with designers if there needs to be a max width, if not it is good! If so apply it with overflow!

src/mixins/_mixins.scss
```diff
@@ -270,31 +270,32 @@
   text-decoration: none;
   display: inline-block;
   transition: all $fd-animation--speed ease-in;
```
InnaAtanasova: missing truncation behavior

src/mixins/_mixins.scss
```diff
@@ -270,31 +270,32 @@
...
-    border-bottom-color: var(--sapLink_Visited_Color);
   }
 
   @include fd-disabled() {
     opacity: 0.4;
     color: var(--sapLinkColor);
-    border-bottom-color: var(--sapLinkColor);
-    cursor: not-allowed;
+    pointer-events: none;
   }
```
InnaAtanasova: missing Active and Inverted state

stefanoScalzo:
We need to apply the dropdown for long  breadcrumbs

InnaAtanasova:
Link component is missing `visited` and `inverted` states as described in the specs (Interactive states)

InnaAtanasova:
The last element in the Breadcrumb should be different than the others (@sapUiContentLabelColor text color, not clickable, no underline)

InnaAtanasova:
Screen Shot 2020-01-14 at 2 28 11 PM

The underline should go to the end (under the arrow-down icon)

InnaAtanasova:
Screen Shot 2020-01-14 at 3 13 49 PM

You prob need to add fd-reset() to the overflow items (Check un-normalize in the Playground)

JKMarkowski:
Comments are solved Inna :) Thanks for noticing

bcullman:
Not sure if this is about this or now, but.... There is something about the styling of the link that cause our alert bar to be incorrectly sized when there is a link inside of it. 

see:
 
image

versus:

image

This is just one place where this causes an issue. A simple paragraph of text with a link within the copy will cause the line with that link to have a different line high as well.

I always thought this was because the style was not using true underline style, but rather a border bottom "faux" underline, was causing the link element to fill out it's box size, throwing off the sizing. 

Removing the border bottom style and using standard underline does help, but it still looks blown out. see:

image

JKMarkowski:
Hi @bcullman thanks for noticing, links have wrong line-height. It should have 1rem(16px) so the alert will be in correct size

PR #489, opened 2019-12-04
fix: move compact modifier class from parent to children elements

InnaAtanasova:
## Description 
The `compact` modifier class was applied on the wrapping div. We agreed it's better to apply it on each child element.
Also, fixed the toggle function for applying the theme in the test playground


estelaxu3:
Missing the grey lines (///) in IE.

Reference from Chrome:
Screen Shot 2019-12-04 at 9 34 58 AM

stefanoScalzo:
Looks good to me! travis keeps failing so try and push it one more time!

InnaAtanasova:
> Missing the grey lines (///) in IE.
> 
> Reference from Chrome:
> Screen Shot 2019-12-04 at 9 34 58 AM

This is browser specific

estelaxu3:
The rest looks great. We can release it.

Issue #485, opened 2019-11-29
Side navigation broken in IE


estelaxu3:
The navigation arrow is not aligned.

estelaxu3:
![image](https://user-images.githubusercontent.com/54589269/70249729-cd716180-174a-11ea-8274-6e73bb3925f4.png)

PR #490, opened 2019-12-05
fix: Align side-nav navigation arrow on IE11

salarenko:
Aligning side-nav navigation arrow on IE11.

## Related Issue
Closes SAP/fundamental-styles#485

## Description
Error occurred because of IE11 breaking `align-items: center` after setting `min-height` of an element. One of the workarounds enabling using `min-height` without breaking flex-box properties is to additionally set `height` < `min-height`.

More about the issue: https://github.com/philipwalton/flexbugs/issues/231

Merge also includes updated side-nav playground template, providing updated html structure and 3 common side-navs usages.


test/templates/side-nav/index.njk
```diff
@@ -7,42 +7,196 @@
 {% set example %}
 <nav class="fd-side-nav">
     <div class="fd-side-nav__main-navigation">
-      <ul class="test-sidenav-list">
-        <li>Item 1</li>
-        <li>Item 1</li>
-        <li>Item 1</li>
-        <li>Item 1</li>
-      </ul>
```
InnaAtanasova: We shouldn't bring the nested list in sidenav. There are tests for Nested list. Inside Sidenav tests we should only test the Sidenav styling because our components are self-contained. You should revert the test changes

salarenko: Restored previous test cases

PR #491, opened 2019-12-05
fix: the PR fixes problem when running backstop tests on local linux

rengare:
## Description
The PR fixes problem when running backstop tests on local linux machine. I tried to upgrade backstop to newest version but it brokes most of tests. This is the only solution that works right now.
Lack of support for host.docker.internal on linux machines is the real rootcause of the issue.
It can be fixed when the flag ```--network host``` is passed to docker command although in current version of backstop it is not posible to override docker command using configuration (which is fixed in newest version).

### Before:
When running backstop tests on local linux machine, task was unable to finish

### After:
The task run fine on linux machine
image


rengare:
The PR supports both platforms linux and mac

backstopJsFix/runDocker.js
```diff
@@ -0,0 +1,41 @@
+const { spawn } = require('child_process');
```
jbadan: could we move this into `scripts/` folder?

rengare: You mean the file? Yes off course I going to move it right now

rengare: Done

Issue #484, opened 2019-11-29
Product switch broken in IE


estelaxu3:
When click on the menu button, there is no menu popping out.

Issue #479, opened 2019-11-29
Button doesn't work properly in IE


estelaxu3:
The Split Button is not interactive when click on dropdown in IE, missing a popover. 

Reference in Chrome
<img width="222" alt="Screen Shot 2019-11-29 at 10 22 48 AM" src="https://user-images.githubusercontent.com/54589269/69878340-ae7a5780-1292-11ea-8ba2-e98ab986fe7f.png">

estelaxu3:
It is now working properly in IE, we can close this. 

![image](https://user-images.githubusercontent.com/54589269/70249019-cb5ad300-1749-11ea-9918-a660e1b76086.png)

PR #498, opened 2019-12-06
docs: removed functionality of es6 because it brokes IE and fixed 2 issues

rengare:
## Related Issue
Closes SAP/fundamental-styles#484
Closes SAP/fundamental-styles#479

## Description
ES6 specification is not supported in IE 11 and ended up in displaying error in console. After fix, aria-controls which are handled by jquery works fine

### After:
after


InnaAtanasova:
I changed the title to `docs` coz it's a fix in the documentation.

Issue #497, opened 2019-12-06
Can't make a button round after migrating to 0.3.0


d067006:
**Describe the bug**
After migrating from 0.1.0 to 0.3.0 version, we were not able to make buttons round using fd-image--circle class.

> What version of the Fundamental Library Styles are you using?
0.3.0

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?
SAP Customer Service Moments

> What front-end framework are you implementing in, e.g., Angular, React, etc.?
Angular version 8

**To Reproduce**

1. Create a button
2. Use **fd-image--circle** as class
3. button is not round

Source code:

`    <button fd-button placement="bottom" [fdPopover]="template" [noArrow]="false" [glyph]="'add'" [(isOpen)]="isOpen"
        class="fd-image--circle"></button>`

**Expected behavior**
On version 0.1.0 this was working fine and after we migrated to 0.3.0 this doesn't work anymore. We would like to know if this is a bug or there is something that has changed and we are not aware of.

**Screenshots**
![image](https://user-images.githubusercontent.com/52275838/70314650-307cfa00-1818-11ea-9b4c-5927160752b4.png)

droshev:
Hi @d067006,
we understand about your issue. On the other side, Fiori 3 doesn't have circled buttons and if we deliver circled buttons then those buttons won't be aligned and also they won't be theme-able.
If you have scenarios where you want to have those buttons then it is safer you if you add a custom class to those buttons in your applications([example](https://stackblitz.com/edit/angular-fjja5h)).

PR #499, opened 2019-12-06
feat: add a new class to Button component to allow rounded buttons

rengare:
## Related Issue
Closes SAP/fundamental-styles#497

## Description
Round css class added for button: ```fd-button--rounded```

### Before:
Before PR a css class ```fd-image--circle``` was used

### After:
after


stefanoScalzo:
Just make sure you document the added class! LGTM

estelaxu3:
Looks consistent across all platform.

rengare:
added a test case for rounded button. I had some troubles with tests but I manage to fix it

rengare:
> You shouldn't have so many updated tests in your PR. You should have only 1 test that is testing a button with rounded corners

I manage to fix this issue, now PR is clean :)

InnaAtanasova:
We keep the PR but we need to discuss more this solution

droshev:
@rengare do we need this PR? I think we agreed for time being not to do this in ngx as it doesn't align with Fiori 3

rengare:
closed because it is not in a fiori 3 spec

PR #501, opened 2019-12-06
fix: line height changed to match the font

rengare:
## Related Issue
Closes SAP/fundamental-styles#467

## Description
Line height changed to match a character which went out of boundaries 

### Before:
before

### After:
It is fixed but I can't upload the screenshot (github returns error)


InnaAtanasova:
Hi Gracjan, 
I renamed it ti `fix` because it's not a breaking change. We use `feat` when we introduce breaking changes (remove classes, change the html, etc).

rengare:
Ok, I'll remember that :)

Issue #417, opened 2019-11-07
button: compact button height is not matching with Fiori3 standards


sKudum:
**Describe the bug**
fundamental-ngx/core compact button height is not as per Fiori3 standards.

> Is this issue related to a specific component?
       Button component

> What version of the Fundamental Styles are you using?
       0.3.0

> What front-end framework are you implementing in, e.g., Angular, React, etc.?
Angular

**Expected behavior**

- compact button height should be 26px including border.

**Screenshots**
In fundamental-ngx

<img width="363" alt="Screen Shot 2019-11-07 at 2 21 09 PM" src="https://user-images.githubusercontent.com/53487468/68374004-ff060580-0169-11ea-80b0-fb29ed85719c.png">

In Fiori3

<img width="579" alt="Screen Shot 2019-11-07 at 2 21 42 PM" src="https://user-images.githubusercontent.com/53487468/68374011-04fbe680-016a-11ea-82bb-4fa4c5c89844.png">

trilodge:
same holds true for inputs.

Based on _settings.scss -> $fd-forms-height--compact: $fd-spacing--base * 7 !default; // 28px 

New modifier would be 6.5 to achieve 26px.

droshev:
@sKudum how about the buttons here https://fundamental-styles.netlify.com/components/button.html

droshev:
@stefanoScalzo could you check @sKudum point?

PR #509, opened 2019-12-11
[WIP] feat: refactor Button component to match Fiori 3 specs

rengare:
## Related Issue
Closes SAP/fundamental-styles#417

## Description
Changes nade to the button component according to fiori3 specs


src/_settings.scss
```diff
@@ -265,8 +265,7 @@ $fd-width--gutter: map-get($fd-spacing, s); // desktop default
 
 // Form Configuration
 $fd-forms-height: $fd-spacing--base * 9 !default; // 36px
-$fd-forms-height--compact: $fd-spacing--base * 7 !default; // 28px
-$fd-forms-height--input-text: $fd-forms-height !default; // deprecate
+$fd-forms-height--compact: $fd-spacing--base * 6.5 !default; // 26px before 28px
```
JKMarkowski: This is also used in `fd-form-select` class (native select component). Should it be changed that way? I saw that we have button configuration below (278-281 lines)
image

src/button.scss
```diff
@@ -220,7 +221,7 @@ a.#{$block} {
 
       // placements
       &::before {
-        margin-right: fd-space(2);
+        margin-right: $fd-button-icon-margin-right;
```
JKMarkowski: Could you also apply these changes to RTL mode?

InnaAtanasova:
There are too many tests updated. Since our components are self-contained you shouldn't update tests that are not testing the buttons.

src/_settings.scss
rengare: I thin so, You can check the message https://github.com/SAP/fundamental-styles/issues/417#issuecomment-550999964

src/button.scss
rengare: sure

src/_settings.scss
InnaAtanasova: I think Jedrzej is right. Changing the form-height affects all forms and hence all the tests that are updated. The best way to "solve" this issue is to create new variables where you can declare the button height in cosy and compact mode

rengare: done, the default value was restored and new variable inside the button file introduced

src/button.scss
rengare: fixed

rengare:
> There are a couple of things that I noticed that are not matching Fiori3 specs:
> Cosy button
> 
>     * Cosy button has a height of 2.25rem and clickable height of 2.75rem (Compact button has a height of 1.625rem and clickable height of 2rem)
> 
>     * Paddings left and right are 0.625rem (or 10px) including border
> 
>     * Icon and Text button has wrong paddings/margins
> 
>     * should set a min width

rengare:
@InnaAtanasova 
-     Cosy button has a height of 2.25rem and clickable height of 2.75rem (Compact button has a height of 1.625rem and clickable height of 2rem) -> **I can't fix it without breaking changes**
-     Paddings left and right are 0.625rem (or 10px) including border -> **fixed**
-     Icon and Text button has wrong paddings/margins -> **fixed**
-     should set a min width -> **fixed**

JKMarkowski:
image
The compact mode buton has not centered text. Let's change line-height to `1rem`, it should help.

InnaAtanasova:
Let's discuss the Breaking Changes with @droshev after the holidays.

stefanoScalzo:
Check the visual regression tests, I don't believe you need all of them

src/button.scss
```diff
@@ -261,6 +269,20 @@ a.#{$block} {
     }
   }
 
+  &--text-alignment {
+    &-left {
```
stefanoScalzo: We should make left as the default and only use the center or right as modifiers

bcullman:
can we see a before and after screen shots on the PR description?

JKMarkowski:
@rengare  There is still to much test images changed. Overflow is also missing, but it's probably gonna be solved in other PR

sKudum:
> @rengare There is still to much test images changed. Overflow is also missing, but it's probably gonna be solved in other PR

Hi @JKMarkowski , PR#600 is closed as per Deno's comments.

rengare:
closed due to problems with build. New PR was made as continuation #624

PR #513, opened 2019-12-13
feat: Introduce Dialog as successor for Modal - Fiori 3 refactor

salarenko:
## Related Issue
Closes SAP/fundamental-styles#512

## Description
Changes introduce:
- new component Dialog replacing Modal
- Dialog component matching Fiori3 standard
- update of Fundamental Styles documentation
- update of developers playground examples
- removed `fd-overlay` support for Modal

Dialog component is composed out of:
- Bar representing dialogs:
    - header
    - subheader
    - footer
- Busy Indicator
    - loader

## Screenshots
image

image

image

image


bcullman:
not sure if this is in scope, but the box shadow on the modal header *seems* wrong here.

mikerodonnell89:
> not sure if this is in scope, but the box shadow on the modal header _seems_ wrong here.

The box shadow from the specs: 
Screen Shot 2020-01-17 at 1 20 20 PM

From this PR: 
Screen Shot 2020-01-17 at 1 24 17 PM


I agree that it does seem too pronounced in the PR.  The spec sheet does say to use `--sapContent_Shadow3` which this PR is doing... however I think the spec sheet might've listed the wrong box shadow value

bcullman:
Does UI5 web components have it wrong here:

https://sap.github.io/ui5-webcomponents/playground/components/Dialog/

(they also show a centered title)

src/modal.scss
```diff
@@ -4,111 +4,190 @@
...
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    height: $fd-modal-header-height;
+    box-shadow: $fd-modal-header-box-shadow;
+    background-color: $fd-modal-header-color;
+
+    .#{$block}__title {
+      @include fd-reset();
+      @include fd-type("1");
```
InnaAtanasova: Should get rid of `fd-type`, it's a hard-coded value and won't be affected by the theme. You can use css variable from sap-theming

src/modal.scss
```diff
@@ -4,111 +4,190 @@
...
-    position: absolute;
-    right: fd-space(2);
+    .#{$block}__close {
+      margin-left: fd-space(reg);
 
-    @include fd-rtl() {
-      left: fd-space(2);
-      right: auto;
+      @include fd-rtl() {
+        margin-right: fd-space(reg);
```
InnaAtanasova: needs to be `rem` values

InnaAtanasova:
Screen Shot 2020-01-27 at 4 13 32 PM
I think something is wrong with the width of the modal. The min-width is 20rem which is 320px. This means that the example should look like this:
Screen Shot 2020-01-27 at 4 15 58 PM

droshev:
> approved although rest of comments have to be fixed

if there

stefanoScalzo:
@droshev Shouldn't we add another class with the dialog name?

stefanoScalzo:
I think that the dynamic paddings should be shown depending on the screen size since the screen sizes are hardcoded. The information is written properly if they are curious. but by including them for the same screen size it is confusing

stefanoScalzo:
Screen Shot 2020-02-12 at 6 35 48 PM
Handle looks broken

stefanoScalzo:
is the draggable dialog in fiori?

stefanoScalzo:
Are we adding the message dialog as well since we took care of busy dialog?

stefanoScalzo:
Screen Shot 2020-02-12 at 6 53 18 PM
Screen Shot 2020-02-12 at 6 53 38 PM
Screen Shot 2020-02-12 at 6 54 00 PM
fix overflows

stefanoScalzo:
Shouldn't there be a line/shadow on top of the subheader?

stefanoScalzo:
Need to add footer with begin and end specs in documentation

salarenko:
@mikerodonnell89 Dialog buttons now also use `.fd-button--emphasized`. To be honest in case of buttons I don't see much difference between docs screenshots and current implementation.

@stefanoScalzo please don't review PRs in [WIP] and use review mode for comments.
1. Modal has been renamed into Dialog
2. In terms of responsive padding I followed [Dialog with responsive padding](https://sapui5.hana.ondemand.com/#/entity/sap.m.Dialog/sample/sap.m.sample.Dialog). Here padding is applied depending on Dialog size. I think this is more logical solution than applying paddings based on view-port size, but it is to be discussed on daily meeting.
3. Has been fixed
4. It doesn't exist in style specs, but again I was suggested by [Dialog (Draggable)](https://sapui5.hana.ondemand.com/#/entity/sap.m.Dialog/sample/sap.m.sample.Dialog)
5. From what I've consulted with other developers, Message Dialog is going to be introduced as separate component (Message Box)
6. I've seen that issue but I dont know if it shouldn't be also fixed on Bar component side. To be discussed on daily meeting.
7. This is Bar component issue
8. Fixed

src/dialog.scss
```diff
@@ -0,0 +1,170 @@
...
+      max-height: calc(100vh - #{$fd-dialog-content-margin-y});
+    }
+
+    &--resizable {
+      @include fd-icon("resize-corner", "m", "after") {
+        color: $fd-dialog-resize-handle-color;
+        position: absolute;
+        cursor: nwse-resize;
+        right: 0;
+        bottom: -1px;
```
rengare: the specification says that bottom should be 0px, did you set bottom to -1 because of incorrect visual representation?

src/dialog.scss
```diff
@@ -0,0 +1,170 @@
...
+    &--no-vertical-padding {
+      padding-top: 0;
+      padding-bottom: 0;
+    }
+  }
+
+  &__footer {
+    .#{$block}__decisive-button {
+      min-width: $fd-dialog-footer-button-min-width;
+    }
```
InnaAtanasova: Since it's a requirement for all buttons inside Popover and Dialogs maybe it's better to add this requirement in the Bar component. And to "force" this rule we can target directly button elements. Because if the user omits this class the button won't be forced to apply the 4rem min width. Just a suggestion :)

salarenko: The thing is this should be applied only to "Begin" and "End" buttons. Docs are written as the suggest there can be other buttons inside footer which shouldn't be modified.

image

InnaAtanasova: We can always target the first and the last child. My only concern is that the user can easily omit this class and to be honest it's easy to avoid this class by targeting the first and the last child of the footer/floating footer. Up to you to decide :)

salarenko:
Because of pipeline problems, this feature has been moved from fork branch to origin branch `feat/dialog-fiori3_2` and merged with #687

PR #517, opened 2019-12-16
docs: Example fixes on firefox

JKMarkowski:
## Description
On smoke tests, I found some of components examples doesn't look good on Firefox. So:
* (All) I removed `multi-input` link from component group - it was redirecting to wrong url. Place for `multi-input` is in Patterns section.
* ShellLayout example with `Alert` wasn't up to date with latest changes.  
* Busy Indicator - Flex wasn't good idea there, `<br>` Didn't work inside `flex` on firefox
 

### Before:
busyindbad
image

### After:
busyindgood
image

### DIFF
--- a/docs/pages/layouts/shell-layout.md
+++ b/docs/pages/layouts/shell-layout.md
@@ -188,8 +188,8 @@ The container visibility can be toggled with the `aria-hidden` attribute.
     </div>
     <div class="fd-shell__overlay fd-overlay fd-overlay--alert" aria-hidden="false">
         <div class="fd-alert fd-alert--warning fd-alert--dismissible" role="alert" id="4Nolz351">
-          <button class="fd-alert__close" aria-controls="4Nolz351" aria-label="Close"></button>
-          Lorem ipsum dolor sit amet, consectetur adipisicing elit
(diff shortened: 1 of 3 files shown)


docs/pages/layouts/shell-layout.md
```diff
@@ -188,8 +188,8 @@ The container visibility can be toggled with the `aria-hidden` attribute.
     </div>
     <div class="fd-shell__overlay fd-overlay fd-overlay--alert" aria-hidden="false">
         <div class="fd-alert fd-alert--warning fd-alert--dismissible" role="alert" id="4Nolz351">
-          <button class="fd-alert__close" aria-controls="4Nolz351" aria-label="Close"></button>
-          Lorem ipsum dolor sit amet, consectetur adipisicing elit
+          <button class="fd-button fd-button--light fd-button--compact fd-alert__close" aria-controls="4Nolz351" aria-label="Close"></button>
```
stefanoScalzo: We need to keep the fd-alert__close class to keep it self contained.

JKMarkowski: It is there actually, but we use the scss from `fd-button`.

PR #519, opened 2019-12-16
fix: placeholder vertical alignment in safari

rengare:
## Description
PR fixes placeholder issue in safari

### Before:
image

### After:
image

### DIFF
--- a/src/input.scss
+++ b/src/input.scss
@@ -36,7 +36,10 @@ $block: #{$fd-namespace}-input;
 
     line-height: $fd-form-input-height;
     font-style: italic;
-    color: var(--sapField_PlaceholderTextColor);
+  }
+
+  &::-webkit-input-placeholder {
+    line-height: normal;
   }
 
   &--compact {
@@ -49,6 +52,10 @@ $block: #{$fd-namespace}-input;
 
     &::placeholder {
       line-height: $fd-form-input-height--compact;
+    }
+
+    &::-webkit-input-placeholder {
+      line-height: normal;
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/input.scss
```diff
@@ -32,11 +32,12 @@ $block: #{$fd-namespace}-input;
   }
 
   &::placeholder {
-    @include fd-reset();
-
     line-height: $fd-form-input-height;
     font-style: italic;
-    color: var(--sapField_PlaceholderTextColor);
```
JKMarkowski: Do we wan't to remove `color` from the placeholder?

stefanoScalzo: No we don't

InnaAtanasova: Actually we do! The color of the placeholder is set in _forms.scss

InnaAtanasova: font-style is also not necessary

estelaxu3:
I think the text is not aligned properly, it should move down 1 or 2 pixel. 
Screen Shot 2019-12-16 at 11 46 17 AM

src/input.scss
rengare: right, removed by mistake, revert is comming

InnaAtanasova:
I opened a PR just to be able to test the Netlify deploy in IE11 but if you remove all `placeholder` rules in `src/input.scss` everything is fixed. There is no need of setting the line-height in either cozy or compact mode

https://github.com/SAP/fundamental-styles/pull/521/files

rengare:
fixed provided by #521

PR #521, opened 2019-12-16
fix: fix the alignment of the placeholder in inputs

InnaAtanasova:


rengare:
Please check safari cause the problem was there

stefanoScalzo:
> Please check safari cause the problem was there

It is good in all browsers

rengare:
so I have closed a #519

Issue #526, opened 2019-12-17
Can write inside disabled input


salarenko:
**Describe the bug**
Disabled input presented in Forms / Inputs documentation section is not disabled.
input can be accessed after clicking on label. Input element has no _disabled_ attribute.

Issue appears on Chrome and Edge (probably also on Firefox and Safari, but not on IE11).
Issue is related to Forms / Input component.
Fundamental styles version: v0.4.0-rc.21

To reproduce issue go to: _"/components/form.html"_
![Forms-Fundamental-Styles-Google-Chrome](https://user-images.githubusercontent.com/17496353/70982210-e1a44f80-20b6-11ea-96bb-061c35ed3b06.gif)

PR #537, opened 2019-12-17
fix: Can write inside disabled input

salarenko:
## Related Issue
Closes SAP/fundamental-styles#526

## Description
Disabled property was missing in _Disabled Input_

### DIFF
--- a/docs/pages/components/form.md
+++ b/docs/pages/components/form.md
@@ -81,7 +81,7 @@ Do not use the input field if:
     <br />
     <div class="fd-form-item">
         <label class="fd-form-label" for="input-05">Disabled Input:</label>
-        <input class="fd-input is-disabled" type="text" id="input-05" disabled placeholder="Field placeholder text">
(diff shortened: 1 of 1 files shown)


docs/pages/components/form.md
```diff
@@ -81,7 +81,7 @@ Do not use the input field if:
     <br />
     <div class="fd-form-item">
         <label class="fd-form-label" for="input-05">Disabled Input:</label>
-        <input class="fd-input is-disabled" type="text" id="input-05" placeholder="Field placeholder text">
+        <input class="fd-input is-disabled" type="text" id="input-05" disabled placeholder="Field placeholder text">
```
InnaAtanasova: you can remove `is-disabled` class. The idea was that the user could use either `is-disabled` class or `disabled` attr interchangeably

salarenko: Ok, `is-disabled` class removed.

I don't think there is a point of creating `is-disabled` css class to actually disable a button, link or input. CSS wont cover things like _Tab_ key navigation or automatic focus on label click.

Issue #538, opened 2019-12-17
Pickers in IE11


stefanoScalzo:
**Describe the bug**
The width of the pickers is too small
> Is this issue related to a specific component?
Time and date picker
> What version of the Fundamental Library Styles are you using?
latest

PR #539, opened 2019-12-17
fix: input-group-pickers

stefanoScalzo:
## Related Issue
Closes SAP/fundamental-styles#538

## Description
Fixes the sizing in IE11 for input groups with time picker and date picker.
Currently no screeenshots, using netlify for IE testing

### DIFF
--- a/src/input-group.scss
+++ b/src/input-group.scss
@@ -60,7 +60,9 @@ $fd-input-border-radius: var(--sapField_BorderCornerRadius);
     @include fd-reset-child-spacing();
 
     border: none;
-    flex: 1;
+    flex: 0;
+    flex-grow: 1;
+    width: auto;
     padding-right: 0.25rem;
     padding-left: 0.25rem;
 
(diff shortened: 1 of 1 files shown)


src/input-group.scss
```diff
@@ -60,7 +60,10 @@ $fd-input-border-radius: var(--sapField_BorderCornerRadius);
     @include fd-reset-child-spacing();
 
     border: none;
-    flex: 1;
+    flex: none;
```
salarenko: `flex: none` is a shorthand for: `flex-grow: 0; flex-shrink: 0; flex-basis: auto;` and then grow and shrink values are additionally overwritten. 
I'd suggest using: `flex: 1 1 auto`

src/input-group.scss
```diff
@@ -60,7 +60,8 @@ $fd-input-border-radius: var(--sapField_BorderCornerRadius);
     @include fd-reset-child-spacing();
 
     border: none;
-    flex: 1;
+    flex: 1 1 auto;
+    width: auto;
```
InnaAtanasova: Do we need the `width: auto` when we have `flex: 1 1 auto`?

stefanoScalzo: apparently not...fixed

PR #540, opened 2019-12-18
fix: Add focus state to input group

JKMarkowski:
## Description
As we discussed, there was a need to add `is-focus` state on `InputGroup` component.

### Before:
There was not focus before.

### After:
image


src/input-group.scss
```diff
@@ -11,6 +11,12 @@ $block: #{$fd-namespace}-input-group;
 
 $fd-input-border-radius: var(--sapField_BorderCornerRadius);
 
+%fd-input-group-focus {
+  box-shadow: none;
+  outline-offset: -0.1875rem;
+  outline: var(--sapContent_FocusWidth) var(--sapContent_FocusStyle) var(--sapContent_FocusColor);
+}
```
InnaAtanasova: You can declare this as a mixin with @include for more options

Issue #528, opened 2019-12-17
Input states are placed behind the input


salarenko:
**Describe the bug**
Issue is related to Forms / Inputs states component.

Input states (Error, Warning, Valid...) are displayed behind input.

Issue appears on IE11.
Fundamental styles version: v0.4.0-rc.21

**To Reproduce**
Go to: _"/components/form.html"_, section Input States
Focus any input

**Expected behavior**
Messages displayed in popover under input component (like in Chrome)

**Screenshots**
Chrome
![image](https://user-images.githubusercontent.com/17496353/70983688-727c2a80-20b9-11ea-964e-cdbc370b8065.png)

IE11
![image](https://user-images.githubusercontent.com/17496353/70983533-31841600-20b9-11ea-9666-09089b957d33.png)

This subject has been partially covered in https://github.com/SAP/fundamental-styles/issues/483#issuecomment-562619461

PR #542, opened 2019-12-19
fix: Remove flex property from input-message-group IE11

JKMarkowski:
## Related Issue
Closes https://github.com/SAP/fundamental-styles/issues/528

## Description
So after some research, I came up to think, that `flex: 1` on IE11 cause problem with message-group container. It's changed to `display:flex`. Now also the example looks better, before the container was interpreted like it has `0px height`

### Before:
image

### After:
image

### DIFF
--- a/src/form-input-message-group.scss
+++ b/src/form-input-message-group.scss
@@ -9,6 +9,6 @@
 $block: #{$fd-namespace}-form-input-message-group;
 
 .#{$block} {
-  flex: 1;
+  display: flex;
   position: relative;
 }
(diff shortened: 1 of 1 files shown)


src/form-input-message-group.scss
```diff
@@ -9,6 +9,5 @@
 $block: #{$fd-namespace}-form-input-message-group;
 
 .#{$block} {
-  flex: 1;
```
salarenko: I think you forgot to add `display: flex`.
You additionally need to specify `flex-grow: 1`, to stretch inline labels and inputs.

image

Issue #527, opened 2019-12-17
Read-Only Input styles are not applied


salarenko:
**Describe the bug**
Issue is related to Forms / Input component.

Input component in Read-only state is not styled as on Chrome, Firefox..

Issue appears on IE11.
Fundamental styles version: v0.4.0-rc.21

**To Reproduce**
Go to: _".../components/form.html"_ Inputs section

**Screenshots**
Chrome 
![image](https://user-images.githubusercontent.com/17496353/70982723-c7b73c80-20b7-11ea-98fd-7c9f205589b4.png)

IE 11
![image](https://user-images.githubusercontent.com/17496353/70982652-a22a3300-20b7-11ea-9d35-482331ad710b.png)

PR #543, opened 2019-12-19
fix: Add hardcoded RGBA calue for readonly on forms

JKMarkowski:
## Related Issue
Closes https://github.com/SAP/fundamental-styles/issues/527

## Description
Problem with IE11 and the rgba values are well known. On our case, IE11 doesn't like variables, so it's taken in callback and transformed to raw value. Unfortunately one of the variables has rgba() value. The solution is to add hardcoded value, in case the variable/callback are not supported by explorer.

### Before:
image

### After:
image

### DIFF
--- a/src/mixins/_forms.scss
+++ b/src/mixins/_forms.scss
@@ -86,6 +86,7 @@
 
   &[readonly],
   &.is-readonly {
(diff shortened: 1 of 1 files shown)


InnaAtanasova:
I created an [issue](https://github.com/SAP/fundamental-styles/issues/547) for this and sent an email to the theme designer team

Issue #515, opened 2019-12-15
Inconsistent Styles for Checkbox and Radio (Safari/Chrome)


fkolar:
##BUG

**Describe the bug**

On Fundamentals-styles v0.3.0 that is currently deployed to the github pates, I see different visual behavior, Please see screenshots

<img width="1155" alt="Screen Shot 2019-12-15 at 8 23 07 PM" src="https://user-images.githubusercontent.com/17149942/70867728-be738600-1f78-11ea-9d95-b8402e2ec9ab.png">
<img width="1071" alt="Screen Shot 2019-12-15 at 8 22 51 PM" src="https://user-images.githubusercontent.com/17149942/70867730-c0d5e000-1f78-11ea-84c0-c7481374e961.png">

JKMarkowski:
Hello @fkolar,

We are in the middle of big refactor of forms and forms related component. It will be fixed on `0.4.0` release soon.

fkolar:
@JKMarkowski, cool. Just want to keep this for the records as in order version there is a issue and want to make sure that in 0.4.0 is fixed.

bcullman:
It is unclear which browser each screen shot is from. Please capture screen shots that include the browser shell/chrome as well.

PR #546, opened 2019-12-19
feat: checkbox and radio pseudo elements

JKMarkowski:
## Related Issue
Closes https://github.com/SAP/fundamental-styles/issues/515

## Description
Because of IE11, there was no way to add modify `:after` and `::before` pseudo elements inside input[checkbox/radio]. So now pseudo elements belongs to labels. 
The structure of inputs are changed now, before it was: 
```html
<label class="fd-form-label fd-form-label--radio" for="pDidh761">
    <input type="radio" class="fd-radio" id="pDidh761" name="radio">Field label
</label>
```
And after:
```html
<input type="radio" class="fd-radio" id="pDidh761" name="radio">
<label class="fd-radio__label" for="pDidh761">
    Field label
</label>
```

The `id` and `for` attributes are mandatory since now, because it's the only way we cah connect label with input.


src/checkbox.scss
```diff
@@ -12,195 +12,134 @@ $fd-checkbox-dimensions-compact: 1rem;
...
+      display: inline-flex;
+      justify-content: center;
+      align-items: center;
+      font-family: "SAP-icons";
+      background-color: var(--sapField_Background);
+      margin: $fd-checkbox-margin;
+      color: var(--sapSelectedColor);
+      border: var(--sapField_BorderWidth) solid var(--sapField_BorderColor);
+      border-radius: var(--sapField_BorderCornerRadius);
+      font-style: normal;
```
salarenko: Setting default values?
`font-style: normal;`
`font-weight: normal;`

src/mixins/_forms.scss
```diff
@@ -139,19 +139,35 @@
...
+  }
+}
 
-    box-shadow: none;
-    outline: none;
+@mixin fd-form-label() {
+  @include fd-reset();
+  @include fd-ellipsis();
+
+  width: fit-content;
```
salarenko: `width: fit-content;` is not supported in IE11 and Edge.
If label is always going to be flex child I recommend to use `align-self: flex-start`

src/radio.scss
```diff
@@ -15,147 +15,122 @@ $fd-radio-outer-circle-margin-compact: 0.5rem;
...
-      width: $fd-radio-inner-circle-diameter-compact;
-      height: $fd-radio-inner-circle-diameter-compact;
+  @extend %fd-radio-input-hidden;
+
+  &__label {
+    @include fd-form-label();
+
+    display: flex;
+    align-items: center;
+    cursor: pointer;
```
salarenko: Be consequent, use `@include action-cursor();` or `cursor: pointer;`, not both

src/radio.scss
```diff
@@ -15,147 +15,122 @@ $fd-radio-outer-circle-margin-compact: 0.5rem;
...
+      display: inline-flex;
+      justify-content: center;
+      align-items: center;
+      font-family: "SAP-icons";
+      background-color: var(--sapField_Background);
+      margin: $fd-radio-outer-circle-margin;
+      color: var(--sapSelectedColor);
+      border: var(--sapField_BorderWidth) solid var(--sapField_BorderColor);
+      border-radius: 50%;
+      font-style: normal;
```
salarenko: Default values?

JKMarkowski:
Thanks @salarenko for review, I added some changes due to your comments.

stefanoScalzo:
We have an issue for dark mode with the checkboxes
Screen Shot 2019-12-19 at 10 51 53 AM

src/radio.scss
```diff
@@ -7,155 +7,128 @@
...
+    @include action-cursor();
+
+    display: flex;
+    align-items: center;
+
+    &::before {
+      content: "";
+      height: $fd-radio-outer-circle-diameter;
+      width: $fd-radio-outer-circle-diameter;
+      font-size: 0.6875rem;
```
InnaAtanasova: Shouldn't this be `$fd-radio-inner-circle-diameter` ?

src/radio.scss
```diff
@@ -7,155 +7,128 @@
...
+    + .#{$block}__label {
+      pointer-events: none;
+      opacity: 0.4;
+    }
+  }
+
+  &--compact {
+    + .#{$block}__label {
+      &::before {
+        font-size: 0.4375rem;
```
InnaAtanasova: Maybe `$fd-radio-inner-circle-diameter-compact` ?

src/checkbox.scss
```diff
@@ -7,200 +7,136 @@
...
 
-    &:indeterminate {
-      &::after {
-        background-color: var(--sapField_SuccessColor);
-      }
-    }
+  &:indeterminate + .#{$block}__label::before {
+    padding-bottom: 0.0625rem;
+    content: "\e17b";
+    font-size: 0.875rem;
```
InnaAtanasova: It should be 0.75 x 0.75 rem sized square

src/checkbox.scss
```diff
@@ -7,200 +7,136 @@
...
 
-    @include fd-hover() {
-      @include fd-disabled() {
-        border-color: var(--sapField_SuccessColor);
-      }
-    }
+  &:indeterminate + .#{$block}__label--compact::before {
+    content: "\e17b";
+    padding-bottom: 0;
+    font-size: 0.5625rem;
```
InnaAtanasova: It should be 0.5 x 0.5 rem

InnaAtanasova:
There's padding-right and margin-right 
Screen Shot 2019-12-19 at 4 39 28 PM
Screen Shot 2019-12-19 at 4 39 10 PM
Screen Shot 2019-12-19 at 4 33 47 PM
Screen Shot 2019-12-19 at 4 20 11 PM

InnaAtanasova:
Fieldset legend has a pointer cursor on hover

src/checkbox.scss
JKMarkowski: It should be and it is. In general `font-size: 0.75rem` doesn't look good. It looks same as before, when we add 0.75rem + 0.0625rem (1px)

src/checkbox.scss
JKMarkowski: As above, 0.5rem doesn't look good.

JKMarkowski:
Hello @InnaAtanasova padding is removed now, there i only margin left. Legend cursor is auto, instead of pointer.

stefanoScalzo:
Screen Shot 2019-12-20 at 1 16 52 PM
The focus must be on the label and not full width

InnaAtanasova:
> Hello @InnaAtanasova padding is removed now, there i only margin left. Legend cursor is auto, instead of pointer. IE11 offset is fixed.

Why do we need the margin? A few months ago a team reported this as an issue. It's very noticeable when there's no label text 
Screen Shot 2019-12-20 at 1 50 00 PM

InnaAtanasova:
The Breaking changes need to be documented

Issue #531, opened 2019-12-17
Product switch for small screens is broken


salarenko:
**Describe the bug**
Error is related to Product Switch component in _mobile_ mode.

Product switch in mobile mode is broken. Products are collapsed and no title and subtitle is visible.

Error occurs on IE11.
Fundamental styles version: v0.4.0-rc.21

**To Reproduce**
Go to: _"/components/product-switch.html"_, Smallest Screen section

**Screenshots**
Chrome
![image](https://user-images.githubusercontent.com/17496353/70986576-76f71200-20be-11ea-81c9-0959ce2457bd.png)

IE11
![image](https://user-images.githubusercontent.com/17496353/70986508-516a0880-20be-11ea-801a-edf0f53602dd.png)

PR #550, opened 2019-12-20
fix: Broken product switch on mobile devices - IE11

salarenko:
## Related Issue
Closes SAP/fundamental-styles#531

## Description
IE11 has problem with `flex-basis` value. For some reason IE11 doesn't respect elements `height` value if `flex-basis: 25%` is used.
Since in mobile mode products are aligned as a list, my solution is changing to `flex-basis: auto` on mobiles.

To be honest product-switch styles are a little bit messy. I'd recommend small refactor in the future.

### Before:
image

### After:
image

### DIFF
--- a/src/product-switch.scss
+++ b/src/product-switch.scss
@@ -79,6 +79,7 @@ $block: #{$fd-namespace}-product-switch;
       padding: 1rem;
       text-align: left;
       border-radius: 0;
+      flex-basis: auto;
 
       &.selected {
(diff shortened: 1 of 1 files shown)


InnaAtanasova:
> 
> To be honest product-switch styles are a little bit messy. I'd recommend small refactor in the future.
> 

@salarenko , could you work on this after the holidays? Any improvements are more than welcome

Issue #554, opened 2019-12-20
button as addon in input group is cut


droshev:
It came as a result of PR: #539 
<img width="1147" alt="Screen Shot 2019-12-20 at 4 51 25 PM" src="https://user-images.githubusercontent.com/4380815/71295250-19c2c100-2349-11ea-90b1-5b9e282f9bf1.png">

We should create a visual test for that as part of the fix

PR #558, opened 2019-12-30
fix: (InputGroup) Add min-width property only for ie10+, fix addon truncation

JKMarkowski:
## Related Issue
Closes https://github.com/SAP/fundamental-styles/issues/554

## Description
We had some bug on input-group component, which was caused by setting `auto` property for `flex-basis`. It was mostly because of IE11, which doesn't have any native `min-width` for `inputs` 

### Before:
IE11:
image

All browsers:
image

### After:
IE11:
image

All browsers:
image

### DIFF
--- a/src/input-group.scss
+++ b/src/input-group.scss
@@ -72,7 +72,7 @@ $fd-input-border-radius: var(--sapField_BorderCornerRadius);
 
     background-color: transparent;
     border: none;
-    flex: 1 1 auto;
+    flex: 1;
     padding-right: 0.25rem;
     padding-left: 0.25rem;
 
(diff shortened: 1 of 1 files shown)


src/input-group.scss
```diff
@@ -72,7 +72,7 @@ $fd-input-border-radius: var(--sapField_BorderCornerRadius);
 
     background-color: transparent;
     border: none;
-    flex: 1 1 auto;
+    flex: 1 1;
```
InnaAtanasova: flex: 1;

JKMarkowski: Done

PR #559, opened 2020-01-03
docs: landing page

InnaAtanasova:
This is the implementation of the landing page. There are a few minor things to be changed when Leo gets back from vacation (update 1-2 logos and the background images in the first section and "More About Fundamental Library Setion). Here are the screenshots from Zeplin (the visuals):

Desktop:
Screen Shot 2020-01-03 at 1 56 12 PM
Screen Shot 2020-01-03 at 1 57 39 PM
Screen Shot 2020-01-03 at 1 57 34 PM
Screen Shot 2020-01-03 at 1 57 21 PM

Mobile:
Screen Shot 2020-01-03 at 1 56 50 PM
Screen Shot 2020-01-03 at 1 56 43 PM
Screen Shot 2020-01-03 at 1 56 37 PM
Screen Shot 2020-01-03 at 1 56 28 PM


estelaxu3:
It looks awesome, thank you for bring it to life!

Couple things I noticed:
1. Key feature content should be left aligned instead of centered. 
Zeplin file:
Screen Shot 2020-01-03 at 2 13 04 PM

2. Could we add hover effect to the library tiles? All four tiles are currently outlined and have shadows, maybe we can take out the shadow when the tile is not selected so it is more interactive and responsive.

Vanessa-Cusmich:
Agreed, it looks great!

I was also thinking we should remove the links below the social media logos and hyperlink the images.

Thanks!
Vanessa

InnaAtanasova:
> It looks awesome, thank you for bringing it to life!
> 
> Couple things I noticed:
> 
> 1. Key feature content should be left aligned instead of centered.
>    Zeplin file:
>    Screen Shot 2020-01-03 at 2 13 04 PM
> 2. Could we add hover effect to the library tiles? All four tiles are currently outlined and have shadows, maybe we can take out the shadow when the tile is not selected so it is more interactive and responsive.

Done

InnaAtanasova:
> Agreed, it looks great! I was also thinking we should remove the links below the social media logos and hyperlink the images. Thanks! Vanessa >.

Done

bcullman:
pls. recheck layout on mobile - footer text is overlapping.

InnaAtanasova:
> pls. recheck layout on mobile - footer text is overlapping.

It's fixed now. Was a Safari issue. Thanks for catching it

docs/_sass/_landingPage.scss
```diff
@@ -0,0 +1,721 @@
...
+
+@media (max-width: 1024px){
+    .features-section {
+        &__container {
+            grid-template-columns: auto auto;
+        }
+    }
+}
+
+@media (max-width: 768px){
```
JKMarkowski: What about adding only one media query per `max-width`?
From largest (1024px) to smallest (600px)

JKMarkowski: The rest is perfect.

InnaAtanasova: Doesn't look good without this breakpoint. O.w. it's not a problem to change it from 768 to 600px

InnaAtanasova: This is how the section looks like at 630px:
Screen Shot 2020-01-07 at 8 17 48 AM

This is how the section looks like compared to the other sections:
Screen Shot 2020-01-07 at 8 18 12 AM
Screen Shot 2020-01-07 at 8 18 32 AM

estelaxu3:
LGTM

Issue #493, opened 2019-12-05
Add grid layout


stefanoScalzo:
Feature requests must include:
Add form layout class for grid 471 and 12 column layout

**Is your feature request related to a problem? Please describe.**
There is no horizontal layout for forms and it does not give structure
> Is this issue related to a specific component?
Forms
> What browser are you working in?
chrome
> What version of Fundamental Library Styles are you using?
latest

Issue #297, opened 2019-08-27
Make layout-grid component IE11 compatible 


jbadan:
Blocked by https://github.com/SAP/fundamental-styles/issues/296

This article series shows how to do this. https://css-tricks.com/css-grid-in-ie-debunking-common-ie-grid-misconceptions/

The build changes to the playground in #296 will make this work much easier to accomplish.

InnaAtanasova:
This is being solved here https://github.com/SAP/fundamental-styles/pull/889

PR #585, opened 2020-01-10
feat: add preset layout grid with nesting 

stefanoScalzo:
## Related Issue
Closes SAP/fundamental-styles#297

## Description
Adding preset grids for users to use up to 16 columns and compliance to fiori 3 specs. In progress for IE

fd-layout-grid--gap-1 class removed

### Before:
Nothing before

### After:
Screen Shot 2020-01-10 at 11 38 59 AM


mikerodonnell89:
Needs visual tests updated

src/layout-grid.scss
```diff
@@ -8,12 +8,28 @@
...
-  grid-gap: var(--fd-width-gutter, $fd-width--gutter);
+  grid-template-columns: 100%;
+  grid-gap: 1rem;
+
+  @media only screen and (max-width: 599px) {
+    grid-gap: 0.5rem;
+  }
+
+  * {
+    height: fit-content;
```
salarenko: We shouldn't use `fit-content` in our styles. It is [not supported](https://caniuse.com/#search=fit-content) in many browsers

JKMarkowski:
@stefanoScalzo The main problem there is ofc IE11, which doesn't entirely support [grid](https://caniuse.com/#search=grid). Maybe let's follow the bootstrap 4. There is grid system based on flex. It should work on ie11.

stefanoScalzo:
Goign to fixed in a pr that does not use css grids

Issue #204, opened 2019-07-24
Refactor Shellbar to match Fiori 3 specs and the sap-theming


InnaAtanasova:
**Description of the bug**

> Is this issue related to a specific component?
ShellBar

**Expected behavior**
- Container padding L: padding is 0 20px 0 32px; Fiori 3 padding is 0 2rem;
- Container padding S: padding is 0 12px 0 16px; Fiori 3 padding is 0 1rem;
- ShellBar buttons: in Fiori 3, icon buttons, icon button-image, menu button have focus, regular, hover, active, and disabled states. Buttons in Fundamental-styles have only regular state. The spacing for menu button is incorrect. 
- Second title: The font-size in Fundamental-styles is .85714rem which results in 13.71424px. In Fiori 3 the size is @sapMFontSmallSize (12px). 
- Search: In Fundamental-styles the Search input and icon button is one element, while in Fiori 3 we have 3 separate elements - Input, Icon Button and Text Button. The placeholder text is different. The input element has only regular state in Fundamental-styles, while Fiori 3 has regular, hover, active, disabled states, focus, and clear button.
- Search overlay - When the Search is triggered the content is covered with an Overlay:
`background: fade(@sapUiBlockLayerBackground, 60)`; This overlay is missing in Fundamental-styles.

InnaAtanasova:
Also consider this [issue](https://github.com/SAP/fundamental-styles/issues/478)

PR #586, opened 2020-01-12
feat: refactor Shellbar component to match Fiori 3 specs

mikerodonnell89:
## Related Issue
Closes SAP/fundamental-styles#204

## Description
Update shellbar to match Fiori 3 specs

### Before:
Screen Shot 2020-01-12 at 12 23 02 PM
Screen Shot 2020-01-12 at 12 22 53 PM
Screen Shot 2020-01-12 at 12 22 47 PM
Screen Shot 2020-01-12 at 12 22 42 PM

### After:
Screen Shot 2020-01-12 at 9 56 54 PM
Screen Shot 2020-01-16 at 5 30 49 PM

Screen Shot 2020-01-12 at 12 21 30 PM
Screen Shot 2020-01-12 at 12 21 22 PM


docs/_sass/_landingPage.scss
```diff
@@ -631,13 +631,6 @@ body {
     color: $home-page-action-color;
 }
 
-.sap-icon--decline::before {
-    font-size: 20px;
-    transition: opacity .3s, transform .3s;
-    transform: rotate(-180deg);
-    color: $home-page-action-color;
-}
-
```
InnaAtanasova: Why do you remove this? It's part of the landing page

src/shellbar.scss
```diff
@@ -14,23 +14,34 @@ $block: #{$fd-namespace}-shellbar;
...
+  $fd-shellbar-color: var(--sapShell_TextColor); // White
+  $fd-shellbar-background-color: transparent;
+  $fd-field-border-color: var(--sapField_BorderColor);
+  $fd-shell-active-text-color: var(--sapShell_Active_TextColor);
+
+  // Background
+  $fd-shellbar-background: var(--sapShellColor);
+  $fd-shell-hover-background: var(--sapShell_Hover_Background);
+  $fd-shell-active-background: var(--sapShell_Active_Background);
+  $fd-content-badge-background: var(--sapContent_BadgeBackground);
```
InnaAtanasova: You should have !default after the variables

src/shellbar.scss
```diff
@@ -14,23 +14,34 @@ $block: #{$fd-namespace}-shellbar;
...
+
+  // Box shadow
+  $fd-shellbar-box-shadow: inset 0 -0.0625rem 0 0 var(--sapShellBorderColor);
 
   @include fd-reset();
   @include fd-var-color("background-color", $fd-shellbar-background-color, --fd-color-shell-1);
```
InnaAtanasova: You need to remove all occurrences of `fd-var-color` and `fd-color`...

src/shellbar.scss
```diff
@@ -41,23 +52,23 @@ $block: #{$fd-namespace}-shellbar;
...
 
   @include fd-screen(xl) {
-    padding: $fd-shellbar-outer-spacing-tb $fd-shellbar-outer-spacing-lr-large;
+    padding: $fd-shellbar-outer-spacing-tb $fd-shellbar-outer-spacing-lr;
   }
 
   @include fd-screen(xxl) {
-    padding: $fd-shellbar-outer-spacing-tb $fd-shellbar-outer-spacing-lr-large;
+    padding: $fd-shellbar-outer-spacing-tb $fd-shellbar-outer-spacing-lr;
   }
```
InnaAtanasova: looks like the padding doesn't change, you can remove this and just add `padding: 0 1rem;`

src/shellbar.scss
```diff
@@ -75,12 +86,36 @@ $block: #{$fd-namespace}-shellbar;
     &--copilot {
       justify-content: center;
       order: 2;
+
+      .fd-shellbar__button {
```
InnaAtanasova: .#{$block}__button

src/shellbar.scss
```diff
@@ -90,16 +125,32 @@ $block: #{$fd-namespace}-shellbar;
     @include fd-var-color("color", fd-color("shell", 2), --fd-color-shell-2);
```
InnaAtanasova: this should be removed

src/shellbar.scss
```diff
@@ -114,6 +165,8 @@ $block: #{$fd-namespace}-shellbar;
 
     @include fd-var-color("color", $fd-shellbar-color, --fd-color-text-5);
```
InnaAtanasova: this should be removed
Also 
@include fd-type("1");
@include fd-weight("bold");
line-height: 1.1;
These should be removed and replaced by values coming from the specs

src/shellbar.scss
```diff
@@ -229,5 +286,37 @@ $block: #{$fd-namespace}-shellbar;
...
+
+        &:hover {
+          background-color: $fd-shell-hover-background;
+          border-color: $fd-field-border-color;
+        }
+
+        .fd-input-group__input {
+          border: none;
+          color: $fd-shellbar-color;
+        }
```
InnaAtanasova: `fd` should be replaced by `#{$fd-namespace}`
#{$fd-namespace}-counter--notification

src/shellbar.scss
```diff
@@ -151,13 +204,17 @@ $block: #{$fd-namespace}-shellbar;
...
+  &__title {
+    padding-left: 0.5rem;
+  }
+
   &__logo {
     @include fd-reset();
 
     display: flex;
     align-items: center;
     vertical-align: middle;
```
InnaAtanasova: I don't think you need these 3 lines

InnaAtanasova:
Screen Shot 2020-01-13 at 5 00 20 PM
In the playground you need to add a ShellBar specific class to the `input-group-input-playground`, like the button case: `<button class="btn-playground fd-shellbar__button">%%</button>`

InnaAtanasova:
Please add some css variables to `theme-ugly.css` to test the theming in the Playground

InnaAtanasova:
Screen Shot 2020-01-13 at 5 03 13 PM

Check the element spacing in RTL

docs/_sass/_landingPage.scss
mikerodonnell89: It was causing issues with the decline button on the shellbar search input, and I could not find where this was actually being used in the docs so I thought it might've been leftover... where would this icon be found on the landing page?  I'm not seeing it

JKMarkowski:
Title is not vertically aligned
image

mikerodonnell89:
> For MegaMenu Button you don't need `fd-product-menu__control` class. You need to use the Menu Button component `<button class="fd-button--light fd-button--menu">Product Title</button>` and use the same overwrite rules to apply the dark background and remove the border on hover.

@InnaAtanasova I may be mistaken, but isn't the `_product-menu.scss` code specifically meant for this button?  Should I rename everything in that file to match fd-button--menu (fd-shellbar__button--menu or something like that)

bcullman:
i am concerned that the red counter is appearing slightly outside the borders of the shellbar.

The shellbar will often be the top element on the screen. I would hope this does not push the shellbar down.

mikerodonnell89:
@bcullman good catch, looks like my version did not match the specs, I have updated the code and updated the screenshot

src/shellbar.scss
```diff
@@ -14,50 +14,56 @@ $block: #{$fd-namespace}-shellbar;
...
 
   @include fd-screen(xl) {
-    padding: $fd-shellbar-outer-spacing-tb $fd-shellbar-outer-spacing-lr-large;
+    padding: 0 1rem;
   }
 
   @include fd-screen(xxl) {
-    padding: $fd-shellbar-outer-spacing-tb $fd-shellbar-outer-spacing-lr-large;
+    padding: 0 1rem;
   }
```
InnaAtanasova: The padding is the same for all screens. Can we remove these lines and just add `padding: 0 1 rem;` to the `fd-shellbar` class?

src/shellbar.scss
```diff
@@ -149,15 +186,38 @@ $block: #{$fd-namespace}-shellbar;
...
-    vertical-align: middle;
-    max-height: fd-space(10);
+    padding-right: 0;
+    padding-left: $fd-shellbar-item-spacing;
+
+    @include fd-rtl() {
+      padding-right: $fd-shellbar-item-spacing;
+      padding-left: 0;
+    }
+
```
InnaAtanasova: The logo doesn't have padding-left. All elements are "separated" by `Element Spacing` which is element G in the specs. If there's no Back Button (A1) you don't have padding-left. If there's a Back Button, the Back Button itself should have padding-right playing the role of an Element Spacing
Screen Shot 2020-01-21 at 11 16 45 AM

src/shellbar.scss
```diff
@@ -229,5 +289,64 @@ $block: #{$fd-namespace}-shellbar;
...
+        &:hover {
+          background-color: $fd-shell-hover-background;
+        }
+      }
+    }
+  }
+
+  &__button--menu {
+    &::after {
```
InnaAtanasova: The font-size of the Menu Button is not correct. This is an issue not related to this PR, it comes from the Button component where for some reason the correct font-size is overwritten to 14px. It should be @sapMFontMediumSize which is .875rem. If you want you can specify the correct font-size here. It should follow the Button component specs

src/shellbar.scss
```diff
@@ -75,45 +81,79 @@ $block: #{$fd-namespace}-shellbar;
...
   &__subtitle {
     @include fd-reset();
-    @include fd-type("1");
-    @include fd-weight("bold");
-
-    line-height: 1.1;
-
-    @include fd-var-color("color", $fd-shellbar-color, --fd-color-text-5);
 
+    font-size: var(--sapFontSmallSize);
```
InnaAtanasova: The font-size is not the same for title and subtitle

InnaAtanasova:
Screen Shot 2020-01-21 at 11 46 22 AM
The button's border radius is not correct. It shouldn't have border-radius on the left side

InnaAtanasova:
Screen Shot 2020-01-21 at 11 52 21 AM
The icon is not in the middle

InnaAtanasova:
Screen Shot 2020-01-21 at 11 57 59 AM
Screen Shot 2020-01-21 at 11 56 10 AM
Screen Shot 2020-01-21 at 11 56 21 AM

The popover body in these cases should not have arrow. Only the Product Switch popover body has arrow

src/shellbar.scss
mikerodonnell89: Should I add an example of a shellbar with a back button?

mikerodonnell89:
> Screen Shot 2020-01-21 at 11 52 21 AM
> 
> The icon is not in the middle and the size of the button doesn't look right

@InnaAtanasova where are you seeing this?  Can't recreate on chrome or safari

mikerodonnell89:
> Screen Shot 2020-01-21 at 11 52 21 AM
> 
> The icon is not in the middle and the size of the button doesn't look right

@InnaAtanasova This is caused by `<span class="fd-shellbar__title"></span>`'s `::after`, I think in this case if the dev wanted just the down caret with no title, they would use a button, not an empty title span?

stefanoScalzo:
Need to add ellipses
Screen Shot 2020-01-23 at 9 33 40 AM
Screen Shot 2020-01-23 at 9 27 57 AM
Screen Shot 2020-01-23 at 9 28 24 AM


 and fix the color for the focus on the logo

stefanoScalzo:
Screen Shot 2020-01-23 at 11 56 26 AM
Shouldn't the buttons be following the light button specs? Looks weird how the focus color changes and isn't that visible

mikerodonnell89:
> Screen Shot 2020-01-23 at 11 56 26 AM
> 
> Shouldn't the buttons be following the light button specs? Looks weird how the focus color changes and isn't that visible

The specs reference shellbar-specific color variables so I think this is intentional

src/shellbar.scss
```diff
@@ -75,45 +62,78 @@ $block: #{$fd-namespace}-shellbar;
...
 
-    @include fd-active-pressed-selected() {
-      background-color: hsl(217, 21%, 21%);
+    &:hover {
+      background: $fd-shell-hover-background;
     }
 
-    @include fd-hover() {
-      background-color: hsl(217, 21%, 25%);
+    &:active {
```
salarenko: Shouldn't we use `@fd-hover`,  `@fd-active`... mixins here? I'm not sure so that's rather the question for more experienced ones with styles (@InnaAtanasova). Anyway I think we should choose one approach, using mixin or explicit selectors.

docs/pages/components/shellbar.md
```diff
@@ -89,62 +89,49 @@ This example includes the product menu for navigating to applications within the
...
-                    <li><a href="#" class="fd-menu__item"><strong>Sear</strong>ch Result B</a></li>
-                    <li><a href="#" class="fd-menu__item"><strong>Sear</strong>ch Result C</a></li>
-                    <li><a href="#" class="fd-menu__item"><strong>Sear</strong>ch Result D</a></li>
-                </ul>
-              </nav>
+    <div class="fd-shellbar__action fd-shellbar__action--desktop">    
+      <div class="fd-popover__control">
+        <div aria-label="Image label" aria-controls="F4GcX348b" aria-expanded="false" aria-haspopup="true">
+          <div class="fd-input-group fd-shellbar__input-group">
+            <input type="text" class="fd-input fd-input-group__input fd-shellbar__input-group__input" id="F4GcX348b1" value="Sear" placeholder="Search...">
```
salarenko: Please remove or complete the value from search inputs, it looks weird in the documentation (`value="Sear"`).

salarenko:
There is also a problem with this particular example in IE11. The rest of examples presented in documentation looks correctly.

image

PR #587, opened 2020-01-14
docs: Change usage of checkbox in multi input example

JKMarkowski:
## Description
There was wrong checkbox usage inside multi-input pattern.

### Before:
image

### After:
image


InnaAtanasova:
That's a good catch! It's nice you caught in before the release

Issue #590, opened 2020-01-16
form messages not placed properly


stefanoScalzo:
**Describe the bug**
The margin on the text area is shifting the message downwards. Also due to the border width, the message margin at the top must be recalculated

> Is this issue related to a specific component?
text area
> What version of the Fundamental Library Styles are you using?
latest
> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

> What front-end framework are you implementing in, e.g., Angular, React, etc.?
angular

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

PR #591, opened 2020-01-16
fix: form messages margin fix for textarea

stefanoScalzo:
## Related Issue
Closes SAP/fundamental-styles#590

## Description
Putting margins to make form messages and shadow fiori 3 compliant. Shadow doesn't have the same visuals as UI5 BUT is right according to specs. Did not include input screenshots because since it is behind the border they won't be visible

### Before:
Screen Shot 2020-01-16 at 4 27 36 PM

### After:
Screen Shot 2020-01-16 at 4 34 16 PM


docs/pages/components/form.md
```diff
@@ -156,27 +156,28 @@ The state of the input field can reflect validity of the data entered, whether t
 Along with Invalid and Warning, error messages should be displayed below the field so the user can correct the error and move forward.
 
 {% capture inputs %}
-<div class="fd-form-item fd-has-margin-none fd-popover">
+<div class="fd-form-item fd-popover">
```
InnaAtanasova: You actually don't need `fd-popover`. This example doesn't have a message

InnaAtanasova:
@stefanoScalzo is off Monday so I pushed some changes into his PR (removed classes that we don't need, add `cursor: text` for textarea. 
I think it looks good. @JKMarkowski could you also take a look at it and if it's ok you can merge it.

PR #593, opened 2020-01-17
fix: update modal docs to remove .modal-demo-bg

jbadan:
## Description
This class was only referenced in `customstyles.css`, which is not used on the docs site (it is only called in `head_print.html`.

This was causing confusion as to whether that div was actually necessary.  

### Before:
Screen Shot 2020-01-17 at 10 01 51 AM

### After:

Screen Shot 2020-01-17 at 10 02 15 AM


docs/css/customstyles.css
```diff
@@ -1102,11 +1102,6 @@ h4.panel-title {
   text-align: center;
 }
 
-.modal-demo-bg{
-    background-color:rgba(0, 0, 0, 0.5);
-    padding: 40px;
-}
-
```
jbadan: @InnaAtanasova do you think we can just remove this file entirely and cleanup `head_print.html`? Or maybe even delete `head_print`? It's referencing a ton of css that doesn't exist anymore and I'm not sure we still need it.

Issue #562, opened 2020-01-07
Max width for button


stefanoScalzo:
**Describe the bug**
There should be a max-width on the button or else it may go out of the container with overflow.
> Is this issue related to a specific component?
Button
> What version of the Fundamental Library Styles are you using?
Latest
> What front-end framework are you implementing in, e.g., Angular, React, etc.?

Angular

**To Reproduce**
1. Go to '...' Button component
2. Make text super long

**Expected behavior**
There should be ellipses but there isn't

**Screenshots**
![Screen Shot 2020-01-07 at 10 37 42 AM](https://user-images.githubusercontent.com/50607147/71907483-22353d00-313a-11ea-9840-80fc77750ba6.png)

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] ios
 - Browser [e.g. chrome, safari] : chrome

sKudum:
Could you please elaborate the issue? Are we expecting truncation and Ellipsis ?  fixed width?

PR #600, opened 2020-01-22
fix(button): 562 button responsive long text

sKudum:
## Related Issue
Closes SAP/fundamental-styles#562

## Description
long text on the button will be truncated and shown in ellipse formate based on the responsive mode.

### Before:

Screen Shot 2020-01-22 at 10 21 26 PM

### After:
Screen Shot 2020-01-22 at 10 22 51 PM

### DIFF
--- a/docs/pages/components/button.md
+++ b/docs/pages/components/button.md
@@ -240,3 +240,14 @@ Button with multiple actions
 <button class="fd-button--compact fd-button--menu">Compact</button>
 {% endcapture %}
 {% include display-component.html component=button-menu %}
+
+## Button With Text Overflow
+
+Content overflow in the button will be shown with ellipsis
+
+{% capture button %}
+<div style="width:319px;">
+<button class="fd-button--compact fd-button--positive">Compact button with looooooooooooooooooooooooooooooooooooooooooooooonger text</button>
+</div>
+{% endcapture %}
+{% include display-component.html component=button %}
--- a/src/button.scss
+++ b/src/button.scss
@@ -2,9 +2,18 @@
 @import "./mixins";
 @import "./functions";
 
+@mixin fd-button-responsive() {
+  white-space: nowrap;
+  max-width: 100%;
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
+
 @mixin isDisabled() {
   opacity: 0.4;
   cursor: not-allowed;
+
+  @include fd-button-responsive();
 
   &:focus {
     outline: none;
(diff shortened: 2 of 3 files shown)


stefanoScalzo:
Why is it that for popover it still overflows?

sKudum:
> Why is it that for popover it still overflows?

Could you please elaborate -Popover.
Because this fix is to truncate only long text present inside a button.

sKudum:
Hi @InnaAtanasova ,

I did not test the animation on the button for this change in the playground. 
Because below is my understanding.
1). A component that has an animation effect that has to be tested with an Animation option.
2).Themes: in case I did any color changes then I need to test it with different themes.
3).Touch: needs to tested when my component has a touch feature as well. 
Other options like normalize, unnormalized, RTL, compact are tested.

docs/pages/components/button.md
```diff
@@ -240,3 +240,13 @@ Button with multiple actions
 <button class="fd-button--compact fd-button--menu">Compact</button>
 {% endcapture %}
 {% include display-component.html component=button-menu %}
+
+## Button With Text Overflow
+
+Content overflow in the button will be shown with ellipsis
+
+{% capture button %}
+<button class="fd-button">Default button with longer text</button>
```
droshev: I don't see ellipsis in this example
Screen Shot 2020-01-23 at 8 33 10 AM

sKudum: Hi Deno, 
     As requested in the bug truncation is applied based on the parent container width.
For example.
When we decrease screen width in inspect mode mobile view or tab with or with lesser screen width we can see the truncation effect.

sKudum: Sure, In button.md I will remove the duplicate example and update with a different examples.

droshev: > Hi Deno,
> As requested in the bug truncation is applied based on the parent container width.
> For example.
> When we decrease screen width in inspect mode mobile view or tab with or with lesser screen width we can see the truncation effect.

What is the value of adding another example if we cannot show it in the documentation.
Btw the menu button spec has more requirements in terms of `max-width`

sKudum: Added example with lesser parent width which shows the truncation.

sKudum: As per firori when we have a menu in the split button then the max-width is 12rem after that truncation has to be applied. For this, we have a bug.
https://github.com/SAP/fundamental-styles/issues/602.

droshev:
@sKudum can you wait for the PR https://github.com/SAP/fundamental-styles/pull/509 to be merged as it is refactoring the button. Also the playground tests don't bring much value

sKudum:
> @sKudum can you wait for the PR #509 to be merged as it is refactoring the button. Also the playground tests don't bring much value

Sure, will wait for #509 and regarding playground test cases- we don't need to have them at least for few button options

sKudum:
> Does it make sense to double the visual tests for that? also i don't see ellipsis anywhere

Changes done

droshev:
> > Does it make sense to double the visual tests for that? also i don't see ellipsis anywhere
> 
> Changes done

I still see 120 files for the visual tests

droshev:
@sKudum I think it is better to close the PR until we get #509 merged

sKudum:
> @sKudum I think it is better to close the PR until we get #509 merged

Hi @droshev , May I know the reason please.

droshev:
> > @sKudum I think it is better to close the PR until we get #509 merged
> 
> Hi @droshev , May I know the reason please.

The `button` component is under refactoring and most probably what you want to achieve will done with  #509. 
Also, the fix in this PR is not covered by any test and it doesn't work for the `menu button`

sKudum:
Closing this PR as suggested by Deno.

PR #601, opened 2020-01-22
fix: product switch list style

stefanoScalzo:
## Related Issue
#403 

## Description
To add a default list style being none to remove dots being used outside of styles

## Screenshots
In styles default does not show bullets, therefore, will be shown in other libraries


stefanoScalzo:
Will be fixed in implementation

PR #603, opened 2020-01-23
docs: updated and improved landingpage

trilodge:
## Description 

- added semantics
- added headline structure
- fixed markup
- improved acccessibility
- improved performance from 2.8mb page size down to under 1mb
- optimized css
- removed doubled main navigation for mobile
- optimized images via imageOptim and image formats
- replaced framework images with svgs


InnaAtanasova:
Screen Shot 2020-01-23 at 9 40 32 AM

In this case it's probably better to use margin-right instead of padding-left to add space between the links, o.w. the focus looks a bit weird

InnaAtanasova:
Screen Shot 2020-01-23 at 9 42 16 AM
Screen Shot 2020-01-23 at 9 40 20 AM

Some links are underlined and some are not when they are focused. It's better to make them consistent.

stefanoScalzo:
Should focused elements be underlined? 
Screen Shot 2020-01-23 at 9 45 27 AM

trilodge:
I added the underline for :focus, because the CSS was lacking focus styles at all. The outline isn't enough. If there is a guideline on how :focus and :active should look like i would update the css accordingly.

PR #605, opened 2020-01-23
[WIP] feat: responsive grid layouts

stefanoScalzo:
## Related Issue

## Description
This pull request will change the layout grid's since the ones we currently have are 1-not responsive and 2- work with css grids which are not supported in IE. These concepts use width, padding, flex and margin's to create the grids. Each of these classes will be applied to that column and does not represent the whole grid. It is also associated to the fiori specs for there responsive grid's hence why I am defining a size for the grid's and they are not being defined per application.

Classes removed: fd-layout-grid--gap-1
fd-layout-grid__span-row-(2,3,4,5,6)
fd-layout-grid__span-column-(2,3,4,5,6)
## Screenshots
No screenshots as it is the markup and application that is changed


src/layout-grid.scss
```diff
@@ -8,39 +8,139 @@
...
+.#{$block}-container {
+  @include fd-reset();
+
+  display: flex;
+  flex-wrap: wrap;
+  flex-direction: row;
+  max-width: 100%;
+  position: relative;
+
+  @media (min-width: 0) {
```
mikerodonnell89: You can use the variable `$fd-breakpoints` for these media query values

Issue #604, opened 2020-01-23
Refactor Combobox to match latest Fiori3 specs


JKMarkowski:
Combobox needs to match the latest Fiori3 designs.
The design has been updated since we implemented this component. The current implementation of the component is using the css variables from @sap-theming.

Issue #596, opened 2020-01-20
Multi-input list not wrapping  for very long text 


sunnyyadav100:
**Describe the bug**

Multi-Input component list is not gracefully handling very long text.

<img width="1470" alt="image" src="https://user-images.githubusercontent.com/9139037/72709700-794ffe80-3b8b-11ea-88c6-bd8b889d0321.png">

> Is this issue related to a specific component?

https://sap.github.io/fundamental-styles/patterns/multi-input.html

> What version of the Fundamental Library Styles are you using?

0.4.1

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

> What front-end framework are you implementing in, e.g., Angular, React, etc.?

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

JKMarkowski:
Hi @sunnyyadav100, currently we are working on adapting dropdowns (multi-input, combobox, select) To fiori3 specifications. It's supposed to be pushed in v0.6.0 in next 2 weeks.

PR #609, opened 2020-01-27
feat: Combobox, MultiInput, Select fiori3

JKMarkowski:
## Related Issue
fixes: https://github.com/SAP/fundamental-styles/issues/596
fixes: https://github.com/SAP/fundamental-styles/issues/572
fixes: https://github.com/SAP/fundamental-styles/issues/573
fixes: https://github.com/SAP/fundamental-styles/issues/604

## Description
In this PR I removed old `dropdown.scss` and `multi-input.scss`. It got replaced by new modifiers for `list` component, which are `--multi-input` and `--dropdown`.

### Dropdown Component usage:
For combobox, it will use `popover`, `input-group` and **`list with --dropdown modifier`** component
Multi Input - now it uses `popover`, `token`, `tokenizer`, `input-group` and **`list with --mult-input modifier`** components
Select has it's own styling, so it uses `popover`, `select` and  **`list with --dropdown modifier`** components

### New featues:
To adapt some components, that can be used along with Combobox, MultiInput and Select, I had to add some modifiers.

**Select**
Select got new class `fd-select__button`. Before it was pseudo element, but we should get rid of  pseudo elements unnecessary usage.

**InputGroup**
`fd-input-group` class got new modifier `fd-input-group--control`, to reacts on `hover`/`active` effects.
`fd-input-group__button` should have `is-expanded` class, when popover is expanded.

**Popover**
To add limitations of width, specified in documentation, there is new modifier for class `fd-popover__body`, it's `fd-popover__body--dropdown`.

**Form Messages**
Before messages were used only inside `popover` component. When it's used with Select, MultiInput, Combobox, it should stay visible all the time, so there is added `fd-form-message--static` modifier.

### Docs
JS functionality on our docs page had to be changed, actions with `aria-controls` for dropdowns will propagate to all elements with the same `aria-controls ID`.
 
### Test Cases
Tests for `dropdown` and `multi-input` are inside `list` component.

### Removed classes
* `fd-dropdown` class and file is completely removed
* `fd-multi-input` class and file is completely removed

### Before:

Combobox
image

Compact Combobox
image

MutliInput
image

Compact MultiInput
image

Select
image

Compact Select
image

### After:

Combobox:
image

Compact Combobox with more text:
image

Combobox with Message:
image

Combobox Disabled:
image

combobox readonly:
image

Multi Input  with more text
image

Multi Input Compact:
image

REadonly Multi Input with Popover
image

Disabled Multi Input
image

Multi Input with Message
image

Select:
image

Compact Select with more text:
image

Two Columns Select:
image

Disabled Select
image

Readonly Select:
image

Semantic States with Messages
image

image


estelaxu3:
Visuals looking good to me.

JKMarkowski:
Thanks @estelaxu3, Anyway the PR's code and preview are ready for review.

src/input-group.scss
```diff
@@ -201,6 +201,31 @@ $fd-input-border-radius: var(--sapField_BorderCornerRadius);
 
     border: none;
     width: 100%;
+
```
salarenko: Please update component structure description (Same for the rest of edited components)
```
/*!
.fd-input-group+(--inline)
    .fd-input-group__addon+(--button)
*/
```

src/input-group.scss
```diff
@@ -201,6 +201,31 @@ $fd-input-border-radius: var(--sapField_BorderCornerRadius);
...
+    }
+  }
+
+  &--control {
+    @include fd-hover() {
+      .#{$block}__button {
+        background-color: var(--sapButton_Lite_Hover_Background);
+
+        &.is-expanded {
+          background-color: var(--sapButton_Active_Background);
```
salarenko: Shouldn't this be refactored to use SCSS variables initialized with CSS variables or values at the top of this selector?  (Same for the rest of edited components)

JKMarkowski:
> * Disabled Select, Combobox and Multi-input can be opened using keyboard
>   image

I don't think if there should be anything preventing from opening popover in such a case.

> * Combobox and Multi-input input button doesn't take full height
>   image

It happens only, when you zoom in/out

> * Missing class that makes the input wide (also disabled example)
>   image

It's fixed now, there is no wide class anywhere anymore

>   image
> * Please describe classes used for setting components in specified visual state. Is there no for example `is-disabled`? We should only base on `aria`?

:+1:

src/input-group.scss
JKMarkowski: Done

src/input-group.scss
JKMarkowski: Done

InnaAtanasova:
Screen Shot 2020-02-03 at 4 55 18 PM

The content of the "body" is cutting the border-radiuses of the popover

InnaAtanasova:
Screen Shot 2020-02-03 at 4 54 42 PM
The outline should be white?

InnaAtanasova:
Screen Shot 2020-02-03 at 4 58 49 PM

The dropdown body should have border-radius only at the bottom.

JKMarkowski:
> I'm not seeing the bottom border on the selected item for this theme in the specs
> 
> Screen Shot 2020-02-03 at 5 27 18 PM

We had a talk with Inna about it and it should be there.
Combobox and multi input should share body with select/list

droshev:
@estelaxu3 could you take a look on this component?

estelaxu3:
In multi input, the icon is styles is not matching Fiori specs:
Styles:
Screen Shot 2020-02-06 at 10 06 02 AM

Fiori3:
Screen Shot 2020-02-06 at 10 06 21 AM

estelaxu3:
Also, I noticed that in Search input, the visuals is not completely matched. The first image is from Fiori3 and second is from Styles.
Screen Shot 2020-02-06 at 10 21 48 AM

JKMarkowski:
@estelaxu3 search input is deprecated on our site, icon will be change in example.

estelaxu3:
The rest looks good to me.

JKMarkowski:
> In multi input, the icon on the right corner in Styles is not matching Fiori specs:
> Styles:
> Screen Shot 2020-02-06 at 10 06 02 AM
> 
> Fiori3:
> Screen Shot 2020-02-06 at 10 06 21 AM

image

it's from docs, I used good icon, some of icons in docs are different than others

droshev:
@JKMarkowski could you list all the class names that were removed with this PR?

JKMarkowski:
> @JKMarkowski could you list all the class names that were removed with this PR?
> @Vanessa-Cusmich Could you review the documentation wording?
> 
> I think it is important to write in which case which component should be used (e.g. for less than 12 items then the developer should use `select` for more than 12 items the developer can use `combobox`, etc.)
> 
> Here are my comments regarding:
> `Select`
> 
> * can we make the documentation of the example a little bit more verbose(e.g. explaining what 2 columns is, etc.) The idea is to add some description what the examples are for. You did that for Select text wrapping.

Added 

> * Can you add an example of select with text and icon in the menu?

Added

> * I am not sure whether the button of the select should get the focus when it is clicked. I couldn't find this on the documentation. The focus there is on the entire field including the button.

I'm not sure either, it's about adding `tabindex='-1'` argument to example

> 
> Screen Shot 2020-02-09 at 11 14 34 AM
> 
> * (Question) Should we open a separate issue for select rendered on mobile?

I think we do it in separated issue

> * (Question) Does the select more like a pattern in our case?

I would say select is component, because it has it's own styles.

> 
> `Combobox`
> 
> * I think it is good to write what the component is about (e.g. `combobox` allows users to select an item from a predefined list and it provides an editable input field for filtering the list, and a dropdown menu with a list of the available options. If the entries are not validated by the application, users can also enter custom values
> * It is hard to understand:
> 
> ```
> The combobox input component is an opinionated composition of the input group, popover and list components. The popover is shared between multi-input and select component. In Select there are much more examples of modifications that can be provided also to combobox component.
> ```
> 
> Also the combobox is a pattern and not a component
> 
> * Can you change the first example the field and menu to be with the same width?
> * (Question) Should we open a separate issue for select rendered on mobile?
> * I am not sure whether the button of the combobox should get the focus when it is clicked. I couldn't find this on the documentation. The focus there is on the entire field including the button.
> * Can you add examples with Two-Column Layout and Grouping?
> 
> `Multi Input`
> 
> * I think the description here is a copy from the combobox:
> 
> ```
> The combobox input component is an opinionated composition of the input group, popover and list components. The popover is shared between multi-input and select component. In Select there are much more examples of modifications that can be provided also to combobox component.
> ```
> 
> Can you update it and see my comment for `combobox`?
> 
> * Can you make at least the first examples the token data to match the menu underneath?
> 
> Screen Shot 2020-02-09 at 11 54 24 AM
> 
> * Can we add a description what the pattern is about (e.g. A multi-input field allows the user to enter multiple values, which are displayed as tokens. To help the user enter a valid value, you can enable the suggestions feature and the value help option.)
> * (Question) Should we open a separate issue for select rendered on mobile?
> * Can you add an example with Grouping?

JKMarkowski:
hi @droshev I think I added all of your hints. Please take a look on it and we can merge it

droshev:
@JKMarkowski let me know once you apply the comments that were discussed during the code review

JKMarkowski:
Hello @droshev  changes are provided anr PR is ready to check

Issue #579, opened 2020-01-07
Rename Alert component to Message strip


InnaAtanasova:
Alert component should be renamed to Message strip to match the naming in Fiori3. As of January 8th, the component reflects the latest design and theme parameters. The only change needs to be done is the renaming and documenting the Breaking Changes

PR #611, opened 2020-01-27
fix: Rename Alert to Message Strip | Mark Alert as deprecated

salarenko:
## Related Issue
Closes SAP/fundamental-styles#579

## Description
This PR:
-  introduces new component `Message Strip` which codebase has been copied from `Alert` component

-  `Alert` component has been marked as Deprecated, it is still supported with fundamental-styles css classes

- Example demo with `Alert` has been removed in favor of `Message Strip` component since its the default component used to displaying alert messages


src/_settings.scss
```diff
@@ -127,7 +127,7 @@ $fd-colors: map-merge(
     ),
     "status": (
       1: hsl(147, 85.3%, 26.7%), // #107E3E Semantic Colors positive
-      2: hsl(28, 90.2%, 48%), // #E9730C Semantic Colors alert
+      2: hsl(28, 90.2%, 48%), // #E9730C Semantic Colors message strip
       3: hsl(0, 100%, 36.7%), // #BB0000 Semantic Colors negative
       4: hsl(210, 2.8%, 42.7%), // #6A6d70 Semantic Colors neutral text
       5: hsl(210, 90.9%, 42.9%) // #6A6D70 Semantic Colors neutral border
```
droshev: same as above

src/_settings.scss
```diff
@@ -193,8 +193,8 @@ $fd-link-color--disabled: rgba($fd-link-color, 0.4) !default;
 
 $fd-color--negative: map-get($fd-colors-status, 3) !default;
 $fd-color--error: $fd-color--negative !default;
-$fd-color--alert: map-get($fd-colors-status, 2) !default;
-$fd-color--warning: $fd-color--alert !default;
+$fd-color--message-strip: map-get($fd-colors-status, 2) !default;
```
droshev: same as above

src/_settings.scss
```diff
@@ -206,7 +206,7 @@ $fd-colors-background-states: map-merge(
     "selected-hover": change-color($fd-color--action, $alpha: 0.1),
     "positive": map-get($fd-colors-background, 3),
     "negative": map-get($fd-colors-background, 5),
-    "alert": map-get($fd-colors-background, 4),
+    "message-strip": map-get($fd-colors-background, 4),
```
droshev: same as above

src/badge.scss
```diff
@@ -54,7 +54,7 @@ $block: #{$fd-namespace}-badge;
   }
 
   &--warning {
-    @include fd-var-color("background-color", fd-color-state("alert"), --fd-color-background-4);
+    @include fd-var-color("background-color", fd-color-state("message-strip"), --fd-color-background-4);
```
droshev: same as above

src/functions/_color.scss
```diff
@@ -21,7 +21,7 @@
   }
 }
 
-// $state: hover, selected, disabled, visited, negative, positive, alert
+// $state: hover, selected, disabled, visited, negative, positive, message strip
```
droshev: same as above

src/overlay.scss
```diff
@@ -11,6 +11,7 @@ $block: #{$fd-namespace}-overlay;
   $fd-overlay-background-color: rgba(#4a505c, 0.3) !default;
 
   --fd-overlay-background-color: #{$fd-overlay-background-color};
+  $fd-overlay-message-strip-top-margin: fd-space(7);
```
droshev: none of the `fd` variables should be used

src/root.scss
```diff
@@ -16,7 +16,7 @@
   --fd-color-background-selected-hover: #{map-get($fd-colors-background-states, "selected-hover")};
   --fd-color-background-information: #{map-get($fd-colors-background-states, "information")};
   --fd-color-background-positive: #{map-get($fd-colors-background-states, "positive")};
-  --fd-color-background-alert: #{map-get($fd-colors-background-states, "alert")};
+  --fd-color-background-alert: #{map-get($fd-colors-background-states, "message-strip")};
```
droshev: same as above

src/root.scss
```diff
@@ -31,7 +31,7 @@
   // status colors
   --fd-color-positive: #{$fd-color--positive};
   --fd-color-negative: #{$fd-color--negative};
-  --fd-color-alert: #{$fd-color--alert};
+  --fd-color-alert: #{$fd-color--message-strip};
```
droshev: same as above

test/templates/message-strip/index.njk
```diff
@@ -0,0 +1,336 @@
...
+        padding: 20px;
+        font-size: initial;
+        line-height: initial;
+    }
+</style>
+
+<h1>Message Strip</h1>
+
+<!-- output the component example and the code snippet -->
+{% set example %}
```
droshev: I know this is not part of this ticket, but can we optimize the test page and add loops instead of copying the same code for each message strip?
This will reduce the code in the test file

droshev:
I guess since we are not changing anything in alert the PR can be just `fix`, no?

salarenko:
> I guess since we are not changing anything in alert the PR can be just `fix`, no?

Even if this is copy paste I think we should still mark this as feature since we still introduce new component.

src/overlay.scss
salarenko: This is out of scope of this task. If we want to refactor overlay we should create separate task for this.

droshev: If we touch it, we fix it

droshev: If we touch it, we fix it

droshev:
> > I guess since we are not changing anything in alert the PR can be just `fix`, no?
> 
> Even if this is copy paste I think we should still mark this as feature since we still introduce new component.

at the same time we don't introduce a breaking change so it has to be `fix`

droshev:
> @salarenko I don't see the changes I commented. Can you refactor and remove all the variables related to alert (e.g. `_settings.scss` like `$fd-color--alert`, `--fd-color-alert`)...
> Those kind of changes will make the PR `feat`

We agreed to tackle the color refactoring in a separate PR

docs/_sass/_docs-display-component.scss
```diff
@@ -112,6 +112,12 @@
                 margin-top: -410px;
             }
         }
```
InnaAtanasova: Shouldn't we get rid of this?

docs/_sass/_docs-general.scss
```diff
@@ -217,6 +217,10 @@ pre {
   margin: 0 0 30px 0;
 }
 
+.docs-message-strip {
```
InnaAtanasova: If we are not using the alert classes we should prob remove these

src/fundamental-styles.scss
```diff
@@ -5,6 +5,7 @@
 @import "./icon";
 @import "./layout";
 @import "./alert";
+@import "./message-strip";
```
InnaAtanasova: We are listing the components in an alphabetical order.

PR #615, opened 2020-01-28
[WIP] fix: make popover header, subheader and footer reusable mixins

InnaAtanasova:
`WIP - we are waiting for designers to respond to our questions regarding the alignment of the Footer content.` 

## Description

Popover header, subheader and footer follow the `Page` specs. The same is valid for Modal component. The code is moved to mixins and this way it can be reused in multiple components.


InnaAtanasova:
We will implement Bar component which will bring the Header/Subheader and the Footer

PR #621, opened 2020-01-29
docs: apply unified approach for loading images in the documentation

InnaAtanasova:
## Description
The images in the documentation were loaded in two different ways. We agreed to use only one approach which is using `https://placeimg.com` online library. For this reason the "static" images were replaced by `https://placeimg.com/400/400/nature` dynamic images.


InnaAtanasova:
> Can we also remove unused images from the project?

Most of them are used for the test. The images we load from Internet are "dynamic"- they change every few minutes which will cause our tests to fail

PR #622, opened 2020-01-31
fix: Rename Alert to Message Strip | Mark Alert as deprecated

salarenko:
## Related Issue
Closes SAP/fundamental-styles#579

## Description
This PR:
-  introduces new component `Message Strip` which codebase has been copied from `Alert` component

-  `Alert` component has been marked as Deprecated, it is still supported with fundamental-styles css classes

- Example demo with `Alert` has been removed in favor of `Message Strip` component since its the default component used to displaying alert messages

(reopened #611)


salarenko:
@droshev Hey Deno, I've applied the changes we've been talking about.

src/overlay.scss
```diff
@@ -11,6 +11,7 @@ $block: #{$fd-namespace}-overlay;
   $fd-overlay-background-color: rgba(#4a505c, 0.3) !default;
 
   --fd-overlay-background-color: #{$fd-overlay-background-color};
+  $fd-overlay-message-strip-top-margin: fd-space(7);
```
droshev: Can you revert any changes in this file?

docs/_sass/_docs-display-component.scss
```diff
@@ -112,6 +112,12 @@
                 margin-top: -410px;
             }
         }
+
```
droshev: can we remove this

docs/_sass/_docs-display-component.scss
```diff
@@ -112,6 +112,12 @@
                 margin-top: -410px;
```
droshev: can we remove this as we agreed

src/overlay.scss
salarenko: These are the changes needed for messages strip to work.

docs/_sass/_docs-display-component.scss
salarenko: We can't do that. This class is used in documentation when message strip is positioned inside overlay example.

PR #637, opened 2020-02-05
fix: Specify background-image: none; for product-menu control

skvale:
## Related Issue
N/A

## Description

A background-image from global styling can creep into the fundamental control:

### Before:
Screen Shot 2020-02-05 at 11 29 31 AM

### After:
Screen Shot 2020-02-05 at 11 29 40 AM

### DIFF
--- a/src/_product-menu.scss
+++ b/src/_product-menu.scss
@@ -26,6 +26,7 @@ $block-product-menu: #{$fd-namespace}-product-menu;
   &__control {
     @include fd-var-color("background-color", fd-color("shell", 2), --fd-color-shell-2);
 
+    background-image: none;
     position: relative;
 
     @include fd-button-reset();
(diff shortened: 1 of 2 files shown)


src/_product-menu.scss
```diff
@@ -26,6 +26,7 @@ $block-product-menu: #{$fd-namespace}-product-menu;
   &__control {
     @include fd-var-color("background-color", fd-color("shell", 2), --fd-color-shell-2);
 
+    background-image: none;
```
jbadan: Can we add this to the button reset instead so we only need to fix it in one place? 
https://github.com/SAP/fundamental-styles/blob/master/src/mixins/_mixins.scss#L184

salarenko:
Hey @skvale make sure you sign "Contributor License Agreement" and we will be ready to merge :)

droshev:
@skvale in case the pipeline task status doesn't get updated could you create a regular(non-forked) PR?

PR #640, opened 2020-02-06
docs: remove alert/message strip outdated demo pages

droshev:
## Description
- revert the deprecation label for `alert`
- remove message strip overlay demo pages


droshev:
> Shouldn't we support Alert as it has been before Message Strip introduction? (with overlay and overlay demo examples)
> Also `Alerts provide inline messages` is not true, that's Message Strip functionality. Alert should still keep its original functionality and in the future be replaced with `Message Toast` component.

We don't change the functionality, but I think it is good to stop misleading the stakeholders in terms of behaviour. Whoever is using the `alert` already won't get affected anyway.

droshev:
update the markup to match the original styling
Screen Shot 2020-02-10 at 9 25 29 AM

droshev:
Screen Shot 2020-02-10 at 2 46 26 PM

Issue #634, opened 2020-02-04
New Bar component


InnaAtanasova:
Popover and Modal components need Header, Subheader and Footer which are part of the Bar component. We need to implement the Bar component and then reuse its parts in the other components that have Header, Subheader and Footer.

PR #641, opened 2020-02-07
feat: new Bar component

InnaAtanasova:
## Related Issue
Closes SAP/fundamental-styles#634

Bar component is a new component which is a building block for Header, SubHeader, Footer, and Floating Footer in Page and containers. 

This PR also include:
* Add Page Background Design to Page and update of the documentation
* Update the Popover usage of Header, SubHeader and Footer. Some classes are removed since we nolonger need them. We keep the `fd-popover__body-header` and `fd-popover__body-footer` to apply to rounded corners when the Bar element is inside the Popover Body.
It can't be done with `:first-child` and `:last-child` since Popover Body with arrow uses the before and after.

Breaking Changes:
### Page
* Removed class `fd-page__header`. Bar component is now used for Page/Container header.

### Popover
* Removed class `fd-popover__body-header--with-subheader`. For both cases (header with or without subheader use `fd-popover__body-header`). The Bar element is handling the case when Header is followed by a Subheader. 
* Removed class `fd-popover__body-subheader`. 
* Removed all `--compact` modifier classes: `fd-popover__body-header--compact`, `fd-popover__body-subheader--compact`, `fd-popover__body-footer--compact`.


estelaxu3:
There seems to have some spacing issues in following examples:
1. The space Field placeholder text box  is off
Screen Shot 2020-02-07 at 11 23 34 AM

2. The icons should move down 1 or 2 pixels
Screen Shot 2020-02-07 at 11 23 45 AM

Screen Shot 2020-02-07 at 11 23 56 AM

src/bar.scss
```diff
@@ -0,0 +1,176 @@
...
+
+  &--home-page-S {
+    @include responsive-page($padding-x: 0.5rem);
+
+    &.#{$block}--floating-footer {
+      @include responsive-floating-footer();
+    }
+  }
+
+  &--page-M_L,
```
salarenko: Should be lower case separated with `-`
```suggestion
  &--page-m-l,
```
Same for different sizes in `src/bar.scss` and `src/page.scss`

test/templates/bar/index.njk
```diff
@@ -0,0 +1,500 @@
+{% extends "layout.njk" %}
+{% from "./../format.njk" import format %}
+
+{% block content %}
+
+
+<h2>Bar - Desktop</h2>
```
salarenko: Since we noticed how much number of test images reflects on pipe performance, could we somehow group our test cases? Like Header + all modification cases, Subheader + all modification cases etc.

InnaAtanasova: Will do this. Makes sense

src/bar.scss
InnaAtanasova: will change to s, m, l, xl

PR #654, opened 2020-02-12
fix: refactor form-message inside popover

InnaAtanasova:
I added a modifier class in Popover component to handle the case when we have input group with messages. 
- The `fd-popover__control` is no longer used at the same level as the `fd-input` in the documentation.
- The `fd-popover__body` is no longer used at the same level as the `fd-message` in the documentation.

Because `fd-popover` scss file was imported first in `src/fundamental-styles.scss` the problem wasn't noticeable. We shouldn't rely on the order of file import. 

Please take a look at the code and we will discuss it tomorrow

### DIFF
--- a/src/form-message.scss
+++ b/src/form-message.scss
@@ -9,30 +9,20 @@ $block: #{$fd-namespace}-form-message;
 
 .#{$block} {
   @include fd-reset();
+  @include fd-ellipsis();
 
-  z-index: 3;
-  clear: both;
-  height: auto;
-  width: fit-content;
+  padding: 0.5rem;
   min-width: 6rem;
   max-width: 22rem;
-  padding: 0.5rem;
-  margin-top: -0.25rem;
+  width: 100%;
   font-size: var(--sapFontSmallSize);
   font-weight: normal;
   line-height: normal;
   color: var(--sapTextColor);
-  overflow: hidden;
-  text-overflow: ellipsis;
-  border-radius: 0 !important;
   white-space: pre-line !important;
 
   @include fd-rtl() {
     right: 0;
-  }
-
-  &--compact {
-    margin-top: -0.1875rem;
   }
 
   &--error {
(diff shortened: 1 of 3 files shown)


stefanoScalzo:
Would it be possible to create an example in the documentation, please?

src/popover.scss
```diff
@@ -226,6 +235,12 @@ $block: #{$fd-namespace}-popover;
       }
     }
 
+    &--input-message-group {
+      box-shadow: none;
+      border: none;
+      margin-top: -0.25rem !important;
```
stefanoScalzo: is there a way to do this without using important?

src/form-message.scss
```diff
@@ -9,32 +9,22 @@ $block: #{$fd-namespace}-form-message;
...
-  text-overflow: ellipsis;
-  border-radius: 0 !important;
   white-space: pre-line !important;
 
   @include fd-rtl() {
     right: 0;
   }
 
-  &--compact {
-    margin-top: -0.1875rem;
```
stefanoScalzo: Where are you taking care of the compact version?

stefanoScalzo:
Screen Shot 2020-02-12 at 6 02 16 PM
We need zindex for this reaosn

src/popover.scss
InnaAtanasova: no

src/form-message.scss
InnaAtanasova: Screen Shot 2020-02-12 at 8 48 40 PM

no need to have a specific value for compact. -0.25rem is handling both cases. I included an example with compact input in the documentation

src/popover.scss
JKMarkowski: I'm also no fan of using `!important`. It's in modifier, so it shouldn't be a problem, I saw that bootstrap 3/4 uses `!imrpotant` in some places. After removing important, it works actually
Without:
image

With:
image

InnaAtanasova: How do you test popper.js inside fundamental-styles?

PR #663, opened 2020-02-14
feat: update switch component

InnaAtanasova:
## Description
- Toggle has been completely renamed to switch in the documentation
- Removed from the documentation the example with labels on sides
- Updated the hard-coded colors from the sap-theming (lighten/darken)
- The default switch is now `Toggle without Labels on Sides` so the code has been refactored (basically moving rules)
- BREAKING CHANGE - since the  `Toggle without Labels on Sides` is now the default switch I removed `fd-switch--no-text` modifier class
- Updated the tests

### Before:
Screen Shot 2020-02-14 at 6 12 23 PM

### After:
Removed from the documentation


docs/pages/components/switch.md
```diff
@@ -1,67 +1,24 @@
 ---
-title: Toggle / Switch
-id: toggle
-keywords: toggle, swtich
```
droshev: maybe we can keep the keyword `toggle` for developers looking for that component. It will redirect them to `switch`?

InnaAtanasova: done

InnaAtanasova:
> Icons in Switch component are off the center
> image
> 
> Can we fix this within this PR?
> I couldn't find what causes this misalignment, but we can use flex-box to center the icon, or change icon to `display: block`.

This issue is in NGX, should be fixed with the new release

PR #691, opened 2020-02-21
fix: Minimal input width on IE11 

salarenko:
## Related Issue
Closes SAP/fundamental-styles#690

## Description
On IE11 when container uses `display: flex` `min-width` property is ignored. Basic width can be set using `flex-basis` property.

Before:
image

After:
image

### DIFF
--- a/src/input-group.scss
+++ b/src/input-group.scss
@@ -13,6 +13,7 @@ $fd-input-border-radius: var(--sapField_BorderCornerRadius);
 
 .#{$block} {
 
+  $fd-input-width-basis: 10rem;
   $fd-input-group-add-on-border: 0.0625rem;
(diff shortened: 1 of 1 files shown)


salarenko:
> Please check my comments in Slack regarding the solution Brad from Concur provided.

Then I think we can set `flex-basis` for each browser. This won't change current behavior of input elements and will fix IE11 issue.

droshev:
Let's remove the IE specific fix and leave the width issue to the app developer.
In the documentation we can set `width` to `100%` so the date picker will be consistent

salarenko:
> Let's just remove the IE specific fix and leave the width issue to the app developer.
> In the documentation we can set `width` to `100%` so the date picker will be consistent

I don't think this is a valid solution. TBD during daily.

Issue #571, opened 2020-01-07
Refactor Menu to match Fiori 3 specs and the sap-theming


InnaAtanasova:
Refactor Menu component to match Fiori 3 specs and the sap-theming

- For the colors, remove all occurrences of hard-coded values and replace them with the corresponding values from @sap-theming 
- Where possible, use css variables from @sap-theming for the borders, border-radius, font-size, font-family, etc.
- Update the paddings, margins, font-sizes, line-heights, etc to match 100% the Fiori 3 specs
- If necessary, rename the component and its classes accordingly 
- Update the project Playground tests if necessary and test the component. See [here](https://github.com/SAP/fundamental-styles#download-and-installation) more information about how to run the development Playground
- Update the visual regression tests
- Test the component in all browsers (cross-browser support)
- a11y
- Document the Breaking Changes
- Provide a detailed description of the changes that are made in the Pull Request and attach screenshots with `Before` and `After`
- In the Pull Request link to the related issue [see the PR Guidelines](https://github.com/SAP/fundamental-styles/wiki/Contribution-Guidelines#make-a-pull-request)

Please refer to the following links for more information about the project:
- [Development Guidelines](https://github.com/SAP/fundamental-styles/wiki/Development-Guidelines)
- [Visual Testing](https://github.com/SAP/fundamental-styles/wiki/Visual-Testing)

kavya-b:
Fiori3 link: https://wiki.wdf.sap.corp/wiki/pages/viewpage.action?pageId=2061501144 
Few things I see that are missing and might be needed from Fundamentals' side:
1. Removal of group header(and associated title class) from styles, playground and documentation
2. Addition of add-on icon as a secondary icon to the right-side for affordance arrow to show a sub-menu. It appears that this secondary icon is to be used only with a sub-menu but it is not known whether only the affordance-arrow-icon is to be used or any icon can be used.
<img width="515" alt="image" src="https://user-images.githubusercontent.com/53509521/73234945-4a9cde00-41b2-11ea-9287-c49d97cb2479.png">
This will need addition of new class, and new example in the documentation and playground pages.

3. Using separators to visually represent groups instead of applying them on every item. This probably needs only documentation and playground changes. 
[...]

PR #702, opened 2020-02-25
feat: Menu refactor to Fiori3

kavya-b:
## Related Issue
Closes SAP/fundamental-styles#571

## Description
Update Menu styles, documentation and tests to align with Fiori3 specs

- removed `fd-menu__group` classes, their playground tests and documentation examples related to group headers.
- added new classes for compact and mobile modes: `fd-menu--compact` and `fd-menu--mobile`. Default will be cozy tablet mode.
- moved `fd-menu__item--separated` from `<ul>` level to its own element `fd-menu__separator` in a `<span>` since separators are now used instead of group headers to show grouping.
- using existing class `fd-menu__title` for the text that will be used in a `<span>`
- added new class for  secondary icon `fd-menu__addon-after` that will be used similar to `fd-menu__addon-before`
- added a modifier class for the affordance arrow when used in a submenu `fd-menu__addon-after--submenu` to get the right(or left in RTL) navigation arrow in a smaller icon size. 
- added new class for submenu `fd-menu__sublist` at the `<ul>` level corresponding to `fd-menu__list` of the parent menu.
- added new class `fd-menu__link` that wraps the item name and associated icons, if any.
- added new class `has-child` that can be applied on the parent `__link` class to apply styles for the parent element that has a submenu.
- moved `fd-menu__item` class to the parent `<li>` tag.

## Breaking changes
- Removed `fd-menu__item--separated`. Instead use `fd-menu__separator` in its own `<span>` after the `<li>` item where you want the separator.
- Removed `fd-menu__group`. Separator itself acts a visual cue for grouping.
- Removed `fd-menu--addon-before`(not `fd-menu__addon-before`) as it is unnecessary now and will be covered by existing classes.
- Moved `fd-menu__item` class to the parent `<li>` tag. Use `fd-menu__link` class on the `<a>` tag instead.

For new classes added, please refer to the above section.

### Before:
image
image

### After:
image
image

### Updated screenshots:
image

image


src/menu.scss
```diff
@@ -5,165 +5,303 @@
...
-    @include fd-weight("normal");
+    &--separated {
+      border-bottom-style: solid;
+      border-bottom-width: 0.0625rem;
+      border-bottom-color: $fd-menu-item-separator-color;
+    }
 
-    text-decoration: none;
+    @include fd-hover() {
+      .#{$block}__option-name {
```
JKMarkowski: Could you change this element name to `__title`?

src/menu.scss
```diff
@@ -5,165 +5,303 @@
...
+  &__list,
+  &__sublist {
     @include fd-reset();
 
+    margin: 0;
+    padding: 0;
     list-style: none;
-    margin-bottom: 0;
-    padding-left: 0;
+    max-width: 20rem;
```
JKMarkowski: Let's put `max-width` to parent `fd-menu`  container

src/menu.scss
```diff
@@ -5,165 +5,303 @@
...
+    }
+
+    @include fd-screen(l) {
+      width: 2rem; // to account for item right padding of 12px or 0.75rem so 2rem-0.75rem = 1.25rem
+      height: 1.25rem; // to account for item top padding of 6px or 0.375rem so 2rem-0.375rem-0.375rem = 1.25rem
+      margin: 0 0 0 -0.5rem;
+
+      @include fd-rtl() {
+        margin: 0 -0.75rem 0 0; // since before and after icons switch places, their margins also switch
+      }
```
JKMarkowski: Let's not use `fd-screen`, we let users to add size modes by themselves on `fd-menu` level. so ex. `fd-menu--compact`.

src/menu.scss
```diff
@@ -5,165 +5,303 @@
...
+
+      @include fd-hover() {
+        color: $fd-menu-item-color--active;
+      }
+    }
+
+    flex: 0 0 auto;
+    display: flex;
     background-color: transparent;
     float: left;
```
JKMarkowski: Try to remove floats, it's out of control. There are better alternatives

src/menu.scss
```diff
@@ -5,165 +5,303 @@
...
+    @include fd-selected() {
+      color: $fd-menu-primary-icon-color;
+
+      @include fd-hover() {
+        color: $fd-menu-item-color--active;
+      }
+    }
+
+    background-color: transparent;
+    float: right;
```
JKMarkowski: Again float, which is not needed

src/menu.scss
```diff
@@ -5,165 +5,303 @@
...
+
+    @include fd-selected() {
+      color: $fd-menu-primary-icon-color;
+
+      @include fd-hover() {
+        color: $fd-menu-item-color--active;
+      }
+    }
+
+    flex: 0 0 auto;
```
JKMarkowski: It's also not needed I think.

src/menu.scss
kavya-b: I was handling cozy, compact modes for tablet, desktop and mobile through this. If we provide option to user to add size modes, then we won't have control if they decide to use compact mode on phone thereby not aligning to Fiori3, right?

kavya-b:
> ### The menu item has text decorator this doesn't match fiori 3
> image

**This is coming from the styles to <a> tag in documentation itself. It appears fine in playground:**
image
 
> ### wrong width in IE 11
> image

I will work on this; however, in the meanwhile, do you think this is related to https://github.com/SAP/fundamental-styles/pull/691 ?

> ### Could you add RTL example to documentation?

I will add this.

kavya-b:
> * Let's add cozy mode by default, because right now there is only `compact` mode.

Do you mean, I should have the cozy mode applied by default? But for desktop, it is mentioned as Compact mode, right?
> * There are also differencted between Compact/Cozy(tablet)/Cozy(Phone)
>   image

Currently, the styles of cozy, compact are applied when the screen is resized. However, from your other comment, it seems that these styles are better provided as separate `fd-menu-cozy` `fd-menu-compact` classes. I will work on changing this.
> * Background color of `fd-menu` is not set

I assumed that since fd-menu will be part of a popover, the background color, box shadow and border attributes of the container should be part of the popover. I have verified that the appropriate values are applied to the popover class. 
> * Remove underline from `<a>` if used with `__title`
>   image

This is coming from the styles `<a>` tag in documentation itself. It appears fine in playground:
image
Currently same thing is happening for side-navigation component as well.

src/menu.scss
JKMarkowski: Actually, in all of cases, when we have more than 1 `size mode` such as cozy/compact, we let users decide on application level, which one they want to use it.

JKMarkowski: Actually, in all of cases, when we have more than 1 `size mode` such as cozy/compact, we let users decide on application level, which one they want to use it.

rengare:
> I will work on this; however, in the meanwhile, do you think this is related to #691 ?
Frankly, I don't know, we know that it is due to the popover and input used together but maybe it is related. I would have to check that

src/menu.scss
kavya-b: removed

src/menu.scss
kavya-b: removed

src/menu.scss
kavya-b: removed

src/menu.scss
kavya-b: added `fd-menu--compact-desktop` and `fd-menu--mobile` classes. Tablet cozy is default.

src/menu.scss
```diff
@@ -5,165 +5,303 @@
...
-    @include fd-disabled() {
-      @include fd-var-color("outline-color", fd-color-state("disabled", "action"), --fd-color-action-disabled);
-      @include fd-var-color("color", $fd-forms-color--disabled, --fd-color-text-4);
+    @include fd-item-responsive-padding();
 
-      cursor: not-allowed;
+    .#{$block}__option-name {
+      @include fd-reset();
+
+      flex: 3 3 10%;
```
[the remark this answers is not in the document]
kavya-b: we need this to ensure the primary, secondary icons and title inside a menu item are properly placed. If removed, the secondary icon comes to the end of the title and misaligns.

src/menu.scss
kavya-b: have moved to fd-menu

src/menu.scss
kavya-b: done

kavya-b:
There are 2 known issues in IE as of now that I couldn’t find a fix for but I also see the same issue is present in components that use List component(Select, Multi Input etc)-
1. focus border outline-offset property is not applied properly in IE due to which it sometimes does not appear properly
2. clicking on the parent <li>  area alone activates the states (active, selected etc). If you click on an area that is a child of <li> the states don’t show up.

stefanoScalzo:
Screen Shot 2020-03-04 at 8 05 15 PM
disabled should not change color on hover and on click

kavya-b:
> Screen Shot 2020-03-04 at 8 05 15 PM
> disabled should not change color on hover and on click

fixed

src/menu.scss
```diff
@@ -5,165 +5,295 @@
...
   }
 
   &__item {
     @include fd-reset();
-    @include fd-ellipsis();
 
-    display: block;
+    display: flex;
+    align-items: center;
     transition: all $fd-animation-speed ease-in;
```
JKMarkowski: There is no mention about transition in menu specs

src/menu.scss
```diff
@@ -5,165 +5,295 @@
...
+      @include fd-hover() {
+        color: $fd-menu-item-color--active;
+      }
+    }
+
+    display: flex;
+    background-color: transparent;
+
+    @include fd-icon-size("m", "before");
+    @include fd-icon-size("m", "after");
```
InnaAtanasova: This mixin makes the icon 16px. We should use `rem`
Screen Shot 2020-03-09 at 1 50 57 PM
Screen Shot 2020-03-09 at 1 50 49 PM

src/menu.scss
```diff
@@ -5,165 +5,295 @@
...
     }
 
     background-color: transparent;
-    float: left;
-    clear: left;
-    min-height: 1px;
-    width: 30px;
+
+    @include fd-icon-size("s", "before");
+    @include fd-icon-size("s", "after");
```
InnaAtanasova: This size should be applied only if the user wants affordance arrow. It shouldn't be applied for all addon-after. Maybe create a modifier class to handle the affordance arrow case

kavya-b: so, would you say `fd-menu__addon-after` by default takes any icon in the normal font-size of 1rem but for affordance arrow I should keep a class like `fd-menu__addon-after--submenu` and handle the smaller size case here?

src/menu.scss
kavya-b: Okay, I will just use `var(--sapFontLargeSize)` directly instead

src/menu.scss
kavya-b: Missed this. Will remove it, thanks for pointing it out!

kavya-b:
> Screen Shot 2020-03-09 at 2 01 41 PM
> 
> I don't think it's a good approach to add the separator to the item. It should be a separate element. In the case where we have an item that is a separator and is selected, the selected border overwrites the separator. It should be like this: (has both lines)
> Screen Shot 2020-03-09 at 2 01 28 PM

I see. I will add a separate element independent of the item that is placed after the item.
> 
> * `fd-menu--compact-desktop` should be `fd-menu--compact` for consistency with the other components

Will change this
> * In Mobile mode how do we display the cascading (the second menu level) ?
> 
> Screen Shot 2020-03-09 at 2 05 50 PM
> 
The spec says _'Both controls are cascading, except for the sap.m.Menu on phone, where it opens fullscreen (with drill-drown navigation).'_ So the submenu will look the same way as the parent menu. 
> * For the documentation part - would you please remove the `Affordance arrow` if there is no second level. It gives the impression that something is not working because clicking on it does not open anything.

I will remove the arrows and instead show other icons to show how secondary icons look.

kavya-b:
Addressed @JKMarkowski and @InnaAtanasova's comments: 
- removed redundant right-arrow icons when submenu was not really present. For submenu, use `fd-menu__addon-after--submenu` class modifier to get the smaller size for affordance arrow:
`<span class="fd-menu__addon-after fd-menu__addon-after--submenu sap-icon--navigation-right-arrow"></span>`
- instead used other icons to showcase secondary icon size
image
- using separator in a different element
image
- removed 'desktop' from modifier class name
- removed transition property
- applied font-size on icon directly
- updated playground tests and screenshots accordingly - added selected-separated usecase to visual test and documentation
image

InnaAtanasova:
Screen Shot 2020-03-11 at 11 08 11 AM

Hi @kavya-b , 
`fd-menu__item--separated` is contradicting the BEM methodology. The separator is not an menu item, which means we can not use `fd-menu__item`. Since it's not a menu item and is not a modifier of this item we can not use the `--separated`. You can rename it to something like `fd-menu__separator`.
For more info about BEM you can check [this](http://getbem.com/introduction/)

kavya-b:
> Screen Shot 2020-03-11 at 11 08 11 AM
> 
> Hi @kavya-b ,
> `fd-menu__item--separated` is contradicting the BEM methodology. The separator is not a menu item, which means we can not use `fd-menu__item`. Since it's not a menu item and is not a modifier of this item we can not use the `--separated`. You can rename it to something like `fd-menu__separator`.
> For more info about BEM you can check [this](http://getbem.com/introduction/)

Okay, have changed the name to `fd-menu__separator` now.

droshev:
@kavya-b Few comments:
- can you update the documentation and create examples without edge cases like long text items; hybrid items of text and icons - those can be added as part of the playground so can have testing examples there. The goal of the documentation is to show correct and recommended layouts without too long text or inconsistent icon/texts
- would you mind documenting the class names that the menu supports (e.g. `fd-menu__addon-before`, `fd-menu__addon-after`, `fd-menu__separator`,  all state classes, `fd-menu__sublist`)

src/menu.scss
```diff
@@ -5,165 +5,294 @@
...
+
+      background-color: $fd-menu-item-background-color--active;
+
+      .#{$block}__addon-before,
+      .#{$block}__addon-after {
+        color: $fd-menu-primary-icon-color-active-selected;
+      }
+
+      .#{$block}__addon-after {
+        // border-bottom-color: var(--sapList_BorderColor); // TODO change this when fade() evaluates to a value
```
droshev: you can remove this comment

kavya-b:
> @kavya-b Few comments:
> 
> * can you update the documentation and create examples without edge cases like long text items; hybrid items of text and icons - those can be added as part of the playground so can have testing examples there. The goal of the documentation is to show correct and recommended layouts without too long text or inconsistent icon/texts

okay, so for documentation, instead of showing every combination of single icon/double icon/long text, I keep only one format and use the same even for mobile mode and rtl mode? Should I keep the example with different states(active, hover, selected etc) or remove it as well?

> * would you mind documenting the class names that the menu supports (e.g. `fd-menu__addon-before`, `fd-menu__addon-after`, `fd-menu__separator`,  all state classes, `fd-menu__sublist`) - you can get inspired from here: https://fundamental-styles.netlify.com/components/bar.html

had added some of these in the PR description, but will update to the documentation description as well.
> * in the PR add a section `BREAKING CHANGES` and add all class names that were removed

Okay, will do this.

src/menu.scss
kavya-b: done. have pushed in the other changes as well, please check.

droshev:
@kavya-b thanks for the updates.
Could you try to review your component following the points from https://github.com/SAP/fundamental-styles/wiki/Self-Contained-Styles. There is something wrong with the separator:
Screen Shot 2020-03-18 at 2 31 03 PM

kavya-b:
> https://github.com/SAP/fundamental-styles/wiki/Self-Contained-Styles

Hi @droshev, I tried to figure out if I am missing something that makes the unnormalized mode look like that, but cannot point to anything.
`fd-menu__separator` is in its own element in order to cover the edge case where we have a selected item and also need to show the separator, as pointed by @InnaAtanasova  in this comment: https://github.com/SAP/fundamental-styles/pull/702#pullrequestreview-371413725

droshev:
@kavya-b We need to check how the nested menu behaves and looks on mobile. I guess it may behave like the nested list in side navigation. Can you follow up on that?
Meanwhile i fixed the self-contained issue I had raised before.

kavya-b:
> @kavya-b We need to check how the nested menu behaves and looks on mobile. I guess it may behave like the nested list in side navigation. Can you follow up on that?

Hi @droshev , the [spec](https://wiki.wdf.sap.corp/wiki/pages/viewpage.action?pageId=2061501144#Menu(Fiori3)-MenuInteractionStates) says "_On Phone the Submenu opens in a new page, leaving the Menu option behind._ " The same visual design will apply for the submenu; on clicking of the parent item, the submenu will open in the same dialog, replacing the parent menu options, and the device's back button will bring it back to the parent menu. 
Does this need to be added to the documentation? I am not sure how to show a full screen replacement of parent menu in the same example using only aria-controls. I have something like this now: 
image
Please let me know if this suffices, I will push in the changes.
> Meanwhile i fixed the self-contained issue I had raised before.

Thank you for this.

kavya-b:
@droshev As discussed, the documentation for mobile mode has been modified to reflect full screen:
submenu_mobile

Issue #709, opened 2020-02-28
minor button issues


droshev:
- [x] Alert - close button
Before:
<img width="1271" alt="Screen Shot 2020-02-28 at 3 20 11 PM" src="https://user-images.githubusercontent.com/4380815/75584544-ff3e4f80-5a3d-11ea-942f-4a2e5078e9a2.png">
After:
<img width="1277" alt="Screen Shot 2020-02-28 at 3 20 23 PM" src="https://user-images.githubusercontent.com/4380815/75584586-0ebd9880-5a3e-11ea-9dd2-7cd9081efabf.png"> Addressed with https://github.com/SAP/fundamental-styles/pull/714

- [x] Message Strip: Close button - border on hover and vertical position
<img width="1266" alt="Screen Shot 2020-02-28 at 3 22 39 PM" src="https://user-images.githubusercontent.com/4380815/75584661-3876bf80-5a3e-11ea-9125-53ec252e59cc.png">Addressed with https://github.com/SAP/fundamental-styles/pull/714

- [x] Bar - update doc with button split instead of button group
<img width="1412" alt="Screen Shot 2020-02-28 at 3 29 38 PM" src="https://user-images.githubusercontent.com/4380815/75585035-319c7c80-5a3f-11ea-9fea-0b3b8ac759af.png">  fixed in https://github.com/SAP/fundamental-styles/pull/710

- [x] Notification close button on hover
<img width="988" alt="Screen Shot 2020-02-28 at 3 33 33 PM" src="https://user-images.githubusercontent.com/4380815/75585325-cc955680-5a3f-11ea-9e65-78522df861dd.png"> addressed in https://github.com/SAP/fundamental-styles/pull/716

- [x] Side Navigation - selected/clicked item - get undelrline
<img width="285" alt="Screen Shot 2020-02-28 at 3 36 38 PM" src="https://user-images.githubusercontent.com/4380815/75585481-36156500-5a40-11ea-8e4c-a668280743a2.png">
<img width="276" alt="Screen Shot 2020-02-28 at 3 36 28 PM" src="https://user-images.githubusercontent.com/4380815/75585483-36adfb80-5a40-11ea-9505-cc74323ffb1a.png">

- [x] Select button alignment and background color is wrong
Before:
[...]

PR #713, opened 2020-02-28
fix: message strip dismiss button styling

droshev:
fix: message strip dismiss button styling
https://github.com/SAP/fundamental-styles/issues/709

Before:
Screen Shot 2020-02-28 at 3 22 39 PM
After:
Screen Shot 2020-02-28 at 5 20 30 PM

### DIFF
--- a/src/alert.scss
+++ b/src/alert.scss
@@ -78,10 +78,8 @@ $block: #{$fd-namespace}-alert;
 
   // Elements
   &__close {
-    @include fd-button-reset();
     @include fd-alert-close-btn-container();
     @include fd-icon("decline", "m");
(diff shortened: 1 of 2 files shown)


bcullman:
looks like the button is closer to the bottom than then top.

due to setting the `top: 0.625 rem` 

Unsetting that will cause the button to properly center vertically.

droshev:
> looks like the button is closer to the top than then the bottom.
> 
> due to setting the `top: 0.625 rem`
> 
> Unsetting that will cause the button to properly center vertically.

the spec is positioning the button top/right as the content can grow vertically

PR #725, opened 2020-03-02
fix: underline issue on side navigation

rengare:
## Related Issue
reported #709 

## Description
Underline removed from side navigation

### Before:
see #709 

### After:
image


InnaAtanasova:
Looks good but we should avoid pushing reformatted code and moving code around because it's really hard to track what has been done in terms of fixing the issue.

PR #744, opened 2020-03-02
fix: make Shellbar button styles more specific

mikerodonnell89:
## Related Issue
fixes: #743 

## Description

Shellbar and button classes were having some conflicts, this should fix it.  The issue was made evident by the stackblitz for the shellbar example in the fd-ngx documentation.  Refer to this issue: https://github.com/SAP/fundamental-ngx/issues/2098

PR #754, opened 2020-03-03
fix: Tokenizer Input + Shellbar button improve specificity

jbadan:
## Related Issue

## Description
Specificity of (0,0,0) is not allowing `fd-shellbar__button` to override `fd-button`, same with `fd-tokenizer__input` and `fd-input`. These `!important` allow these to override. This is not seen on the fundamental styles docs site due to the order in the fundamental-styles.css file.

In Fundamental-react:
Screen Shot 2020-03-03 at 12 37 27 PM
Screen Shot 2020-03-03 at 12 38 06 PM

### DIFF
--- a/src/shellbar.scss
+++ b/src/shellbar.scss
@@ -99,26 +99,26 @@ $block: #{$fd-namespace}-shellbar;
     @include fd-icon-size("m", "before");
     @include fd-icon-size("m", "after");
 
-    border-color: transparent;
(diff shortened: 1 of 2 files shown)


mikerodonnell89:
I don't think using !important is improving specificity, it's just forcing an override.  I think this would be a better solution (in the example of the shellbar) https://github.com/SAP/fundamental-styles/pull/744

PR #760, opened 2020-03-04
feat: screen-reader only utility class

jacobdevera:
## Description
Adds a utility class to make content only visible to screen readers. This is useful for e.g. providing instructions to screen readers users how to navigate more complex components.

No visual changes expected. New reference screenshots were a result of `is-active` styles not being applied correctly before.


jbadan:
Can you change the title to fix? 

Should this be used in any of the components we already have so we can include it inside that component's scss file? That way we won't have to include all the helpers in each component that needs `screen-reader-only` class.

jbadan:
Will you merge this as a `feat` because we're changing underlying html and update https://github.com/SAP/fundamental-styles/wiki/Breaking-Changes with the details for the implementation libraries?

PR #761, opened 2020-03-05
fix: added support for text truncation if button width is set

rengare:
## Related Issue
[ngx #2110](https://github.com/SAP/fundamental-ngx/issues/2110)
## Description
Button text is now truncated (as described in fiori3) when button width is set

image

### Before:
image

### After:
image

### DIFF
--- a/src/button.scss
+++ b/src/button.scss
@@ -61,6 +61,9 @@ $block: #{$fd-namespace}-button;
   max-height: $fd-button-height;
   min-width: $fd-button-min-width;
   line-height: $fd-button-line-height;
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis;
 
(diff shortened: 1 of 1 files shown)


src/button.scss
```diff
@@ -61,6 +61,9 @@ $block: #{$fd-namespace}-button;
   max-height: $fd-button-height;
   min-width: $fd-button-min-width;
   line-height: $fd-button-line-height;
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis;
```
InnaAtanasova: Not a big deal but you can use @mixin fd-ellipsis instead :)

rengare:
> Hi Team, this matches Fiori3, is it possible to have example just like how we have for the menu button.

I'll create another PR with updates for docs

PR #769, opened 2020-03-05
fix: add li directly with list style

stefanoScalzo:
## Related Issue
Closes SAP/fundamental-styles #768

## Description
This pull request is to help target product switch that enables drag and drop. The drag and drop features changes the structure of the DOM, therefore, we need to explicitly target the li to remove bullets
## Screenshots
###Before
Screen Shot 2020-03-05 at 1 30 12 PM
###After
Screen Shot 2020-03-05 at 1 29 23 PM

### DIFF
--- a/src/product-switch.scss
+++ b/src/product-switch.scss
@@ -57,6 +57,10 @@ $block: #{$fd-namespace}-product-switch;
     }
 
     width: 47rem;
+
+    > li {
+      list-style-type: none;
+    }
   }
 
   &__item {


src/product-switch.scss
```diff
@@ -57,6 +57,10 @@ $block: #{$fd-namespace}-product-switch;
     }
 
     width: 47rem;
+
+    > li {
+      list-style-type: none;
```
InnaAtanasova: Could you add this rule to `&__item` too?

stefanoScalzo:
> Can you maintain the issue attributes and move the rule to the class?

What do you mean by issue attributes?

src/product-switch.scss
stefanoScalzo: If we do this why don't we just target all li under list and make them list-style-type:none

stefanoScalzo:
Needs to be done on ngx side

PR #770, opened 2020-03-06
docs: step input component into patterns

jbadan:
## Related Issue

## Description

Fundamental react had a contributor create a "StepInput" component that matches design specs using existing Fundamental Styles styles - https://github.com/SAP/fundamental-react/pull/914

Adding this patterns page to the documentation website so the libraries stay in sync.


docs/_data/sidebars/left-navigation-sidebar.yml
```diff
@@ -66,6 +66,10 @@ entries:
       url: /patterns/search-input.html
       output: web
 
+    - title: Step Input
+      url: /patterns/step-input.html
+      output: web
+
```
jbadan: How do we make the pictures for the index page?

droshev: @estelaxu3 do you think you can help in this?

InnaAtanasova: I make them in Adobe XD :D :D :D But I am terrible at this

bcullman:
these would be more useable if the buttons were both on the same side. is this possible?

jbadan:
@bcullman that wouldn't match the design spec though

bcullman:
have the designers actually tried using the stuff they design? :)

we should bring it up with them

Issue #743, opened 2020-03-02
Shellbar button styles need to be more specific


mikerodonnell89:
See https://github.com/SAP/fundamental-ngx/issues/2098

PR #772, opened 2020-03-09
fix: use buttonReset mixin on shellbar buttons

mikerodonnell89:
## Related Issue
Closes SAP/fundamental-styles #743 

## Description

Use buttonReset mixin and remove fd-button from all shellbar buttons


docs/_includes/demo-page-components/shell-header.html
```diff
@@ -56,18 +56,18 @@
       </div>
     </div>
          <div class="fd-shellbar__action fd-shellbar__action--desktop">
-            <button class="fd-button fd-shellbar__button sap-icon--bell" aria-label="Notifications">
+            <button class="fd-shellbar__button sap-icon--bell" aria-label="Notifications">
```
jbadan: Does this change still work if `fd-button` class is applied? For react, we'd need to make a new "button" component just for this use case. The whole point of using `fd-button` and then overriding it with `fd-shellbar__button` was to make these components as reusable as possible.

rengare: thats why fd-button uses mixin, which can be used to combine new buttons (which fd-shellbar__button is not a type of fiori 3 spec for buttons)

jbadan: I understand that from the perspective of fundamental-styles library, but for an implementation library (especially fundamental-react), going down this path leaves us 2 options:
1) create second button component specifically that's not reusable. No thanks.
2) make the current button "smart" so that it knows when it's being used in shellbar and to remove the default `fd-button` class - which is used on every other component that is using the button. I prefer not to introduce unnecessary logic into a component that should be simple. 

For those 2 reasons, we established the pattern of extending/overriding the base classes for each building block component.

bcullman: when in isolation, @rengare mixin *seems* like the right answer, (and I may have even advocated for such a solution in the past)

But I am now firmly in agreement with @jbadan here - fundamental-styles needs to look further down the road on how the implementation libraries will use this CSS. 

A mix-in will force us to create a shellbar-only button. Thats all kinds of wrong.

greg-a-smith: The whole point of self-contained styles (and components) is to define a baseline set of styles for the elements and modifiers so that we can build a single, resuable component in the implementation libraries.  Alternate uses of these components can override some basic styling (colors, for example) by injecting a class owned by a different "block" (the **B** in BEM).  The button "block" is `fd-button` and there are multiple elements and modifiers that extend it.  In this use case, `Shellbar` would inject a class `fd-shellbar__button` (or something like that) to add its styling overrides.

rengare: Well, I agree with your opinion and at the same time I want to avoid the usage of `! Important`. Lets take a look at bootstrap, they added in few places `! Important` modifier and it's a bad thing for application developer (I struggled a few times with that). Of course, I'm open to any proposal and if we agree that `! Important` will be needed then I don't mind

JKMarkowski:
Hello @jbadan I would say that it's more `feature`. Styles will stay ok, even if user doesn't change markup.

docs/_includes/demo-page-components/shell-header.html
bcullman: injecting an *additional* class as suggested here will allow for selector specificity without the need for an `! important`

mikerodonnell89:
closing in favor of https://github.com/SAP/fundamental-styles/pull/744

PR #773, opened 2020-03-09
chore: add GitHub assistant to Fundamental-styles library

rengare:
## Related Issue

## Description
Github-assistant rewriten in typescript. Script will always display release notes information or error during requests to github api.

- the script was rewritten to typescript with error log
- the script uses `async-await` to avoid nested chain of promisses
- the script should be used as old one (one thing which was changed was logging message when necessary
- fundamental-styles publish.sh was updated to run script from source (in future we want to release executable file but for now we want to test release script)


droshev:
@jbadan @greg-a-smith do you think you can take a look on this PR? We know this should not be part of the source of the repository, but we want to tackle the problems first, once we solve them all then we can think where to put the script.

greg-a-smith:
This seems like a lot of work to fix one small issue.  The reason this isn't working for releases is because the standard-version output isn't including the leading "v" in front of the new version tag.  It creates a new tag (_e.g._ `v1.2.3`) and that part is working fine.  When outputting that, standard-version is just showing `1.2.3`.  This "tag" is being passed in to github-assistant and since that tag never existed, it isn't grabbing the proper commit log nor is it updating the release for the proper tag.  In past research, this seems to be a combination of standard-version's version and the repo itself.  Bottom line, if you fix the screen scraping script to add a leading "v" if one isn't found, this should all work correctly.

greg-a-smith:
And to @jbadan's point, _IF_ changes are desired, make them in the `github-assistant` repo.  Don't start making multiple copies of this.

droshev:
we will close it for now

Issue #774, opened 2020-03-11
tokenizer inner compact width should be "auto", make cozy mode horizontally scrollable


mikerodonnell89:
**Describe the bug**

Currently, for the desktop/compact mode, resizing can cause issues where the inner element doesn't observe the margins of the fd-tokenizer container.

Also, the tokenizer should be horizontally scrollable in cozy mode.

![Screen Shot 2020-03-11 at 4 39 38 PM](https://user-images.githubusercontent.com/2471874/76471045-440ba400-63b7-11ea-876c-0257d64c4822.png)

Setting the `fd-tokenizer__inner` class to 'auto' fixes this

PR #775, opened 2020-03-11
fix: make compact tokenizer inner width: auto and make cozy mode horizontally scrollable

mikerodonnell89:
## Related Issue
Closes SAP/fundamental-styles# #774


JKMarkowski:
Hello @mikerodonnell89 is it done by purpose?
Opera:
image

Ie11:
image

InnaAtanasova:
Screen Shot 2020-03-13 at 8 46 16 AM
- I think the width of the tokens is wrong. 
- The focus outline is cut at the right side 
- The naming of the PR title is not following the guidelines
- Would you please add Assignee, Projects and Milestones?

src/tokenizer.scss
```diff
@@ -10,18 +10,24 @@ $block: #{$fd-namespace}-tokenizer;
...
   display: flex;
   align-items: center;
 
   &__inner {
+    &::-webkit-scrollbar {
+      display: none; // hide scrollbar but keep element scrollable for Chrome/Safari/Opera
+    }
+
+    padding: 0 $fd-tokenizer-spacing;
+    overflow: hidden;
```
mikerodonnell89: The interaction spec states that the cozy (phone and tablet) tokenizer should be scrollable by swiping the tokens left and right.  So I think here we do need overflow-x: scroll

docs/pages/components/token.md
```diff
@@ -280,3 +285,75 @@ To show focus on whole element, it's needed to add `.is-focus` into container.
 </div>
 {% endcapture %}
 {% include display-component.html component=compact-tokenizer %}
+
+## Tokenizer - Scrollable
+The Compact tokenizer is for large screens.
```
mikerodonnell89: This is the cozy tokenizer.  And the user won't have tab keys in cozy mode (phone/tablet).  In compact (desktop) mode they will have the "____more" text which when clicked opens the popover list containing the hidden tokens.

JKMarkowski: In my my approach, there is added `--scrollable`, which can be added in cozy and compact mode.

JKMarkowski:
@mikerodonnell89 
This is for desktop
tokenizerdesktop

This one is on mobile/tablets/touch devices
tokenizer

src/tokenizer.scss
```diff
@@ -38,7 +44,7 @@ $block: #{$fd-namespace}-tokenizer;
   &__input {
     border: none !important;
     min-width: 3rem !important;
-    padding: 0 $fd-tokenizer-spacing 0 0 !important;
+    padding: 0 !important;
```
jbadan: Did we want to keep with the `important!` pattern here or switch to the new pattern we did with shellbar buttons?

JKMarkowski: Now it's handled like on shellbar.

Issue #664, opened 2020-02-15
switch component latest design


droshev:
implement the missing features for switch component

PR #777, opened 2020-03-13
feat: add internal label to switch component

JKMarkowski:
## Related Issue
Closes https://github.com/SAP/fundamental-styles/issues/664

## Description
There are some cleanups with switch component. We had `&__text` leftover. Now this element should be used outside of the switch. 
Focus stays only for switch, even if text is there
image

`&__icon` element is standalone, it doesn't need `&__text` anymore.
There is also 1 test added with switch + text and 1 section example with switch + text.

I changed also `left/right` css properties to `transform: translate(x,y)`. Thanks to this component is better interpreted by browsers and is not very dependent on relative parent. Also this approach is much cleaner and easier to maintain.

Tested on IE11

### Before:
-

### After:
image


InnaAtanasova:
Could you please add more meaningful title for the PR? Thank you :)

Issue #681, opened 2020-02-20
menu list inside popover has wrong border


stefanoScalzo:
**Describe the bug**
Menu list inside popover cuts into borders
> Is this issue related to a specific component?
every list that uses popover
> What version of the Fundamental Library Styles are you using?
latest

> What front-end framework are you implementing in, e.g., Angular, React, etc.?
all
![Screen Shot 2020-02-20 at 10 17 35 AM](https://user-images.githubusercontent.com/50607147/74957815-19d94e80-53d6-11ea-83df-ba51eca4df1e.png)

PR #792, opened 2020-03-20
fix: popover overflow and truncation for breadcrumb

stefanoScalzo:
## Related Issue
Closes SAP/fundamental-styles#681

## Description
This PR removes overflow from popover so that the corners do not look weird. It also applies proper fiori 3 styling using dropdown styling for the truncation.
### Before:
Screen Shot 2020-03-20 at 10 36 58 AM

### After:
Screen Shot 2020-03-20 at 10 35 20 AM


src/list.scss
```diff
@@ -357,11 +365,11 @@ $block: #{$fd-namespace}-list;
...
-        border-bottom-right-radius: $fd-list-last-item-bottom-border-radius;
-      }
-    }
+    // > * {
+    //   &:last-child {
+    //     border-bottom-left-radius: $fd-list-last-item-bottom-border-radius;
+    //     border-bottom-right-radius: $fd-list-last-item-bottom-border-radius;
+    //   }
+    // }
   }
```
InnaAtanasova: Please remove this if we don't need it

stefanoScalzo:
> 1. The body of the popover should have rounded corners.
>    Here is how it looks now:
> 
> Screen Shot 2020-03-20 at 12 08 11 PM
> 
> Here is how it should look like (as per specs)
> Screen Shot 2020-03-20 at 12 04 15 PM
> 
> 1. The focus outline is not correct. Please check `Focus Visual Concept ` in the specs for more info.
>    Here is how it looks in your PR:
> 
> Screen Shot 2020-03-20 at 12 08 11 PM
> 
> How it looks in Select and MultiInput. All these components follow the List specs:
> Screen Shot 2020-03-20 at 12 09 00 PM
> 
> 1. Please maintain the right section of the PR in GitHub. I did it for you now but I can't do it for every PR
> 
> Screen Shot 2020-03-20 at 12 10 00 PM

Even though the image on the breadcrumb page has rounded corners it says: Breadcrumbs are responsive. If there is insufficient horizontal space, the links in the breadcrumb trail collapse into a dropdown menu (sap.m.Select); The select component does not have rounded top corners.

More detail on what is the issue with the focus please, it has the proper styling, offset and width

src/list.scss
stefanoScalzo: fixed

InnaAtanasova:
> > 1. The body of the popover should have rounded corners.
> >    Here is how it looks now:
> > 
> > Screen Shot 2020-03-20 at 12 08 11 PM
> > Here is how it should look like (as per specs)
> > Screen Shot 2020-03-20 at 12 04 15 PM
> > 
> > 1. The focus outline is not correct. Please check `Focus Visual Concept ` in the specs for more info.
> >    Here is how it looks in your PR:
> > 
> > Screen Shot 2020-03-20 at 12 08 11 PM
> > How it looks in Select and MultiInput. All these components follow the List specs:
> > Screen Shot 2020-03-20 at 12 09 00 PM
> > 
> > 1. Please maintain the right section of the PR in GitHub. I did it for you now but I can't do it for every PR
> > 
> > Screen Shot 2020-03-20 at 12 10 00 PM
> 
> Even though the image on the breadcrumb page has rounded corners it says: Breadcrumbs are responsive. If there is insufficient horizontal space, the links in the breadcrumb trail collapse into a dropdown menu (sap.m.Select); The select component does not have rounded top corners.
> 
> More detail on what is the issue with the focus please, it has the proper styling, offset and width

@stefanoScalzo  If there is discrepancy in the specs the approach is to reach the designers and ask them for confirmation which one - the visuals or the text is correct.

For the offset - please check the offset distance. From what I see it's bigger than it should be

PR #793, opened 2020-03-20
fix: fixing release notes

rengare:
## Description
PR checks if std-version returns tag with 'v' or without.


InnaAtanasova:
Please maintain this section. Thanks 
Screen Shot 2020-03-20 at 12 15 52 PM

PR #794, opened 2020-03-20
fix: added workaround for creating release notes

rengare:
## Description
PR adds a workaround for creating release notes

## Screenshots
image


droshev:
@rengare it would be good to merge this before we release version `0.8.0` so we can validate it during the release

PR #798, opened 2020-03-25
fix: busy indicator circles blurry in Safari

droshev:
busy indicator circles blurry in Safari


estelaxu3:
This is what it looks like so far when I zoom in 300%. But edges looks clean without zoom in. 
Screen Shot 2020-03-24 at 9 09 22 PM

droshev:
@estelaxu3 I don't expect someone to zoom that much 😄. The goal is on 100% in Safari not to see blur.

Issue #807, opened 2020-03-27
New component - Message Box


InnaAtanasova:
Introduce a new component - Message Box. 
Message Box inherits the appearance and behaviour from the Dialog component.
![Screen Shot 2020-03-26 at 8 18 14 PM](https://user-images.githubusercontent.com/39598672/77708626-f25e3e80-6f9e-11ea-9703-0d597bc5fa18.png)

PR #808, opened 2020-03-27
fix: introduce new component Message Box

InnaAtanasova:
## Related Issue
Closes SAP/fundamental-styles#807

## Description
Introduce new component to the library called Message Box. 
Message Box inherits the look and the behaviour of Dialog and for this reason certain blocks of code that are reused in both components are moved into  `_dialog-placeholders` file.


estelaxu3:
For information message box, the correct icon HTML code is &#xe05c
For confirmation message box, the correct icon HTML code is &#xe090;

stefanoScalzo:
> For information message box, the correct icon HTML code is &#xe05c
> For confirmation message box, the correct icon HTML code is &#xe090

The issue is that she used the icon given in the specs... So maybe we have to contact designers to update that spec?

docs/pages/components/message-box.md
```diff
@@ -0,0 +1,440 @@
...
+id: message-box
+keywords: message message-box
+sidebar: left-navigation-sidebar
+toc: false
+permalink: components/message-box.html
+folder: components
+summary:
+---
+
+Message Box is used to display simple messages to the user. These messages could be Standard (Default), Confirmation, Error, Success, Warning and Information (Neutral). The Message Box component inherits the look and the behaviour of the [Dialog]({{site.baseurl}}/components/dialog.html) component.
```
salarenko: > The Message Box component inherits the look and **the behaviour** of the Dialog

I'm not sure about that part. From my understanding Message Box is a simplified version of the Dialog + semantic states that makes a perfect fit for displaying alerts/confirmation widows etc. I don't think it will provide such features as dragging/resizing/responsiveness. Current description shows Message Box as an extension of the Dialog.

src/message-box.scss
```diff
@@ -0,0 +1,126 @@
...
+.#{$block} {
+  @extend %dialog;
+
+  display: none;
+
+  // ELEMENTS
+  &__content {
+    @extend %dialog-content;
+
+    overflow: hidden;
```
salarenko: This is going to prevent popovers from overflowing Message Box window. Dialog (with changes brought in this PR) doesn't prevent `__content` overflow.

src/message-box.scss
```diff
@@ -0,0 +1,126 @@
...
+  &__body {
+    @include fd-reset();
+
+    @extend %dialog-body;
+
+    padding: 1rem;
+  }
+
+  &__title {
+    @include fd-reset();
```
salarenko: I think we should reuse Dialogs styles for `title`.

docs/pages/components/message-box.md
bcullman: perhaps "...a subset of it's behavior..."

Issue #785, opened 2020-03-20
Fiori 3 Design Review: Form popup message should have box shadow


estelaxu3:
Form Input State should have box shadow

https://sap.github.io/fundamental-styles/components/form.html

<img width="269" alt="Screen Shot 2020-03-19 at 9 00 42 PM" src="https://user-images.githubusercontent.com/54589269/77127944-b833f080-6a24-11ea-9948-543c372a385b.png">

PR #810, opened 2020-03-27
fix: Add box shadow to form messages

JKMarkowski:
## Related Issue
Closes https://github.com/SAP/fundamental-styles/issues/785

## Description
There is theming parameter for shadow added for form messages.

### Before:
image

### After:
image


stefanoScalzo:
the thing is the shadow looks like it has a border which doesn't look the same as ui5

Issue #786, opened 2020-03-20
Fiori 3 Design Review: Missing Subtle Link from Link component


estelaxu3:
Missing Subtle Link from Link component

In Fiori 3 wiki: https://wiki.wdf.sap.corp/wiki/pages/viewpage.action?pageId=2057741636

<img width="976" alt="Screen Shot 2020-03-19 at 9 03 21 PM" src="https://user-images.githubusercontent.com/54589269/77128052-1b258780-6a25-11ea-9374-63506dde8b42.png">

JKMarkowski:
This theming parameter is still mising on https://github.com/SAP/theming-base-content

droshev:
`--sapLink_SubtleColor`

PR #815, opened 2020-03-29
fix: add subtle link, improve the tests

droshev:
fixes: #786 
Screen Shot 2020-03-29 at 3 29 03 PM


src/mixins/_mixins.scss
```diff
@@ -308,6 +308,10 @@
     color: var(--sapLink_InvertedColor);
   }
 
+  &--subtle {
```
salarenko: Shouldn't this `link` mixin be extracted to separate link component file?

JKMarkowski:
Thsi variable doesn't have value, cause there is no `--sapLink_SubtleColor` in sap theming repo.

src/mixins/_mixins.scss
```diff
@@ -308,6 +308,10 @@
...
+    color: var(--sapLink_SubtleColor);
```
InnaAtanasova, line 312: No such variable in @sap-theming

droshev:
> This variable doesn't have value, because there is no `--sapLink_SubtleColor` in sap theming repo.

it has been released now

src/mixins/_mixins.scss
droshev: it has been released

src/mixins/_mixins.scss
droshev: @salarenko we have an inverted one here. How do you see the subtle to be in another file?

Issue #789, opened 2020-03-20
Fiori 3 Design Review: Boarder List should have underline 


estelaxu3:
In border list there should be an underline after it is selected for accessibility purpose: 

https://sap.github.io/fundamental-styles/components/list.html

Issue #762, opened 2020-03-05
bug: IE11 Focus ring does not appear on all sides in Select Mobile Mode


kavya-b:
**Describe the bug**

> Is this issue related to a specific component?
Yes, Select using List component

> What version of the Fundamental Library Styles are you using?
0.7.0

> What front-end framework are you implementing in, e.g., Angular, React, etc.?
Angular

**To Reproduce**
1. Go to 'https://sap.github.io/fundamental-styles/components/select.html' in IE
2. Scroll down to 'Select Mobile Mode'
3. click on any of the items
4. See error: the left and right side of focus ring is missing

**Expected behavior**
Focus ring should be visible in all sides. Ideally it should be an inner focus, but IE11 does not support outline-offset property which is why inner focus anyway does not work. But focus border should at least appear on all sides.

**Screenshots**
![image](https://user-images.githubusercontent.com/53509521/75979894-788ae600-5f07-11ea-8a5a-b9e184b052dc.png)

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] Windows
 - Browser [e.g. chrome, safari] IE
 - Version [e.g. 22] 11

droshev:
@JKMarkowski can you check whether we can reproduce it?

InnaAtanasova:
I checked and I could reproduce the issue

PR #823, opened 2020-03-30
fix: add border to the selected state in borderless design and fix outline-offset issue in IE11

InnaAtanasova:
## Related Issue
Closes SAP/fundamental-styles#789
Closes SAP/fundamental-styles#762

## Description
Add border bottom to the selected state in borderless design;
Added `::before` attribute to `fd-list__item` to play the role of an outline since we need an outline offset and this property is not supported in IE11

### Before:
Screen Shot 2020-03-30 at 3 01 04 PM

### After:
Screen Shot 2020-03-30 at 3 00 34 PM


src/list.scss
```diff
@@ -11,6 +11,24 @@
...
+      position: absolute;
+      display: block;
+      border: var(--sapContent_FocusWidth) var(--sapContent_FocusStyle) $outlineColor;
+      top: 0.1875rem;
+      left: 0.1875rem;
+      right: 0.1875rem;
+      bottom: 0.1875rem;
+    }
+  }
+}
```
JKMarkowski: I like this solution, it can be used always, when there are no pseudo elements.

Issue #788, opened 2020-03-20
Fiori 3 Design Review: Side Navigation should exclude Condensed Mode


estelaxu3:
There should not have condensed mode in Side Navigation with Group Header cannot have Condensed Mode. 

https://sap.github.io/fundamental-styles/components/side-navigation.html

droshev:
I think the solution of this issue is just to update the documentation and write what is the recommended way.

estelaxu3:
I think it is a good idea. We can keep the condensed mode for now and I can propose this to visual core designers.

Issue #787, opened 2020-03-20
Fiori 3 Design Review: Side Navigation


estelaxu3:
In Side Navigation, Items with sub-level should be clickable (The item and the arrow button should have separate click events). 

https://sap.github.io/fundamental-styles/components/side-navigation.html

PR #825, opened 2020-03-31
feat: side navigation expand icon

JKMarkowski:
## Related Issue
Closes https://github.com/SAP/fundamental-styles/issues/787
Closes https://github.com/SAP/fundamental-styles/issues/788
## Description
There is added new icon element, which replaced old arrow pseudo element. Icon is represented by `fd-nested-list__expand-icon` element.

Icon also has been aligned with fiori3 newest specs from side nav specs.

Group headers are removed from condensed mode. This rule has been added to description.

Breaking changes:
Old:
```html
<li class="fd-nested-list__item">
    <a class="fd-nested-list__link has-child">
        <span class="fd-nested-list__title">Level 1 Item</span>
    </a>
</li>
```

New:
```html
<li class="fd-nested-list__item">
    <div class="fd-nested-list__content has-child" tabindex="0">
        <a class="fd-nested-list__link" tabindex="-1">
            <span class="fd-nested-list__title">Level 1 Item</span>
        </a>
        <a class="fd-nested-list__expand-icon" tabindex="-1"></a>
    </div>
</li>
```

### Before:
image
image

### After:
image
image


InnaAtanasova:
It's not related to the scope of this PR but maybe it's good idea to add some animation when we expand the second/third level of the Side Nav. Now it "jumps" and it's not the best UX behaviour. We can probably reach the designers for animation values @estelaxu3 @droshev

docs/pages/components/side-navigation.md
```diff
@@ -373,13 +383,11 @@ The lists in both sections (Main and Utility) should have the `fd-nested-list--c
 
 ## Side navigation with icons - condensed state
 Use the `fd-side-nav--condensed` modifier class for `condensed` state.
+Group headers are forbidden in `condensed` state.
```
InnaAtanasova: I think we can replace this text with:
The following requirements must be met:
- all first level items of the Side Navigation should have icons
- the Side Navigation should not have Group Headers

@Vanessa-Cusmich

estelaxu3:
I agree. We can propose this to designers.

docs/pages/components/side-navigation.md
stefanoScalzo: agreed

Vanessa-Cusmich: Agreed +1

JKMarkowski: Descriptions are changed.

droshev:
@JKMarkowski can you open an issue in NGX for bringing this change? Thanks!

Issue #819, opened 2020-03-30
Component Message Toast


droshev:
Introduce new component Message Toast

InnaAtanasova:
This component is dependent on [Message View](https://github.com/SAP/fundamental-styles/issues/854)

PR #826, opened 2020-03-31
fix: introduce new Message Toast component

InnaAtanasova:
## Related Issue
Closes SAP/fundamental-styles#819

## Description
- Introduce a new component to the library - Message Toast


InnaAtanasova:
> 1. I dont know why but there is an empty `<p></p>` which makes this space looking to big.
>    image

I wanted an empty line and added `<br>` but apparently because it's an `md` file it wrapped it with a `<p>` tag. I removed the `<br>`.

InnaAtanasova:
> 2. I'm not sure if moving @fd-reset to placeholders is a good idea, lets imagine scenario:
> 
> ```
> .component-class {
>     padding: $component-padding;
>     font-size: $component-font-size;
> 
>    @extends %dialog;
> }
> ```
> 
> And all code above is overwritten. I understood placeholders as chunks of code with reusable styling. `@fd-resize` somehow belongs to this 'reusable' definition but it doesn't bring desired styles, it resets unwanted.
> 
> I'd love to ear others opinion on that.

@salarenko I removed the reset from the placeholders

InnaAtanasova:
> > 1. I'm not sure if moving @fd-reset to placeholders is a good idea, lets imagine scenario:
> > 
> > ```
> > .component-class {
> >     padding: $component-padding;
> >     font-size: $component-font-size;
> > 
> >    @extends %dialog;
> > }
> > ```
> > 
> > 
> > And all code above is overwritten. I understood placeholders as chunks of code with reusable styling. `@fd-resize` somehow belongs to this 'reusable' definition but it doesn't bring desired styles, it resets unwanted.
> > I'd love to ear others opinion on that.
> 
> @salarenko I removed the reset from the placeholders

Usually all the @include and @extend functions happen at the beginning (right after you open the block) and then you add the css rules specific for the class. 
`.component-class {
   @extends %dialog;
    padding: $component-padding;
    font-size: $component-font-size;
}`
But you are right, if you don't follow this practice and we somehow miss it in the PR review things might go wrong. For this reason I removed the reset from the placeholders.

Issue #837, opened 2020-04-03
feat: DisplayOnly Mode checkbox styles


DeepakSap14:
Display Only checkbox needs to be created in Core checkbox. Currently styles for Displayonly mode for checkbox is not created. This PR submits styles changes for Display only checkbox.

**Is your feature request related to a problem? Please describe.**

> Is this issue related to a specific component?
Checkbox

> What browser are you working in?
Chrome, Firefox, Microsoft edge, safari

> What version of Fundamental Library Styles are you using?
0.8.1

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

> What front-end framework are you implementing in, e.g., Angular, React, etc.?
Angular

**Describe the solution you'd like**
Display only checkbox should be created.

**Describe alternatives you've considered**

**Additional context**

zaheral:
Hi @DeepakSap14 ,
I'm also interested on this issue. I saw that it is a while ago, since the last update. Do we have some news on this ticket?

DeepakSap14:
Hi @Zaheral,
Currently this is not in priority for us. we had a PR but it is closed without merge because of clarifications pending.
if you wish you can work on it. you can see related comments on.
https://github.com/SAP/fundamental-styles/pull/838.

droshev:
duplicated https://github.com/SAP/fundamental-styles/issues/1743

PR #838, opened 2020-04-03
fix: checkbox styles for display only mode

DeepakSap14:
## Related Issue
https://github.com/SAP/fundamental-styles/issues/837

## Description
This is Styles changes for Display Only checkbox.

### After:
Display only checkbox


JKMarkowski:
Hello @DeepakSap14, could you add example of usage to documentation page `checkbox.md`?

src/checkbox.scss
```diff
@@ -169,3 +170,61 @@ $block: #{$fd-namespace}-checkbox;
     }
   }
 }
+
+.#{$displayonlyblock} {
```
JKMarkowski, line 174: I would go more for `.is-readonly / readonly` state instad of having another block class. So class: `fd-checkbox is-readonly`. And on css `&.is-readonly + fd-checkbox__label {}`

src/checkbox.scss
```diff
@@ -169,3 +170,61 @@ $block: #{$fd-namespace}-checkbox;
...
+
+  $fd-displaycheckbox-margin: 0.5rem;
+
+  &__label {
+    width: fit-content;
+    min-width: fit-content;
+    overflow: hidden;
+    display: flex;
+    pointer-events: none;
```
JKMarkowski, line 183: fit-content doesn't work on ie11, Width doesn't need to be changed

src/checkbox.scss
```diff
@@ -6,6 +6,7 @@
 .fd-checkbox
 */
 $block: #{$fd-namespace}-checkbox;
+$displayonlyblock: #{$fd-namespace}-displayonly-checkbox;
```
InnaAtanasova, line 9: I don't think this should be a component. Display-only mode should be a modifier class. Example:
`fd-checkbox--display-only`

jbadan:
This should be marked `fix` if you've made no changes to the existing underlying html - see https://github.com/SAP/fundamental-styles/wiki/Automation-Processes---Versioning-and-Publishing#zeroth-version-library

DeepakSap14:
Hi,
i have updated checkbox.md file as well. but while creating example in checkbox , issue with indeterminate checkbox is found and styles not working as expected.(previous implementation as well) please check once.
Screenshot

docs/pages/components/checkbox.md
```diff
@@ -0,0 +1,199 @@
...
+{% include display-component.html component=checkbox-rtl %}
+
+## Display Only Checkbox
+
+{% capture checkbox-rtl %}
+    <input type="checkbox" class="fd-checkbox--display-only" id="1tn4O446" checked style="display:none">
+	<label class="fd-checkbox--display-only__label" for="1tn4O446">Label</label>
+  
+    <input type="checkbox" class="fd-checkbox--display-only" id="StRHi735"  style="display:none">
+	<label class="fd-checkbox--display-only__label" for="StRHi735">Label</label>
```
jbadan, line 176: I think this is a little backwards, I would think it should be something like `fd-checkbox__label fd-checkbox__label--display-only`.

docs/pages/components/checkbox.md
```diff
@@ -0,0 +1,199 @@
+---
```
jbadan, line 1: If we're going to move checkboxes to it's own page (👍 ), we should remove the examples from the "Form" page and provide a link to this page in it's place.

docs/pages/components/checkbox.md
```diff
@@ -0,0 +1,199 @@
...
+        <label class="fd-checkbox__label" for="h4z5N521">Label</label>
+    </div>
+{% endcapture %}
+
+{% include display-component.html component=checkbox-rtl %}
+
+## Display Only Checkbox
+
+{% capture checkbox-rtl %}
+    <input type="checkbox" class="fd-checkbox--display-only" id="1tn4O446" checked style="display:none">
```
jbadan, line 172: This should be `fd-checkbox fd-checkbox--display-only`. It should always have `fd-checkbox`

docs/pages/components/checkbox.md
```diff
@@ -0,0 +1,310 @@
...
+{% include display-component.html component=checkbox %}
+
+## Display Only Checkbox
+{% capture display-checkbox %}
+<fieldset class="fd-fieldset">
+    <legend class="fd-fieldset__legend">Display Only Checkboxes</legend>
+    <div class="fd-form-group">
+        <div class="fd-form-item">
+            <input type="checkbox" class="fd-checkbox fd-checkbox--display-only" id="1tn4O446" checked>
+            <label class="fd-checkbox__label--display-only" for="1tn4O446">Label</label>
```
jbadan, line 278: These should also have class `fd-checkbox__label`:  `fd-checkbox__label fd-checkbox__label--display-only`

DeepakSap14:
@jbadan @InnaAtanasova 
In fiori for Display only mode checkbox, as font-family, HTML Unicode etc values are defined. so i picked it to develop as Css. (same way for other checkboxes also these values are defined).
If we go by icon approach, we have to make major changes in fundamental-ngx/core checkbox implementation to accommodate Display Only checkbox icon.

jbadan:
What is the purpose of display only checkboxes? The spec isn't very clear to me. My assumption is that because read-only checkboxes are not considered an accessible pattern, that they've introduced this concept of display only. If that is the case, what is your plan to make implementing this accessible + not allowing interaction with the checkbox?

InnaAtanasova:
Closing this PR for now as we need further discussions with designers and UX team.

PR #839, opened 2020-04-03
fix: display list item new component styles as per Fiori 3

sKudum:
## Description
The list component of Fiori3 has the below-mentioned items types.

Screen Shot 2020-04-03 at 7 45 31 PM

This PR has the style related to the display list item component as per Fiori Design Guidelines.
Screen Shot 2020-04-03 at 7 47 43 PM


droshev:
Screen Shot 2020-04-05 at 9 48 11 AM

sKudum:
> Is it possible to simply extend `fd-list` instead of inventing a new component?

Hi jbadan, 

As per Fiori 3 List and all its subtypes and item types should be exposed as independent components.
That's is why we have started with the below approach.

Our approach will be.

1. List.scss=> will have list container styles
2. List-item-base.scss=> styles which are commanly used among all the item types of list.
3. <component>.scss=> component level css as per Fiori specfication.

By this, components won't be tightly coupled, users can use components as their requirement, fewer attributes to be filled by them and in future, any changes we want to take up it will be easy and will cause zero/less effect on other components.

sKudum:
> Is it possible to simply extend `fd-list` instead of inventing a new component?
> 
> This should also be marked `fix`, please see: https://github.com/SAP/fundamental-styles/wiki/Automation-Processes---Versioning-and-Publishing#zeroth-version-library

> Changes done.

jbadan:
@sKudum As discussed privately, we should talk about this with the group on our Thursday meeting.

My proposal is to use `fd-list` as the base of the component and add additional classes where necessary to achieve the different styles: 
`<ul class="fd-list"><li class="fd-list_item fd-list_item--display" tabindex="0">`, etc.  

This would achieve the desired outcome for the implementation libraries - not needing to repeat logic for the "n" number of lists that come out fiori3.

sKudum:
> @sKudum As discussed privately, we should talk about this with the group on our Thursday meeting.
> 
> My proposal is to use `fd-list` as the base of the component and add additional classes were necessary to achieve the different styles:
> `<ul class="fd-list"><li class="fd-list_item fd-list_item--display" tabindex="0">`, etc.
> 
> This would achieve the desired outcome for the implementation libraries - not needing to repeat logic for the "n" number of lists that come out fiori3.

Sure Jenna, will discuss with Inna and team and take the action accordingly, Meanwhile I forwarded the mail with the requirement details.

sKudum:
Hi @jbadan , 

As discussed I created it as a modifiers in list.scss ==>https://github.com/SAP/fundamental-styles/pull/847
it has action list item and display list item styles in it. 

If you are ok with this approach then I will close #839(current PR) and continue working on #847

jbadan:
@sKudum - yes that's great!

sKudum:
As per Jenna's suggestion created new component action and display list item styles as modifier in https://github.com/SAP/fundamental-styles/pull/849 in consideration with react. So closing this PR.

Issue #779, opened 2020-03-17
Refactor StepInput to match Fiori 3 specs


InnaAtanasova:
The StepInput pattern in Fundamental styles should be refactored to match the Fiori3 specs. This might require creating a StepInput component. Currently it's a pattern since it's composed by other components and doesn't have specific classes. 

Current implementation can be found  [here](https://deploy-preview-770--fundamental-styles.netlify.com/patterns/step-input.html)

prsdthkr:
It think it is because of the [Fiori 3 spec](https://wiki.wdf.sap.corp/wiki/pages/viewpage.action?pageId=2048477546), but the focus ring is not visible on the individual step up and step down buttons as seen in the patterns example. I can see this becoming an accessibility issue.

Patterns example shows focus ring on individual buttons:
![image](https://user-images.githubusercontent.com/43764832/81961400-2b8e5480-95c7-11ea-8e07-6fe898064489.png)


![image](https://user-images.githubusercontent.com/43764832/81961469-4791f600-95c7-11ea-9988-b2efa765e17c.png)

PR #841, opened 2020-04-05
fix: Create StepInput to match Fiori 3 specs

manu-kr:
## Related Issue

https://github.com/SAP/fundamental-styles/issues/779

## Description
The StepInput pattern in Fundamental styles should be refactored to match the Fiori3 specs. This might require creating a StepInput component. Currently it's a pattern since it's composed by other components and doesn't have specific classes

add example for step input in the documetation.

Create play ground for test step input.

### Before:
New component. so no before screent shot

### After:

image


jbadan:
Additionally, this pr should be marked `fix`, please see https://github.com/SAP/fundamental-styles/wiki/Automation-Processes---Versioning-and-Publishing#zeroth-version-library

jbadan:
@manu-kr I believe you and Inna have connected about the requested changes. Just a heads up - if you edit my comment to reply to me, I won't get a notification and will probably never see it 😄 It's better to leave a new comment.

manu-kr:
Hi Jenna,

I have discussed with Inna. Regarding your comments , I have few question on this. I will discuss with you on this today. Let me know once you’re available.

Thanks,
Manu

docs/_data/components.yml
```diff
@@ -78,6 +78,9 @@
     - title: Status Indicators
       meta: status-indicators
 
+    - title: Step Input
+      meta: step-input
+
```
jbadan, line 83: Can you remove the patterns step-input page?

jbadan:
@JKMarkowski we've gone back and forth on this - I do this it does need to be a component to match the spec (it will need overrides on `fd-input` on the bare minimum for width -it is weird it doesn't just match the other input components but that's not my decision). That said, I think we could simplify this further.

src/step-input.scss
```diff
@@ -0,0 +1,153 @@
+@import "./settings";
```
jbadan, line 1: `./settings` should be removed

src/step-input.scss
```diff
@@ -0,0 +1,153 @@
...
+$block: #{$fd-namespace}-step-input;
+
+.#{$block} {
+  $fd-step-input-border-thin-width: 0.0625rem;
+  $fd-step-input-border-thick-width: 2 * $fd-step-input-border-thin-width;
+  $fd-input-group-button-active-focus-color: var(--sapContent_ContrastFocusColor);
+  $fd-button-icon-font-size: 0.875rem;
+  $fd-button-icon-font-size-cozy: 1rem;
+
+  @mixin iconSize($size) {
```
jbadan, line 16: Do we need to create a mixin that we're only using once?

docs/pages/components/step-input.md
```diff
@@ -0,0 +1,181 @@
...
+
+The Step Input component is an opinionated composition of the Input, Button components.
+It is typically used to enter numbers.The value can be increased or reduced with the provided controls.
+{: .docs-intro}
+
+## Step Input with compact
+
+{% capture step-input-compact %}
+
+<div class="fd-step-input">
```
jbadan, line 19: Playing around with the branch, I'm failing to see why `<div class="fd-step-input">` is necessary. Can you explain why?

src/step-input.scss
```diff
@@ -0,0 +1,153 @@
...
+  @include fd-reset();
+  @include fd-rtl();
+
+  &__content-wrapper {
+    width: auto;
+    min-width: 2.75rem;
+    margin-top: 0.25rem;
+    margin-bottom: 0.25rem;
+    padding: 0;
+    box-sizing: border-box;
```
jbadan, line 40: `padding: 0` and `box-sizing: border-box` is already part of `fd-input-group` and can be removed here

src/step-input.scss
```diff
@@ -0,0 +1,153 @@
...
+    vertical-align: middle;
+    display: inline-block;
+    position: relative;
+
+    &.is-readonly {
+      outline: none;
+    }
```
jbadan, line 47: Should this just be for step-input or moved to input-group?

src/step-input.scss
```diff
@@ -0,0 +1,153 @@
...
+  &__content-wrapper:focus-within {
+    box-shadow: none;
+    outline-offset: -0.1875rem;
+    outline-width: var(--sapContent_FocusWidth);
+    outline-color: var(--sapContent_FocusColor);
+    outline-style: var(--sapContent_FocusStyle);
+
+    &.is-readonly {
+      outline: none;
+    }
```
jbadan, line 59: Same for this section? Is this specific to step input?

src/step-input.scss
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  &__button-container {
+    width: 2.125rem;
+    min-width: 2rem;
+    vertical-align: middle;
+    display: inline-block;
+    outline: none;
+    height: 100%;
```
jbadan, line 68: This height 100% seems unnecessary

src/step-input.scss
```diff
@@ -0,0 +1,153 @@
...
+
+    &.is-readonly {
+      display: none;
+    }
+
+    @include fd-hover() {
```
jbadan, line 74: Shouldn't this just be part of `fd-input-group__addon--button`?

src/step-input.scss
```diff
@@ -0,0 +1,153 @@
...
+    }
+  }
+
+  &__button-container--compact {
+    width: 1.5rem;
+    min-width: 1.5rem;
+  }
+
+  .#{$fd-namespace}-button.#{$block}__button {
+    @include fd-button-reset();
```
jbadan, line 85: We can remove ` @include fd-button-reset();` because that's already included in `fd-button`

src/step-input.scss
```diff
@@ -0,0 +1,153 @@
...
+    &.is-readonly {
+      display: none;
+    }
+
+    @include fd-hover() {
+      @include buttonContainerHover();
+    }
+  }
+
+  &__button-container--compact {
```
jbadan, line 79: You can add this like:

```
&__button-container {
    &--compact {
   }
}
```

src/step-input.scss
jbadan: Now that I play with it more, a bunch of the focus stuff needs to go in general. Try using this component with your keyboard - it's not possible to tab to the buttons as currently written.

InnaAtanasova:
@manu-kr 
Screen Shot 2020-04-16 at 1 00 59 PM

Please make sure you have all the necessary resets.

src/step-input.scss
InnaAtanasova: It is. The `@include buttonContainerHover();` is not needed

src/step-input.scss
```diff
@@ -0,0 +1,153 @@
+@import "./settings";
+@import "./new-settings";
+@import "./mixins";
+@import "./functions";
+@import "./button";
```
InnaAtanasova, line 5: Please don't import the button. This file is huge! You don't need `@include buttonContainerHover();`. Even if you need a mixin like this it's better to copy the mixin instead of importing a big file like button

src/step-input.scss
```diff
@@ -0,0 +1,153 @@
...
+
+  @include fd-fiori-focus() {
+    outline-color: $fd-input-group-button-active-focus-color;
+  }
+
+  @include fd-reset();
+  @include fd-rtl();
+
+  &__content-wrapper {
+    width: auto;
```
InnaAtanasova, line 35: IE11 has problems with `auto`. We need to support IE11

src/step-input.scss
```diff
@@ -0,0 +1,153 @@
...
+    @include iconSize($fd-button-icon-font-size);
+  }
+
+  .#{$fd-namespace}-button.#{$block}__button--compact {
+    width: 1.5rem;
+    min-width: 1.5rem;
+    height: 1.5rem;
+  }
+
+  &__inputtext {
```
InnaAtanasova, line 111: Should be `inputText`, better just `text` if it's not already used

src/step-input.scss
```diff
@@ -0,0 +1,153 @@
...
+    width: auto;
```
InnaAtanasova, line 112: Avoid using `auto` because of IE11 problems

JKMarkowski:
IE11 - There is `x` on input, I'm not sure if it's on purpose
image

src/step-input.scss
```diff
@@ -0,0 +1,153 @@
...
+    vertical-align: middle;
+    display: inline-block;
+    position: relative;
+
+    &.is-readonly {
+      outline: none;
+    }
+  }
+
+  &__content-wrapper:focus-within {
```
JKMarkowski, line 50: focus within is not supported by IE11, but there is no other way to handle focus, without JS.

src/step-input.scss
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    @include iconSize($fd-button-icon-font-size);
+  }
+
+  .#{$fd-namespace}-button.#{$block}__button--compact {
+    width: 1.5rem;
+    min-width: 1.5rem;
+    height: 1.5rem;
+  }
```
JKMarkowski, line 109: I would go for 
```scss
.#{$fd-namespace}-button {
    &.#{$block}__button {
        ....
        &--compact {
            ....
        }
    }
}
```

src/step-input.scss
JKMarkowski: And In general try to use `kebab-case`

src/step-input.scss
```diff
@@ -0,0 +1,153 @@
...
+      background-color: var(--sapSelectedColor);
+    }
+  }
+
+  &__content-wrapper--compact {
+    min-width: 2rem;
+    margin-top: 0.1875rem;
+    margin-bottom: 0.1875rem;
+  }
+
```
JKMarkowski, line 135: Let's put it into `&__content-wrapper` with `&--compact {}`

src/step-input.scss
JKMarkowski: Also please put it to `__content-wrapper` with `&:focus-within {}`

InnaAtanasova:
Closing the PR now as discussed with the team. We will re-open it later

PR #851, opened 2020-04-09
fix: breadcrumb and popover overflow

stefanoScalzo:
## Related Issue
Closes SAP/fundamental-styles#681

## Description
This PR removes overflow from popover so that the corners do not look weird. It also applies proper fiori 3 styling using dropdown styling for the truncation. Also fixes the link color of the arrow for the dropdown
### Before:
image

### After:
image


src/popover.scss
```diff
@@ -217,8 +225,8 @@ $block: #{$fd-namespace}-popover;
       display: -ms-grid;
       -ms-grid-columns: max-content;
       width: max-content;
-      border-top-left-radius: 0;
-      border-top-right-radius: 0;
+      border-top-left-radius: 0.125rem;
+      border-top-right-radius: 0.125rem;
```
InnaAtanasova, line 229: Should be a variable not value

jbadan:
This is a `feat` because you changed underlying html - see https://github.com/SAP/fundamental-styles/wiki/Automation-Processes---Versioning-and-Publishing#zeroth-version-library

src/popover.scss
```diff
@@ -217,8 +225,8 @@ $block: #{$fd-namespace}-popover;
...
 
       &-fill {
         max-width: 37.5rem;
```
rengare, line 232: please move value to variable

src/list.scss
```diff
@@ -379,12 +387,5 @@ $block: #{$fd-namespace}-list;
     .#{$block}__icon {
       line-height: 1rem;
```
rengare, line 388: please move value to variable

stefanoScalzo:
> I don't think `fd-list__title--anchor` is a good composition. Right now you support only title as anchor excluding other elements composed within `fd-list__item`, for example such scenario:
> 
> ```
> <ul class="fd-list">
>   <li tabindex="0"  class="fd-list__item">
>       <span class="fd-list__icon sap-icon--lightbulb"></span>
>       <a class="fd-list__title fd-list__title--anchor" tabindex="-1" href="#">Link Text</a>                    
>       <span class="fd-list__icon sap-icon--menu"></span>
>   </li>
> </ul>
> ```
> 
> I would rather see it like that:
> 
> ```
> <ul class="fd-list">
>   <li tabindex="0"  class="fd-list__item">
>       <a class="fd-list__link" tabindex="-1" href="#">
>           <span class="fd-list__icon sap-icon--lightbulb"></span>
>           <span class="fd-list__title">Link Text</a>                    
>           <span class="fd-list__icon sap-icon--menu"></span>
>       </a>                    
>   </li>
> </ul>
> ```
> 
> I don't know if your markup is enforced by the documentation but it seems to create usage limitations and to be Breadcrumb dedicated solution.
> 
> * We currently rather use `link` name for classes styling `<a>` tags.
> * Please update documentation if you create new features

It was dedicated to breadcrumb because this was the only component I saw wit this mistake. The regular dropdown doesn't have this but we can discuss during tomorrow's meeting!

src/popover.scss
stefanoScalzo: fixed

src/popover.scss
stefanoScalzo: done

src/list.scss
stefanoScalzo: done

docs/pages/components/breadcrumb.md
```diff
@@ -41,19 +41,19 @@ there should be popover with missing options.
...
                 </a>
             </div>
-            <div class="fd-popover__body fd-popover__body--no-arrow" aria-hidden="true" id="breadcrumb1">
-                <ul class="fd-list fd-list--compact">
-                    <li class="fd-list__item">
-                      <a class="fd-list__title" href="#">Link Text</a>
+            <div class="fd-popover__body fd-popover__body--no-arrow fd-popover__body--no-overflow fd-popover__body--dropdown" aria-hidden="true" id="breadcrumb1">
+                <ul class="fd-list fd-list--compact fd-list--dropdown" role="listbox">
+                    <li class="fd-list__item fd-list__item--anchor" tabindex="-1" role="option">
+                      <a class="fd-list__item fd-list__title fd-list__title--anchor" tabindex="0" href=""><span class="">Link Text</span></a>
```
InnaAtanasova, line 47: The solution with `anchor` would work only in the case where a li item has ONLY text. But list item can have icons (before/after), it could be a whole container like in the case of Byline. I am currently working on List component so I suggest you park this issue for now and wait for my PR to be merged. If the issue still exists after that you can continue working on it

stefanoScalzo: Hmm alright, i think the li would have to hold a div then so ill wait for your PR.

Issue #843, opened 2020-04-07
Menu links and icons have underline


meganaconley:
The `menu.scss` file applies `text-decoration: none` to the `fd-menu__title` element, but this class is assigned to the span within the anchor, which results in the anchor still displaying an underline.

In Fundamental React, the menu looks like this when only including `menu.css`:

<img width="364" alt="Screen Shot 2020-04-07 at 11 37 47 AM" src="https://user-images.githubusercontent.com/57502177/78706622-3d901f80-78c4-11ea-81c5-1a09d5a49e90.png">

The reason the Fundamental Styles docs version of the menu does not have an underline is because the style is coming from the `docs-all.css` file:

<img width="719" alt="Screen Shot 2020-04-07 at 11 28 36 AM" src="https://user-images.githubusercontent.com/57502177/78706107-6532b800-78c3-11ea-91b1-5e0b88ac6f21.png">

To fix this, `text-decoration: none` needs to move from `fd-menu__title` to `fd-menu__link`.

Issue #817, opened 2020-03-30
Component Menu - shortcuts features


droshev:
The Menu got a new feature: shortcuts - listing the keyboard combination for activating that item

PR #855, opened 2020-04-11
fix: Introduce menu shortcut | Improve A11y | Fix Menu issues

salarenko:
## Related Issue
Closes SAP/fundamental-styles#843
Closes SAP/fundamental-styles#817
Closes SAP/fundamental-styles#110

## Description
This PR:
- Refactors Menu code
- Fixes menu states bugs
- Fixes menu underline bug
- Adds missing menu styles like box shadow and rounded corners
- Introduces Menu A11y roles
- Introduces Menu Shortcut features
- Updates documentation and styles

### Before:

ols-sublist

old-states

### After:
image

new-sublist

new-states


src/menu.scss
```diff
@@ -5,352 +5,331 @@
...
+  $fd-menu-icon-width-compact: 2rem !default;
+  $fd-menu-icon-width-mobile: 2.75rem !default;
+  $fd-menu-icon-color: var(--sapContent_NonInteractiveIconColor) !default;
+  $fd-menu-icon-color-active: var(--sapList_Active_TextColor) !default;
+  $fd-menu-icon-color-selected: var(--sapContent_NonInteractiveIconColor) !default;
+
+  // Menu Title
+  $fd-menu-title-font-size: var(--sapFontSize) !default;
+
+  // Menu SHortcur
```
InnaAtanasova, line 45: Typo :)

src/menu.scss
```diff
@@ -5,352 +5,331 @@
...
+    left: 100%;
     z-index: 2;
-    margin: 0.25rem 0 0 -0.25rem; // offset of submenu
+    margin: $fd-menu-sublist-offset;
+    width: 100%; // to fix IE issue
 
     @include fd-rtl() {
       right: 100%;
       left: initial;
       margin: 0.25rem -0.25rem 0 0;
```
InnaAtanasova, line 111: Since you are moving everything to variables this one can be a var too

src/menu.scss
```diff
@@ -5,352 +5,331 @@
...
+      }
+
+      @include fd-rtl() {
+        .#{$block}__addon-before {
+          margin: 0 -0.75rem 0 0; // since before and after icons switch places, their margins also switch
+        }
+
+        .#{$block}__addon-after {
+          margin: 0 0 0 -0.5rem; // since before and after icons switch places, their margins also switch
+        }
```
InnaAtanasova, line 150: These should be vars, especially because they are reused in RTL and if the values change in future it will be easy to forget to update the values in the RTL block

src/menu.scss
```diff
@@ -5,352 +5,331 @@
...
+        .#{$block}__addon-before {
+          margin: 0 -1rem 0 0; // since before and after icons switch places, their margins also switch
+        }
 
-    @include fd-hover() {
-      background-color: $fd-menu-item-background-color--selected-hover;
+        .#{$block}__addon-after {
+          margin: 0 0 0 -1rem; // since before and after icons switch places, their margins also switch
+        }
+      }
```
InnaAtanasova, line 194: Values need to be moved in vars

src/menu.scss
```diff
@@ -5,352 +5,331 @@
...
-    }
 
-    .#{$block}__addon-before {
-      min-width: 2.75rem; // to account for item right padding of 16px or 1rem so 2.75rem-1rem = 1.75rem
-      width: 2.75rem; // to account for item right padding of 16px or 1rem so 2.75rem-1rem = 1.75rem
-      height: 1.25rem; // to account for item top padding of 12px or 0.75rem so 2.75rem-0.75rem-0.75rem = 1.25rem
-      margin: 0 0 0 -1rem;
-    }
+      .#{$block}__link {
+        padding: 0.75rem 1rem 0.75rem 1rem; // line-height=20px so 2.75rem or 44px-20px=24px/2= 12px or 0.75rem each
```
InnaAtanasova, line 169: padding: 0.75rem 1rem;

src/menu.scss
```diff
@@ -5,352 +5,331 @@
...
-
     &[aria-hidden='true'] {
       display: none;
     }
   }
 
   &__list > li {
     position: relative; // submenu's positioning relative to parent item
-    width: 100%;  // to ensure in mobile mode the width of parent is taken fully
+    width: 100%; // to ensure in mobile mode the width of parent is taken fully
```
rengare, line 121: can be moved to variable

src/menu.scss
```diff
@@ -5,352 +5,331 @@
...
   }
 
   &--compact {
-    @include fd-reset();
-
     .#{$block}__link {
-      // for desktop mode
       padding: 0.375rem 0.75rem 0.375rem 0.5rem; // line-height=20px so 2rem or 32px-20px=12px/2= 6px or 0.375rem each
```
rengare, line 126: you can move those to variables

src/menu.scss
```diff
@@ -5,351 +5,306 @@
...
+  // Menu Link
+  $fd-menu-icon-width: 2.25rem !default; // to account for item right padding of 16px or 1rem so 2.25rem-1rem = 1.25rem
+  $fd-menu-icon-width-compact: 2rem !default;
+  $fd-menu-icon-width-mobile: 2.75rem !default;
+  $fd-menu-icon-after-margin: 0 -1rem 0 0 !default;
+  $fd-menu-icon-before-margin: 0 0 0 -0.75rem !default;
+  $fd-menu-icon-before-margin-compact: 0 0 0 -0.5rem !default;
+  $fd-menu-icon-after-margin-compact: 0 -0.75rem 0 0 !default;
+  $fd-menu-icon-before-margin-mobile: 0 0 0 -1rem !default;
+  $fd-menu-icon-after-margin-mobile: 0 -1rem 0 0 !default;
```
JKMarkowski, line 23: I don't like those kind of variables. I would go more for `$fd-menu-icon-margin(-mobile)` with `-x`, or `-y`.
so ex. `$fd-menu-icon-margin-x: 1rem`

src/menu.scss
JKMarkowski: `0.25rem` should be in var

src/menu.scss
```diff
@@ -5,351 +5,306 @@
...
-    position: relative;   // so that sub-menus align relatively to parent menu item
-  }
-
   &__link {
     @include fd-reset();
+    @include action-cursor();
+    @include fd-fiori-focus();
 
-    border-radius: inherit;
+    text-decoration: none;
```
JKMarkowski, line 127: Maybe we should consider having someting like `fd-reset-link()`, Our docs styles often change colors/underline on hover with some weird transition time.

src/menu.scss
```diff
@@ -5,351 +5,306 @@
...
-    @include fd-rtl() {
-      margin: 0 -1rem 0 0; // since before and after icons switch places, their margins also switch
-    }
+  &--full-width {
+    max-width: 100%;
+  }
+
+  &--compact {
+    .#{$block}__link {
+      padding: 0.375rem 0.75rem; // line-height=20px so 2rem or 32px-20px=12px/2= 6px or 0.375rem each;
```
JKMarkowski, line 240: Maybe this should also be in variables

JKMarkowski: Also could you remove this comment? I don't think if it describes anything there.

src/menu.scss
JKMarkowski: Let's have a talk about it

salarenko: I guess that one is resolved according to our conversation.

src/menu.scss
salarenko: This value is used only in this one place.

src/menu.scss
salarenko: Good idea, but maybe we should move this to `fd-reset()`?

JKMarkowski: I'm not sure, I wouldn't add anything to `fd-reset`, it will increase size of our lib. Let's keep it as it is right now. Things like that should be included in separated PR.

InnaAtanasova:
@salarenko , one of your tests for menu is failing

droshev:
@salarenko is this still considered `feat`? if so can you document what we changed(markup or removed some classes)?

Issue #860, opened 2020-04-13
Tokenizer - broken padding in RTL mode


mikerodonnell89:
Need to add RTL support for tokenizer
![Screen Shot 2020-04-13 at 2 32 20 PM](https://user-images.githubusercontent.com/2471874/79158552-c6f79400-7d93-11ea-9dac-117084f0434a.png)

PR #861, opened 2020-04-13
fix: add RTL support for Tokenizer

mikerodonnell89:
## Related Issue
Closes SAP/fundamental-styles# #860 

## Description
Add rtl support for tokenizer

### Before:
Screen Shot 2020-04-13 at 2 32 20 PM

### After:
Screen Shot 2020-04-13 at 2 54 42 PM

### DIFF
--- a/src/tokenizer.scss
+++ b/src/tokenizer.scss
@@ -32,6 +32,10 @@ $block: #{$fd-namespace}-tokenizer;
 
   .fd-token {
     margin-right: $fd-tokenizer-spacing;
+
+    @include fd-rtl() {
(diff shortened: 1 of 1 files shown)


src/tokenizer.scss
```diff
@@ -79,6 +83,10 @@ $block: #{$fd-namespace}-tokenizer;
 
     .fd-token {
       margin-right: $fd-tokenizer-compact-spacing;
+
+      @include fd-rtl() {
+        margin: 0 0 0 $fd-tokenizer-spacing;
```
InnaAtanasova, line 88: I would move this to a variable

Issue #873, opened 2020-04-17
Side Navigation RTL support


droshev:
**Describe the bug**
At the moment Side Navigation has few RTL issues

> Is this issue related to a specific component?
Side Navigation

![image](https://user-images.githubusercontent.com/4380815/79598979-c708d400-80b2-11ea-8b33-db94f269617b.png)
![image](https://user-images.githubusercontent.com/4380815/79598986-ca9c5b00-80b2-11ea-8202-e3d148251227.png)
![image](https://user-images.githubusercontent.com/4380815/79598998-cff9a580-80b2-11ea-8ff9-b8aba5c49ea9.png)

PR #877, opened 2020-04-20
fix: Apply fixes to Side nav's rtl

JKMarkowski:
## Related Issue
fixes: https://github.com/SAP/fundamental-styles/issues/873

## Description
So right now border are on left side, when RTL. Also for first level of indentation, there is padding-left and padding-right

### Before:

image

### After:
image


InnaAtanasova:
I think you forgot to push something because I can see in your screenshots the little triangle is on the other side but in the preview it doesn't change the position

Screen Shot 2020-04-20 at 5 26 08 PM

InnaAtanasova:
Screen Shot 2020-04-20 at 5 25 45 PM

Not in the scope of this PR but the icon is not in the right position in RTL in this example

mikerodonnell89:
> Screen Shot 2020-04-20 at 5 25 45 PM
> 
> Not in the scope of this PR but the icon is not in the right position in RTL in this example

Noticed this as well but thought i should point out that it is fine on the examples with icons (cozy and compact mode)
Screen Shot 2020-04-21 at 1 50 22 PM

mikerodonnell89:
> I think you forgot to push something because I can see in your screenshots the little triangle is on the other side but in the preview it doesn't change the position
> 
> Screen Shot 2020-04-20 at 5 26 08 PM

Ok so after building this pr locally and copying the dist over to fundamental-ngx, the arrow is on the correct side.  However, in fundamental-styles, adding `dir="rtl"` to the parent element, I'm seeing what Inna is seeing here.

JKMarkowski:
So I fixed RTL 
image
 now additional padding is only on condensed rtl mode
image

JKMarkowski:
> Icon on IE 11 is not mirrored
> image

IE11 is finally supported

Issue #862, opened 2020-04-14
Radio button alignment in RTL


droshev:
**Describe the bug**
the inline radio button has different margin in RTL compared to not inline

> Is this issue related to a specific component?
![Screen Shot 2020-04-13 at 11 09 09 PM](https://user-images.githubusercontent.com/4380815/79182107-fb8f3e00-7ddb-11ea-8c4a-ba268f4a02e3.png)
![Screen Shot 2020-04-13 at 11 08 51 PM](https://user-images.githubusercontent.com/4380815/79182109-fc27d480-7ddb-11ea-8c32-be828143697d.png)

> What version of the Fundamental Library Styles are you using?
latest
> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

> What front-end framework are you implementing in, e.g., Angular, React, etc.?

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

JKMarkowski:
![image](https://user-images.githubusercontent.com/26483208/79969005-31de5480-8491-11ea-82d9-697fef968d45.png)
Checkbox also has this issue. It's because of `fd-form-group--inline` rtl support

PR #884, opened 2020-04-22
fix: Add Rtl support for form group

JKMarkowski:
## Related Issue
fixes: https://github.com/SAP/fundamental-styles/issues/862

## Description
There is removed margin from inline form-group

### Before:
image

### After:
image


src/form-group.scss
```diff
@@ -21,13 +21,26 @@ $block: #{$fd-namespace}-form-group;
   &--inline {
     display: flex;
     justify-content: flex-start;
+    margin-right: 1rem;
```
rengare, line 24: value should be moved to variable since its used in rtl mode as well

rengare: Also visual tests are failed

JKMarkowski: I decided to remove this margin, it's leftover after old `forms`. Now it's used only to make inline checkboxes/radio, which has labels with paddings/margins.

Issue #755, opened 2020-03-04
 IE11 - issue with Popover body overflow 


InnaAtanasova:
The Popover body does not take into account the width of the Bar component.

![image (3)](https://user-images.githubusercontent.com/39598672/75796035-1e402880-5d41-11ea-86da-3f1947a52868.png)

PR #886, opened 2020-04-23
fix: IE11 popover with bar component

JKMarkowski:
## Related Issue
fixes: https://github.com/SAP/fundamental-styles/issues/755

## Description
There was some issue on popover using bar. The bar component didn't expand container.

### Before:
image

### After:
image

### DIFF
--- a/src/bar.scss
+++ b/src/bar.scss
@@ -44,7 +44,8 @@ $block: #{$fd-namespace}-bar;
 
     display: flex;
     align-items: center;
-    flex: 1;
+    flex: 1 1 auto;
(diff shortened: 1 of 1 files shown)


stefanoScalzo:
we need to take into account overflow
Screen Shot 2020-04-23 at 12 14 53 PM

JKMarkowski:
> we need to take into account overflow
> Screen Shot 2020-04-23 at 12 14 53 PM

You are right, overflow works very bad there (like before those changes). Tehre is no way to handle it in our lib side.

Issue #852, opened 2020-04-10
List Enhancement - Navigation Indication


InnaAtanasova:
Our List component is missing Navigation Indication functionality that is present in Fiori 3 specs. This functionality is required for the List View of Message View component.

PR #887, opened 2020-04-24
fix: introduce Navigation Indication functionality to List component

InnaAtanasova:
## Related Issue
Closes SAP/fundamental-styles#852

## Description
- Add Navigation Indication functionality to List component. 
- Add list items with links. 
- Hover, Active, Selected, etc states are applied only to clickable (links) items. The states have been removed from `li`


src/list.scss
```diff
@@ -11,6 +11,53 @@
     .fd-list+(__item, __label, _input,  __title, __message, __footer)+(--no-wrap)
 */
 
+$block: #{$fd-namespace}-list;
+$form-item: #{$fd-namespace}-form-item;
+
+$fd-list-item-padding-x: 1rem;
+$fd-dropdown-vertical-padding: 0.75rem;
+$fd-dropdown-vertical-compact-padding: 0.5rem;
```
salarenko, line 19: no `!default` ?

InnaAtanasova: I just moved the existing variables outside of the block, didn't notice they were missing the `!default`. I addded it to all vars that were missing it. Thanks for noticing it.

InnaAtanasova:
> Implementation and examples look good. My other observations:
> 
> * Size of list and its variations is getting huge. Keeping all this code in single `.scss` file might cause future problems with maintenance. Also documentation has grown rapidly. Perhaps we should consider splitting this implementation into separate files and think about creating sub-levels for our documentation?
> * List elements are missing `is-disabled` state. I know that documentation doesn't say anything about it, but I think disabling list item is a common scenario. Perhaps we should contact designers or just add this state?

Point 1 - it's not in the scope of this PR. You have a point, we can discuss it internally and then with the other teams consuming the fund-styles.
Point 2 - the specs and the UX guideline do not mention the disabled states. Maybe @estelaxu3 could help us? It does not concern the current PR, it can be a separate issue.

docs/pages/components/list.md
```diff
@@ -74,23 +76,58 @@ In SAP Fiori, we distinguish between tables and lists. Both usually contain homo
...
+
+{% capture list-with-links-navigation %}
+<ul class="fd-list fd-list--navigation fd-list--navigation-indication" role="list">
+  <li tabindex="-1" role="listitem" class="fd-list__item fd-list__item--link">
+      <a tabindex="0" class="fd-list__link fd-list__link--navigation-indicator" href="#">
+        <span class="fd-list__title">List item 1</span>
+      </a>
+  </li>
+  <li tabindex="-1" role="listitem" class="fd-list__item fd-list__item--link">
+      <a tabindex="0" class="fd-list__link fd-list__link--navigation-indicator is-selected" href="#">
```
JKMarkowski, line 90: I like the idea with `--link` modifier

JKMarkowski:
image
What do you think about putting normal element there, instead of pseudo element? Also it could take whole height of container

src/list.scss
```diff
@@ -573,21 +637,277 @@ $form-item: #{$fd-namespace}-form-item;
...
+
+  // NAVIGATION INDICATION
+  &--navigation-indication {
+
+    .#{$block}__item {
+      &::after {
+        min-width: $fd-list-navigation-indicator-width - 1rem;
+        content: '';
+      }
+    }
```
JKMarkowski, line 651: Is there any reason, why we have it ?

InnaAtanasova: Screen Shot 2020-04-28 at 8 35 28 AM

Actually the answer to your 2 questions is one. Take a look at `List Item 4` and `List Item 5`. The elements that are on the right side should always be aligned, with or without navigation indicator. How I achieved is by using pseudo element that has a min-width which is the width of the nav indication. This space is now reserved for nav indication and the content to the left is aligned. If you have the class indicating that this element has nav indication you change the content of the pseudo element to >.

PR #888, opened 2020-04-24
docs: reduce some button examples(icon and text)

### LINKED ISSUE #881: Button - documentation page - remove button (text + icon)
Since Fiori 3 recommends only buttons with icon or with text then we should remove the examples from the doc where we show buttons with icons and text.
We should keep them as part of the playground at the same time

droshev:
fixes #881

### DIFF
--- a/docs/pages/components/button.md
+++ b/docs/pages/components/button.md
@@ -110,12 +110,12 @@ When the user activates the button, the menu opens. This is the default type.
 <button class="fd-button fd-button--negative fd-button--menu" aria-disabled="true" disabled>Negative Button</button>
 <button class="fd-button fd-button--attention fd-button--menu" aria-disabled="true" disabled>Attention Button</button>
 <br><br>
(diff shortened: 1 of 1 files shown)


estelaxu3:
The documentation should be revised, there is no **Icon and Text buttons** anymore. It should only be text only or icon only. 
Screen Shot 2020-04-24 at 12 39 15 PM

InnaAtanasova:
The title is not following the guidelines

droshev:
> The documentation should be revised, there is no **Icon and Text buttons** anymore. It should only be text only or icon only.
> Screen Shot 2020-04-24 at 12 39 15 PM

that is correct
it is either icon or text

estelaxu3:
> > The documentation should be revised, there is no **Icon and Text buttons** anymore. It should only be text only or icon only.
> > Screen Shot 2020-04-24 at 12 39 15 PM
> 
> that is correct
> it is either icon or text

Oh... I was ppbl having a stroke and thought Or is And. Sorry. 
Look good to me.

PR #889, opened 2020-04-24
fix: responsive grid layout with new docs

### LINKED ISSUE #297: Make layout-grid component IE11 compatible

### LINKED ISSUE #970: Layout Grid Breakpoints do not match Fiori3 spec
**Describe the bug**
Layout Grid Breakpoints do not match Fiori3 spec

our breakpoints are currently set here:
https://github.com/SAP/fundamental-styles/blob/master/docs/_sass/_include-media.scss#L32-L37
```
  'phone': 320px,
  'tablet': 768px,
  'desktop': 1024px,
  'display': 1920px
```

according to the SAP Fiori3 specification here:
https://wiki.wdf.sap.corp/wiki/pages/viewpage.action?pageId=2038570628 and
https://wiki.wdf.sap.corp/wiki/pages/viewpage.action?pageId=2038567803

we should be using the following:

| Name    | Container Width                                  | Cols. | Col. Gutter | Row Gutter |
|---------|--------------------------------------------------|-------|-------------|------------|
| Size XL | @media (min-width: 1440px)                       | 16    | 1rem        | 1rem       |
| Size L  | @media (min-width: 1024px) & (max-width: 1439px) | 12    | 1rem        | 1rem       |
| Size M  | @media (min-width: 600px) & (max-width: 1023px)  | 8     | 1rem        | 1rem       |
| Size S  | @media (max-width: 599px)                        | 4     | 0.5rem      | 0.5rem     |

stefanoScalzo:
BREAKING CHANGES: 
fd-col--span removed
fd-layout-grid removed

## Related Issue
Closes #970 
Closes #297

## Description
These new grid layouts will be compatible with IE11 and fiori3.
Removed Classes `fd-col--span`, `fd-layout-grid`


src/layout-grid.scss
```diff
@@ -2,44 +2,156 @@
...
 */
 
-$block: #{$fd-namespace}-layout-grid;
+$block: #{$fd-namespace}-col;
+$blockContainer: #{$fd-namespace}-container;
+$blockRow: #{$fd-namespace}-row;
 
-.#{$block} {
-  @include fd-reset();
+@media (min-width: 0) {
```
JKMarkowski, line 12: You can remove this selector

src/layout-grid.scss
```diff
@@ -2,44 +2,156 @@
...
+
+  .#{$blockRow} {
+    margin-bottom: 1rem;
+
+    .#{$block} {
+      .#{$blockRow} + .#{$blockRow} {
+        margin-top: 1rem;
+      }
     }
   }
```
JKMarkowski, line 116: This is overcomplicated, I would go for 
```scss
.#{$blockRow} {
    margin-bottom: 1rem;
    &:last-chlid {
        margin-bottom: 0;
    }
}
```

And if you set margin for sides, use `margin-left` and `margin-right`

docs/pages/components/form.md
```diff
@@ -851,3 +851,239 @@ Do not use the checkbox control if:
...
+<br>
+
+## Form with 12 column Layout
+
+{% capture form-edit %}
+<div class="fd-edit-mode-container">
+    <div class="fd-form-edit-mode-row fd-form-edit-mode-row--vertical">
+        <div class="fd-form-edit-mode fd-form-edit-mode--12">
+            <div class="fd-form-edit-mode-flex">
+                <label class="fd-form-label" for="input-1">Default Input:</label>
```
JKMarkowski, line 1051: There are repeated ids, let's change it

docs/pages/components/form.md
```diff
@@ -851,3 +851,239 @@ Do not use the checkbox control if:
...
+{% endcapture %}
+{% include display-component.html component=form-edit %}
+
+<br>
+
+## Form with 12 column Layout
+
+{% capture form-edit %}
+<div class="fd-edit-mode-container">
+    <div class="fd-form-edit-mode-row fd-form-edit-mode-row--vertical">
```
JKMarkowski, line 1048: I don't get, why you need to put there markup different than for grids
image
Look at 2 last examples
This is what I achieved it with grid (Only one thing you need to do is lower label's margin/center it vertically)

JKMarkowski: That's the code
```
<div class="fd-container">
    <div class="fd-row">
        <div class="fd-col fd-col--4" style="text-align: right">
            <label class="fd-form-label" for="grid-input-1">Default Input:</label>
        </div>
        <div class="fd-col fd-col--7">
            <input class="fd-input fd-input--compact" type="text" id="grid-input-1" placeholder="Field placeholder text">
        </div>
    </div>
    <div class="fd-row">
        <div class="fd-col fd-col--4" style="text-align: right">
            <label class="fd-form-label" for="grid-input-1">Default Input:</label>
        </div>
        <div class="fd-col fd-col--7">
            <div class="fd-row">
                <div class="fd-col fd-col--6">
                    <input class="fd-input fd-input--compact" type="text" id="grid-input-1" placeholder="Field placeholder text">
                </div>
                <div class="fd-col fd-col--6">
                    <input class="fd-input fd-input--compact" type="text" id="grid-input-1" placeholder="Field placeholder text">
                </div>
            </div>
        </div>
    </div>
</div>```

File: src/layout-grid.scss (hunk 2 above)
Comment (stefanoScalzo) [reply src/layout-grid.scss:116]: We cannot use the solution provided because the first child from a nested row cannot have a margin bottom. And yes i tried using first child selector, that solution does not work either. It is to specifically target a row with has older/higherDOM siblings

File: docs/pages/components/form.md (hunk 4 above)
Comment (stefanoScalzo) [reply docs/pages/components/form.md:1048]: You are right, ill need to import the layout grid and fix this! BUT will still require addition markup for text alignment

File: docs/pages/components/form.md (hunk 3 above)
Comment (stefanoScalzo) [reply docs/pages/components/form.md:1051]: fixed

File: src/layout-grid.scss (hunk 1 above)
Comment (stefanoScalzo) [reply src/layout-grid.scss:12]: sounds good

File: docs/pages/components/form.md (hunk 4 above)
Comment (stefanoScalzo) [reply docs/pages/components/form.md:1048]: Lets just verify with the team it is ok, i had done it this way for self-contained but well see what they say

File: src/layout-grid.scss [hunk 5]
```diff
@@ -2,44 +2,147 @@
...
+
+    .#{$blockRow} {
+      margin: 0 -0.5rem;
     }
   }
+}
+
+@media (min-width: 1024px) {
+  .#{$block} {
+    @for $n from 1 through 12 {
```
rengare, line 131: You can simplify this storing 12, 16 to variable, creates a map function which gets number and return letters `l, xl` etc.

stefanoScalzo: fixed

docs/pages/components/form.md
```diff
@@ -851,3 +851,133 @@ Do not use the checkbox control if:
...
 
 <br>
+
+## Form with 4-7-1 Layout
+
+{% capture form-edit %}
+<div class="fd-container fd-form-layout-grid-container">
+    <div class="fd-row">
+        <div class="fd-col fd-col--4 fd-form-layout-grid-flex fd-form-layout-grid-flex--right">
+          <label class="fd-form-label" for="input-1">Default Input:</label>
```
JKMarkowski, line 861: Change Id, because it's redundant

test/templates/index.njk
```diff
@@ -18,12 +18,13 @@
...
           <li><a href="form-group">.fd-form-group</a></li>
           <li><a href="form-label">.fd-form-label</a></li>
+          <li><a href="form-layout-grid">.fd-form-layout-grid</a></li>
           <li><a href="form-item">.fd-form-item</a></li>
           <li><a href="form-message">.fd-form-message</a></li>
           <li><a href="form-select">.fd-form-select</a></li>
           <li><a href="icon">.fd-icon</a></li>
           <li><a href="image">.fd-image</a></li>
-		  <li><a href="info-label">.fd-info-label</a></li>
+		      <li><a href="info-label">.fd-info-label</a></li>
```
JKMarkowski, line 27: Please remove additional spacing

docs/pages/components/form.md
```diff
@@ -851,3 +851,133 @@ Do not use the checkbox control if:
...
+        <div class="fd-col fd-col--4 fd-form-layout-grid-flex fd-form-layout-grid-flex--right">
+          <label class="fd-form-label" for="input-2">2 Inputs: </label>
+        </div>
+        <div class="fd-col fd-col--7">
+            <div class="fd-row">
+                  <div class="fd-col fd-col--6">
+                      <input class="fd-input" type="text" id="input-2" placeholder="Field placeholder text">
+                  </div>
+                  <div class="fd-col fd-col--6">
+                      <input class="fd-input" type="text" id="input-3" placeholder="Field placeholder text">
```
jacobdevera, line 885: I'm not sure if this is just to demonstrate what 2 inputs look like in the layout, but `input-3` would need a label.

docs/pages/components/form.md
```diff
@@ -851,3 +851,133 @@ Do not use the checkbox control if:
...
+        <div class="fd-col fd-col--4 fd-form-layout-grid-flex fd-form-layout-grid-flex--right">
+          <label class="fd-form-label" for="input-1">Default Input:</label>
+        </div>
+        <div class="fd-col fd-col--7">
+            <input class="fd-input fd-input--compact" type="text" id="input-1" placeholder="Field placeholder text">
+        </div>
+    </div>
+    <div class="fd-row">
+        <div class="fd-col fd-col--4 fd-form-layout-grid-flex fd-form-layout-grid-flex--right">
+          <label class="fd-form-label fd-form-label--required" for="input-1c">Required Input: </label>
```
jacobdevera, line 911: The `id` `input-1c` doesn't seem to map to any `input`. I think you might've forgot to change the `id` for the corresponding inputs next to it that currently have `input-1`.

There are several of these in this PR

docs/pages/components/form.md
stefanoScalzo: fixed

docs/pages/components/form.md
stefanoScalzo: fixed

docs/pages/components/form.md
stefanoScalzo: fixed

mikerodonnell89:
Nested grid items overlapping on really small screens:
Screen Shot 2020-06-15 at 5 26 04 PM

stefanoScalzo:
> As @mikerodonnell89 this has some minor problems ad it is missing storybook, but we can tackle them down in next iterations.
> Maybe we should also consider replacing Grid with Layout Grid? Its weird to have them both.

We are replacing it though. Ive deleted all files with our regular grid. I will address the nesting, and yes storybook will be in another pr

stefanoScalzo:
> I think this is a good start. Let's try to do few iterations with some changes.
> Maybe you wanna squash all commits into one so you can minimize the work of solving conflicts.
> First let's start with the documentation:
> 
> * can you document all the classes that are part of the grid in the beginning
> * maybe it is worth documenting the inflection points
> * let's remove the example `Layout Grid with nesting` for now. It is confusing
> * Can we create those examples:
>   
>   * 4 columns which are equally distributed in the biggest inflection point:
> 
> ```
> <cell 1 (3 columns)><cell 2 (3 columns)><cell 3 (3 columns)><cell 4 (3 columns)>
> ```
> 
> ```
> <cell 1 (6 columns)><cell 2 (6 columns)>
> <cell 3 (6 columns)><cell 4 (6 columns)>
> ```
> 
> ```
> <cell 1 (12 columns)>
> <cell 2 (12 columns)>
> <cell 3 (6 columns)><cell 4 (6 columns)>
> ```
> 
> ```
> <cell 1>
> <cell 2>
> <cell 3>
> <cell 4>
> ```
> 
> After the documentation let's try to include some flex-box features
> 
> * for vertical alignment (like _align items_)
> * for horizontal alignment (like _justify content_)

fixed! we cannot add flexbox properties as they are not part of fiori 3 and they may not be what they user wants to accomplish

stefanoScalzo:
> Nested grid items overlapping on really small screens:
> Screen Shot 2020-06-15 at 5 26 04 PM
> 
> otherwise this is looking good to me

fixed

droshev:
> > I think this is a good start. Let's try to do few iterations with some changes.
> > Maybe you wanna squash all commits into one so you can minimize the work of solving conflicts.
> > First let's start with the documentation:
> > 
> > * can you document all the classes that are part of the grid in the beginning
> > * maybe it is worth documenting the inflection points
> > * let's remove the example `Layout Grid with nesting` for now. It is confusing
> > * Can we create those examples:
> >   
> >   * 4 columns which are equally distributed in the biggest inflection point:
> > 
> > ```
> > <cell 1 (3 columns)><cell 2 (3 columns)><cell 3 (3 columns)><cell 4 (3 columns)>
> > ```
> > 
> > 
> > ```
> > <cell 1 (6 columns)><cell 2 (6 columns)>
> > <cell 3 (6 columns)><cell 4 (6 columns)>
> > ```
> > 
> > 
> > ```
> > <cell 1 (12 columns)>
> > <cell 2 (12 columns)>
> > <cell 3 (6 columns)><cell 4 (6 columns)>
> > ```
> > 
> > 
> > ```
> > <cell 1>
> > <cell 2>
> > <cell 3>
> > <cell 4>
> > ```
> > 
> > 
> > After the documentation let's try to include some flex-box features
> > 
> > * for vertical alignment (like _align items_)
> > * for horizontal alignment (like _justify content_)
> 
> fixed! we cannot add flexbox properties as they are not part of fiori 3 and they may not be what they user wants to accomplish

I don't see anything done

stories/layout-grid/layout-grid.stories.js
```diff
@@ -1,451 +1,320 @@
...
-`;
+- Use the <code class="docs-code">fd-container--no-gap</code> modifier to remove the gutter spacing between all columns.
+
+- Use the <code class="docs-code">fd-row</code> to seperate rows
+
+-Use the code class="docs-code">fd-row</code> to seperate rows and have an automatic gutter in between columns when they column wraps to the next line.
+
+- Use the <code class="docs-code">fd-col</code> to create a column of a certain width. The default value will create a column the full width, being 12 columns of the parent at size small and above unless specified another size 
+
+- Use the <code class="docs-code">fd-col--x</code> modifier will make the column spread x-rows/12
```
droshev, line 18: Screen Shot 2020-07-21 at 7 32 03 PM

stories/layout-grid/layout-grid.stories.js
```diff
@@ -1,451 +1,320 @@
...
+<h2>Breakpoints</h2>
+- Small 600px <br>
+<code class="docs-code">fd-col--x</code> will be used until the next specified minimum breakpoint, column takes x/12 space of the row
+- Medium 600px<br>
```
droshev, line 22: duplicated 600pc

src/form-layout-grid.scss
```diff
@@ -0,0 +1,61 @@
+@import "./mixins";
+@import "./layout-grid.scss";
+
+$block: #{$fd-namespace}-form-layout-grid;
+$blockCol: #{$fd-namespace}-col;
+$blockContainer: #{$fd-namespace}-container;
```
droshev, line 6: maybe it is worth bringing the file `_container.scss` and merge it with this file?

InnaAtanasova:
I am still struggling to understand why we have to do `fd-col fd-col--4--xl`. BEM does not recommend chaining elements or modifiers like in the case of `fd-col--4--xl`. Honestly I've never met classes named like this. But it's probably just me and the other have the same opinion as you. As a developer I'd rather see something like `fd-col-4--xl` or `fd-col-4 fd-col--xl` than double modifiers. 
Anyways, the code and the documentation part look good. Great job!

src/form-layout-grid.scss
```diff
@@ -0,0 +1,165 @@
...
+
+      .#{$blockCol} {
+        .fd-form-label {
+          padding-bottom: 0;
+        }
+
+        @for $n from 0 through $cols {
+          &--#{$n}--m {
+            text-align: right;
+            justify-content: flex-end;
```
InnaAtanasova, line 91: You are using this block of code rather often. I'd put it in a mixin like 
```
@mixin something($align, $justify) {
    text-align: $align;
    justify-content: $justify;
}
```

and then use it like this:
```
@include @something('right', 'flex-end') 
```

InnaAtanasova:
Screen Shot 2020-08-03 at 6 05 50 PM

Is this ok in RTL?

droshev:
> I am still struggling to understand why we have to do `fd-col fd-col--4--xl`. BEM does not recommend chaining elements or modifiers like in the case of `fd-col--4--xl`. Honestly I've never met classes named like this. But it's probably just me and the other have the same opinion as you. As a developer I'd rather see something like `fd-col-4--xl` or `fd-col-xl--4` or `fd-col-4-xl` or `fd-col-xl-4` than double modifiers.
> Anyways, the code and the documentation part look good. Great job!

I like the suggestion. @stefanoScalzo do you think you can rename the classes?

stefanoScalzo:
> > I am still struggling to understand why we have to do `fd-col fd-col--4--xl`. BEM does not recommend chaining elements or modifiers like in the case of `fd-col--4--xl`. Honestly I've never met classes named like this. But it's probably just me and the other have the same opinion as you. As a developer I'd rather see something like `fd-col-4--xl` or `fd-col-xl--4` or `fd-col-4-xl` or `fd-col-xl-4` than double modifiers.
> > Anyways, the code and the documentation part look good. Great job!
> 
> I like the suggestion. @stefanoScalzo do you think you can rename the classes?

@InnaAtanasova @droshev I could rename the classes.
I had used --4 because by default, fd-col has a 12 column width, once you apply fd-col--widthWhatver it changes the width.
Also in the case that fd-col--4--wtvSize it is still overwriting fd-col and fd-col--widthWhatever

jbadan:
Shouldn't sizes be separate - `fd-col`, `fd-col--4`, `fd-col--xl` as 3 different classes. 

For example, something along the lines of:
```
        @for $n from 0 through $cols {
          &--#{$n} {
              .#{$blockCol}--m {
                    @include placement('right', 'flex-end');
                     margin-bottom: 0;
             }
          }
        }
``

stefanoScalzo:
Branch is failing due to button PR's which is also caused in master

droshev:
> Branch is failing due to button PR's which is also caused in master

Did you address @InnaAtanasova 's comments?

stefanoScalzo:
> > Branch is failing due to button PR's which is also caused in master
> 
> Did you address @InnaAtanasova 's comments?

yes @droshev class names have been changed

PR #894, opened 2020-04-27
feat: Modify examples with doubled menu box shadow

salarenko:
In the aftermath of: #855

## Description
New Menu implementation brings by default menu box-shadow. Examples using menu within popover need to be modified to prevent using double box shadows coming separately from Menu and Popover.

### Before
image

### After
image

- [x] [Breaking Changes Wiki](https://github.com/SAP/fundamental-styles/wiki/Breaking-Changes)

### DIFF
--- a/docs/pages/components/action-bar.md
+++ b/docs/pages/components/action-bar.md
@@ -164,7 +164,7 @@ When there are several main actions for a page, consider displaying them under a
                 </div>
                 <div class="fd-popover__body fd-popover__body--right" aria-hidden="true" id="wgxzK859">
                     <nav class="fd-menu" id="">
-                        <ul class="fd-menu__list">
+                        <ul class="fd-menu__list fd-menu__list--no-shadow">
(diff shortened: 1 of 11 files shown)


InnaAtanasova:
> I think this should be marked as `feat` because it's going to need to be called out to the implementation libraries that they will need to add an additional class name to the menu.
> 
> If this was just missed in the previous breaking pr, I'd link that pr here and call out it out in the pr title. Otherwise, for me personally, when I'm updating fundamental-styles in fundamental-react, I'd have never looked at this pr and wouldn't have added this class to these components.

I agree with this. Changed the title to `feat` and added `Update Breaking Changes` check in the template. Please update the wiki and check it when done. Thanks

salarenko:
@InnaAtanasova - I updated breaking changes

rengare:
@stefanoScalzo you can close your PR #917 since this one will fix the problem

Issue #876, opened 2020-04-17
The content of the Popover body is cutting the border radius of the Popover


droshev:
**Describe the bug**
The content of the Popover body is cutting the border radius of the Popover
> Is this issue related to a specific component?
popover
![image](https://user-images.githubusercontent.com/4380815/79600623-6fb83300-80b5-11ea-8288-d78c49cabd59.png)
https://github.com/SAP/fundamental-styles/pull/682

PR #895, opened 2020-04-27
fix: remove overflow from popover

stefanoScalzo:
## Related Issue
Closes SAP/fundamental-styles#876

## Description
This Pr will remove the overflow from the inner content in popover body
### Before:
Screen Shot 2020-04-27 at 11 11 39 AM

### After:
Screen Shot 2020-04-27 at 11 11 54 AM


stefanoScalzo:
> You can't use this trick because then you are breaking the popover with arrow. You need the overflow to show the arrows. And no, we can't change the implementation of the arrow because popper.js is using the same approach.
> 
> Screen Shot 2020-04-27 at 2 12 43 PM

So then we need to go with my initial approach of having border-radius on the dropdown?

JKMarkowski:
@stefan-sto This issue, is more complicated. If you put `border-radius` only inside popover, it won't affect menu items.

src/list.scss
```diff
@@ -133,6 +133,12 @@ $semantic-color: (
   list-style: none;
   width: 100%;
 
+  &--no-borderBot {
```
InnaAtanasova, line 136: What is `borderBot`?

src/popover.scss
```diff
@@ -231,6 +231,10 @@ $block: #{$fd-namespace}-popover;
       opacity: 0;
       visibility: hidden;
     }
+
+    &__no-overflow {
```
InnaAtanasova, line 235: This is not BEM. Since it's a modifier it should be `--no-overflow`

docs/pages/patterns/multi-input.md
```diff
@@ -586,7 +594,8 @@ To add text in the `body` of the component, simply include your text in the `fd-
         </div>
     </div>
     <div class="fd-popover__body fd-popover__body--no-arrow fd-popover__body--dropdown fd-popover__body--dropdown-fill" aria-hidden="true" id="F4GcKJH8a">
-        <ul class="fd-list fd-list--multi-input fd-list--has-message">
+      <div class="fd-popover__body__no-overflow">
```
InnaAtanasova, line 597: This cannot work. Something called `fd-popover__body--something`  should be applied on the `fd-popover__body` element.

src/list.scss
stefanoScalzo: borderBottom

docs/pages/patterns/multi-input.md
stefanoScalzo: this was the wrapper which we had discussed about

docs/pages/components/breadcrumb.md
```diff
@@ -42,20 +42,22 @@ there should be popover with missing options.
...
-                    </li>
-                    <li class="fd-list__item">
-                      <a class="fd-list__title" href="#">Link Text</a>
-                    </li>
-                    <li class="fd-list__item">
-                      <a class="fd-list__title" href="#">Link Text</a>
-                    </li>
-                </ul>
+              <div class="fd-popover__body__no-overflow">
+                <ul class="fd-list fd-list--no-borderBot fd-list--compact">
```
mikerodonnell89, line 46: IMO no bottom border should be default behavior so no extra class is needed

stefanoScalzo: I will create another PR for this and put this one as blocked

stefanoScalzo:
> We agreed to target only the rounder corners in this PR

It is though? I said the border will be in another PR @droshev

JKMarkowski:
I'm not entirely sure if we should use `feat` there. We introduce new way of dealing with some issues. Keeping old markups won't change anything.

mikerodonnell89:
> I'm not entirely sure if we should use `feat` there. We introduce new way of dealing with some issues. Keeping old markups won't change anything.

The markup has changed with the wrapper element

docs/pages/patterns/combobox-input.md
```diff
@@ -34,22 +34,57 @@ If the entries are not validated by the application, users can also enter custom
...
+              <span class="fd-list__title">Kiwi</span>
+          </li>
+          <li role="option" tabindex="0" class="fd-list__item">
+              <span class="fd-list__title">Kiwi</span>
+          </li>
+          <li role="option" tabindex="0" class="fd-list__item">
+              <span class="fd-list__title">Kiwi</span>
+          </li>
+      </ul>
+    </div>
```
InnaAtanasova, line 87: Could you use different fruits? And if you want to show the scroll case you prob want to add max-height

stefanoScalzo:
> Screen Shot 2020-05-14 at 4 23 42 PM
> Since Breadcrumb is now using the popover wrapper do we need this class?



> Screen Shot 2020-05-14 at 4 23 42 PM
> Since Breadcrumb is now using the popover wrapper do we need this class?

You bring a good point, yes we need the class for the border-radius of the popover body to be modified. It also made realize that we need to adjust the inner wrapper when it is inside breadcrumb so it also has the proper radius as I did for dropdown

docs/pages/patterns/combobox-input.md
stefanoScalzo: fixed

stefanoScalzo:
Waiting for designers to update us with message for the breadcrumb border radius

docs/pages/patterns/combobox-input.md
```diff
@@ -34,22 +34,57 @@ If the entries are not validated by the application, users can also enter custom
...
-                <span class="fd-list__title">Orange</span>
-            </li>
-            <li role="option" tabindex="0" class="fd-list__item">
-                <span class="fd-list__title">Banana</span>
-            </li>
-            <li role="option" tabindex="0" class="fd-list__item">
-                <span class="fd-list__title">Kiwi</span>
-            </li>
-        </ul>
+    <div class="fd-popover__wrapper docs-max-height">
```
droshev, line 37: here I still see the issue, which btw in master is not there
Screen Shot 2020-05-22 at 4 24 35 PM

src/popover.scss
```diff
@@ -219,6 +224,11 @@ $block: #{$fd-namespace}-popover;
       border-top-left-radius: 0;
       border-top-right-radius: 0;
 
+      .#{$block}__wrapper {
```
droshev, line 227: If you introduce a new class for popover I would expect to have it documented on the popover page.

docs/pages/patterns/combobox-input.md
JKMarkowski: I think we agreed to keep it as it is right now, there is no way to hide this part of scrollbar. Other solutions (hiding whole/customizing) were abandoned.

docs/pages/components/popover.md
```diff
@@ -24,6 +24,7 @@ The basic wrapping structure of a popover.
 * `fd-popover` - wrapper for the whole container
 * `fd-popover__control` - control element to toggle the display of the popover
 * `fd-popover__body` - wrapper that contains the popover content for CSS-only implementations
+* `fd-popover-body__wrapper` - wrapper for the body to prevent overflow
```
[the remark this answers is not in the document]
stefanoScalzo: It wouldn't make sense to re add an example because they are only to be used with things like combobox so it would confuse user to have combobox in popover.

docs/pages/patterns/combobox-input.md
```diff
@@ -288,15 +332,16 @@ To add text in the `body` of the component, simply include your text in the `fd-
...
 </div>
 <span class="fd-form-message fd-form-message--static fd-form-message--success">Success message</span>
 {% endcapture %}
 {% include display-component.html component=semantic %}
 
 ## Combobox Mobile Mode
-For mobile devices, or tablets, combobox component should be displayed in fullscreen mode.
-So instead of using popover and dropdown, it should be wrapped in `dialog` and `bar` components.
+For mobile devices or tablets, the Combobox component should be displayed in fullscreen mode.
+Instead of using popover and dropdown, it should be wrapped in `dialog` and `bar` components.
```
droshev, line 344: popover and dropdown are components too. Should we capitalize them too as it was done in the previous line?

stefanoScalzo: not part of the scope of this PR but i will do the first one

PR #902, opened 2020-04-27
chore: delete unused code, add size testing and improve travis build times

jbadan:
## Description

This pr uses `bundlesize` to track the size of `fundamental-styles.css`. If we like this, we can improve it in the future to track the files most consumers are actually using. Although most don't use `fundamental-styles.css` it's a good starting point/overview of what all the files combined would be in size.

Otherwise, removed unused/repetitive code


.travis.yml
```diff
@@ -3,16 +3,20 @@ node_js: 10.19.0
 dist: trusty
 services:
   - docker
-before_install:
-- docker pull backstopjs/backstopjs:3.9.2
```
jbadan, line 7: Moving this into just the testing job saves us ~40 seconds for our other jobs that don't need it

config/bundlesize.json
```diff
@@ -0,0 +1,8 @@
+{
+    "files": [
+        {
+            "path": "./dist/fundamental-styles.css",
+            "maxSize": "70 kB"
```
jbadan, line 5: This is the file that would need to be changed as the bundle grows - it's currently at ~68kb.

PR #906, opened 2020-04-28
chore: update readme

jbadan:
## Related Issue

## Description
Update the readme to reflect what breaking changes are in fundamental-styles. Here, unlike in the implementation libraries, we should be treating `Non-visual HTML attribute changes/additions (such as `role`, `aria-*`, `data-*`)` as breaking changes as they require developer intervention in the implementation libraries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ The following circumstances will be considered a MAJOR or BREAKING change:
 * Droppping existing classnames, css variables, color names, color groups, spacing parameters
 * The existing underlying HTML markup of a component is altered
 * Non-visual HTML attribute changes/additions (such as `role`, `aria-*`, `data-*`)
-* An existing unit test is altered to account for either of the above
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -92,13 +92,12 @@ Merges to the `master` branch will be published as a prerelease. Prereleases wil
 The following circumstances will be considered a MAJOR or BREAKING change:
 * Droppping existing classnames, css variables, color names, color groups, spacing parameters
 * The existing underlying HTML markup of a component is altered
+* Non-visual HTML attribute changes/additions (such as `role`, `aria-*`, `data-*`)
```
bcullman, line 95: perhaps an additional clarification sentence above, something like 

"Fundamental Styles provides CSS directly, and HTML as reference to consumers. Because of thes reference relationship of the HTML seen in fundamental styles, we want to be very clear when we alter that reference so that it is properly reflected in JS implementation libraries. Because of this, even non-visual changes will be treated as breaking."

rengare: @jbadan I think that what @bcullman wrote is correct, could you add this as well?

jbadan: @bcullman @rengare I also removed the note about altering unit tests.... because we don't have them here 😆 I think that line was more tailored to the implementation libraries.

Issue #844, opened 2020-04-09
New Component - Action List Item


sKudum:
### Component name
Action List Item

Create Action list item as per Fiori3 standards.

<img width="1234" alt="Screen Shot 2020-04-09 at 10 13 12 AM" src="https://user-images.githubusercontent.com/53487468/78858677-c2a54100-7a4a-11ea-8de8-60c9b334342b.png">

PR #910, opened 2020-04-28
fix: add ActionListItem functionality to List component

InnaAtanasova:
## Related Issue
Closes SAP/fundamental-styles#844

## Description
- Add ActionListItem functionality to List component
- Provide a solution for :active state not working properly in IE11


docs/pages/components/list.md
```diff
@@ -111,6 +111,55 @@ If only a subset of the list items are navigable you should indicate those by di
 
 {% include display-component.html component=list-with-links-navigation %}
 
+## Standard List with Action List Item
+
+{% capture standard-list-action %}
+<ul class="fd-list" role="list">
+  <li role="listitem" tabindex="0" class="fd-list__item fd-list__item--action">
+      <span class="fd-list__title">Forward</span>
```
meganaconley, line 119: It seems as if these should be `button`s instead of `span`s if the idea is they will trigger an action. Reading these with VoiceOver, they're currently indistinguishable from the plain list items.

InnaAtanasova:
> **Firefox double focus**
> image
> 
> **Buttons**
> I'm not sure if button element is a good choice for this since on ngx we made the assumption that this is anchor tag

Did you take a look at Megan's comment regarding a11y? I will work on the double focus line but I agree with her point.

meganaconley:
> **Buttons**
> I'm not sure if button element is a good choice for this since on ngx we made the assumption that this is anchor tag

To add to Inna's comment, the only thing that should ever be an anchor is an actual navigation link. If it doesn't trigger navigation to a new page it should be a button. I believe there are other instances of this in other components we'll need to follow up on.

PR #916, opened 2020-04-28
fix: shellbar product menu padding bug

### LINKED ISSUE #705: Shellbar - collapsed product menu hover sizing is weird
When the shellbar is collapsed to its smallest size, the product menu text disappears and only a down caret displays.  The hover state area is sized weird and not centered over the caret.

Screen Shot 2020-02-26 at 6 05 12 PM

Also while we're here, the docs mention 'fd-product-menu' which has been deprecated, so let's remove that.

mikerodonnell89:
fixes #705 

I am not sure if this is the proper way to get the `1.937rem` value which comes from button.scss.  Should that variable be imported from that file?


InnaAtanasova:
1. The space between the logo and the product menu disappears in mobile. Is it according to the specs?
Screen Shot 2020-05-04 at 2 34 46 PM

2. In RTL the product menu button is very close to the logo

Screen Shot 2020-05-04 at 2 32 33 PM

3. The logo has padding-right in RTL which should prob go on the other side
Screen Shot 2020-05-04 at 2 32 50 PM

JKMarkowski:
Could you rebase it, or merge current master? Shellbar's buttons has been changed. Maybe it's better to test them with those changes.

Issue #915, opened 2020-04-28
Docs: menu list in popover body on mobile screens has bad margin


mikerodonnell89:
On mobile screens, opening a popover body with a menu list has a margin of 48px, causing this issue: (coming from docs specific styling)
![Screen Shot 2020-04-28 at 3 42 44 PM](https://user-images.githubusercontent.com/2471874/80540792-390cd300-8967-11ea-8721-467f8eeb58a4.png)

PR #917, opened 2020-04-28
docs: remove margin with list

stefanoScalzo:
## Related Issue
Closes SAP/fundamental-styles#915

## Description
This pr will remove the extra padding which is no longer necessary in tablet mode. Design approved by @estelaxu3. 

### Before:
Screen Shot 2020-04-28 at 7 08 30 PM

### After:
Screen Shot 2020-04-28 at 7 08 26 PM

### DIFF
--- a/docs/_sass/_docs-main-content.scss
+++ b/docs/_sass/_docs-main-content.scss
@@ -25,54 +25,13 @@
     margin: 10px 0 10px 0;
   }
 
-  ul {
-    @media screen and (max-width: 950px) {
-      padding-left: 0;
-      margin-left: 3rem;
-    }
-  }
-
-<<<<<<< HEAD
-<<<<<<< HEAD
   > h2 {
     @extend .docs-header-h2;
-=======
-    > ul {
-        @media screen and (max-width: 950px) {
-            padding-left: 0;
-            margin-left: 3rem;
-        }
-    }
->>>>>>> docs:add more specfic class to ul
-
-    @media screen and (max-width: 950px) {
-      padding-left: 0;
-      margin-left: 1.5rem;
-=======
-    > h2 {
-        @extend .docs-header-h2;
-        @media screen and (max-width: 950px) {
-            padding-left: 0;
-        }
->>>>>>> docs: fix margins
-    }
   }
 
-<<<<<<< HEAD
   > h3 {
     @extend .docs-header-h3;
   }
-
-  > p {
-    @media screen and (max-width: 950px) {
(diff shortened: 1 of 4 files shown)


rengare:
Popover with Header and fd-menu looks odd to me @InnaAtanasova @droshev what do you think?

rengare:
@stefanoScalzo you can close this PR because @salarenko made new one (#894) which fix the issue

stefanoScalzo:
> @stefanoScalzo you can close this PR because @salarenko made new one (#894) which fix the issue

This pr is to make sure all ul dont get affected which is still needed for future in case someone uses a ul without a modifier class

docs/_sass/_docs-main-content.scss
```diff
@@ -22,7 +22,7 @@
         margin: 10px 0 10px 0;
     }
 
-    ul{
+    > ul {
         @media screen and (max-width: 950px) {
             padding-left: 0;
             margin-left: 3rem;
```
InnaAtanasova, line 28: I don't think we should style ul in the documentation. In my opinion this should be removed and if the documentation page needs styling for ul it should rely on a specific class

stefanoScalzo: fixed

InnaAtanasova:
Screen Shot 2020-05-01 at 1 56 16 PM
Screen Shot 2020-05-01 at 1 56 05 PM
Screen Shot 2020-05-01 at 1 55 09 PM

stefanoScalzo:
> If this is not in the scope of this PR I think a new issue should be created. Then you put the list inside a popover the borders overlap resulting in a darker border color.
> 
> Screen Shot 2020-05-01 at 1 56 16 PM
> 
> Screen Shot 2020-05-01 at 1 56 05 PM
> 
> Screen Shot 2020-05-01 at 1 55 09 PM

I've also noticed this issue and it's not only for this, we have overlapping issues in many places i will create an issue

docs/pages/components/notification.md
```diff
@@ -8,8 +8,7 @@ permalink: components/notification.html
 folder: components
 summary:
 ---
-
-{: .docs-intro}
+{: .docs-intro .docs-intro--def-margin}
```
droshev, line 11: this should not be part of this PR

stefanoScalzo: The PR changes all the margin-left's in the documentation. There is nothing touching the library so it should be all part of same pr as it is all in the same scope
@droshev

PR #922, opened 2020-04-29
feat: migrate pagination with theming parameters

droshev:
migrate pagination with theming parameters

BREAKING CHAGES:
- `...` element doesn't have more `fd-pagination__link` as class. It is not clickable 
- the class `fd-pagination__link--more` has been renamed to `fd-pagination__more`


src/pagination.scss
```diff
@@ -1,4 +1,4 @@
-@import "./settings";
+@import "./new-settings";
 @import "./mixins";
 @import "./functions";
```
jbadan, line 3: Can we remove functions?

docs/pages/components/pagination.md
```diff
@@ -89,11 +89,11 @@ When the first page is active, the Back arrow should be disabled.
     <a href="#" class="fd-pagination__link fd-pagination__link--previous" aria-label="Previous"
     aria-disabled="false"></a>
     <a href="#" class="fd-pagination__link">1</a>
-    <span class="fd-pagination__link fd-pagination__link--more" role="presentation"></span>
+    <span class="fd-pagination__link--more" role="presentation"></span>
     <a href="#" class="fd-pagination__link">21</a>
     <a href="#" class="fd-pagination__link" aria-selected="true">22</a>
     <a href="#" class="fd-pagination__link">23</a>
-    <span class="fd-pagination__link fd-pagination__link--more" role="presentation"></span>
+    <span class="fd-pagination__link--more" role="presentation"></span>
```
jbadan, line 96: Can we put this back to the way it was before? `"fd-pagination__link fd-pagination__link--more"` That way it follows the established patterns and `--more` would just have that extra before block

Issue #678, opened 2020-02-19
Breadcrumb - Fiori3 update


salarenko:
Breadcrumb Fiori3 specification has changed and components styling needs to be updated.

stefanoScalzo:
Only thing that was changed is the max-width which was already done because we found that out before it got posted by asking designers.

droshev:
One recent change was that the bottom margin is not there anymore

PR #923, opened 2020-04-30
feat: breadcrumb popover with proper fiori 3

stefanoScalzo:
BREAKING CHANGES: 
fd-breadcrumb__popover-body added

## Related Issue
Closes SAP/fundamental-styles#678

## Description
This PR fixes the ellipses/collapsed mode of breadcrumb to fiori 3.
It makes the border-top 0.125rem as discussed with @estelaxu3 and applies the proper list items
The overflow issue is targeted in another PR

### Before:
Screen Shot 2020-04-30 at 3 34 32 PM

### After:
Screen Shot 2020-04-30 at 3 37 01 PM


docs/pages/components/breadcrumb.md
```diff
@@ -41,21 +41,24 @@ there should be popover with missing options.
...
-                      <a class="fd-list__title" href="#">Link Text</a>                    
-                    </li>
-                    <li class="fd-list__item">
-                      <a class="fd-list__title" href="#">Link Text</a>
-                    </li>
-                    <li class="fd-list__item">
-                      <a class="fd-list__title" href="#">Link Text</a>
-                    </li>
-                </ul>
+            <div class="fd-popover__body fd-popover__body--no-arrow fd-popover__body--breadcrumb" aria-hidden="true" id="breadcrumb1">
```
InnaAtanasova, line 44: In my opinion this is not the best approach. Inside popover you shouldn't have classes that are specific for other components like `fd-popover__body--breadcrumb`. Isn't it possible to do it like this:
```
<div class="fd-popover__body fd-popover__body--no-arrow fd-breadcrumb__popover-body" >
```

In this case `fd-breadcrumb__popover-body` will bring the rounded corners and the code will be in the breadcrumb.scss file. Also I'd suggest testing this in NGX since it also involves popper.js and thing not always work out of the box there.

stefanoScalzo: I had thought about that, but i did some future preventative measures thinking that in the future it would need to use this attribute. I've noticed they use either 0.25 or 0.125. Maybe name it fd-popover__body--alt-radius? @InnaAtanasova

InnaAtanasova: `fd-popover__body--breadcrumb` should not be used inside other components. This is breadcrumb specific property (even if other components have the same value for rounded corners).

stefanoScalzo: Alright will fix

stefanoScalzo:
> @stefanoScalzo if we are changing the markup then the PR should be `feat` and listing the BREAKING CHANGES.
> Also the pipeline is failing.

Fixed

Issue #924, opened 2020-04-30
fix list active state background color


stefanoScalzo:
**Describe the bug**
The current background color of the list__item does not change when the item is selected
> Is this issue related to a specific component?
list
> What version of the Fundamental Library Styles are you using?
current
<img width="214" alt="Screen Shot 2020-04-30 at 4 19 38 PM" src="https://user-images.githubusercontent.com/50607147/80755252-7f486a80-8afe-11ea-9863-2b03efab8b55.png">

PR #925, opened 2020-04-30
fix: active state of list item for combobox

stefanoScalzo:
## Related Issue
Closes SAP/fundamental-styles#924

## Description
This Pr will change the background color and the text color of the active state for the list item for combobox. It will not become the appropriate fiori 3 theming color. There is no focus outline as shown in the specs for these items. This example is for combobox but I would keep it under the list styling for future possible purposes

### Before:
Screen Shot 2020-04-30 at 4 19 38 PM

### After:
Screen Shot 2020-04-30 at 4 21 06 PM

### DIFF
--- a/src/list.scss
+++ b/src/list.scss
@@ -383,6 +383,56 @@ $semantic-color: (
     .#{$block}__item {
       height: auto;
       padding: $fd-dropdown-vertical-padding $fd-list-item-padding-x;
+
+      @include fd-disabled() {
+        pointer-events: none;
+        opacity: 0.5;
+      }
+
+      @include fd-selected() {
+        border-bottom-color: var(--sapSelectedColor);
+        background-color: var(--sapList_SelectionBackgroundColor);
+        cursor: grabbing;
+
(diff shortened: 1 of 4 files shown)


src/list.scss
```diff
@@ -909,4 +909,20 @@ $semantic-color: (
       @include navigation-byline-borderless();
     }
   }
+
+  &--interactive {
+    > * {
```
InnaAtanasova, line 914: This shouldn't be the universal selector. You should use `fd-list__item`. If a list item is interactive you need to have styling for all interactive states - hover, selected, selected+hover, active and the cursor should be a pointer.

InnaAtanasova: To add more to my comment: the interaction states should be applied to `fd-list--dropdown`

InnaAtanasova:
Please check if this is fixed with the merge of ActionListItem (already in master)

InnaAtanasova:
Closing this because the issue is no longer reproducible

Issue #938, opened 2020-05-03
Defect Hunting - May 2-4, 2020


droshev:
List of issues found during the defect hunting on May 2nd-4th, 2020

- [x] documentation bleeding into list (Gracjan) #941
![Screen Shot 2020-05-02 at 10 49 03 PM](https://user-images.githubusercontent.com/4380815/80897494-49a2ad80-8cc7-11ea-8b11-7ae5747161f3.png)
![Screen Shot 2020-05-02 at 10 48 57 PM](https://user-images.githubusercontent.com/4380815/80897495-4a3b4400-8cc7-11ea-8087-0bddb178e07c.png)
- [x] List item focus looks not correct(3 pixels inside) - I'm Sure that 3px focus is correct [Jedrzej]
List
![Screen Shot 2020-05-02 at 10 54 23 PM](https://user-images.githubusercontent.com/4380815/80897557-0137bf80-8cc8-11ea-9508-94f77b14c0ed.png)
![image](https://user-images.githubusercontent.com/26483208/80959599-71009400-8e07-11ea-982d-d6a0d880586c.png)

Menu
![Screen Shot 2020-05-02 at 10 54 13 PM](https://user-images.githubusercontent.com/4380815/80897556-fd0ba200-8cc7-11ea-8a92-e2f9a7b73e5a.png)

- [x] Side Navigation RTL the arrow indicating nested levels is on the wrong side (Gracjan) #946
![Screen Shot 2020-05-02 at 11 01 38 PM](https://user-images.githubusercontent.com/4380815/80897648-f9c4e600-8cc8-11ea-9a89-2e8a66742640.png)

- [x] Select RTL - button is off - Jedrzej https://github.com/SAP/fundamental-styles/pull/939
![Screen Shot 2020-05-02 at 11 03 01 PM](https://user-images.githubusercontent.com/4380815/80897655-1e20c280-8cc9-11ea-84be-3664a60c1df0.png)

- [x] Shellbar > Avatar has wrong size(too wide) - Jedrzej https://github.com/SAP/fundamental-styles/pull/940
![Screen Shot 2020-05-02 at 11 13 20 PM](https://user-images.githubusercontent.com/4380815/80897790-9fc52000-8cca-11ea-902c-f08e805686c9.png)

- [x] Popover without arrow next to input - extra padding - Jedrzej **To Discuss**
[...]

PR #941, opened 2020-05-04
docs: fixed margin problem for list

rengare:
## Related Issue
Related #938 

## Description
Doc class removed which fix margin problem
### Before:
image

### After:
image


stefanoScalzo:
Do not merge this there is another bug which will be fixed in https://github.com/SAP/fundamental-styles/pull/917

PR #945, opened 2020-05-04
fix: RTL calendar landscape orientation

JKMarkowski:
## Related Issue
Defect Hunting

## Description
There is removed one workaround done, to achieve floating `close` button on landscape mobile mode for calendar. It also fixes RTL issues there.

### Before:
image

### After:
image


docs/_sass/_docs-display-component.scss
```diff
@@ -89,10 +89,6 @@
         width: 640px;
         height: 360px;
       }
-
-      .fd-calendar__navigation--main {
```
stefanoScalzo, line 93: You are removing the class but the .md still uses it

InnaAtanasova: This is documentation scss, the class is still there in the `calendar.scss`

InnaAtanasova:
https://github.com/SAP/fundamental-styles/issues/952

Opened an issue for the documentation in mobile view.

PR #947, opened 2020-05-04
fix: Add RTL support to Avatar

salarenko:
## Related Issue
Part of: #938

## Description
Add RTL support to Avatar

### Before:
image

### After:
image


stefanoScalzo:
Can you please include an example in the documentation

src/avatar.scss
```diff
@@ -62,6 +62,11 @@ $block: #{$fd-namespace}-avatar;
         bottom: map_get($size-set, "offset");
         right: map_get($size-set, "offset");
 
+        @include fd-rtl() {
+          right: auto;
+          left: map_get($size-set, "offset");
```
stefanoScalzo, line 67: might want to use this as a variable

InnaAtanasova: no need to be a variable, it's using map_get

PR #949, opened 2020-05-04
feat: Remove search input

salarenko:
## Related Issue
Part of: #938
This PR:
- Search Input Pattern documentation page
- Remove `fd-search-input` styles
- Update shellbar tests to use stub `input-group` without `fd-search-input`

## Description
`SearchInput` has been removed. There are still some usages in "demo" layouts, but since they are no longer supported I don't think it's worth spending time on fixing them.

## BREAKING CHANGES


stefanoScalzo:
Our `shell-header.html` still uses fd-search-input

salarenko:
@stefanoScalzo `shell-header.html` is a part of demo pages

PR #950, opened 2020-05-04
docs: Display icon for each message-box example

salarenko:
## Related Issue
Part of: #938

## Description
Show icons for every message-box in "Message Box With Responsive Behavior" example section to keep things consistent.

### Before:
image

### After:
image


stefanoScalzo:
As discussed, we need both examples

PR #986, opened 2020-05-08
fix: Step Input

JKMarkowski:
## Related Issue
fixes: https://github.com/SAP/fundamental-styles/issues/779

### After:
image
image


droshev:
Can you create a storybook instead of njk visual tests?

docs/pages/components/step-input.md
```diff
@@ -0,0 +1,445 @@
...
+title: Step Input
+id: step-input
+keywords: Step Input, Step, Input
+sidebar: left-navigation-sidebar
+toc: false
+permalink: components/step-input.html
+folder: components
+---
+
+The Step Input component is an opinionated composition of the Input, Button components.
```
Vanessa-Cusmich, line 11: Can we use the description from the Fiori Guidelines? I think it sounds better. I also included the usage guidelines in case you want to add them.

The step input control allows the user to change the input values in predefined increments (steps).

Usage
Use the step input if:
The user needs to adjust amounts, quantities, or other values quickly.
The user needs to adjust values for a specific step (for example, in a shopping cart).
Do not use the step input if:
The user needs to enter a static number (for example, postal code, phone number, or ID). In this case, use the regular input field control instead.
You want to display a value that rarely needs to be adjusted and does not pertain to a particular step. In this case, use the regular input field control instead.
You want the user to enter dates and times. In this case, use the date picker, date range selection, time picker, or date/time picker instead.

docs/pages/components/step-input.md
```diff
@@ -0,0 +1,445 @@
...
+It is typically used to enter numbers.The value can be increased or reduced with the provided controls.
+{: .docs-intro}
+
+
+## Step Input with Cozy
+Cozy Mode is the default one, it should be used in mobile and tablet devices.
```
Vanessa-Cusmich, line 17: On smartphones and tablets, the step input is shown in cozy mode (default).

docs/pages/components/step-input.md
```diff
@@ -0,0 +1,445 @@
...
+            </button>
+    </div>
+    
+</div>
+{% endcapture %}
+{% include display-component.html component=step-input-cozy %}
+
+
+## Compact Step Input
+Step Input can be used in compact mode, which is done to be used in desktop, or larger screen devices.
```
Vanessa-Cusmich, line 76: The Step Input should be used in compact mode when using a desktop of devices with large screens. It can be achieved by adding the `--compact` modifier to the main element as well as the button and input elements.

docs/pages/components/step-input.md
```diff
@@ -0,0 +1,445 @@
...
+                type="button">
+            </button>
+    </div>
+    <span class="fd-form-label fd-form-label--unit-description">PC</span>
+</div>
+{% endcapture %}
+{% include display-component.html component=step-input-compact %}
+
+## Step Input with Different states
+Step Input Component also supports semantic states, same as for every form. They can be achieved by adding
```
Vanessa-Cusmich, line 144: The Step Input component also supports semantic states as does every form. The semantic states can be customized by adding the `is-error` | `is-success` | `is-warning` | or `is-information`  into the `fd-step-input` element.

JKMarkowski:
> Can you create a storybook instead of njk visual tests?

For now both are added

src/step-input.scss
```diff
@@ -0,0 +1,115 @@
...
+
+.#{$block} {
+  $fd-button-icon-font-size: 0.875rem;
+  $fd-step-input-cozy-height: 2.25rem;
+  $fd-step-input-compact-height: 1.5rem;
+  $fd-step-input-cozy-button-width: 2.125rem;
+  $fd-step-input-compact-button-width: 1.5rem;
+  $fd-step-input-description-padding: 0.5rem;
+
+  @mixin iconSize($size) {
```
salarenko, line 13: Is it worth to create a mixin if its used only once?

src/step-input.scss
```diff
@@ -0,0 +1,115 @@
...
+  @include fd-form-states();
+  @include fd-fiori-focus-within();
+
+  width: auto;
+  border-style: solid;
+  border-width: var(--sapField_BorderWidth);
+  border-color: var(--sapField_BorderColor);
+  border-radius: var(--sapField_BorderCornerRadius);
+  background-color: var(--sapField_Background);
+  height: calc(#{$fd-step-input-cozy-height} + 2 * var(--sapField_BorderWidth));
```
salarenko, line 31: I don't know if this is correct because there is no exact value in spec, but our standard input height is `36px` and step-input height is `38px`.

src/step-input.scss
```diff
@@ -0,0 +1,115 @@
...
+    border: none;
+    border-radius: 0;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+
+    @include fd-focus() {
+      outline: none;
+
+      &,
```
salarenko, line 50: Shouldn't you use `@include fd-hover()` here?

src/step-input.scss
```diff
@@ -0,0 +1,115 @@
...
+      outline: none;
+    }
+
+    &--compact {
+      height: 100%;
+    }
+  }
+
+  /** Width is different due to different border width  */
+  &.is-information,
```
salarenko, line 87: Missing `success` state

src/step-input.scss
```diff
@@ -0,0 +1,115 @@
...
+  &.is-error,
+  &.is-warning {
+    .#{$block}__button.#{$fd-namespace}-button {
+      width: calc(#{$fd-step-input-cozy-button-width} - 0.125rem);
```
salarenko, line 91: We should move `0.125rem` to a variable

src/step-input.scss
JKMarkowski: success has `1px` from mixin

src/step-input.scss
JKMarkowski: You are right, there should be `36px` with borders included (34px + 2px)

JKMarkowski:
All changes are addressed, it's ready to be reviewed again

mikerodonnell89:
can we just have `--no-number-spinner` be default behavior for our number type inputs so no extra class is needed?

JKMarkowski:
> can we just have `--no-number-spinner` be default behavior for our number type inputs so no extra class is needed?

We decided to have it in other PR

prsdthkr:
It think it is because of the [Fiori 3 spec](https://wiki.wdf.sap.corp/wiki/pages/viewpage.action?pageId=2048477546), but the focus ring is not visible on the individual step up and step down buttons as seen in the patterns example. I can see this becoming an accessibility issue.

Patterns example shows focus ring on individual buttons:
image


image

prsdthkr:
> It think it is because of the [Fiori 3 spec](https://wiki.wdf.sap.corp/wiki/pages/viewpage.action?pageId=2048477546), but the focus ring is not visible on the individual step up and step down buttons as seen in the patterns example. I can see this becoming an accessibility issue.
> 

**Correction:** It is okay that the focus ring is not appearing on the step buttons. This is in line with the `<input type="number">` implementation as seen here: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/step.

But we should also make sure that the buttons inside the input are not tab-able. Currently there is an invisible tab stop. As seen in the above GIF.

JKMarkowski:
@prsdthkr Thanks for pointing it, I will add `tabindex: -1` to all buttons, to make them not tabable

PR #1012, opened 2020-05-12
feat: updates to SideNavigation for accessibility

meganaconley:
## Description
- Updated SideNavigation to be a collection of `nav` sections for multiple groups
- Moved group header out of the list so it can be correctly associated as the title of each nav
- Added aria-hidden to icons except in compact mode, where aria-label was added to replace the text
- Added aria-label to utility menu as it does not have an option for a visible header
- Updated role of nested group header to `button` and removed `href` as it is not a link
- Added skip link to top of navigation list so users can skip to the next content if desired

## Screenshots
No visual changes

### DIFF
--- a/docs/pages/components/side-navigation.md
+++ b/docs/pages/components/side-navigation.md
@@ -142,11 +142,11 @@ On expand, the `is-expanded` class should be propagated also to `__content` elem
                 </a>
             </li>            
             <li class="fd-nested-list__item">
-                <div class="fd-nested-list__content has-child" tabindex="0">
-                    <a class="fd-nested-list__link" role="button" tabindex="-1">
+                <div class="fd-nested-list__content has-child">
+                    <span class="fd-nested-list__link">
                         <span class="fd-nested-list__title">Level 1 Item</span>
-                    </a>
-                    <a class="fd-nested-list__expand-icon" role="button" tabindex="-1" aria-controls="EX100L2" aria-haspopup="true"></a>
+                    </span>
+                    <button class="fd-nested-list__expand-icon" aria-controls="EX100L2" aria-haspopup="true" aria-expanded="false"></button>
                 </div>
                 <ul class="fd-nested-list level-2" id="EX100L2" aria-hidden="true">
                     <li class="fd-nested-list__item">
@@ -221,12 +221,12 @@ On expand, the `is-expanded` class should be propagated also to `__content` elem
                 </a>
             </li>
             <li class="fd-nested-list__item">
-                 <div class="fd-nested-list__content is-selected has-child" tabindex="0">
-                    <a class="fd-nested-list__link" role="button" tabindex="-1">
+                <div class="fd-nested-list__content is-selected has-child">
+                    <span class="fd-nested-list__link">
                         <span aria-hidden="true" class="fd-nested-list__icon sap-icon--employee"></span>
                         <span class="fd-nested-list__title">Level 1 Item</span>
-                    </a>
-                    <a class="fd-nested-list__expand-icon" role="button" aria-controls="EX400L2" tabindex="-1" aria-haspopup="true"></a>
+                    </span>
+                    <button class="fd-nested-list__expand-icon" aria-controls="EX100L2" aria-haspopup="true" aria-expanded="false"></button>
                 </div>
                 <ul class="fd-nested-list fd-nested-list--text-only level-2" id="EX400L2" aria-hidden="true">
                     <li class="fd-nested-list__item">
@@ -317,12 +317,12 @@ The lists in both sections (Main and Utility) should have the `fd-nested-list--c
                 </a>
             </li>
             <li class="fd-nested-list__item">
-                <div class="fd-nested-list__content has-child" tabindex="0">
-                    <a class="fd-nested-list__link" href="#" tabindex="-1">
+                <div class="fd-nested-list__content has-child">
+                    <span class="fd-nested-list__link">
                         <span aria-hidden="true" class="fd-nested-list__icon sap-icon--employee"></span>
                         <span class="fd-nested-list__title">Level 1 Item</span>
-                    </a>
-                    <a class="fd-nested-list__expand-icon" href="#" tabindex="-1" aria-controls="EX500L2" aria-haspopup="true"></a>
+                    </span>
+                    <button class="fd-nested-list__expand-icon" aria-controls="EX100L2" aria-haspopup="true" aria-expanded="false"></button>
                 </div>
                 <ul class="fd-nested-list fd-nested-list--text-only level-2" id="EX500L2" aria-hidden="true">
                     <li class="fd-nested-list__item">
@@ -417,12 +417,12 @@ The following requirements must be met:
                 </a>
             </li>
             <li class="fd-nested-list__item">
-                <div class="fd-nested-list__content is-selected has-child" tabindex="0">
-                    <a class="fd-nested-list__link" href="#" tabindex="-1">
-                        <span aria-label="Level 1 Item" class="fd-nested-list__icon sap-icon--employee"></span>
+                <div class="fd-nested-list__content is-selected has-child">
+                    <span class="fd-nested-list__link">
+                        <span aria-hidden="true" class="fd-nested-list__icon sap-icon--employee"></span>
                         <span class="fd-nested-list__title">Level 1 Item</span>
-                    </a>
-                    <a class="fd-nested-list__expand-icon" href="#" tabindex="-1" aria-haspopup="true"></a>
+                    </span>
+                    <button class="fd-nested-list__expand-icon" aria-controls="EX100L2" aria-haspopup="true" aria-expanded="false"></button>
                 </div>
             </li>
             <li class="fd-nested-list__item">
@@ -481,12 +481,12 @@ The following requirements must be met:
                 </a>
             </li>
             <li class="fd-nested-list__item">
-                <div class="fd-nested-list__content is-selected has-child" tabindex="0">
-                    <a class="fd-nested-list__link" href="#" tabindex="-1">
-                        <span aria-label="Level 1 Item" class="fd-nested-list__icon sap-icon--employee"></span>
+                <div class="fd-nested-list__content is-selected has-child">
+                    <span class="fd-nested-list__link">
+                        <span aria-hidden="true" class="fd-nested-list__icon sap-icon--employee"></span>
                         <span class="fd-nested-list__title">Level 1 Item</span>
-                    </a>
-                    <a class="fd-nested-list__expand-icon" href="#" tabindex="-1" aria-haspopup="true"></a>
+                    </span>
+                    <button class="fd-nested-list__expand-icon" aria-controls="EX100L2" aria-haspopup="true" aria-expanded="false"></button>
                 </div>
             </li>
             <li class="fd-nested-list__item">
--- a/docs/pages/components/side-navigation.md
+++ b/docs/pages/components/side-navigation.md
@@ -226,7 +226,7 @@ On expand, the `is-expanded` class should be propagated also to `__content` elem
                         <span aria-hidden="true" class="fd-nested-list__icon sap-icon--employee"></span>
                         <span class="fd-nested-list__title" href="#">Level 1 Item</span>
                     </a>
-                    <button class="fd-button fd-nested-list__button" aria-controls="EX100L2" aria-haspopup="true" aria-expanded="false" aria-label="Expand submenu"></button>
+                    <button class="fd-button fd-nested-list__button" aria-controls="EX400L2" aria-haspopup="true" aria-expanded="false" aria-label="Expand submenu"></button>
                 </div>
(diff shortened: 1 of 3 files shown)


docs/pages/components/side-navigation.md
```diff
@@ -140,10 +143,10 @@ On expand, the `is-expanded` class should be propagated also to `__content` elem
             </li>            
             <li class="fd-nested-list__item">
                 <div class="fd-nested-list__content has-child" tabindex="0">
-                    <a class="fd-nested-list__link" href="#" tabindex="-1">
+                    <a class="fd-nested-list__link" role="button" tabindex="-1">
```
bcullman, line 146: *could* this be a button?

i could have sworn that IE handles `<a>` without `href` values badly

meganaconley: I figured out a better setup for this to use an actual button.

docs/pages/components/side-navigation.md
```diff
@@ -198,31 +201,32 @@ On expand, the `is-expanded` class should be propagated also to `__content` elem
...
-                    <a class="fd-nested-list__link" href="#" tabindex="-1">
-                        <span class="fd-nested-list__icon sap-icon--employee"></span>
+                <div class="fd-nested-list__content is-selected has-child">
+                    <span class="fd-nested-list__link">
+                        <span aria-hidden="true" class="fd-nested-list__icon sap-icon--employee"></span>
                         <span class="fd-nested-list__title">Level 1 Item</span>
-                    </a>
-                    <a class="fd-nested-list__expand-icon" href="#" aria-controls="EX400L2" tabindex="-1" aria-haspopup="true"></a>
+                    </span>
+                    <button class="fd-nested-list__expand-icon" aria-controls="EX100L2" aria-haspopup="true" aria-expanded="false"></button>
```
jbadan, line 229: In the past, we've used the pattern `<button class='fd-button fd-nested-list__button'`` so that the implementation libraries can reuse their existing buttons without creating one-off functionality

docs/pages/components/side-navigation.md
```diff
@@ -198,31 +201,32 @@ On expand, the `is-expanded` class should be propagated also to `__content` elem
...
+    <a class="fd-side-nav__skip-link" href="#content">Skip navigation</a>
+    <div class="fd-side-nav__group-header" id="EX400H1">
+        Group Header 1
+    </div>
+    <nav class="fd-side-nav__main-navigation">
+        <ul class="fd-nested-list" aria-labelledby="EX400H1">
             <li class="fd-nested-list__item">
                 <a class="fd-nested-list__link" href="#">
-                    <span class="fd-nested-list__icon sap-icon--home"></span>
+                    <span aria-hidden="true" class="fd-nested-list__icon sap-icon--home"></span>
```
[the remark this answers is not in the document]
meganaconley: It is a purely decorative icon so for accessibility purposes it should be hidden.

docs/pages/components/side-navigation.md
```diff
@@ -292,31 +297,32 @@ In compact mode the dimensions of the controls are reduced, allowing more inform
...
-                        <span class="fd-nested-list__icon sap-icon--employee"></span>
+                <div class="fd-nested-list__content has-child">
+                    <span class="fd-nested-list__link">
+                        <span aria-hidden="true" class="fd-nested-list__icon sap-icon--employee"></span>
                         <span class="fd-nested-list__title">Level 1 Item</span>
-                    </a>
-                    <a class="fd-nested-list__expand-icon" href="#" tabindex="-1" aria-controls="EX500L2" aria-haspopup="true"></a>
+                    </span>
+                    <button class="fd-button fd-nested-list__button" aria-controls="EX100L2" aria-haspopup="true" aria-expanded="false"></button>
                 </div>
```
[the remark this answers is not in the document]
meganaconley: It has to be a button as it doesn't navigate anywhere, it triggers the submenu. For accessibility purposes, the only things that should be anchors are links that perform navigation.
Do you mean that the element that triggers the submenu could also be a link, and the submenu can be triggered separately? If so I'll revert the first span to a link.

docs/pages/components/side-navigation.md
```diff
@@ -358,13 +365,13 @@ The lists in both sections (Main and Utility) should have the `fd-nested-list--c
...
-        <ul class="fd-nested-list fd-nested-list--compact">
+    </nav>
+    <nav class="fd-side-nav__utility">
+        <ul class="fd-nested-list fd-nested-list--compact" aria-label="Utility Menu">
             <li class="fd-nested-list__item">
                 <a class="fd-nested-list__link" href="#">
-                    <span class="fd-nested-list__icon sap-icon--compare"></span>
-                    <span class="fd-nested-list__title">Level 1 Item</span>
+                    <span aria-hidden="true" class="fd-nested-list__icon sap-icon--compare"></span>
+                    <span aria-hidden="true" class="fd-nested-list__title">Level 1 Item</span>
```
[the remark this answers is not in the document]
meganaconley: This was a mistake - thanks for catching it

docs/pages/components/side-navigation.md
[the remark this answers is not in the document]
JKMarkowski: Hi @meganaconley, element that triggers submenu can also be a link, there is separation between those 2 elements. The rest is ok.
sidenav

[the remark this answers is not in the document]
jbadan: Regarding removing `fd-button` class, I was under the impression we established a pattern that we are _not_ using novel buttons and are applying style overrides to the base `fd-button` where necessary. If that's changed, it's fine, but we really need to be consistent.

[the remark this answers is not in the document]
jbadan: From the attached gif, it looks like 'Item 2' is acting as button, not a link.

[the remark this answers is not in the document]
meganaconley: I've updated the nested list header to be a link as well. It does seem odd as the hover style applies to the entire line but now there are two separate clickable elements where as the gif demonstrates previously the entire row acted as a button.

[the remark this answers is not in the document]
rengare: > From the attached gif, it looks like 'Item 2' is acting as button, not a link.

That's the problem of anchor because it can also look like a button

docs/pages/components/side-navigation.md
```diff
@@ -325,8 +331,7 @@ The lists in both sections (Main and Utility) should have the `fd-nested-list--c
                         </a>
                     </li>
                     <li class="fd-nested-list__item">
-                        <a class="fd-nested-list__link" 
-                            href="#" aria-controls="EX500L3" aria-haspopup="true">
+                        <a class="fd-nested-list__link">
```
jacobdevera, line 334: Is this meant to have the `href` removed?

docs/pages/components/side-navigation.md
```diff
@@ -457,54 +465,55 @@ The following requirements must be met:
...
+                    <span aria-label="Level 1 Item" class="fd-nested-list__icon sap-icon--home"></span>
                     <span class="fd-nested-list__title">Level 1 Item</span>
                 </a>
             </li>
             <li class="fd-nested-list__item">
                 <a class="fd-nested-list__link" href="#">
-                    <span class="fd-nested-list__icon sap-icon--calendar"></span>
-                    <span class="fd-nested-list__title">Level 1 Item</span>
+                    <span aria-label="Level 1 Item" class="fd-nested-list__icon sap-icon--calendar"></span>
+                    <a class="fd-nested-list__title" href="#">Level 1 Item</a>
```
jacobdevera, line 480: Is this supposed to be an anchor in an anchor, as it's not a child item?

docs/pages/components/side-navigation.md
```diff
@@ -457,54 +465,55 @@ The following requirements must be met:
...
                 </a>
             </li>
             <li class="fd-nested-list__item">
                 <div class="fd-nested-list__content is-selected has-child" tabindex="0">
-                    <a class="fd-nested-list__link" href="#" tabindex="-1">
-                        <span class="fd-nested-list__icon sap-icon--employee"></span>
+                    <span class="fd-nested-list__link">
+                        <span aria-hidden="true" class="fd-nested-list__icon sap-icon--employee"></span>
                         <span class="fd-nested-list__title">Level 1 Item</span>
```
jacobdevera, line 487: Does this inner span need to change into an anchor, as it's a child item?

docs/pages/components/side-navigation.md
```diff
@@ -358,13 +365,13 @@ The lists in both sections (Main and Utility) should have the `fd-nested-list--c
...
-        <ul class="fd-nested-list fd-nested-list--compact">
+    </nav>
+    <nav class="fd-side-nav__utility">
+        <ul class="fd-nested-list fd-nested-list--compact" aria-label="Utility Menu">
             <li class="fd-nested-list__item">
                 <a class="fd-nested-list__link" href="#">
-                    <span class="fd-nested-list__icon sap-icon--compare"></span>
-                    <span class="fd-nested-list__title">Level 1 Item</span>
+                    <span aria-hidden="true" class="fd-nested-list__icon sap-icon--compare"></span>
+                    <a class="fd-nested-list__title" href="#">Level 1 Item</a>
```
jacobdevera, line 374: Should this remain as a span, as it's not a child item, similar to the other list items? And it may be causing a rendering issue:

Capture

docs/pages/components/side-navigation.md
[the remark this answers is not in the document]
meganaconley: Actually looking at this again, does this mean that the condensed state can't have a submenu? Because there is no room for a button, but the header is now a link:

Screen Shot 2020-05-22 at 9 36 52 AM

docs/pages/components/side-navigation.md
meganaconley: Fixed the markup

docs/pages/components/side-navigation.md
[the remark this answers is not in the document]
jacobdevera: Condensed state appears to have submenus as popovers in this UI5 example: https://sapui5.hana.ondemand.com/#/entity/sap.tnt.SideNavigation/sample/sap.tnt.sample.SideNavigation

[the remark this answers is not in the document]
meganaconley: Ok, it does look like in the collapsed state they only behave as buttons, I'll update the markup for that.

stefanoScalzo:
Screen Shot 2020-05-28 at 9 28 56 AM
Not sure if this is the desired reaction when tabbing but it is not in UI5 specs

meganaconley:
@stefanoScalzo Here is some documentation on skip navigation - https://webaim.org/techniques/skipnav/

Specifically here we are using the second type which is invisible unless focused.

bcullman:
common behavior. see:

image

image

image

image

docs/pages/components/side-navigation.md
salarenko: Could you explain why this attribute has been removed? (I unresolved this conversation)

src/side-nav.scss
```diff
@@ -29,6 +30,38 @@ $has-child-content-arrow-margin: 0.125rem;
...
+    &:focus {
+      @include fd-link();
+
+      position: static;
+      width: auto;
+      height: auto;
+    }
+  }
+
+  &__group-header {
```
salarenko, line 46: Where does this styles come from? Is that a copy-paste from list header?

docs/pages/components/side-navigation.md
meganaconley: It has been added back in and hrefs from spans removed

docs/pages/components/side-navigation.md
```diff
@@ -394,54 +401,54 @@ The following requirements must be met:
...
             </li>
             <li class="fd-nested-list__item">
                 <a class="fd-nested-list__link" href="#">
-                    <span class="fd-nested-list__icon sap-icon--calendar"></span>
+                    <span aria-label="Level 1 Item" class="fd-nested-list__icon sap-icon--calendar"></span>
                     <span class="fd-nested-list__title">Level 1 Item</span>
                 </a>
             </li>
             <li class="fd-nested-list__item">
                 <div class="fd-nested-list__content is-selected has-child" tabindex="0">
```
jacobdevera, line 420: I believe the tabindex here can be removed as the button now receives focus (there is currently a duplicate tab stop).

docs/pages/components/side-navigation.md
```diff
@@ -457,54 +464,54 @@ The following requirements must be met:
...
```
jacobdevera, line 483: Same as above, can likely remove tabindex

stefanoScalzo:
> https://webaim.org/techniques/skipnav/

Still think this should be styled in case a regular user somehow manages to click the tab. We need to ask the designers for it

meganaconley:
@stefanoScalzo Are you suggesting we should block this update on a style for the link or could it be an issue to be addressed later? Could you send me the name(s) of who I should contact about getting a design?

meganaconley:
@salarenko Could you please re-review?

Issue #712, opened 2020-02-28
Rework validation messaging for components with List


jbadan:
Validation messaging in components that use `List` in a `Popover` should be reworked (ComboboxInput, MultiInput, SearchInput, Select, etc)

Currently, the validation message is being added as a list item: 

` <li class="fd-list__message fd-list__message--success">Success message</li>`

This unfortunately doesn't read correctly on a screen reader - it makes it seem as though the validation is just part of the list. The validation message needs to be separate from the list in general.

<img width="1380" alt="Screen Shot 2020-02-28 at 2 04 31 PM" src="https://user-images.githubusercontent.com/29607818/75590997-6f93a380-5a33-11ea-9ae4-8d8446661fa6.png">

An option moving forward:

```
<div class="fd-popover__body fd-popover__body--no-arrow fd-popover__body--dropdown fd-popover__body--dropdown-fill" aria-hidden="true" id="foo">
        <div aria-live="assertive" class="fd-form-message--success" role="alert">Success message</div>
        <ul class="fd-list fd-list--has-message fd-list--dropdown" role="listbox">
            <li role="option" tabindex="0" class="fd-list__item is-selected">
                <span class="fd-list__title">
                    <span class="fd-list__bold">App</span>le
                </span>
            </li>
          ....
      </ul>
</div>
```

droshev:
it was done with #1013

PR #1013, opened 2020-05-12
feat: Select component reference markup

jacobdevera:
## Description
Changes to Select component markup:
- The inner `<button>` causes an extra unnecessary tab stop. Arguably, the entire select component should be a singular clickable/tabbable button.
- Each individual `<li>` option is tabbable. They should instead be navigable via arrow keys (which would be coded in the implementation libraries) and use a roving tabindex.
- Change `li` to `div` for validation messages, as they should not be part of the list, as described in https://github.com/SAP/fundamental-styles/issues/712

No visual changes.

## References
https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html
Opting to use a roving tabindex vs. aria-activedescendant: https://zellwk.com/blog/element-focus-vs-aria-activedescendant/

### DIFF
--- a/docs/pages/components/select.md
+++ b/docs/pages/components/select.md
@@ -31,16 +31,16 @@ For lists that require more than 12 options, the <a href="/patterns/combobox-inp
   </div>
   <div class="fd-popover__body fd-popover__body--no-arrow fd-popover__body--dropdown" aria-hidden="true" id="h0C6A325">
      <ul class="fd-list fd-list--dropdown" role="listbox">
-        <li class="fd-list__item is-selected" aria-selected="true" role="option">
+        <li class="fd-list__item is-selected" aria-selected="true" role="option" tabindex="0">
            <span class="fd-list__title">List item 1</span>
         </li>
-        <li class="fd-list__item" role="option">
+        <li class="fd-list__item" role="option" tabindex="-1">
            <span class="fd-list__title">List item 2</span>
        </li>
-        <li class="fd-list__item" role="option">
+        <li class="fd-list__item" role="option" tabindex="-1">
            <span class="fd-list__title">List item 3</span>
         </li>
-        <li class="fd-list__item" role="option">
+        <li class="fd-list__item" role="option" tabindex="-1">
            <span class="fd-list__title">List item 4</span>
         </li>
      </ul>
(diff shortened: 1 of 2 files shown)


docs/pages/components/select.md
```diff
@@ -23,24 +23,24 @@ For lists that require more than 12 options, the <a href="/patterns/combobox-inp
 <div class="fd-popover">
   <div class="fd-popover__control">
      <div class="fd-select">
-         <div class="fd-select__control" tabindex="0" aria-controls="h0C6A325" aria-expanded="false" aria-haspopup="true">
+         <div class="fd-select__control" role="button" tabindex="0" aria-controls="h0C6A325" aria-expanded="false" aria-haspopup="true">
              Select
```
prsdthkr, line 27: Should this `div` be a button instead? `Enter` or `Space` don't seem to open the selection list box. Or would that be implemented by consumers?

jacobdevera: I initially tried changing it to a `button`, but there were a lot of style changes that would need to be overridden and I couldn't easily get the text to be centered. I believe leaving it to the implementation libraries should be fine, as the docs are mainly for showing styles and not functionality. I'm not sure which file these keyboard events would be added if they had to be handled.

jbadan: So this would be a button in the implementation libraries but a div here? That's not generally the pattern we've followed in the past

docs/pages/components/select.md
```diff
@@ -23,24 +23,24 @@ For lists that require more than 12 options, the <a href="/patterns/combobox-inp
...
-             <button class="fd-button fd-button--transparent sap-icon--slim-arrow-down fd-select__button"></button>
+             <span class="fd-button fd-button--transparent sap-icon--slim-arrow-down fd-select__button"></span>
```
jbadan, line 28: We shouldn't do this either - these docs are supposed to be the source of truth for html markup.

docs/pages/components/select.md
jacobdevera: No, this is exactly the markup the implementation libraries would follow. I meant that they would need to implement Enter and Space handling in the JavaScript.

docs/pages/components/select.md
jacobdevera: Also markup they would need to follow, because the `button` adds an extra tab stop.

docs/pages/components/select.md
jbadan: Thanks for clarifying!

Issue #518, opened 2019-12-16
Migrate Panel to Fiori 3


estelaxu3:
### Component name
> Panel
The panel is a container for grouping and displaying information. It can be collapsed to save space on the screen.
It consists of Header (A) and Content area (B). A panel can expand and collapse. The title of the Panel is controlled through the Header text property. By default the background of the content is transparent, but it could be set to solid.

### Detailed description
Fiori Design Guidelines - https://experience.sap.com/fiori-design-web/panel/
Estimates: 3 days

> SNK, Marketing,  Kyma, Cloud Portal

PR #1024, opened 2020-05-13
fix: introduce panel component

JKMarkowski:
## Related Issue
fixes: https://github.com/SAP/fundamental-styles/issues/518
fixes: #1057 

## Description
Introducing Fiori 3 Panel component

### Before:
We didn't have the component before. There are no breaking changes in this PR as we refactored the layout panel in a separate PR.

### After:
image
image
image
image


src/new-panel.scss
```diff
@@ -0,0 +1,86 @@
+@import "./new-settings";
+@import "./mixins";
+
+$block: #{$fd-namespace}-new-panel;
+
+.#{$block} {
+
+  $fd-panel-header-height: 2.75rem;
+  $fd-panel-compact-header-height: 2.75rem;
```
InnaAtanasova, line 9: should be 2.5rem

src/new-panel.scss
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  &__header {
+    height: $fd-panel-header-height;
+    min-height: $fd-panel-header-height;
+    padding-right: $fd-panel-header-padding;
+    background-color: var(--sapGroup_TitleBackground);
+    border-bottom: solid 0.0625rem var(--sapGroup_TitleBorderColor);
+    display: flex;
+    align-items: center;
```
InnaAtanasova, line 27: You can use `fd-flex-vertical-center` mixin

src/new-panel.scss
```diff
@@ -0,0 +1,86 @@
...
+  @include fd-rtl() {
+    .#{$block}__header {
+      padding-right: 0;
+      padding-left: $fd-panel-header-padding;
+    }
+  }
+
+  &--fixed {
+    .#{$block}__header {
+      padding-left: 1rem;
```
InnaAtanasova, line 70: You can put this in variable since it's used in rtl

src/mixins/_mixins.scss
```diff
@@ -327,3 +327,9 @@
   -ms-overflow-style: none; // hide scrollbar but keep element scrollable for IE/Edge
   scrollbar-width: none; // hide scrollbar on firefox
 }
+
+@mixin fd-typography($fd-size) {
+  font-family: var(--sapFontHeaderFamily);
+  color: var(--sapGroup_TitleTextColor);
```
InnaAtanasova, line 333: if you want this mixin to be more reusable you can pass font-color as a parameter and set the default value to `var(--sapGroup_TitleTextColor)`. If you don't have to change it you can omit it

src/new-panel.scss
```diff
@@ -0,0 +1,86 @@
...
+
+    @include fd-flex-center();
+
+    .#{$fd-namespace}-button {
+      @include fd-icon('slim-arrow-right');
+
+      &.is-expanded {
+        @include fd-icon('slim-arrow-down');
+      }
+    }
```
InnaAtanasova, line 58: this is ok but the dependeny  (.#{$fd-namespace}-button) can be avoided by adding `fd-panel__expand-button` or just `fd-panel__button`. In this case you don't have to pass `sap-icon--slim-arrow-down` in `<button class="fd-button fd-button--transparent sap-icon--slim-arrow-down">`

droshev:
> Great job! I think the content is missing animation. @estelaxu3 and @droshev I think we need to contact the designers regarding this (or maybe someone from UX team)

I noticed that too. @estelaxu3 do u think you can help us here?

src/new-panel.scss
JKMarkowski: I can agree with you, there should be added some class to `fd-button`, I think there shouldn't be class collision.

JKMarkowski:
Hi @InnaAtanasova Your comments has been addressed :)

docs/pages/components/panel.md
```diff
@@ -0,0 +1,161 @@
...
+</ul>
+
+
+## Expandable Panel
+Expandable panels are much like fixed panels, except their content can be expanded and collapsed (including the info toolbar, if available).
+{% capture expand-panel %}
+<div class="fd-expand-panel">
+    <div class="fd-expand-panel__header">
+        <div class="fd-expand-panel__expand">
+            <button class="fd-button fd-button--transparent fd-expand-panel__button" aria-expanded="false" aria-haspopup="true" aria-controls="jhqDKYdf"></button>
```
jacobdevera, line 32: All of these "expand/collapse" buttons need to have a label. It seems best to add an `id` to the Panel Header, and an `aria-labelledby` of the `button` set to the `id` of the header.

docs/pages/components/panel.md
```diff
@@ -0,0 +1,161 @@
...
+        <div class="fd-toolbar fd-toolbar--clear fd-toolbar--transparent">
+            <span class="fd-toolbar__spacer fd-toolbar__spacer--auto"> </span>
+            <div class="fd-segmented-button" role="group" aria-label="Group label">
+              <button class="fd-button is-selected" aria-pressed="true">Left</button>
+              <button class="fd-button">Middle</button>
+              <button class="fd-button">Right</button>
+            </div>
+        </div>
+    </div>
+    <div role="region" aria-labelledby="jhqDKYdf" class="fd-expand-panel__content" aria-hidden="true" id="jhqDKYdf">
```
jacobdevera, line 44: The `aria-labelledby` should reference an element that labels the element, not itself. In this case, I'm assuming it's one of the Left / Middle / Right buttons, and this switches depending which one is active (hard to tell from the example).

So an `id` should be added to the segmented buttons, and the `aria-labelledby` on the region should be whichever button is active.

docs/pages/components/panel.md
```diff
@@ -0,0 +1,161 @@
...
+        <div class="fd-toolbar fd-toolbar--clear fd-toolbar--transparent">
+            <span class="fd-toolbar__spacer fd-toolbar__spacer--auto"> </span>
+            <div class="fd-segmented-button" role="group" aria-label="Group label">
+              <button class="fd-button is-selected" aria-pressed="true">Left</button>
+              <button class="fd-button">Middle</button>
+              <button class="fd-button">Right</button>
+            </div>
+        </div>
+    </div>
+    <div role="region" class="fd-expand-panel__content">
```
jacobdevera, line 77: Even though it's always expanded, this region still needs an accessible label as described above. `aria-labelledby` referencing one of the segmented buttons.

mikerodonnell89:
I think we need to override the line-height for the button icon here, the icon is not centered to the text
Screen Shot 2020-05-29 at 1 31 14 PM

mikerodonnell89:
So what is the plan for deprecating the layout panel?  This pr should add a note that layout panel is going away.  Also maybe in the meantime, there should also be a note on component panel about how there are 2 things in the library called "panel" at the moment, to avoid confusion

docs/pages/components/panel.md
```diff
@@ -0,0 +1,161 @@
...
+{% endcapture %}
+{% include display-component.html component=expand-panel %}
+
+
+## Fixed Panel
+Fixed panels are useful for grouping custom content. They include headers and info toolbars.
+To create a fixed panel, add the `--fixed` modifier.
+
+{% capture fixed-panel %}
+<div class="fd-expand-panel fd-expand-panel--fixed">
```
mikerodonnell89, line 65: Not sure I am on board with this naming convention, because this panel doesn't expand at all

droshev:
> So what is the plan for deprecating the layout panel? This pr should add a note that layout panel is going away. Also maybe in the meantime, there should also be a note on component panel about how there are 2 things in the library called "panel" at the moment, to avoid confusion

We will keep both right now as the current panel is a layout and it is heavy used. We will just rename it so this one will keep the proper name.

docs/pages/components/panel.md
```diff
@@ -0,0 +1,161 @@
...
+        <div class="fd-toolbar fd-toolbar--clear fd-toolbar--transparent">
+            <span class="fd-toolbar__spacer fd-toolbar__spacer--auto"> </span>
+            <div class="fd-segmented-button" role="group" aria-label="Group label">
+              <button class="fd-button fd-button--compact is-selected" aria-pressed="true">Left</button>
+              <button class="fd-button fd-button--compact">Middle</button>
+              <button class="fd-button fd-button--compact">Right</button>
+            </div>
+        </div>
+    </div>
+    <div role="region" aria-labelledby="Gekf63D" class="fd-expand-panel__content" aria-hidden="true" id="Gekf63D">
```
jacobdevera, line 109: `aria-labelledby` needs to be set to the active segmented button

docs/pages/components/panel.md
```diff
@@ -0,0 +1,161 @@
...
+        <div class="fd-toolbar fd-toolbar--clear fd-toolbar--transparent">
+            <span class="fd-toolbar__spacer fd-toolbar__spacer--auto"> </span>
+            <div class="fd-segmented-button" role="group" aria-label="Group label">
+              <button class="fd-button is-selected" aria-pressed="true">Left</button>
+              <button class="fd-button">Middle</button>
+              <button class="fd-button">Right</button>
+            </div>
+        </div>
+    </div>
+    <div role="region" aria-labelledby="GJL745SD" class="fd-expand-panel__content" aria-hidden="true" id="GJL745SD" style="height: 100px;">
```
jacobdevera, line 141: `aria-labelledby` needs to be set the active segmented button

docs/pages/components/panel.md
```diff
@@ -0,0 +1,161 @@
...
+{% include display-component.html component=fixed-panel %}
+
+## Compact Panel
+To use a compact panel, add the `--compact` modifier.
+
+{% capture compact-panel %}
+<div class="fd-expand-panel fd-expand-panel--compact">
+    <div class="fd-expand-panel__header">
+        <div class="fd-expand-panel__expand">
+            <button class="fd-button fd-button--compact fd-button--transparent fd-expand-panel__button" aria-expanded="false" aria-haspopup="true" aria-controls="Gekf63D"></button>
```
jacobdevera, line 97: `aria-labelledby` should be set to the `h5`

docs/pages/components/panel.md
```diff
@@ -0,0 +1,161 @@
...
+{% include display-component.html component=compact-panel %}
+
+## Fixed Height Content Panel
+When the height of the panel's content is set to a fixed size, the content area can be scrolled through.
+
+{% capture fixed-height-panel %}
+<div class="fd-expand-panel">
+    <div class="fd-expand-panel__header">
+        <div class="fd-expand-panel__expand">
+            <button class="fd-button fd-button--transparent fd-expand-panel__button" aria-expanded="false" aria-haspopup="true" aria-controls="GJL745SD"></button>
```
jacobdevera, line 129: `aria-labelledby` should be set to the `h5`

Issue #914, opened 2020-04-28
Fiori 3 Table


mikerodonnell89:
Table component: 
- [x] Base Table Component - Header, Body, Footer + sizes + states(hover, active, selected) - Done in https://github.com/SAP/fundamental-styles/pull/1042
- [x] Table with checkboxes + sizes - Done in https://github.com/SAP/fundamental-styles/pull/1042
- [x] Highlight Row (fiori3 site) - Elements on left + text - Done in https://github.com/SAP/fundamental-styles/pull/1042
- [x] Navigation Indication (fiori3 site) - Icon on right - Mike Done in https://github.com/SAP/fundamental-styles/pull/1042
- [ ] Analytical Table (fiori3 site) - Group Table
- [x] Responsive Table (fiori3 site) - Pop-in positioning + Merged Cells Done in https://github.com/SAP/fundamental-styles/pull/1042

JKMarkowski:
Questions:
* What about the active state, how should behave active state:
first:
![table1](https://user-images.githubusercontent.com/26483208/82558082-39b31680-9b6d-11ea-9a5f-1f9a968fb1c5.gif)
second:
![table2](https://user-images.githubusercontent.com/26483208/82558094-3fa8f780-9b6d-11ea-82f9-4e87d974982d.gif)

PR #1053, opened 2020-05-22
feat: table fiori3 navigation indication

mikerodonnell89:
## Related Issue

Part of #914 

## Description

Add table navigation indicator class and examples. Note this is a PR to the larger table PR and not to master.

### After:

Screen Shot 2020-05-22 at 12 40 40 PM

### DIFF
--- a/src/table.scss
+++ b/src/table.scss
@@ -167,6 +167,10 @@ $block: #{$fd-namespace}-table;
     @include fd-last-child() {
       border-right-color: transparent;
     }
+
+    &--navigation {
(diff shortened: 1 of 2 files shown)


src/table.scss
```diff
@@ -167,6 +167,10 @@ $block: #{$fd-namespace}-table;
     @include fd-last-child() {
       border-right-color: transparent;
     }
+
+    &--navigation {
```
InnaAtanasova, line 171: I'd call it indicator-bar or navigated

rengare: vote for navigated

Issue #1052, opened 2020-05-22
Refactor Tile - Line Tile


InnaAtanasova:
The line tiles are a text only version of a launch tile. It acts as a link agglomeration in one container.

PR #1059, opened 2020-05-27
fix: add Line Tile type to Generic Tiles

InnaAtanasova:
## Related Issue
Closes SAP/fundamental-styles#1052

## Description
Adding Line Tiles type to Generic Tiles

- [ ] all items on the PR Review Checklist are addressed :


stefanoScalzo:
Screen Shot 2020-05-27 at 5 40 51 PM
cozy mode should have 2.25rem

InnaAtanasova:
> Screen Shot 2020-05-27 at 5 40 51 PM
> cozy mode should have 2.25rem

No, you shouldn't

docs/pages/components/tile.md
```diff
@@ -917,4 +917,474 @@ In action mode view, the close button is displayed on the top right-hand corner
...
+
+<br>
+<strong>Action Mode</strong> <br><br>
+The controls are wrapped in a container <code>fd-tile__action-container</code> which is right aligned within the tile. There's a 0.25rem spacing between the header and the actions container. On edit mode List View the overflow of the list is done with gradient mask.
+
+{% capture tile-line %}
+<h4>Line Tile - Floating Behaviour, Cosy Mode</h4>
+<div class="docs-section-container">
+    <div class="fd-tile-container"> 
+        <div tabindex="0" class="fd-tile fd-tile--line">
```
jacobdevera, line 975: Every `fd-tile` in this file should have a role of `button` as they are clickable items
UI5 also does this: https://sapui5.hana.ondemand.com/#/entity/sap.m.GenericTile/sample/sap.m.sample.GenericTileAsLaunchTile

Everything else looks good to me.

InnaAtanasova:
> * Do we need this wrapper for badge? It looks like `.fd-tile__header` handles centering items.
>   image
> 

Yes, it's painful :D Believe me, I did my best to avoid it. If Badge and Title are not wrapped inside a container then the children of a flex container will be 3 - badge, title and subtitle. When the title and subtitle can't fit in one line, the subtitle needs to go to the second line and it they are still too long then they truncate with ... For this reason we need to treat badge+title as 1 child and subtitle as a separate child, making it 2 children of a flex container.

docs/pages/components/tile.md
InnaAtanasova: Thank you @jacobdevera  I added role button to every tile

Issue #1060, opened 2020-05-27
Defect Hunting - May 27-28


droshev:
- [x] doc site - Gracjan - #1063
![Screen Shot 2020-05-27 at 5 18 23 PM](https://user-images.githubusercontent.com/4380815/83073294-3375dc00-a03e-11ea-94a9-6f89caa5b5a6.png)
- [ ] styling looks odd https://fundamental-styles.netlify.app/layouts/shell-layout.html - It's completely outdated, should be marked as deprecated
- [ ] double message border - Jedrzej - TBD
![Screen Shot 2020-05-27 at 5 26 14 PM](https://user-images.githubusercontent.com/4380815/83073951-50f77580-a03f-11ea-8a53-6dbc08c7aab6.png)
- [x] link `Show All` not properly used - Mike
![Screen Shot 2020-05-27 at 5 29 48 PM](https://user-images.githubusercontent.com/4380815/83074168-b9465700-a03f-11ea-824c-cd0542e38b78.png)

JKMarkowski:
Header in IE11
![image](https://user-images.githubusercontent.com/26483208/83141875-96c04680-a0f0-11ea-8af6-bac46dd4e96e.png)

PR #1063, opened 2020-05-28
docs: layout styles fixed

rengare:
## Related Issue
Mentioned #1060 

## Description
PR brings style fix for layout

### Before:
Screen Shot 2020-05-27 at 5 18 23 PM

### After:
image


stefanoScalzo:
Screen Shot 2020-05-28 at 9 11 15 AM

This is IE, once my grids are merged well use those, or you can take the same approach as the components page.

PR #1066, opened 2020-05-28
feat: select component a11y changes

jbadan:
## Related Issue

## Description
The design spec for readOnly: `Placeholder text and Button are both NOT displayed in read only mode.`

Added min-width: `Min-Width (Select field version with text):
5 rem to be able to see any text and the icon`

Breaking changes:

1. added `aria-labeledby` to `fd-list ul` connect to `id` on `fd-form-label`, see https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html
1.  moved validation messages `div fd-list__message`  _outside_ of `fd-list ul` - `span` cannot be a descendent of `ul`
1. removed “disabled” attribute from `fd-popover__control` as this is a `div` and cannot have  a `disabled` attribute
1. removed `disabled` attribute in favor of `aria-disabled=“true”`: When the action associated with a button is unavailable, the button has aria-disabled set to true. https://www.w3.org/TR/wai-aria-practices/#wai-aria-roles-states-and-properties-3
1. added `tabIndex=“-1” to `fd-list ul`: https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html
 1. `aria-readonly` is not allowed on `role=“button”`, replaced with “is-readonly” class or readonly attribute https://www.w3.org/TR/2017/REC-wai-aria-1.1-20171214/#aria-readonly
 > NOTE: there will be a separate followup to make readonly state accessible 

Additional comments:
- will update Breaking changes wiki **AFTER** this pr is reviewed and we've come to a consensus on what we want the html to look like

Note there is a deleted test in here - part of https://github.com/SAP/fundamental-styles/pull/1108


docs/pages/components/select.md
```diff
@@ -414,19 +414,17 @@ The disabled state can also be achieved by adding the `.is-disabled` class or th
...
 
-{% capture disabled-select %}
+{% capture readonly-select %}
 <div class="fd-popover">
     <div class="fd-popover__control" aria-expanded="false" aria-haspopup="false" aria-readonly="true" readonly>
         <div class="fd-select">
-            <div class="fd-select__control" role="button" aria-expanded="false" aria-haspopup="false" aria-readonly="true" readonly>
-                Select
-           </div>
+            <div class="fd-select__control" role="button" aria-expanded="false" aria-label="Select" aria-haspopup="false" aria-readonly="true" readonly></div>
```
jbadan, line 421: Should this have an aria-label like I've added here because the placeholder is removed in readOnly state? @meganaconley @jacobdevera @prsdthkr

package.json
```diff
@@ -123,8 +123,7 @@
   "pre-commit": {
     "colors": true,
     "run": [
-      "lint:pre-commit",
-      "storybook:visual-docker"
```
jbadan, line 127: This shouldn't be running storybook on a pre-commit hook

docs/pages/components/select.md
```diff
@@ -25,7 +25,7 @@ For lists that require more than 12 options, the <a href="/patterns/combobox-inp
      <div class="fd-select">
          <div class="fd-select__control" role="button" tabindex="0" aria-controls="h0C6A325" aria-expanded="false" aria-haspopup="true">
              Select
-             <span class="fd-button fd-button--transparent sap-icon--slim-arrow-down fd-select__button"></span>
+             <button class="fd-button fd-button--transparent sap-icon--slim-arrow-down fd-select__button"></button>
```
jbadan, line 28: Do these need any aria-hidden attributes? aria-label? or does the placeholder give enough information?

InnaAtanasova:
1. Could you update the [Breaking Changes ](https://github.com/SAP/fundamental-styles/wiki/Breaking-Changes)
2. The focus outline is cut in compact examples:

Screen Shot 2020-05-28 at 12 19 27 PM
Screen Shot 2020-05-28 at 12 19 22 PM

docs/pages/components/select.md
```diff
@@ -25,7 +25,7 @@ For lists that require more than 12 options, the <a href="/patterns/combobox-inp
      <div class="fd-select">
          <div class="fd-select__control" role="button" tabindex="0" aria-controls="h0C6A325" aria-expanded="false" aria-haspopup="true">
              Select
-             <span class="fd-button fd-button--transparent sap-icon--slim-arrow-down fd-select__button"></span>
```
jacobdevera, line 28: This was changed to a `span` to fix the duplicate tabbing issue (you can observe this now by tabbing into the Select component in this PR's deployment).  I guess you could make an argument that anything with the `fd-button` class should be an actual `button`, but the parent being `role="button"` is sufficient for screen readers and keyboard users to use the component without majorly refactoring the styles.

docs/pages/components/select.md
jacobdevera: Can you show where it says not to include a placeholder? I'm looking at https://experience.sap.com/fiori-design-web/select/#readonly and it seems fine as-is.

I think the confusion may be that it seems to be a placeholder when it should actually be a pre-selected value.

docs/pages/components/select.md
meganaconley: I think we should put a `tabIndex` of -1 on this. Otherwise we get two tab stops. The entire select should be a single tabstop as in this example: https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html 

That way it can be a button or a span, whichever is better for the visuals.

docs/pages/components/select.md
jbadan: For everyone else - sent wiki link privately as it's internal :)

jbadan: We can set this one to be a selected item for the docs though?

jacobdevera: Yeah, the text here could be changed to make clear that it's a selected item rather than a placeholder.

docs/pages/components/select.md
```diff
@@ -414,19 +414,17 @@ The disabled state can also be achieved by adding the `.is-disabled` class or th
...
-{% capture disabled-select %}
+{% capture readonly-select %}
 <div class="fd-popover">
-    <div class="fd-popover__control" aria-expanded="false" aria-haspopup="false" aria-readonly="true" readonly>
+    <div class="fd-popover__control" aria-expanded="false" aria-haspopup="false" aria-readonly="true">
         <div class="fd-select">
-            <div class="fd-select__control" role="button" aria-expanded="false" aria-haspopup="false" aria-readonly="true" readonly>
-                Select
-           </div>
+            <div class="fd-select__control" role="button" aria-expanded="false" aria-label="Select" aria-haspopup="false" aria-readonly="true"></div>
```
prsdthkr, line 421: I think this should have `tabindex="0"` since readonly inputs are focusable for screen-reader purposes. Also I believe you removed the selected value by mistake.
```
<div class="fd-select__control" role="button" aria-expanded="false" aria-label="Select" aria-haspopup="false" aria-readonly="true" tabindex="0">Potato</div>
```

package.json
salarenko: This should be moved to separate PR

jbadan: It already was - otherwise you wouldn't be able to commit anything 😉 
https://github.com/SAP/fundamental-styles/pull/1067

jbadan:
@jacobdevera @prsdthkr @salarenko @meganaconley I made a lot of new changes - see new edited description of the pr please!

docs/pages/components/select.md
```diff
@@ -16,31 +16,52 @@ For lists that require more than 12 options, the <a href="/patterns/combobox-inp
...
-        <li class="fd-list__item" role="option" tabindex="-1">
+        <li class="fd-list__item fd-select__item"
+            role="option"
+            tabindex="-1">
            <span class="fd-list__title">List item 3</span>
         </li>
-        <li class="fd-list__item" role="option" tabindex="-1">
+        <li class="fd-list__item fd-select__item"
+            role="option"
+            tabindex="-1">
```
jbadan, line 64: nothing changed in these - I was just spreading them out so I could see what I was doing. Happy to put them back if anyone is bothered by it

docs/pages/components/select.md
```diff
@@ -411,56 +435,57 @@ The disabled state can also be achieved by adding the `.is-disabled` class or th
...
+{% capture readonly-select %}
 <div class="fd-popover">
-    <div class="fd-popover__control" aria-expanded="false" aria-haspopup="false" aria-readonly="true" readonly>
+    <div class="fd-popover__control">
         <div class="fd-select">
-            <div class="fd-select__control" role="button" aria-expanded="false" aria-haspopup="false" aria-readonly="true" readonly>
-                Select
-           </div>
+            <div class="fd-select__control is-readonly" role="button" aria-expanded="false" aria-label="Select" aria-haspopup="false">Selected Item 2</div>
         </div>
```
jbadan, line 445: This is where I still don't know what to do.. I've got the styling added via `is-readonly` class, but how do we communicate it's readonly? The ui5 one has a hidden input. 

Additionally, I'm not sure how about not being able to open the popover -  "Readonly elements are relevant to the user, and application authors SHOULD NOT restrict navigation to the element or its focusable descendants." https://www.w3.org/TR/2017/REC-wai-aria-1.1-20171214/#aria-readonly

jacobdevera: I've been trying to look for other examples on the internet, and there doesn't appear to be anything like this surprisingly. I agree the language is a little vague on what to do for components with popups. We have a couple steps:

1. Like UI5, we can add a hidden `input` with `aria-readonly` as a child of the button. This is so screen readers can tell it is read-only when focused without have to click the button.

1. (optional?) Allow the popover to open in read only mode, and set `aria-readonly` on the `ul` `listbox`.

We should definitely do 1. 2 is potentially optional. The disappearance of the chevron in readonly suggests that it shouldn't be opened, and I can't find any mandate where the list of options needs to be presented in read-only mode.

meganaconley: I think it's ok to say that a combobox is `aria-readonly` only when it's value can't be set. Technically you can't type in the box, but you can change its value. The definition in the link Jenna provided for `aria-readonly` says `The user can set the value of the element.` vs `The user cannot change the value of the element.` By that I don't think it's limited to actually typing only, but any sort of change in value whether it be selecting an option from a dropdown or a radio group, setting a checkbox, etc.

jbadan: @meganaconley @jacobdevera Stupid question - but I'm not following 😆  What should the html look like?

jacobdevera: @jbadan Yeah sorry, I was thinking out loud and got kinda convoluted 😅

Like UI5 (https://sapui5.hana.ondemand.com/#/entity/sap.m.Select/sample/sap.m.sample.Select), I would try to change to `combobox` instead of `button` because it supports `aria-readonly` :
`<div class="fd-select__control is-readonly" role="combobox" aria-expanded="false" aria-label="Select" aria-haspopup="false" aria-readonly="true">Selected Item 2</div>`

jbadan: Thank you!

jbadan: It looks like if we go `role="combobox"` we need a bunch of other changes, no? For example, including an input that's always aria-readonly (not just for this example)

That leads me to, should this div just be an input with `aria-readonly='true' role='combobox'`?

src/select.scss
```diff
@@ -87,6 +88,10 @@ $block: #{$fd-namespace}-select;
     font-size: var(--sapFontSize);
   }
 
+  &__item[aria-selected=true] {
+    border-bottom: 0 !important;
+  }
+
```
JKMarkowski, line 94: It's not fiori3 compatible. First pictures on specifications are wrong.
image
Anyway, we don't use important at all. If this situation occurs, you would need to use stronger selector

src/select.scss
```diff
@@ -14,6 +14,7 @@ $block: #{$fd-namespace}-select;
 
   display: block;
   padding-right: 0;
+  min-width: 5rem;
```
JKMarkowski, line 17: What's the reason, that you added it there ?

docs/pages/components/select.md
```diff
@@ -472,34 +497,37 @@ To add text in the body of the component, simply include your text in the `fd-li
...
-               <span class="fd-button sap-icon--slim-arrow-down fd-select__button"></span>
+               <span class="fd-button fd-button--transparent sap-icon--slim-arrow-down fd-select__button"></span>
            </div>
        </div>
    </div>
    <div class="fd-popover__body fd-popover__body--no-arrow fd-popover__body--dropdown" aria-hidden="true" id="h07j9978H">
-        <ul class="fd-list fd-list--has-message fd-list--dropdown fd-list--compact" role="listbox">
-           <div aria-live="assertive" class="fd-list__message fd-list__message--error" role="alert">Error message</div>
-           <li class="fd-list__item is-selected" aria-selected="true" role="option" tabindex="0">
+        <div aria-live="assertive" class="fd-list__message fd-list__message--error" role="alert">Error message</div>
```
JKMarkowski, line 514: Message is not visible anymore, it went behind select
image

`.fd-dialog__body` is missing `position: relative`

src/select.scss
JKMarkowski: So you we don't need to have `fd-select__item` on every single item anymore.

jbadan: Which spec do we want to follow as the source of truth? https://experience.sap.com/fiori-design-web/select/
Screen Shot 2020-06-02 at 8 05 40 AM

docs/pages/components/select.md
jacobdevera: I don't think the hidden input in the ui5 example is actually contributing anything. Once the role is set to `combobox`, it shouldn't matter if it's a `div` or `input` other than if it's easier for styling purposes.

jbadan: So this is what I was running into: https://www.w3.org/TR/wai-aria-1.1/#combobox

"Authors MUST ensure an element with role combobox contains or owns a text input element with role textbox or searchbox and that the text input has aria-multiline set to false. "

jacobdevera: Technically, we are already breaking spec because we are not supplying an editable textbox input for the combobox in the first place; we're merely using it as a way to get `aria-readonly` to be supported. The spec mentions that the `input` must be a child of the combobox, not the combobox itself. In our case, if we supply the fake hidden input element like UI5 does, at best, it doesn't affect anything for screen reader callouts, and at worst, could be misleading for when the Select is not marked readonly.

Again, whether the base combobox is an `input` or `div` doesn't affect accessibility because the semantics are being overriden by the combobox role.

My take: I'm okay with adding a hidden input into a combobox like the one in my code snippet above, but only for the readonly state. All other select states should remain with the usual listbox in `button` as in the w3c example.

src/select.scss
JKMarkowski: I can only assume, that we should follow list component there. Same as with multi input or combobox. Let's ask designers maybe, because spec shows 2 different views.

docs/pages/components/select.md
jbadan: But what I'm trying to say - should we even be using `role="combobox"` if we're not following the spec? 

This is starting to feel like too hacky of a workaround. We can also just table this and put an actual story in the backlog to investigate this further since we've spent so much time on it already.

jacobdevera: Ideally, no. Agreed that it can wait for a future story.

jbadan: UIA11Y-2324

jbadan:
@jacobdevera @salarenko @JKMarkowski updated for pr feedback. Note that the readonly state will be handled in a separate followup pr.

jbadan:
@salarenko updated.

Issue #1014, opened 2020-05-12
'High Contrast Black' theme Product Switch states WCAG AAA contrast compliance


prsdthkr:
**Describe the bug**
The Product Switch active and hover state color contrast ratio is not 7:1 or above.

> What version of the Fundamental Library Styles are you using?
v0.9.1-rc.0

**To Reproduce**
1. Go to https://deploy-preview-928--fundamental-styles.netlify.app/components/product-switch.html
2. Check Product Switch active and hover states' color contrast 

**Expected behavior**
If we want WCAG AAA compliance, this contrast ratio should be 7:1 or above.

**Screenshots**
![image](https://user-images.githubusercontent.com/43764832/81755981-49e13c80-946f-11ea-87d7-cf0295a33f30.png)
![image](https://user-images.githubusercontent.com/43764832/81756110-a93f4c80-946f-11ea-96b9-778183110978.png)

![image](https://user-images.githubusercontent.com/43764832/81753968-da1c8300-9469-11ea-8acc-3bd4f2de9a25.png)

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: Chrome 
 - Version: 81.0.4044.129 (Official Build) (64-bit)

PR #1069, opened 2020-05-28
feat: #1014 Inline help rebuild with popover component

athulvraj:
## Related Issue

## Description
The inline help component is rebuilt using Popover Component. The Inline help content is rendered in the Popover and the rendering position is calculated by it internally and adds CSS inline styles for it. 

BREAKING CHANGE: The HTML markup and CSS classes of inline help is changes with Popover. Previously used CSS styles are removed.


src/inline-help.scss
```diff
@@ -44,186 +33,27 @@ $block: #{$fd-namespace}-inline-help;
...
-    @include fd-var-color("background-color", $fd-tooltip-icon-background-color, --fd-inline-help-icon-background-color);
-
+    color: var(--sapContent_ContrastIconColor);
+    background-color: $fd-tooltip-icon-background-color;
     text-align: center;
   }
 
   // ELEMENTS *******************************************
   &__content {
     @include fd-type("-1");
```
InnaAtanasova, line 43: `fd-type` is no longer supported. These were the values before. You can replace them by `font-size: var(--sapFontSmallSize)` and `line-heigh: normal`
Screen Shot 2020-05-28 at 4 11 40 PM

src/inline-help.scss
```diff
@@ -2,29 +2,18 @@
...
 
 .#{$block} {
   // LOCAL VARS
-  $fd-tooltip-icon-background-color: fd-color("status", 4) !default;
-  $fd-tooltip-content-background-color: fd-color("background", 2) !default;
-  $fd-tooltip-content-color: fd-color("text", 2) !default;
+  $fd-tooltip-icon-background-color: var(--sapButton_Neutral_Background) !default;
+  $fd-tooltip-content-background-color: var(--sapTile_Background) !default;
+  $fd-tooltip-content-color: var(--sapTextColor) !default;
   $fd-tooltip-padding: fd-space(xs) !default;
```
InnaAtanasova, line 12: `fd-space` is not supported. The resulted value was 8px so you can replace it with 0.5rem. 
These were the values before (in case you encounter them again in the future):
Screen Shot 2020-05-28 at 4 16 16 PM

InnaAtanasova:
The inline inline content should appear on hover. Now it requires a click on the icon to show the body. 
The body should have RTL support

InnaAtanasova:
Thanks for contributing! You did a really great job!

src/inline-help.scss
```diff
@@ -2,29 +2,18 @@
 @import "./functions";
```
droshev, line 2: if the component is not relying on the old variables, then it has to import `./new-settings` instead `./settings`

athulvraj: Done

mikerodonnell89:
I'm not sure if having to set explicit positioning i.e. `style="left: -185px"` etc on each popper div is a good idea, especially since that value would have to change depending on the amount of content in the popper body

athulvraj:
> LGTM, We had to put inline styling, which usually is added by popper.js, or other overlay lib. In styles we don't have it, so we just hardcoded some properties. Good Job with putting theming parameters. Maybe it's worth to write some explanation in description @athulvraj, that the inline styling is only for presence there.

Sure, I will do that

PR #1074, opened 2020-05-29
feat: shellbar button outlines should follow behavior of transparent buttons

mikerodonnell89:
## Related Issue

part of #1060 

## Description
According to the shellbar designer, shellbar buttons should behave the same as transparent buttons, other than what is specifically mentioned in the shellbar specs.  This PR makes the button outline the same as transparent buttons

Worth noting that the outline was, by default, smaller than the user menu button circle, so I used outline-offset to slightly expand that.  Which is what they are doing in sapui5 for the same problem.  Maybe we can discuss alternative solutions

### Before:
Screen Shot 2020-05-29 at 2 42 01 PM

### After:
Screen Shot 2020-05-29 at 2 43 06 PM

### DIFF
--- a/docs/pages/components/shellbar.md
+++ b/docs/pages/components/shellbar.md
@@ -56,7 +56,7 @@ This example shows the minimum shellbar for a single application product with on
     <div class="fd-shellbar__action">
       <div class="fd-popover fd-popover--right">
         <div class="fd-popover__control">
-          <button class="fd-button fd-shellbar__button fd-shellbar__button--user-menu" aria-controls="WV3AY276" aria-expanded="false" aria-haspopup="true" role="button">
+          <button class="fd-button fd-shellbar__button fd-shellbar__button--user-menu" aria-controls="WV3AY276" aria-expanded="false" aria-haspopup="true">
             <span class="fd-avatar fd-avatar--xs fd-avatar--circle">WW</span>
           </button>
         </div>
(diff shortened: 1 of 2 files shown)


stefanoScalzo:
the focus outline should be white when on a dark color background

JKMarkowski:
Let's maybe change the outline-color on shellbar elements, to inverted one.

docs/pages/components/shellbar.md
```diff
@@ -56,9 +56,9 @@ This example shows the minimum shellbar for a single application product with on
     <div class="fd-shellbar__action">
       <div class="fd-popover fd-popover--right">
         <div class="fd-popover__control">
-          <div class="fd-button fd-shellbar__button fd-user-menu__control" aria-controls="WV3AY276" aria-expanded="false" aria-haspopup="true" role="button">
+          <button class="fd-button fd-shellbar__button fd-shellbar__button--user-menu" aria-controls="WV3AY276" aria-expanded="false" aria-haspopup="true" role="button">
```
jacobdevera, line 59: Now that these have been changed to `button`, you can go ahead and remove `role="button"` for all of these.

InnaAtanasova:
Hey @mikerodonnell89 please update the Wiki page with the Breaking Changes 
https://github.com/SAP/fundamental-styles/wiki/Breaking-Changes

stefanoScalzo:
Screen Shot 2020-06-08 at 12 49 32 PM
So after clicking the focus goes back to black

PR #1080, opened 2020-05-31
chore: remove dependancies to the old settings file

droshev:
remove dependancies for the migrated components to the old settings file


mikerodonnell89:
@droshev what do you mean "migrated components"?  Can we just do this for all components?

droshev:
> @droshev what do you mean "migrated components"? Can we just do this for all components?

The migrated components should not depend to the old settings file

PR #1083, opened 2020-05-31
docs: removed unnecessary class from docs

rengare:
## Description

Small update to toolbar docs

### DIFF
--- a/docs/pages/components/toolbar.md
+++ b/docs/pages/components/toolbar.md
@@ -10,7 +10,7 @@ summary:
 tags: [f3, a11y, theme, design]
 ---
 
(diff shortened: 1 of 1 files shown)


docs/pages/components/toolbar.md
```diff
@@ -24,7 +24,7 @@ The toolbar has a solid background by default.
   <div class="fd-toolbar fd-toolbar--solid">
     <span>Title</span>
     <button class="fd-button fd-button--compact fd-button--transparent">Action1</button>
-    <span class="fd-toolbar__spacer fd-toolbar__spacer--auto"> </span>
+    <span class="fd-toolbar__spacer "> </span>
```
droshev, line 27: the PR should be `feat`, since we are changing the markup

stefanoScalzo:
> I'm not sure what about `feat` title, we don't provide any scss, that forces user to change markup.

Technically there is no usage for the modifier class so it makes sense removing it. Which will need to be adjusted to have the "perfect" solution

PR #1084, opened 2020-05-31
chore: update package versions in package.json

droshev:
update package versions in package.json


package.json
```diff
@@ -114,7 +114,7 @@
     "write-yaml": "^1.0.0"
   },
   "peerDependencies": {
-    "@sap-theming/theming-base-content": "~11.1.17"
+    "@sap-theming/theming-base-content": "~11.1.19"
```
jbadan, line 117: Should we get rid of this now that we've given alternate instructions for fonts/icons? We did in react.

droshev: I saw you are removing it in your PR https://github.com/SAP/fundamental-styles/pull/1096

PR #1096, opened 2020-06-01
chore: cleanup package.json - use cross-env for environment variables, add jest snapshots

jbadan:
## Related Issue

## Description
Cleanup package.json:
- remove @sap-theming from peer Dependencies - we no longer recommend consumers get their fonts and icons from here. It is useful as a dev dependency, but we shouldn't throw a peer dependency warning 

- switch to using cross-env for env variables https://www.npmjs.com/package/cross-env
- remove foreach-cli as it is no longer used

- fix storyshots setup


integration/storyshots.test.js
```diff
@@ -20,11 +20,10 @@ const customizePage = (page) => page.emulate(view);
 const beforeScreenshot = (page) => page.emulate(view);
 
 initStoryshots({
-  suite: "Image storyshots",
   test: imageSnapshot({
```
jbadan, line 23: this cleans up the names a bit

package.json
```diff
@@ -101,6 +99,7 @@
     "postcss-custom-properties": "^9.1.1",
     "postcss-import": "^12.0.1",
     "pre-commit": "^1.2.2",
+    "puppeteer": "3.1.0",
```
jbadan, line 102: necessary - https://github.com/storybookjs/storybook/tree/master/addons/storyshots/storyshots-puppeteer#getting-started

stories/button/button.html
```diff
@@ -1,5 +1,11 @@
...
 
-<button class="fd-button sap-icon--cart"></button>
\ No newline at end of file
+<div dir="{button.dir}">
+  <button
+    class="fd-button
+    {button.icon ? button.icon : ''}
+    {button.emphasized ? 'fd-button--emphasized' : ''}"
+  >{button.text}</button>
+</div>
```
jbadan, line 11: Building out button a little bit to show how we need to do these. When we transition this into docs, each example should be it's own `export const name =...` in the `.stories.js` file. Putting a big glob of html in here won't work well for documentation.

For visual regression tests, we can build this out in the future to have a `button.visual.js` where everything from `button.stories.js` is added to that on one page and that's where the visual image is taken. That's how we've been doing it in react so that the visual regression images don't take forever.

Issue #901, opened 2020-04-27
List component - documentation and code 


InnaAtanasova:
Size of list and its variations is getting huge. Keeping all this code in single .scss file might cause future problems with maintenance. Also documentation has grown rapidly. Perhaps we should consider splitting this implementation into separate files and think about creating sub-levels for our documentation?

PR #1097, opened 2020-06-02
chore: separate List scss into multiple files for easier maintenance

InnaAtanasova:
## Related Issue
Closes SAP/fundamental-styles#901

## Description
The List scss file was getting to long and hard to maintain so we decided to split it into multiple files and have 1 file where we import them.


src/list.scss
```diff
@@ -1,980 +1,8 @@
 @import "./new-settings";
 @import "./mixins";
```
jbadan, line 2: We can probably remove new-settings and mixins here since it's inside of listDefinitions

PR #1099, opened 2020-06-03
fix: Add fd-empty mixin, replace empty selectors

JKMarkowski:
## Description
On Angular we had some issues, when used `&:empty` styling. Sometimes Angular under the hood puts some commented code, especially when there are `*ngIf` statements.
image
So there is a need to be able to put empty programatically, it can be done only, by passing additional class.

### DIFF
--- a/src/button.scss
+++ b/src/button.scss
@@ -98,7 +98,7 @@ $block: #{$fd-namespace}-button;
     }
   }
 
-  &:empty {
+  @include fd-empty() {
     &::before,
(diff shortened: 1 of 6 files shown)


mikerodonnell89:
Checkbox and radio both have this selector which I believe also needs to be addressed `.#{$block}__label:empty`

JKMarkowski:
@mikerodonnell89 Nice Catch, I just changed it

PR #1113, opened 2020-06-08
chore: move to jsdocs for story descriptions

jbadan:
Now story descriptions can be added like this:
```
/** Story description here **/
export const basic = () => (`<button>Test</button`);
```
instead of:
```
export const basic = () => (`<button>Test</button`);
basic.parameters = {
    docs: {
       storyDescription: 'Story description here'
     }
};
```


jbadan:
Well... it affects the future docs. It really doesn't matter.

PR #1120, opened 2020-06-09
chore: add Token and Tokenizer to Storybook

mikerodonnell89:


InnaAtanasova:
Could you also add tags to your PR? Thanks

InnaAtanasova:
Also need to remove the link in the `test/templates/index.njk file`

PR #1124, opened 2020-06-09
chore: Add side-nav to storybook

JKMarkowski:
In this PR There are 
- Added Storybook examples for side navigation with javascript methods to deal with expanded flag
- Added Reference image snapshot for side nav
- Removed nested list and side navigation tests from backstop.js with removed ref images.


stefanoScalzo:
Due to using iframe height the docs description cannot be in comments.
underneath iframeHeight: x. put storyDescription: ''
if you have code blocks use <code></code>

jbadan:
Can you update with the latest merge to master (we fixed the snapshot/visual tests) and update the storybook tests here?

droshev:
I added nested list and rtl examples

PR #1132, opened 2020-06-09
feat: add blank option to Select

meganaconley:
## Description 
- Added an example with a blank option in Select
- Set a minimum width for Select to prevent resizing when selecting options and to support a default blank option. Previously the width was set by only the placeholder text causing a blank default to be extremely narrow and resizing if the options were different lengths from each other or the default text.

### Before:

Screen Shot 2020-06-09 at 12 54 06 PM

### After:

Screen Shot 2020-06-09 at 12 53 44 PM

Screen Shot 2020-06-09 at 12 53 28 PM


docs/pages/components/select.md
```diff
@@ -31,7 +31,7 @@ For lists that require more than 12 options, the <a href="/patterns/combobox-inp
             class="fd-select__control"
             role="button"
             tabindex="0">
-             Select
+             <span class="fd-select__text-content">Select</span>
```
jbadan, line 34: Sorry I forgot to say this earlier - can you add these changes to the select storybook examples?

PR #1141, opened 2020-06-10
chore: add Time component to Storybook

rengare:
## Description
The PR brings time component to storybook 

### After:
image


InnaAtanasova:
I addressed all the points except the a11y violation. Logged an issue for it https://github.com/SAP/fundamental-styles/issues/1152

PR #1146, opened 2020-06-10
chore: create link storybook and delete link njk tests

droshev:
create link storybook and delete link njk tests


droshev:
> Should be `chore` not `docs` to be consistent with the other PRs
> You have 1 a11y violation

done

PR #1147, opened 2020-06-11
feat: update selected page markup and add pagination storybook and remove njk

droshev:
add pagination storybook and remove njk

### BREAKING CHANGES
before(markup for selected page):
`<a href="#" class="fd-pagination__link" aria-selected="true">1</a>`

after:

`<a href="#" class="fd-pagination__link is-selected">1</a>`


droshev:
> Should be `chore` not `docs` to be consistent with the other PRs
> You have 1 a11y violation

done

PR #1149, opened 2020-06-11
[WIP] chore: make all heights 100% for storybook

stefanoScalzo:
## Description
This pr will allow all storybooks to take the height needed including popover heights without specifying the Iframe height. We will need to refactor the iframe heights as it is a lot of redundant code that we dont need! 

This solution is the described solution for specifying class specifics in the storybook files/documentation.

### Before:
Screen Shot 2020-06-11 at 2 36 47 PM

### After:
Screen Shot 2020-06-11 at 12 12 34 PM


jbadan:
Can you remove all the `iframeHeight: x` blocks with this?

PR #1156, opened 2020-06-15
fix: add fix for required form-label in IE11

InnaAtanasova:
## Description
In IE11 the `*` in required form label was not rendered next to `:`

### Before:
Screen Shot 2020-06-15 at 5 24 29 PM

### After:

Screen Shot 2020-06-15 at 5 33 44 PM


InnaAtanasova:
> It won't work with overflow
> image

Now I remembered why I had to add this code :D

InnaAtanasova:
This solution is breaking the overflow so I will close the PR

PR #1169, opened 2020-06-21
feat: update datepicker aria attributes and icon; introduce storybook

droshev:
update datepicker aria attributes and icon; introduce storybook
fixes: #1168 

Breaking changes:
calendar icon button:

- added `aria-label="button"`
- removed `aria-hidden="true"`
- replaced `footer` and `header` with `div`


stefanoScalzo:
Could you add a description to some of the examples please.

droshev:
> * There are still some a11y violations:

the a11y violations cannot be fixed as we have more than one example of banner landmark

droshev:
> Could you add a description to some of the examples please.

the title is the description

stories/datepicker/datepicker.stories.js
```diff
@@ -0,0 +1,1684 @@
...
+
+export const sizes = () => `
+<div class="fd-docs-date-picker">
+	<div class="fd-popover">
+		<div class="fd-popover__control">
+            <label class="fd-form-label" for="asdewq123">Pick a date:</label>
+            <div class="fd-input-group">
+				<input class="fd-input fd-input-group__input" type="text" value="" placeholder="Pick a date" id="asdewq123">
+				<span class="fd-input-group__addon fd-input-group__addon--button">
+				<button class="fd-input-group__button fd-popover__control fd-button fd-button--transparent sap-icon--appointment-2" aria-controls="Itcgq828" aria-expanded="false" aria-haspopup="true" aria-label="button"></button>
```
jbadan, line 25: I don't think the aria-label can just be "button" - why did we move away from aria-hidden?

meganaconley: It shouldn't be aria-hidden since it's a focusable/actionable button. We do need a better aria-label for it than `button` though.

rengare: The site https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html uses `aria-label="Choose Date"`

droshev: updated the label

droshev:
> * Toggling popover with calendar does not work properly if we navigate to the story from within the application. It works only when the application has been loaded with direct URL pointing to specified example.
>   datepicker-popover

This happens to all components like datepicker, popover, select,..

> * What do you think about reusing calendar view template for multiple examples? I don't think its changing for different examples (except `Today` buttons).

stories/datepicker/datepicker.stories.js
```diff
@@ -0,0 +1,1684 @@
...
+            </div>
+            <div class="fd-bar__right">
+                <div class="fd-bar__element">
+                    <button type="button" class="fd-button fd-button--transparent fd-calendar__close-button" aria-label="Close"></button>
+                </div>
+            </div>
+        </header>
+        <div class="fd-dialog__body fd-dialog__body--no-vertical-padding">
+            <div class="fd-calendar fd-calendar--mobile-portrait">
+                <header class="fd-calendar__header">
```
prsdthkr, line 1513: can this be changed to
```html
<div class="fd-calendar__header">
```
There should not be 2 headers on the page. Also since the application page might already have a [header that houses the app logo and navigational content](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/header) - I think all the `<header>` usages in Calendar should be changed to `<div>`.

stories/datepicker/datepicker.stories.js
```diff
@@ -0,0 +1,1684 @@
...
+			</footer>
+		</div>
+	</div>
+</div>
+`;
+
+todaySelectionButton.parameters = {
+    docs: {
+        iframeHeight: 460,
+        storyDescription: 'The today selection button in the footer selects today\'s date in the system or user-preferred timezone and closes the DatePicker popover. This date value then appears in the DatePicker\'s input field. If using a compact datepicker add the <code>fd-button--compact</code> class to this button.'
```
stefanoScalzo, line 585: add our class for the <code> element

stories/datepicker/datepicker.stories.js
```diff
@@ -0,0 +1,1684 @@
...
+                        OK
+                    </button>
+                </div>
+            </div>
+        </footer>
+    </div>
+</div>
+`;
+
+mobileDialog.storyName = 'DatePicker dialog in mobile landscape mode with today navigation';
```
stefanoScalzo, line 950: fix a11y issues

stories/datepicker/datepicker.stories.js
```diff
@@ -0,0 +1,1684 @@
+import '../../dist/calendar.css';
+
+export default {
+    title: 'Patterns/Datepicker',
+    parameters: {
+        description: `The date picker lets users select a localized date using touch, mouse, or keyboard input. It consists of two parts: the date input field and the date picker.
```
stefanoScalzo, line 6: if we are saying it as if it is an enumeration we should enumerate the points like
1. x
2. y

stories/datepicker/datepicker.stories.js
droshev: ?

stories/datepicker/datepicker.stories.js
droshev: ?

PR #1175, opened 2020-06-24
fix: specify DatePicker border radius

prsdthkr:
## Description
This change adds border-radius for fd-calendar and hence the DatePicker popover.

### Before
image

### Before (as consumed in fundamental-react DatePicker):
image

### After
image

### After (as consumed in fundamental-react):
image


droshev:
@prsdthkr maybe we should regenerate the storybook snapshots?

Issue #1172, opened 2020-06-23
Token is not compliant with latest Fiori3 updates


kavya-b:
**Describe the bug**
Some updates were made to paddings and other variables values in April. These need to be changed for the current Token/Tokenizer styles.

> Is this issue related to a specific component?
Token/Tokenizer

> What version of the Fundamental Library Styles are you using?
0.10

> What front-end framework are you implementing in, e.g., Angular, React, etc.?
Angular

**To Reproduce**
1. Go to 'https://sap.github.io/fundamental-styles/components/token.html'
2. Styles are not as per latest Fiori3 design

**Expected behavior**
Styles should be aligned with Fiori3

mikerodonnell89:
Could you be more specific?  What is missing?

droshev:
> **Describe the bug**
> Some updates were made to paddings and other variables values in April. These need to be changed for the current Token/Tokenizer styles.
> 
> > Is this issue related to a specific component?
> > Token/Tokenizer
> 
> > What version of the Fundamental Library Styles are you using?
> > 0.10
> 
> > What front-end framework are you implementing in, e.g., Angular, React, etc.?
> > Angular
> 
> **To Reproduce**
> Steps to reproduce the behavior:
> 
> 1. Go to 'https://sap.github.io/fundamental-styles/components/token.html'
> 2. Styles are not as per latest Fiori3 design
> 
> **Expected behavior**
> Styles should be aligned with Fiori3

@kavya-b can you share more details?

kavya-b:
1. **padding values have changed for token in both cozy and compact mode:**
current: only padding left value given
new: vertical padding in addition to padding left also given 

2. **padding for close icon and readonly class:**
current: only horizontal padding
new: has both vertical and horizontal paddings

3. **min-width added for cozy and compact modes**

4. **padding for tokenizer input**
current: left padding of 0.3125rem provided with assumption that tokens are present
new: left padding of 0.625rem when no tokens are present

before/after images are added in the PR #1176 description.

PR #1176, opened 2020-06-24
fix: modify paddings for token and tokenizer

kavya-b:
## Related Issue
Closes SAP/fundamental-styles#1172

## Description
Modified padding of token and tokenizer with latest values from Fiori design spec.


src/token.scss
```diff
@@ -92,12 +110,13 @@ $block: #{$fd-namespace}-token;
   @include fd-token-focus();
 
   &--compact {
-    padding-left: $fd-token-compact-padding-horizontal;
+    padding: 0 0 $fd-token-compact-padding-vertical $fd-token-compact-padding-horizontal;
     height: $fd-token-compact-height;
+    min-width: $fd-token-compact-min-width;
```
manu-kr, line 115: Is this necessary to create a variable if we aren't using more than one time. because its just increases line of code.

kavya-b: variables using rem values only once have now been removed

PR #1181, opened 2020-06-26
fix: label asterix and colon

JKMarkowski:
## Related Issue
closes: https://github.com/SAP/fundamental-styles/issues/870

## Description
There is added `--colon` modifier to every label.
Also spacing between asterix and label/colon is fixed
IE11 label is fixed

### Before:
image

### After:
image

Also there is fixed required label on IE11
image

### DIFF
--- a/stories/form-label/form-label.stories.js
+++ b/stories/form-label/form-label.stories.js
@@ -1,5 +1,6 @@
 import '../../dist/form-item.css';
 import '../../dist/input.css';
+import { footer } from "../list/standard/standard-list.stories";
 
 export default {
     title: 'Components/Forms/Form Label',
@@ -12,7 +13,7 @@ export default {
 
 export const primary = () => `
     <div class="fd-form-item">
-        <label class="fd-form-label" for="input-1">Default input:</label>
+        <label class="fd-form-label" for="input-1">Default input</label>
         <input class="fd-input" type="text" id="input-1" placeholder="Field placeholder text">
     </div>
 `;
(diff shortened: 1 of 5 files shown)


stefanoScalzo:
tihs lgtm, just make sure you remove in all labels the colon in the html

meganaconley:
How does this work with localization? IE Asian languages use double-width font, will the colon be double-width? Also I believe  French puts a space before colon, etc. This is why we include all punctuation in the localization files so that translators can adapt as necessary.

JKMarkowski:
@meganaconley It won't work well if the colon is changed, it can be of course changed by passing different `content` into `before/after` elements. The problem is absolute positioning, which has hardcoded distance between.

salarenko:
I'm definitely not a fan of this solution, but I cannot see another option (except creating additional wrapper) to achieve it.

Why do we even need this? Lets be honest, if you are creating a form where labels are overflowed with ellipsis then there is something wrong with your design or you are covering some really narrow case.

I don't understand why Fiori3 requires us to by default use ellipsis, I think the desired behavior for most of the cases is displaying the label in two rows instead of cutting it.

stories/form-label/form-label.stories.js
```diff
@@ -1,5 +1,6 @@
 import '../../dist/form-item.css';
 import '../../dist/input.css';
+import { footer } from "../list/standard/standard-list.stories";
```
jbadan, line 3: I think this was imported by accident

mikerodonnell89:
Can you document that this is optional?  Agreed with @salarenko that this is not the best option for everyone, especially regarding internationalization concerns

stories/form-label/form-label.stories.js
JKMarkowski: Changed

Issue #905, opened 2020-04-28
Add span example for bar component


stefanoScalzo:
**Describe the bug**
Add a span with text example for documentation on bar component
> Is this issue related to a specific component?

> What version of the Fundamental Library Styles are you using?
cureent
> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

PR #1183, opened 2020-06-26
docs: add logo text example to bar

stefanoScalzo:
## Related Issue
Closes SAP/fundamental-styles#905

## Description
Add an example that shows how to add text to a bar. There are other examples using buttons alone, therefore it would be smart to have one with a text with a span

### Before:
Screen Shot 2020-06-26 at 4 32 34 PM

### After:
Screen Shot 2020-06-26 at 4 31 14 PM


InnaAtanasova:
It's a good idea, I wouldn't use LOGO as text as most logos are images and will be confusing for the user. You can add inline styling to set max-width and maybe ellipses if you want to show the user how to handle long text and truncation. As far as I remember that was the reason for logging this issue

PR #1189, opened 2020-06-28
chore: bump storybook version

droshev:
bump storybook version


package.json
```diff
@@ -54,22 +54,22 @@
   "license": "Apache-2.0",
   "homepage": "https://sap.github.io/fundamental-styles/",
   "devDependencies": {
-    "@babel/core": "^7.10.2",
-    "@babel/node": "^7.10.1",
+    "@babel/core": "^7.10.3",
+    "@babel/node": "^7.10.3",
     "@babel/plugin-transform-runtime": "^7.10.1",
     "@babel/preset-react": "^7.10.1",
```
jbadan, line 60: If you bump the other babels - can you bump these too? They should be in sync

PR #1194, opened 2020-06-30
feat: introduce Object Marker component

### LINKED ISSUE #868: Object Marker component styles creation as per Frori3
Create Object Marker styles as per Fiori3 recommendations.
Screen Shot 2020-04-16 at 8 30 43 PM

Spec Values:
Screen Shot 2020-04-16 at 8 29 26 PM

Lokanathan-k:
Introduce Object Marker component

## Related Issue
fixes #868 

## Description
Introducing new Object Marker uploader component as per the fiori 3 specification.

### After:

Screenshot 2020-06-30 at 3 19 53 PM
Screenshot 2020-06-30 at 3 19 41 PM
Screenshot 2020-06-30 at 3 19 27 PM
Screenshot 2020-06-30 at 3 19 18 PM


docs/pages/components/object-marker.md
```diff
@@ -0,0 +1,84 @@
...
+{: .docs-intro}
+<br>
+
+# Object Marker Icon Only
+{% capture badge %}
+<span class="fd-object-marker sap-icon--request"></span>
+<span class="fd-object-marker sap-icon--add-favorite"></span>
+<span class="fd-object-marker sap-icon--flag"></span>
+<span class="fd-object-marker sap-icon--user-edit"></span>
+<span class="fd-object-marker sap-icon--private"></span>
```
meganaconley, line 21: If these are being used to indicate some sort of status they will need `aria-label`s.

docs/pages/components/object-marker.md
```diff
@@ -0,0 +1,84 @@
+---
+title: Object Marker
+id:
+keywords: object Marker, Marker, object, object-Marker, clickable
+sidebar: left-navigation-sidebar
+toc: false
+permalink: components/object-Marker.html
```
kavya-b, line 7: in the .yml file url is 'object-marker.html' but here it is 'object-Marker.html'

stories/object-marker/custom.scss
```diff
@@ -0,0 +1,3 @@
+.fd-object-marker {
+  margin-right: 20px;
```
kavya-b, line 2: why px and not rem here?

src/fundamental-styles.scss
```diff
@@ -39,6 +39,7 @@
 @import "./notification";
 @import "./numeric-content";
 @import "./object-status";
+@import "./object-marker";
```
InnaAtanasova, line 42: should be before object status

src/object-marker.scss
```diff
@@ -0,0 +1,85 @@
...
+  // CLICKABLE OBJECT STATUS
+  &--link {
+    display: inline-block;
+    text-decoration: none;
+    color: $fd-object-marker-color;
+    cursor: pointer;
+
+    &::before {
+      position: relative;
+      top: 0.125rem;
```
InnaAtanasova, line 44: why do you add this? I couldn't find this in the specs

docs/_data/sidebars/left-navigation-sidebar.yml
```diff
@@ -176,6 +176,10 @@ entries:
       url: /components/object-status.html
       output: web
 
+    - title: Object Marker
```
InnaAtanasova, line 179: same here

src/object-marker.scss
```diff
@@ -0,0 +1,85 @@
...
+
+  // ICON ONLY MODE
+  @include fd-empty() {
+    justify-content: center;
+    padding-right: 0;
+    padding-left: 0;
+
+    &::before {
+      padding-right: 0;
+      padding-left: 0;
```
InnaAtanasova, line 68: Screen Shot 2020-07-02 at 2 48 46 PM
 this part is missing

stories/object-marker/__snapshots__/object-marker.stories.storyshot
```diff
@@ -0,0 +1,194 @@
...
+
+  <span
+    class="fd-object-marker sap-icon--request"
+  >
+    Request
+  </span>
+  
+
+  <span
+    class="fd-object-marker sap-icon--add-favorite"
```
stefanoScalzo, line 64: should just be sap-icon--favorite

stories/object-marker/object-marker.stories.js
```diff
@@ -0,0 +1,81 @@
...
+    tags: ['f3', 'a11y', 'theme']
+};
+
+/**
+ * Object Marker can display the semantic status of an object: (class: `fd-object-marker`)
+ */
+
+export const iconOnly = () => `
+<span class="fd-object-marker sap-icon--request"></span>
+<span class="fd-object-marker sap-icon--add-favorite"></span>
```
stefanoScalzo, line 20: same here

stories/object-marker/object-marker.stories.js
```diff
@@ -0,0 +1,81 @@
...
+<span class="fd-object-marker">Locked</span>
+`;
+
+/**
+ * Object Marker icon and text
+ */
+
+export const iconAndText = () => `
+<span class="fd-object-marker sap-icon--request">Request</span>
+<span class="fd-object-marker sap-icon--add-favorite">Favourite</span>
```
stefanoScalzo, line 41: same here

stories/object-marker/object-marker.stories.js
```diff
@@ -0,0 +1,81 @@
...
+<span class="fd-object-marker sap-icon--flag">Flag</span>
+<span class="fd-object-marker sap-icon--user-edit">Draft</span>
+<span class="fd-object-marker sap-icon--private">Locked</span>
+
+<div dir="rtl">
+<h4>RTL Support</h4>
+<span class="fd-object-marker sap-icon--request">Request</span>
+<span class="fd-object-marker sap-icon--add-favorite">Favourite</span>
```
stefanoScalzo, line 49: same here

stories/object-marker/custom.scss
stefanoScalzo, line 2: can we use rem here. Also the measurements are incorrect, I  think you should use :after instead of before since the space is usually after the marker and add extra spacing if after that icon, there is another

InnaAtanasova: I think he is using this just to add some spacing between the object markers in the documentation, it's not part of the component styling

src/object-marker.scss
```diff
@@ -0,0 +1,83 @@
...
+
+  // ICON ONLY MODE
+  @include fd-empty() {
+    justify-content: center;
+    padding-right: 0;
+    padding-left: 0;
+
+    &::before {
+      padding-right: 0;
+      padding-left: $fd-object-marker-padding-left;
```
InnaAtanasova, line 67: This is not correct. 
Screen Shot 2020-07-07 at 2 42 28 PM

The specs specify that when two Object Markers are next to each other (adjacent siblings) the space between them should be 0.5 rem. As an example is given Icon and Icon but this does not exclude the other combinations. Which means that you should apply this spacing not only in "empty" state of the component. 
In your case this padding is always applied and it shouldn't be the case. It's only when you have 2 elements that are siblings.
Also it should be a margin, not a padding as padding is influencing the width of the element. I suggest you check if you can use the `Adjacent Sibling Selector` and apply margin left to the Object Marker that is following another Object Marker.

src/object-marker.scss
```diff
@@ -0,0 +1,98 @@
...
+
+    @include fd-empty() {
+      &::before {
+        padding-right: 0;
+        padding-left: 0;
+      }
+    }
+  }
+
+  &:first-of-type + span {
```
InnaAtanasova, line 85: wouldn't this work with ANY span?

src/object-marker.scss
```diff
@@ -0,0 +1,98 @@
...
+    &:empty {
+      margin-left: 0.5rem;
```
InnaAtanasova, line 87: Again, this spacing is not only when the Object Marker is an icon only  ("empty"). This is valid for ALL 2 Object markers that are adjacent. Basically if `.fd-object-marker + .fd-object-marker`, the second item (could be a span, could be a div) should have a margin-left of 0.5rem

src/object-marker.scss
Lokanathan-k: the class finally renders as .fd-object-marker:first-of-type + span:empty so it doesnt apply to any other span instead it applies to respective empty icon span tag.

InnaAtanasova:
Screen Shot 2020-07-09 at 10 41 04 AM

Only the second Object Marker has margin-left

docs/pages/components/object-marker.md
```diff
@@ -0,0 +1,121 @@
+---
```
droshev, line 1: @Lokanathan-k this is not rebased

docs/_data/sidebars/left-navigation-sidebar.yml
```diff
@@ -0,0 +1,261 @@
+# Left navigation
```
droshev, line 1: @Lokanathan-k this is not rebased

docs/_data/components.yml
```diff
@@ -0,0 +1,109 @@
+## Data for componnets list display page.
```
droshev, line 1: @Lokanathan-k this is not rebased

stories/object-marker/object-marker.stories.js
```diff
@@ -0,0 +1,109 @@
...
+<span class="fd-object-marker sap-icon--user-edit"></span>
+<span class="fd-object-marker sap-icon--private"></span>
+`;
+
+/**
+ * Object Marker Text only
+ */
+
+export const markerText = () => `
+<span class="fd-object-marker">draft</span>
```
stefanoScalzo, line 30: change to Draft

stories/object-marker/object-marker.stories.js
```diff
@@ -0,0 +1,109 @@
...
+`;
+clickableObjectMarker.parameters = {
+    docs: {
+        iframeHeight: 150
+    }
+};
+
+export const ObjectMarkerWithTwoIcon = () => `
+<div>
+<span class="fd-object-marker sap-icon--favorite"></span>
```
jbadan, line 84: This one needs an aria label

InnaAtanasova:
@Lokanathan-k looks great! Can you fix the failing build and we can merge it. Thanks

stories/object-marker/object-marker.stories.js
jbadan: Not sure why this was marked resolved - it's still missing an aria label

src/object-marker.scss
```diff
@@ -0,0 +1,93 @@
...
+.fd-object-marker
+*/
+$block: #{$fd-namespace}-object-marker;
+
+.#{$block} {
+  // ICON VARIABLES
+  $fd-object-marker-font-size: 1rem !default;
+  $fd-object-marker-color: var(--sapContent_MarkerIconColor) !default;
+  $fd-object-marker-text-color: var(--sapContent_LabelColor) !default;
+  $fd-object-marker-padding-left: 0.5rem !default;
```
mikerodonnell89, line 14: this variable isn't used anywhere

src/object-marker.scss
```diff
@@ -0,0 +1,93 @@
...
+  $fd-object-marker-padding-right: 0.25rem !default;
+  // TEXT VARIABLES
+  $fd-object-marker-font-family: var(--sapFontFamily) !default;
```
JKMarkowski, line 17: This one is not needed, it's in reset

src/object-marker.scss
```diff
@@ -0,0 +1,93 @@
...
+
+  @include fd-reset();
+
+  font-family: $fd-object-marker-font-family;
+  font-size: $fd-object-marker-font-size;
```
JKMarkowski, line 22: font size for text is sapMFontMediumSize (default), only icon has 1rem size

InnaAtanasova:
Hi @Lokanathan-k , I am a bit confused. 
Why do you implement clickable ObjectMarker when the specs say to use Link control if the ObjectMarker is clickable?
Screen Shot 2020-07-15 at 4 32 41 PM

Why did you add this example?
Screen Shot 2020-07-15 at 4 30 38 PM

InnaAtanasova:
Also something regarding scss variables that is worth mentioning. You create a scss variable if the value is used in more than 1 places in the code. 
For example: `$fd-object-marker-icon-size: 1rem !default;` is used only once in line 30. You don't have to put this value in a variable. In line 30 you just write: `font-size: 1rem;`. The idea of having scss variables is that if in future the value is changed you won't have to update all occurrences of this value, you will only update the scss variable. But this does not make sense when you have only 1 occurrence.

src/object-marker.scss
```diff
@@ -0,0 +1,93 @@
...
+  $fd-object-marker-color: var(--sapContent_MarkerIconColor) !default;
+  $fd-object-marker-text-color: var(--sapContent_LabelColor) !default;
+  $fd-object-marker-padding-left: 0.5rem !default;
+  $fd-object-marker-padding-right: 0.25rem !default;
+  // TEXT VARIABLES
+  $fd-object-marker-font-size: var(--sapFontSize) !default;
+
+  @include fd-reset();
+
+  font-size: $fd-object-marker-font-size;
```
JKMarkowski, line 21: Let's remove ` font-size: $fd-object-marker-font-size;`. It's already in `fd-reset()` mixin. You can also remove this variable.
image

You can see that font-size is duplicated

stories/object-marker/object-marker.stories.js
```diff
@@ -0,0 +1,86 @@
...
+/**
+ If the object marker is used as a link, a hover effect is shown on non-touch devices.
+ If the object marker is shown using a combination of icon and text, there is no hover effect for the icon.
+ If Object marker has to be clicked/tabbed by the user add the `fd-object-marker--link` modifier class.
+ */
+
+export const clickableObjectMarker = () => `
+<a class="fd-object-marker--link">
+<span class="fd-object-marker sap-icon--private"></span>
+Locked</a>
```
InnaAtanasova, line 69: You can't have such markup. 
`fd-object-marker--link` is a modifier class, it needs to modify something. You need something like `fd-object-marker fd-object-marker--link` where `fd-object-marker--link` modifies `fd-object-marker`

stories/object-marker/object-marker.stories.js
```diff
@@ -0,0 +1,86 @@
...
+<div dir="rtl">
+<h4>RTL Support</h4>
+<a class="fd-object-marker--link">
+	<span class="fd-object-marker sap-icon--private"></span>
+Locked</a>
+<a class="fd-object-marker--link"> <span class="fd-object-marker sap-icon--user-edit">
+</span>Unsaved Changes</a>
+</div>
+`;
+clickableObjectMarker.parameters = {
```
stefanoScalzo, line 82: when adding iframeHeight you must also specify the description here as well and not comments or storybook wont show it

stories/object-marker/object-marker.stories.js
```diff
@@ -0,0 +1,86 @@
+import '../../dist/object-marker.css';
+
+export default {
+    title: 'Components/ObjectMarker',
```
stefanoScalzo, line 4: use 'Components/Object Marker' to add the space in the side nav

InnaAtanasova:
Hi Lokanathan, 
with the latest changes the icon and the text are not aligned vertically
Screen Shot 2020-07-23 at 3 16 40 PM

I am not able to click in the Clickable Object Marker:
Screen Shot 2020-07-23 at 3 18 41 PM

InnaAtanasova:
Hi @Lokanathan-k , you haven't fixed the vertical alignment issue :) Also the icon gets the underline on hover:
Screen Shot 2020-07-27 at 9 50 31 AM

Not vertically aligned icon and text (current implementation):
Screen Shot 2020-07-27 at 9 35 26 AM

Vertically aligned (how it should be):
Screen Shot 2020-07-27 at 9 35 11 AM

InnaAtanasova:
@sKudum , @droshev , @mikerodonnell89 could you review the PR please and remove the requested changes if everything you pointed out is fixed. Thanks

droshev:
@Lokanathan-k travis is failing

stefanoScalzo:
LGTM

dimamarksman:
In case if link has `href` attribute there is underline below icon.
image

Lokanathan-k:
> @Lokanathan-k when changes are introduced you must request review again. In this case we approved something and you introduced something completely different. You could'v emerged this which is not ok. Also, the implementation we discussed without a span worked fine. Why did you change it?

@InnaAtanasova 
because the Span was needed for it work as expected in IE11 too. the text-decoration: underline was applied to the Icon as well in Internet Explorer, so this span tag was necessary to achieve the same look and feel in all browser.

InnaAtanasova:
Hi @Lokanathan-k , 
My personal opinion is that this solution will complicate the ngx, react and vue implementation. 
I think a better option is to use the following markup (this is just a simplified code to illustrate the idea).
For normal Object marker:
```
<span class="fd-object-marker">
    <span class="sap-icon--* fd-object-marker__icon"></span>
    Optional text here
</span>
```

For clickable Object marker:
```
<a class="fd-object-marker fd-object-marker--link">
    <span class="sap-icon--* fd-object-marker__icon"></span>
    Optional text here
</a>
```

Let me know what you think about it.

InnaAtanasova:
Screen Shot 2020-08-05 at 12 27 40 PM
Screen Shot 2020-08-05 at 12 27 47 PM

Hi @Lokanathan-k , please check my comment above again. Clickable and non-clickable (normal) Object Marker should have the same markup with the only difference that clickable object marker will have a modifier class. 

Normal Object Marker:
```
<span class="fd-object-marker"><i class="sap-icon--private fd-object-marker__icon"></i>Locked</i></span>
```

Clickable Object Marker:
```
<a class="fd-object-marker fd-object-marker--link"><i class="sap-icon--private fd-object-marker__icon"></i>Locked</i></a>
```

Differences between the two:
1. Normal Object Marker is a `<span>` and Clickable is an `<a>`
2. Clickable has `fd-object-marker--link` modifier class

While your approach works fine it will complicate the ngx/react/vue implementation

Lokanathan-k:
> Screen Shot 2020-08-07 at 8 21 56 AM
> 
> Hi @Lokanathan-k ,
> 
> 1. Object Marker with icon only should also follow the markup
> 
> ```
> <span><i></i></span>
> ```
> 
> 1. The size of the icon is not correct. As per specs it should be 1rem (16px).
> 
> Screen Shot 2020-08-07 at 8 19 44 AM
> 
> Screen Shot 2020-08-07 at 8 19 54 AM
> 
> 1. The line-height of the text and the icon should be `normal` not 1.

HI @InnaAtanasova 

when i do this '<span><i></i></span>' approach there is a additional space appending at the icon right which cannot be removed

Screenshot 2020-08-10 at 10 05 03 PM

the Height of the icon has been addressed.

src/object-marker.scss
```diff
@@ -0,0 +1,109 @@
...
+        text-decoration: none;
+      }
+    }
+
+    &:visited {
+      text-decoration: none;
+    }
+  }
+
+  &__icon {
```
JKMarkowski, line 57: Let' add `font-size: 1rem`. Because now it's `0.875`.

InnaAtanasova:
Screen Shot 2020-08-14 at 2 20 16 PM
Screen Shot 2020-08-14 at 2 19 45 PM

In couple of places you are closing the `<i>` multiple times:
<a class="fd-object-marker fd-object-marker--link"><i class="sap-icon--private fd-object-marker__icon"></i>Locked**</i>**</a>

stefanoScalzo:
You need to add focus

stefanoScalzo:
Needs to have hover colors when. hovering

Lokanathan-k:
> Needs to have hover colors when. hovering

@stefanoScalzo this change should go in fundamental style.

PR #1211, opened 2020-07-03
feat: remove fd-nav as it is unused

droshev:
Removing `fd-nav` as it seems to be an old leftover.
Removed classes: `fd-nav`, `fd-nav--vertical`, `fd-nav__item`, `fd-nav__link`


droshev:
I will close this as it is part of a bigger PR #1212

PR #1213, opened 2020-07-04
fix: remove button.scss from file uploader components

droshev:
Remove the include of the `button.scss` into file-uploader.scss`
This was including the whole button file as part of the file uploader.

### DIFF
--- a/src/file-uploader.scss
+++ b/src/file-uploader.scss
@@ -2,53 +2,29 @@
 @import "./new-settings";
 
 $block: #{$fd-namespace}-file-uploader;
+$input: #{$fd-namespace}-input;
+
+$fd-file-uploader-padding: 0.25rem;
 
 .#{$block} {
-  $fd-file-uploader-padding: 0.25rem;
-
   @include fd-reset();
+  @include fd-flex-vertical-center();
 
-  display: inline-flex;
-  align-items: center;
-  width: auto;
-
-  &__input.#{$fd-namespace}-input {
+  &__input.#{$input} {
     width: auto;
     margin-right: $fd-file-uploader-padding;
-
-    @include fd-ellipsis();
-
+    background-color: var(--sapField_Background);
     user-select: none;
+    cursor: pointer;
+
+    &::placeholder {
+      color: var(--sapTextColor);
+    }
 
     @include fd-rtl() {
       margin-left: $fd-file-uploader-padding;
       margin-right: 0;
     }
-
-    cursor: text;
-
-    @include fd-readonly() {
-      background-color: var(--sapField_Background);
-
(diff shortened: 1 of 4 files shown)


src/file-uploader.scss
```diff
@@ -38,19 +37,12 @@ $block: #{$fd-namespace}-file-uploader;
 
     @include fd-focus() {
       outline: none;
-
-      ~ .#{$block}__button {
```
droshev, line 42: @manu-kr what was the reason for that?

src/file-uploader.scss
```diff
@@ -38,19 +37,12 @@ $block: #{$fd-namespace}-file-uploader;
...
-        @include buttonContainerFocus();
-      }
     }
 
     @include fd-hover() {
       outline: none;
       cursor: pointer;
       border-color: var(--sapField_BorderColor);
-      ~ .#{$block}__button {
```
droshev, line 51: @manu-kr and that?

src/file-uploader.scss
manu-kr: @droshev I have included button.scss because of i was needed below mixins i.e defined in button.scss.

@mixin buttonContainerFocus() {
  box-shadow: none;

  @include standardFocus();

  &.#{$block}--ghost {
    @include ghostFocus();
  }

  &.#{$block}--positive {
    @include positiveFocus();
  }

  &.#{$block}--negative {
    @include negativeFocus();
  }

  &.#{$block}--attention {
    @include attentionFocus();
  }

  &.#{$block}--emphasized {
    @include negativeFocus();
  }
}



@mixin buttonContainerHover() {
  @include standardHover();

  &.#{$block}--ghost {
    @include ghostHover();
  }

  &.#{$block}--positive {
    @include positiveHover();
  }

  &.#{$block}--negative {
    @include negativeHover();
  }

  &.#{$block}--attention {
    @include attentionHover();
  }

  &.#{$block}--emphasized {
    @include emphasizedHover();
  }
}

These mixins are already defined in button.scss so i have included this just for reusing.

droshev: @manu-kr why would you need them? for the button?

jbadan:
I agree button.scss should be removed from the import - we also need to cleanup the docs versus stories. The html doesn't match...

src/file-uploader.scss
stefanoScalzo: I think it would be for validation before uploading a file
@manu-kr can you confirm

InnaAtanasova:
> I agree button.scss should be removed from the import - we also need to cleanup the docs versus stories. The html doesn't match...

Done! I also deleted more code :D

PR #1219, opened 2020-07-06
fix: select combobox role

jacobdevera:
## Description
The W3C recently created an example for [select-only comboboxes](https://w3c.github.io/aria-practices/examples/combobox/combobox-select-only.html), which is used for inspiration
- Use role `combobox` to support `aria-readonly`
- `aria-haspopup` should be `listbox`

### Deviations
- Not using `aria-activedescendant`; instead our markup is meant to support a roving tabindex for focus management which is functionally equivalent
- Not using `aria-labelledby`; our documentation features the Select in isolation. This attribute would be used in tandem with a form label

### Testing
- [x] JAWS + Chrome / IE

### Known Issue
- VoiceOver ❌  VO currently does not currently handle combobox very well. See https://github.com/w3c/aria-practices/pull/1396#issuecomment-627730571 . This is also reproducible in OpenUI5 Select


prsdthkr:
Should we move the popover ids from `<div>` to `<ul>`? The [WAI-ARIA v1.2 specification for combobox role](https://www.w3.org/TR/2019/WD-wai-aria-1.2-20191218/#combobox) says:
>  Authors MUST ensure the popup element associated with a combobox has a role of listbox, tree, grid, or dialog. Authors MUST set aria-controls on a combobox element to a value that refers to the combobox popup element.

jacobdevera:
> Should we move the popover ids from `<div>` to `<ul>`? The [WAI-ARIA v1.2 specification for combobox role](https://www.w3.org/TR/2019/WD-wai-aria-1.2-20191218/#combobox) says:

I added an `id` to the `listbox` and set `aria-controls` accordingly, but the id for the popover is still needed for triggering the open/close

Issue #610, opened 2020-01-27
Missing ellipsis in checkbox label | Too long label collapses checkbox input width


salarenko:
**Describe the bug**
This issue is related to Checkbox component

When checkbox label content is too long to fit in label container it shrinks checkbox input width and doesn't wrap with ellipsis.
Checkbox label also uses `width: fit-content` value which shouldn't be used because of no support on IE11.

**Expected behavior**
Checkbox doesn't collapse, label wraps with ellipsis.

**Screenshots**
![image](https://user-images.githubusercontent.com/17496353/73194347-5cd43900-412c-11ea-9d3f-8c25bc595f92.png)

Version: 0.13.1
Win 10, Chrome v79

PR #1220, opened 2020-07-07
feat: Checkbox label ellipsis overflow

salarenko:
## Related Issue
Closes SAP/fundamental-styles#610

## Description
In previous implementation text inside `.fd-checkbox__label` was not able to properly overflow with ellipsis, because of `display: flex`;

This PR introduces new wrapper which should be used inside `.fd-checkbox__label` to provide proper overflow for the text values.

### DIFF
--- a/docs/pages/components/form.md
+++ b/docs/pages/components/form.md
@@ -615,18 +615,20 @@ Do not use the checkbox control if:
         <div class="fd-form-item">
             <input type="checkbox" class="fd-checkbox" id="Ai4ez611">
             <label class="fd-checkbox__label" for="Ai4ez611">
-                Field label
+                <span class="fd-checkbox__text">Field label</span>
             </label>
         </div>
         <div class="fd-form-item">
             <input type="checkbox" class="fd-checkbox" id="Ai4ez612" checked>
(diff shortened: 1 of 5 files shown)


meganaconley:
A11y looks good - full label is read out when truncated. Seconding request for an example but otherwise looks good.

salarenko:
@InnaAtanasova I've also noticed this behavior in storybbok examples, but this is not related to checkbox itself.
Checkbox examples are grouped in `fieldset` and fieldsets are displayed in flex container. None of flexbox examples wrappers up to `class="root"` container has set width limitations, that's why example is stretching as much as it wants to, no matter if checkbox text/labels have some percentage based width limitation defined.

`.fd-checkbox__label` has `max-width: 100%` limitation. I could at it to `.fd-checkbox__text` but I don't think this is necessary because there is no other way of using `.fd-checkbox__text` than putting it inside `.fd-checkbox__label`.

This is how it works if you remove `display: flex` or define `max-width` from wrapping containers:
checkbox-overflow

salarenko:
@InnaAtanasova, @meganaconley
- I've updated sorybook styling, so now you can easily reproduce wrapping behavior
- I've also updated Breaking Changes

PR #1235, opened 2020-07-10
chore: popover click fix

rengare:
## Description
PR brings fix when popover is clicked. Before it throws console error because function was missing

### DIFF
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -304,6 +304,19 @@ function addClass(elementId, className) {
     }
 }
 
+function onPopoverClick(id) {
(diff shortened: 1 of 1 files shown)


.storybook/preview-head.html
```diff
@@ -304,6 +304,19 @@
     }
 }
 
+function onPopoverClick(id) {
+    let elementRef = document.getElementById(id);
+    let toggl = elementRef.getAttribute('aria-hidden');
```
InnaAtanasova, line 309: Is it intentional that `toggle` is `toggl`?

rengare: typo :)

PR #1236, opened 2020-07-10
chore: remove and fix old css variable dependencies

droshev:
emove and fix old css variable dependencies


src/tree.scss
```diff
@@ -160,7 +160,7 @@ $block: #{$fd-namespace}-tree;
 
     position: absolute;
     top: calc(50% - #{$fd-tree-control-width} / 2);
-    margin: -8px fd-space(3) 0 -36px;
```
droshev, line 163: this is in `px` as we will refactor it

Issue #1237, opened 2020-07-10
Button is broken


droshev:
- [x] Menu button #1252
<img width="272" alt="Screen Shot 2020-07-10 at 7 16 07 PM" src="https://user-images.githubusercontent.com/4380815/87209958-d8443400-c2e1-11ea-88a6-b5a7bd6a08a0.png">

 - [x] Split Menu Button #1252
<img width="169" alt="Screen Shot 2020-07-10 at 7 16 30 PM" src="https://user-images.githubusercontent.com/4380815/87209970-e2fec900-c2e1-11ea-9983-6322a714f75e.png">

- [ ] Segmented Button in Popover 
<img width="215" alt="Screen Shot 2020-07-11 at 5 09 18 PM" src="https://user-images.githubusercontent.com/4380815/87233866-5e6f8180-c399-11ea-8daa-6f568c8bfcf5.png">

rengare:
@droshev Segmented button in popover looks that it works
![image](https://user-images.githubusercontent.com/10849982/87409187-e71e3580-c5c4-11ea-960b-6397415176be.png)

PR #1252, opened 2020-07-14
fix: fixing button issues

rengare:
## Related Issue
Realted to #1237

## Description
The PR removes fd-reset from fd-icon since it broke button's example

### Before:
Check the issue

### After:
image
image

### DIFF
--- a/src/icon.scss
+++ b/src/icon.scss
@@ -3,7 +3,6 @@
 $block: sap-icon;
 
 [class*="#{$block}"] {
-  @include fd-reset();
   @include fd-icon-base();
 
   display: inline-block;


salarenko:
> Icons should still have a reset - you can override whatever style from the reset is causing the problem with `style: inherit`

I don't think `inherit` can be successfully used in this structure (siblings). The things we can do is changing icon implementation from this:
```
$block: sap-icon;

[class*="#{$block}"] {
  @include fd-reset();
  @include fd-icon-base();

  display: inline-block;
}

.#{$block} {
  @each $key, $value in $fd-icons {
    &--#{$key} {
      @include fd-icon-glyph($key);
    }
  }
}
``` 
To this:
```
%icon-base {
  @include fd-reset();
  @include fd-icon-base();

  display: inline-block;
}

.#{$block} {
  @each $key, $value in $fd-icons {
    &--#{$key} {
      @extend %icon-base;
      @include fd-icon-glyph($key);
    }
  }
}
```
But this still won't resolve all of the problems (file order import and selector strength "draw").

I think the only "solid" solution here is doing what @InnaAtanasova did in #1253. Not beautiful, but solves all of the problems.

Issue #1238, opened 2020-07-10
Defect Hunting - July 10-15


droshev:
- [x] Button issues - reported in #1237

- [x] do we have compact popover? There are some variables not being used [link](https://github.com/SAP/fundamental-styles/blob/a12ee3fe3dc0b472fbb6f0bb208577d95eac6000/src/popover.scss#L31) - the height variables are not being used

- [x] toolbar storybook missing #1245

- [x] combobox storybook missing #1245

- [x] context menu storybook missing #1245

- [x] multi input storybook missing #1245

- [x] time picker storybook missing #1245

- [x] input storybook missing #1245

salarenko:
- [ ] Info label with icon missing border (icons issue)
![image](https://user-images.githubusercontent.com/17496353/87412245-18990000-c5c9-11ea-92a7-dd39b5141738.png)

rengare:
- [ ] theme in IE 11:
![image](https://user-images.githubusercontent.com/10849982/87426913-43db1980-c5e0-11ea-9225-8348d526a852.png)
- [ ] storybook is super slow in IE 11 vs old documentation

PR #1253, opened 2020-07-14
feat: refactor Info Label

InnaAtanasova:
## Related Issue
Part of SAP/fundamental-styles#1238

## Description
This PR fixes one of the issues reported during [Defect Hunting](https://github.com/SAP/fundamental-styles/issues/1238) regarding Info Label with Icon missing border. 
- SCSS variables that were used only once in the code have been removed
- The `fd-info-label--only-icon` class has been removed. 

### Before:
Screen Shot 2020-07-14 at 4 52 19 PM

### After:
Screen Shot 2020-07-14 at 4 51 31 PM

### DIFF
--- a/src/mixins/_mixins.scss
+++ b/src/mixins/_mixins.scss
@@ -342,3 +342,23 @@
   color: $color;
   font-size: $fd-size;
 }
+
+@mixin fd-icon-overwrite {
+  &[class*=sap-icon] {
+    @content;
+  }
+}
+
+@mixin fd-icon-before-overwrite {
+  &[class*=sap-icon]::before,
+  &::before {
+    @content;
+  }
+}
+
+@mixin fd-icon-after-overwrite {
+  &[class*=sap-icon]::after,
+  &::after {
+    @content;
+  }
+}
--- a/src/info-label.scss
+++ b/src/info-label.scss
@@ -8,6 +8,7 @@ $fd-info-label-display-height: 1rem !default;
 $fd-info-label-icon-text-spacing: 0.1875rem !default;
 $fd-info-label-border-width: 0.063rem !default;
 $fd-info-label-border-style: solid !default;
+$fd-info-label-font-size: var(--sapFontSmallSize) !default;
 
 $color-accents: (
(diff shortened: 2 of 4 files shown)


src/mixins/_mixins.scss
```diff
@@ -342,3 +342,23 @@
   color: $color;
   font-size: $fd-size;
 }
+
+@mixin fd-icon-overwrite {
```
salarenko, line 346: Not sure about naming. I think this is a similar use case to our other mixins and `fd-icon-overwrite` does not overwrite anything by itself, so this should rather be just `fd-icon`.

src/mixins/_mixins.scss
```diff
@@ -342,3 +342,23 @@
...
+  &[class*=sap-icon] {
```
salarenko, line 347: I've noticed this before in Icon implementation, and I think we should switch `[attribute*=value]` selector to `[attribute^=value]` because of performance. But we can leave this for later talk.

src/mixins/_mixins.scss
InnaAtanasova: we added `reset` to icons so now we need to overwrite properties like `border:0`, `margin:0`, etc. `

InnaAtanasova:
> * [ ]  Why did you removed `--icon-only` class? I remember Info Label had to use some specific padding if used only with an icon.

The "default" padding is `padding: 0 0.625rem;` 
These are the paddings when you have Text + Icon and Icon Only:
Screen Shot 2020-07-15 at 9 07 17 AM

It's clear that in both cases they are the same. I don't see the need of having an additional class doing nothing :)
Screen Shot 2020-07-15 at 9 05 15 AM
Screen Shot 2020-07-15 at 9 07 31 AM

src/mixins/_mixins.scss
InnaAtanasova: `fd-icon` is already used

InnaAtanasova:
> * [x]  There is a difference in font size depending whether the icon has been used
>   image

@salarenko I fixed this

src/mixins/_mixins.scss
InnaAtanasova: That's not in the scope of the PR

Issue #1263, opened 2020-07-15
Visual Defect Hunting - July 15, 2020


droshev:
- [x] Input - https://github.com/SAP/fundamental-styles/pull/1265
<img width="1102" alt="Screen Shot 2020-07-15 at 5 26 20 PM" src="https://user-images.githubusercontent.com/4380815/87597757-575db180-c6c0-11ea-8c74-a1e64e9ec5dc.png">

- [x] Page - most probably the same as the Input - https://github.com/SAP/fundamental-styles/pull/1265
Before: 
<img width="1792" alt="Screen Shot 2020-07-15 at 5 22 44 PM" src="https://user-images.githubusercontent.com/4380815/87597579-0a79db00-c6c0-11ea-9d11-d711a14767d6.png">

After: 
<img width="1636" alt="Screen Shot 2020-07-15 at 5 22 57 PM" src="https://user-images.githubusercontent.com/4380815/87597594-106fbc00-c6c0-11ea-8613-93d1808e85af.png">

- [x] Checkbox - https://github.com/SAP/fundamental-styles/pull/1264
<img width="426" alt="Screen Shot 2020-07-15 at 5 28 52 PM" src="https://user-images.githubusercontent.com/4380815/87597939-aefc1d00-c6c0-11ea-8496-1b01430a2772.png">

- [x] Shellbar buttons - https://github.com/SAP/fundamental-styles/pull/1266
before:
<img width="178" alt="Screen Shot 2020-07-15 at 7 48 53 PM" src="https://user-images.githubusercontent.com/4380815/87610783-4dde4480-c6d4-11ea-9b66-7446f5089e4c.png">
<img width="242" alt="Screen Shot 2020-07-15 at 7 48 49 PM" src="https://user-images.githubusercontent.com/4380815/87610785-4dde4480-c6d4-11ea-9bd2-23a9d23c55c6.png">
After
<img width="173" alt="Screen Shot 2020-07-15 at 7 50 04 PM" src="https://user-images.githubusercontent.com/4380815/87610819-664e5f00-c6d4-11ea-8581-2bdd11ee3392.png">
<img width="170" alt="Screen Shot 2020-07-15 at 7 50 01 PM" src="https://user-images.githubusercontent.com/4380815/87610821-664e5f00-c6d4-11ea-82eb-b6886b8b0060.png">

Before:
[...]

PR #1271, opened 2020-07-17
fix: center Panel expand button icon

salarenko:
## Related Issue
Part of: #1263

## Description
Alternative to #1268

Removes unnecessary styling causing icon misalignment.

### DIFF
--- a/src/mixins/_icons.scss
+++ b/src/mixins/_icons.scss
@@ -6,8 +6,6 @@
   @include fd-icon-base($position);
   @include fd-icon-glyph($key, $position);
 
-  font-family: "SAP-icons";
-
   &::#{$position} {
     @content;
   }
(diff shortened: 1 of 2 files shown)


src/mixins/_icons.scss
```diff
@@ -6,8 +6,6 @@
   @include fd-icon-base($position);
   @include fd-icon-glyph($key, $position);
 
-  font-family: "SAP-icons";
```
InnaAtanasova, line 9: why do you remove it?

InnaAtanasova:
I like this solution more than mine. I will close mine in favour of this one.

src/mixins/_icons.scss
salarenko: Take a look how this `@mixin` is used:
```
some-button {
   @include fd-icon('slim-arrow-right') {
      font-size: 1rem;
   }
}
```

Previous implementation was adding `font-family: "SAP-icons"` to both: `.some-button` and `.some-button::before`.
I don't think that modifying styling of `.some-button` is a valid thing because icon exists only on its pseudo-elements.

PR #1272, opened 2020-07-17
chore: update the deploy script

droshev:
update the deploy script


ci-scripts/publish.sh
```diff
@@ -30,5 +30,6 @@ npm publish
 # run this after publish to make sure GitHub finishes updating from the push
 npm run release:create -- --repo $TRAVIS_REPO_SLUG --tag $release_tag --branch master
 
-npm run docs:prod
+npm run build:prod
```
jbadan, line 33: npm run storybook-static then ✅

PR #1278, opened 2020-07-20
chore: 1204 new version of std version

### LINKED ISSUE #1204: migrate to standard version 8.0.0
At the moment the repository is using `standard-version` version `4.2.0` and the goal is to migrate to version `8.0.2`.

We are facing the same issue as https://github.com/conventional-changelog/standard-version/issues/604

[migrating](https://github.com/SAP/fundamental-styles/pull/1248) to standard-version 8.0.0 is blocked

rengare:
## Related Issue
Closes #1204 

## Description
PR migrate std version to 8.0.2 and upgrades nodejs ver to 12.18.2 in order to fix std versions for styles.
PR brings a nodejs script as a workaround till official response and actions are taken by owner of standard-version liblary


.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: node_js
-node_js: 10.21.0
```
jbadan, line 2: I'm curious how this fixes it! Nice find! Is there a link to a test repo in travis showing it working?

If we need to change node version, we should also do it here: https://github.com/SAP/fundamental-styles/blob/master/.nvmrc

rengare: I'm still testing it so I'll find out if this works ;)

PR #1292, opened 2020-07-23
docs: updating the avatar documentation

jessmetivier:
Updating the avatar documentation with more context to the component types, usage and sizes.

(Not sure about the additional information that needs to be included here, will ask during my meeting with deno about PRs)

### DIFF
--- a/stories/avatar/avatar.stories.js
+++ b/stories/avatar/avatar.stories.js
@@ -16,6 +16,7 @@ export default {
 | L | 5 rem | App headers for normal screen sizes | --l modifier |
 | XL | 7 rem | App headers for large screen sizes | --xl modifier |
 
+
 ### Usage
 Use avatar to display:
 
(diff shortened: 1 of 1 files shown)


InnaAtanasova:
> I didn't understand * as a placeholder character, maybe this is widely known in the developer world. Just in case it's not recognized by designers, I decided to remove it.

In the case of `fd-avatar--accent-color-*` * means it can be 1, 2, 3, 4, ...10
`fd-avatar--accent-color-1`, `fd-avatar--accent-color-2`, `fd-avatar--accent-color-3`...`fd-avatar--accent-color-10`

In all cases where * is used it means everything up to the * is repeated and the * means a single word/number/character can be added.

InnaAtanasova:
closing in favour of https://github.com/SAP/fundamental-styles/pull/1310

Issue #756, opened 2020-03-04
Button: use aria-disabled instead of disabled attribute


jbadan:
> ### Description
> For accessibility purposes, it will be better to use the aria-disabled attribute than the built-in `disabled` attribute to differentiate disabled Buttons.
> 
> * Use aria-disabled instead of disabled
> * Handle click/key events when disabled
> * Ensure styles are same
> * Add aria-live region to announce button state change
> 
> ### References
> https://axesslab.com/disabled-buttons-suck/
> and
> https://a11y-101.com/development/aria-disabled

Copied from https://github.com/SAP/fundamental-react/issues/889

Changes will need to be made in this library first to:
1. ensure styles are not affected
2. update documentation

PR #1293, opened 2020-07-24
feat: add screen-reader-only styles to button

Mike-Diaz:
## Related Issue
Closes SAP/fundamental-styles#756

## Description
- Add Styling
added `__instructions` with `screen-reader-only` styling to alert message for a `disabled` button with `allowFocusOnDisable` set to **true**. This text is used for screen readers to alert a user when a button is disabled or when it becomes enabled. 
This message should not be visible to a sighted user not using a screen reader. The button retains all styling previously in place. 

- Add aria-disabled
By using `aria-disabled="true"` without the `disabled` prop, a screen reader will be able to read the state of the button since the button will now be focusable and reachable by key navigation. The screen reader will read out the state of the button informing the user the button is disabled

Related to [Fix: focusable disabled button #1134](https://github.com/SAP/fundamental-react/pull/1134)

### Before:
Button_not_focusable_no_aria

### After:
- Focusable
Disabled_Button_Focusable1

### DIFF
--- a/src/button.scss
+++ b/src/button.scss
@@ -50,7 +50,7 @@ $block: #{$fd-namespace}-button;
   cursor: not-allowed;
 
   &:focus {
-    outline: none;
+    outline: solid;
   }
 }
 
--- a/src/button.scss
+++ b/src/button.scss
@@ -428,6 +428,10 @@ $block: #{$fd-namespace}-button;
     @include half();
   }
 
+  &__instructions {
+    @include fd-screen-reader-only();
+  }
+
   @include fd-hover() {
     @include buttonContainerHover();
   }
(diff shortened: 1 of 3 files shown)


jacobdevera:
There is currently no focus ring on the focusable disabled button. I think a focus ring of the same style as the other buttons should be added to it.

jbadan:
@Mike-Diaz I changed your title to `feat`. This repo was recently upgraded to use standard-version@8 🎉  That means we can use sane titles again,  feat = feature, fix = fix, chore = chore, docs = docs, what a world! 😆

jbadan:
@stefanoScalzo This is a feat. Like I said in my comment, we've upgraded to standard version@8. Therefore, a feature should be labeled as such.

Mike-Diaz:
@jacobdevera Focus ring is now enabled with the use of `is-disabled` and the absence of `disabled.

Mike-Diaz:
changed focus ring style from `solid` to `buttonFocus` for aria-disabled buttons.

Issue #1291, opened 2020-07-23
bug: Storybook website links don't work


Vanessa-Cusmich:
Some of the links aren't clickable. The issue is caused by the offset in Storybook.

Specifically the Contributing, Support and SLA sections.
https://sap.github.io/fundamental-styles/?path=/docs/introduction-overview--page

PR #1294, opened 2020-07-24
chore: Remove page header overflow

salarenko:
## Related Issue
Closes: #1291

## Description
Page headers had a `before` pseudo-element which was overflowing page content preventing user from interaction with covered page elements.

I'm not sure what was the purpose to introduce this styling, but I don't see a value, so my proposition is just to remove it.


droshev:
we should merge in master and downport it to version `0.11`

jbadan:
That was a relic from when the headings were not anchors 😄  The underlying issue was probably fixed when we upgraded storybook

Issue #1051, opened 2020-05-21
button focus goes outside on IE11


stefanoScalzo:
**Describe the bug**
focus goes outside of the button
> Is this issue related to a specific component?
button
> What version of the Fundamental Library Styles are you using?
latest
> What front-end framework are you implementing in, e.g., Angular, React, etc.?
all

**To Reproduce**
1. Click button

**Expected behavior**
Focus should be on the inside of the button
**Screenshots**
![Screen Shot 2020-05-21 at 3 53 04 PM](https://user-images.githubusercontent.com/50607147/82599985-2b6bf700-9b7b-11ea-9f12-b367f1e90f78.png)

rengare:
a common issue in IE 11 which can't be fixed

droshev:
another button issue #1055

Issue #1055, opened 2020-05-23
button on press down doesn't show the focus dots(Safari, Firefox)


sKudum:
**Description**
          On click on buttons dotted border focusing is not happening for Firefox, Safari.

> Is this issue related to a specific component?
          :  Button

> What version of the Fundamental Library Styles are you using?
            0.9

> What front-end framework are you implementing in, e.g., Angular, React, etc.?
            Angular

**To Reproduce**

1. In Mac Machine open Browser: Safari, firefox
2. Enter the fundamental-styles url
3. Click on Button
4. It is not focused on the dotted border.

**Expected behavior**
        Mouse click on button focusing with dotted border should  happen even in Firefox and Safari

<img width="290" alt="Screen Shot 2020-05-23 at 2 21 26 PM" src="https://user-images.githubusercontent.com/53487468/82726329-073a2280-9d01-11ea-933e-d70b6ef68e84.png">

droshev:
a button related issue #1051

InnaAtanasova:
As discussed with @rengare this fix would require `before` or `after` attributes which are already used for the icons. Currently we don't have a solution for this issue.

droshev:
We discussed the issue and agreed to do a breaking change and introduce a `i` element for the icon so we can address the focus and press down issue

PR #1305, opened 2020-07-30
feat: moving icons from pseudoelements to <i> tag in button

rengare:
## Related Issue
closes https://github.com/SAP/fundamental-styles/issues/1051

## BREAKING CHANGE
- the button icon is moved to a separate child `<i>` element
- Menu button: the text is wrapped in a `<span>`

## Description
The button's markup with icon has been changed. 
Before:
```html
<button class="fd-button sap-icon--*"></button>
```
After:
```html
<button class="fd-button">
    <i class=sap-icon--*></i>
</button>
```

Sap Icons inside button now has inherit color, same as pseudo elements, which are responsible to keep icon.
Also `fd-button__text` element has been introduced, for now it's only for menu button. It's made to let icon know, that there is any other element (plain text is not treated as element), so it can add margin.

The PR brings new icon syntax to button's also there button usage examples changed in 
- action bar,
- bar
- calendar
- combobox
- multi input
- product switch
- select
- shellbar
- step input
- time
- time picker
- toolbar

- popover
- counter 
- datepicker
- form grid
- input group
- notification
- page

## Breaking change
image


stories/button/button.stories.js
```diff
@@ -16,19 +16,19 @@ export default {
 
 export const primary = () => `
     <button class="fd-button">Create</button>
-    <button class="fd-button sap-icon--cart"></button>
+    <button class="fd-button"><i class="sap-icon--cart fd-button__icon-left"></i></button>
```
jacobdevera, line 19: This would be a good opportunity to add an `aria-label` to all the `<button>` elements with no inner text (similar to what is done in line 173).

rengare: Done

stories/button/button.stories.js
```diff
@@ -49,9 +60,9 @@ types.parameters = {
 
 export const segmentedButton = () => `
     <div class="fd-segmented-button" role="group" aria-label="Group label">
-        <button class="fd-button sap-icon--survey"></button>
-        <button class="fd-button sap-icon--pie-chart is-selected" aria-pressed="true"></button>
-        <button class="fd-button sap-icon--pool"></button>
+        <button aria-label="More" class="fd-button"><i class="sap-icon--survey fd-button__icon-left"></i></button>
+        <button aria-label="More" class="fd-button is-selected" aria-pressed="true"><i class="sap-icon--pie-chart fd-button__icon-left"></i></button>
+        <button aria-label="More" class="fd-button"><i class="sap-icon--pool fd-button__icon-left"></i></button>
```
jacobdevera, line 65: Rather than all having the same `aria-label` `"More"`, it should indicate the purpose of the button. You can make up ones for demo purposes based on what the icon represents e.g. "Survey", "Pie Chart", "Pool", "Cart"

jbadan:
This is the same solution we use internally at Concur - https://hig.concur.com/pages/develop/web/components/nui-widgets/v13/buttondropdown.html, https://hig.concur.com/pages/develop/web/components/nui-widgets/v13/button.html

We also add an additional `<span>` for text inside the button -ex: `<button class="fd-button"><i class="sap-icon--cart fd-button__icon-left"></i><span>button</span></button>` 

I don't think we should make a workaround solution just to avoid breaking changes, unless there's a compelling reason to do this a different way - a11y, etc. We're 0th for a reason and shouldn't be afraid of breaking changes nor make development decisions based on avoiding them.

stories/button/button.stories.js
JKMarkowski: @jacobdevera Done

JKMarkowski:
@jbadan Let's keep the `<i>` elements then. It will free pseudo elements, that can be used to render properly focus indicator, even on IE11. Overflow/position are good for now, so I'm not into introducing additional <span> element for text.

src/shellbar.scss
```diff
@@ -283,6 +283,16 @@ $block: #{$fd-namespace}-shellbar;
     padding: 0;
     outline-color: var(--sapContent_ContrastFocusColor);
 
+    > i {
```
jbadan, line 286: Can we avoid using element selectors?

JKMarkowski:
It's ready to be reviewed, all of components have new markup now.

stories/feed-input/feed-input.stories.js
```diff
@@ -55,11 +57,13 @@ export const notEmpty = () => `
...
                   aria-required 
                   rows="1">"Textarea" with a text makes submit button active</textarea>
-        <button class="fd-button sap-icon--feeder-arrow 
+        <button class="fd-button 
                 fd-feed-input__submit-button" 
                 aria-label="Send"
                 aria-disabled="false" 
-                title="Send"></button>
+                title="Send">
+            <i class="sap-icon--feeder-arrow"></i>
```
stefanoScalzo, line 65: there are 2 icons per feed input

JKMarkowski:
> What does this have to do with #1055 ? I have addressed that issue without breaking changes with pr #1337

It's more about fixing https://github.com/SAP/fundamental-styles/issues/1051, Idk why 1055 was as a scope.
It will release pseudo elements for additional padding top/bottom clickable area and focus indicator working even on IE11

JKMarkowski:
@droshev 
> Can we also address the clickable area
> Screen Shot 2020-08-17 at 8 32 11 PM

I would do it in other PR, same as changing focus outline, let's have a talk about it on meeting

JKMarkowski:
@InnaAtanasova 
I wanted to not have higher button, right now it doesn't take more space.  Should it?
As you can see clickable area is expanded, 4 pixels are just not enough to notifce it
button

JKMarkowski:
@InnaAtanasova it has been fixed. Feel free to check it again

src/button.scss
```diff
@@ -83,321 +78,295 @@ $block: #{$fd-namespace}-button;
...
+    > .#{$block}__icon--right {
+      margin-left: 0;
+      margin-right: $fd-button-icon-margin;
     }
   }
 }
 
+@mixin iconStyles {
+  @include iconStylesLtr();
+  @include iconStylesRtl();
```
salarenko, line 346: I don't think RTL should be splitted from LTR like that. Using `@include iconStylesRtl();` is misleading. I'd expect it to bring full icon styling for RTL mode, but in fact this brings only a modifier and to properly work it musts always be used together with LTR mixin.

src/mixins/button/_emphasized.scss
```diff
@@ -23,11 +23,15 @@
 }
 
 @mixin emphasizedPressedFocus() {
-  outline-color: var(--sapContent_ContrastFocusColor);
+  &::after {
```
salarenko, line 26: Maybe it's worth introducing so we can more easily control element used as focus host?

```
@mixin buttonFocusLocation() {
  &::after {
    @content
  }
}
```

salarenko (changes requested):
Looks, good. Just two minor concerns:

src/mixins/button/_emphasized.scss
JKMarkowski: Let's postpone it to buttons refactor

src/button.scss
JKMarkowski: Great Idea, it really can be misleading

src/mixins/button/_emphasized.scss
InnaAtanasova: Aren't we refactoring Buttons in this PR? I think we should try to finalize the Button refactoring and stop introducing Breaking Changes for this component.

JKMarkowski: It has been addressed - I removed some redundant mixins, which didn't bring any value

salarenko:
Changes introduced by me [(commit):](https://github.com/SAP/fundamental-styles/pull/1305/commits/5f0fb6e1d55101f9e3a4d44140a4c8bcef2843e8)
- Removed unused variables
- Removed `iconSize` which was referring to styling the icons on `before`, `after` elements
- Merged `iconStylesLtr` and `iconStylesRtl` into one code-block `iconStyles` for better readability
- Removed `--half` modifier, a leftover from button--half used for recently removed input-number with spinner
- Removed the need of having `fd-button__icon--right`, `fd-button__icon--right` classes

InnaAtanasova:
Screen Shot 2020-08-25 at 2 30 29 PM
Screen Shot 2020-08-25 at 2 30 23 PM

I think the focus offset is not correct. @estelaxu3  and @jessmetivier could you check the visuals and confirm? Thanks

jessmetivier:
@InnaAtanasova Looks like you're right about the offset being "off" hehe (Excuse my terrible pun 😅 )

From the looks of it, the padding is slightly larger than what's illustrated in SAPUI5:

**SAP UI5 button offset**
SAPUI5 button offset

**Here are the specs from UI5:** 
Screen Shot 2020-08-25 at 2 55 34 PM

src/mixins/button/_button-helper.scss
```diff
@@ -7,8 +7,18 @@ $block: #{$fd-namespace}-button;
 }
 
 @mixin buttonFocus() {
-  outline-style: dotted;
-  outline-width: var(--sapContent_FocusWidth);
-  outline-color: var(--sapContent_FocusColor);
-  outline-offset: -0.1875rem;
+  $fd-focus-offset: 0.125rem;
```
InnaAtanasova, line 10: should be 0.1875rem

PR #1332, opened 2020-08-05
fix: button active outline for safari and ff

### LINKED ISSUE #1055: button on press down doesn't show the focus dots(Safari, Firefox)

mikerodonnell89:
fixes #1055 

Problem is that when clicking a <button> on firefox or safari, specifically running OSX, :focus selector is not applied, only :active.  Previous implementation didn't apply outline unless the button was both focused and active.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#Clicking_and_focus

### DIFF
--- a/src/button.scss
+++ b/src/button.scss
@@ -329,22 +329,27 @@ $block: #{$fd-namespace}-button;
     @include standardPressedFocus();
 
     &.#{$block}--ghost {
+      @include ghostPressed();
       @include ghostPressedFocus();
     }
 
     &.#{$block}--positive {
+      @include positivePressed();
       @include positivePressedFocus();
     }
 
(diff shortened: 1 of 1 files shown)


stefanoScalzo:
The focus disappears after the click

PR #1338, opened 2020-08-06
feat: Title component

meganaconley:
## Description
Introduce Fiori 3 Title component and replace existing titles in other components with new Title component for consistency.

## Screenshots

Screen Shot 2020-08-06 at 11 33 18 AM

Screen Shot 2020-08-06 at 11 33 23 AM

Screen Shot 2020-08-06 at 11 33 29 AM

Screen Shot 2020-08-06 at 11 33 46 AM


stories/action-bar/action-bar.stories.js
```diff
@@ -19,7 +20,9 @@ export const backBtn = () => `
         <div class="fd-action-bar__back">
             <button aria-label="button"class="fd-button fd-button--transparent fd-button--compact sap-icon--navigation-left-arrow"></button>
         </div>
-        <h3 class="fd-action-bar__title">Page Title</h3>
+        <div class="fd-action-bar__title">
+            <h1 class="fd-title fd-title--h3">Page Title</h3>
```
InnaAtanasova, line 24: Shouldn't the tag match the modifier class? `<h3>` ... `--h3`, `<h1>`... `--h1` ? I'm afraid `<h1 class="fd-title fd-title--h3">` will be very confusing for the user.

meganaconley: The spec does state the visual and semantic levels can be different. While it's confusing at first, it's a big win for accessibility. Often a design wants a smaller or larger heading than actually belongs on the page. This way developers can set the correct levels while styling them differently if needed.

stefanoScalzo: > The spec does state the visual and semantic levels can be different. While it's confusing at first, it's a big win for accessibility. Often a design wants a smaller or larger heading than actually belongs on the page. This way developers can set the correct levels while styling them differently if needed.

This is fine but me but I think this should be specified in documentation so users are not confused!

meganaconley: I've updated documentation in this component and fixed the closing tag.

PR #1397, opened 2020-08-17
fix: List footer as span

### LINKED ISSUE #1396: List footer should not be a list item
Since a list's footer is not semantically a part of the list it should not be a `<li>` within the `<ul>`.

List

v0.12.0

React

### Suggested Markup
```html
<ul class="fd-list" role="list">
      <li tabindex="0" class="fd-list__item"><span class="fd-list__title">Apple</span></li>
      <li tabindex="0" class="fd-list__item"><span class="fd-list__title">Banana</span></li>
      <li tabindex="0" class="fd-list__item"><span class="fd-list__title">Durian</span></li>
      <li tabindex="0" class="fd-list__item"><span class="fd-list__title">Grape</span></li>
   </ul>
<span class="fd-list__footer">Updated 3 minutes ago</span>
```

prsdthkr:
## Related Issue
Closes #1396 

## Description
In this change, we update the List's markup so that the List's footer is no longer a list item. The footer appears as a span after the unordered list.

## Screenshots
No visual changes

### DIFF
--- a/stories/dialog/__snapshots__/dialog.stories.storyshot
+++ b/stories/dialog/__snapshots__/dialog.stories.storyshot
@@ -375,17 +375,17 @@ exports[`Storyshots Components/Dialog Select Dialog Example 1`] = `
                 
           </li>
           
-                
-          <li
-            class="fd-list__footer"
-          >
-            
-                    2 items selected
-                
-          </li>
-          
             
         </ul>
+          
+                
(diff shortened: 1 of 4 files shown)


droshev:
@prsdthkr can we add breaking change when we merge so we don't forget to update the markup in the implementation libraries?

Issue #1383, opened 2020-08-14
Create Component Message Page


droshev:
Message pages give feedback to the user when an app or list is empty, or when an error has occurred. There are different use cases for showing a message page. The layout is the same, but the text and icon can change depending on the use case. You can use the message page in the following situations:

- Filtering with no results
- Searching with no results
- Empty app
- Too many items
- Item saved as a tile that is not longer available
- Error

More info: visual spec and interaction

PR #1402, opened 2020-08-18
feat: new component Message Page

InnaAtanasova:
## Related Issue
Closes SAP/fundamental-styles#1383

## Description
Introducing new component Message Page


stories/message-page/message-page.stories.js
```diff
@@ -0,0 +1,212 @@
...
+        </div>
+    </div>
+</div>
+`;
+
+filter.storyName = 'Message Page with Filter';
+filter.parameters = {
+    docs: {
+        iframeHeight: messageBoxHeight,
+        storyDescription: `Use case: The user has set a filter and there are no results.
```
jessmetivier, line 56: "Use the filter message page for when a user has set a filter and there are no results. This message then prompts the user to remove or re-evaluate the filter they set for the page."

stories/message-page/message-page.stories.js
```diff
@@ -0,0 +1,212 @@
...
+        </div>
+    </div>
+</div>
+`;
+
+search.storyName = 'Message Page with Search';
+search.parameters = {
+    docs: {
+        iframeHeight: messageBoxHeight,
+        storyDescription: `Use case: The user has searched for something but there are no results.
```
jessmetivier, line 84: "Use the search message page for when a user has searched for something, but there are no results for that inquiry."

stories/message-page/message-page.stories.js
```diff
@@ -0,0 +1,212 @@
...
+        </div>
+    </div>
+</div>
+`;
+
+noItems.storyName = 'Message Page with No Items';
+noItems.parameters = {
+    docs: {
+        iframeHeight: messageBoxHeight,
+        storyDescription: `Use case: The app contains no items.
```
jessmetivier, line 112: "Use the no items message page for when the user has viewed a page that is supposed to contain items, but there are none.

Mandatory: Display a **product icon**, or any other icon that matches your use case."

stories/message-page/message-page.stories.js
```diff
@@ -0,0 +1,212 @@
...
+        </div>
+    </div>
+</div>
+`;
+
+error.storyName = 'Message Page with an Error';
+error.parameters = {
+    docs: {
+        iframeHeight: messageBoxHeight,
+        storyDescription: `Use case: The app cannot show any content due to an error.
```
jessmetivier, line 143: Use the error message page for when the user attempts to view a page but an error has occurred, resulting in no available content. If you can, provide a link to the app start screen in the error message.

Mandatory: Add the <code>.sap-icon--document</code> class to the <code>fd-message-page___icon-container</code> to display a page icon.

stories/message-page/message-page.stories.js
```diff
@@ -0,0 +1,212 @@
...
+        </div>
+    </div>
+</div>
+`;
+
+buttons.storyName = 'Message Page with Formatted Text and Buttons';
+buttons.parameters = {
+    docs: {
+        iframeHeight: messageBoxHeight,
+        storyDescription: `The text on the Message Page can be formatted (such as bold, italic, underline, and bullet points), and also buttons can be added in the Actions area. It's recommended to not use more than 2 buttons in the Actions area.
```
jessmetivier, line 208: "Use the message page with formatted text and buttons when you want to add formatted text (bold, italic, underline and/or bullet points) and buttons to the action area. It's highly recommended not to use more than two buttons."

stories/message-page/message-page.stories.js
jessmetivier: I bolded product icon because I assume that this is a known icon name in Fiori? Can you confirm?

jessmetivier: Otherwise don't bold it in the documentation!

stories/message-page/message-page.stories.js
InnaAtanasova: we need the "...in the Actions area" at the end of the sentence. There could be more buttons (like the Show More button) but in the Actions area it's recommended to have at most 2

stories/message-page/message-page.stories.js
InnaAtanasova: Mandatory: Use the <code class="docs-code">.sap-icon--document</code> icon.

I left this part like this. `.sap-icon--document` is inside the `fd-message-page___icon-container`, the class is not applied to it.

stories/message-page/message-page.stories.js
InnaAtanasova: Done

stories/message-page/message-page.stories.js
```diff
@@ -0,0 +1,198 @@
...
+const messageBoxHeight = 800;
+
+export const filter = () => `
+<div style="height: 600px; width: 100%;">
+    <div class="fd-message-page">
+        <div class="fd-message-page__container">
+            <div class="fd-message-page__icon-container">
+                <span class="sap-icon sap-icon--filter fd-message-page__icon"></span>
+            </div>
+            <div class="fd-message-page__content">
```
meganaconley, line 39: We could use a `role="status"` and `aria-live="polite"` on this message so the message is announced to the a screen reader user when it appears.

https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA22.html
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions

InnaAtanasova: Thanks @meganaconley I added the role and aria-live to `<div class="fd-message-page__content">`
Thank you for providing the links.

src/message-page.scss
```diff
@@ -0,0 +1,88 @@
...
+
+    width: 100%;
+    height: 8rem;
+  }
+
+  &__icon {
+    font-size: 6rem;
+    color: var(--sapContent_NonInteractiveIconColor);
+    opacity: 50%;
+  }
```
JKMarkowski, line 41: Could you make stronger selector, since it's going to be used in with icons, that have reset mixin?

JKMarkowski (commented):
Hello @InnaAtanasova , there is only one question about some minor thing. Rest LGTM

src/message-page.scss
```diff
@@ -0,0 +1,86 @@
...
+
+  &__container,
+  &__content,
+  &__title,
+  &__subtitle,
+  &__actions,
+  &__more {
+    @include fd-reset();
+
+    max-width: 30rem;
```
salarenko, line 25: Setting up max-width on the container wouldn't be enough?

salarenko (commented):
Looks good, two minor concerns:

src/message-page.scss
InnaAtanasova: nope :) it's for IE11

src/message-page.scss
```diff
@@ -0,0 +1,86 @@
+@import "./new-settings";
+@import "./mixins";
+
+$fd-message-page: 0 !default;
```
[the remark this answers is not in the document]
InnaAtanasova: I removed it :) I forgot to delete this, it's not used anywhere, I just created it as a placeholder in case I need variables

src/message-page.scss
InnaAtanasova: Done

PR #1403, opened 2020-08-19
chore: Improve storybook-prep.js code

salarenko:
## Description
This is a minor improvement of `sorybook-prep.js` code readability. If you are wondering about performance of calling `readFileSync` as a template literal, it doesn't have any significant negative impact (after rework code is the same/a bit faster than previous version).


salarenko:
> This looks cleaner, but if you click the deploy preview there's no longer an introduction page.

@jbadan Yup, sorry some rookie React mistake (and I was wondering why did you add new lines after component tags 😛). Should be fine now.

PR #1405, opened 2020-08-19
fix: Add additional spacing to buttons as a pseudo element

JKMarkowski:
## Description
There is added additional spacing for buttons. . So now clickable area is expanded with invisible indicators.
Pseudo elements are absolute positioned.

### DIFF
--- a/src/button.scss
+++ b/src/button.scss
@@ -27,8 +27,8 @@ $fd-button-compact-padding-x: calc(0.5rem - #{$fd-button-border-width});
 $fd-button-compact-min-width: 2rem;
 $fd-button-compact-line-height: calc(#{$fd-button-compact-height} - 2 * #{$fd-button-border-width});
 
-$fd-button-additional-spacing: 0.25rem;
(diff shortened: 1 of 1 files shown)


src/button.scss
```diff
@@ -89,6 +105,11 @@ $block: #{$fd-namespace}-button;
   padding-left: $fd-button-compact-padding-x;
   padding-right: $fd-button-compact-padding-x;
   line-height: $fd-button-compact-line-height;
+
+  &::before {
+    top: -($fd-button-compact-additional-spacing + 0.0625rem);
+    bottom: -($fd-button-compact-additional-spacing + 0.0625rem);
```
InnaAtanasova, line 111: I think this 0.0625rem can be a variable as well :D

InnaAtanasova:
Check my comment and you can merge. Looks good

src/button.scss
JKMarkowski: Variable name has been changed :)

PR #1406, opened 2020-08-19
fix: button move focus indicator

JKMarkowski:
## Description
There is moved button focus indicator. From outline to ::after pseudo element with proper offset.

### DIFF
--- a/src/button.scss
+++ b/src/button.scss
@@ -63,6 +63,7 @@ $block: #{$fd-namespace}-button;
   min-width: $fd-button-min-width;
   overflow: hidden;
   text-overflow: ellipsis;
+  position: relative;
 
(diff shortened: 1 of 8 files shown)


InnaAtanasova:
@salarenko can you check in IE? I checked the other browsers and it looks ok. Thanks

InnaAtanasova:
I know it's Gracjan solution but why do we have so many mixins doing the same thing? Can we have 1 mixin for the focus and reuse it in all types?

JKMarkowski:
> I know it's Gracjan solution but why do we have so many mixins doing the same thing? Can we have 1 mixin for the focus and reuse it in all types?

LEt's address it in other PR

Issue #1368, opened 2020-08-13
Object Number


droshev:

olshansky:
Took to implementation.
_____
Fiori Design Guideline - https://experience.sap.com/fiori-design-web/object-display-elements/#object-number 
UI5 example - https://sapui5.hana.ondemand.com/#/entity/sap.m.ObjectNumber/sample/sap.m.sample.ObjectNumber

PR #1419, opened 2020-08-20
feat: Object Number new component

olshansky:
## Related Issue
Closes SAP/fundamental-styles#1368

## Description
Introducing new Object Number component

### After:
Screenshot 2020-08-21 at 15 05 32


src/object-number.scss
```diff
@@ -0,0 +1,63 @@
...
+
+  @include fd-reset();
+
+  font-family: $fd-object-number-font-family;
+  color: $fd-object-number-neutral;
+  max-width: inherit;
+
+  &__text,
+  &__unit {
+    max-width: inherit;
```
JKMarkowski, line 34: Let's change max-width to 100%, because of overflow/truncation

src/object-number.scss
```diff
@@ -0,0 +1,63 @@
...
+  $fd-object-number-neutral: var(--sapNeutralTextColor);
+
+  // LARGE DESIGN VARIABLES
+  $fd-object-number-font-size-large: 1.5rem;
+
+  @include fd-reset();
+
+  font-family: $fd-object-number-font-family;
+  color: $fd-object-number-neutral;
+  max-width: inherit;
```
JKMarkowski, line 30: Let's change max-width to 100%, because of overflow/truncation

JKMarkowski (changes requested):
Check my comments regarding overflow. 
Also could you create example with bold numbers? I bet you can use `<b>` instead of spans there

stories/object-number/object-number.stories.js
```diff
@@ -0,0 +1,57 @@
+import '../../dist/object-number.css';
+import './custom.scss';
+
+export default {
+    title: 'Components/Object Number',
+    parameters: {
+        description: `The Object Number displays number and number unit properties for an object.
+        The number can be displayed using semantic colors to provide additional meaning about the object to the user.`
+    },
+    tags: ['f3', 'a11y', 'theme']
```
jbadan, line 10: @meganaconley @prsdthkr @jacobdevera @Mike-Diaz Should this qualify for `a11y` tags if it's explicitly saying to use only colors to convey additional meaning? 

What else could we do here to convey the semantic states?

meganaconley: Yes, either adding a visible message or we could consider a screen reader message with additional information. Although that doesn't help people who are colorblind, so ideally we would have something else visual such as the icon.

src/object-number.scss
```diff
@@ -0,0 +1,63 @@
...
+  $fd-object-number-text-font-size: var(--sapFontMediumSize) !default;
+  $fd-object-number-unit-font-size: var(--sapFontSmallSize) !default;
+  $fd-object-number-neutral: var(--sapNeutralTextColor);
+
+  // LARGE DESIGN VARIABLES
+  $fd-object-number-font-size-large: 1.5rem;
+
+  @include fd-reset();
+
+  font-family: $fd-object-number-font-family;
```
InnaAtanasova, line 28: You don't need this line of code. `fd-reset` mixin is bringing it.

src/object-number.scss
```diff
@@ -0,0 +1,63 @@
...
+  "positive": ("color": var(--sapPositiveTextColor)),
+  "critical": ("color": var(--sapCriticalTextColor)),
+  "negative": ("color": var(--sapNegativeTextColor)),
+  "informative": ("color": var(--sapInformativeTextColor))
+);
+
+.#{$block} {
+  // TEXT VARIABLES
+  $fd-object-number-font-family: var(--sapFontFamily) !default;
+  $fd-object-number-text-font-size: var(--sapFontMediumSize) !default;
```
InnaAtanasova, line 19: These variable does not exists in sap-theming. Please check `css_variables.css` file from `node_modules/@sap-theming/theming-base-content/content/Base/baseLib/sap_fiori_3/css_variables.css`. If you need medium size you can use `--sapFontSize` corresponding to 0.875rem

src/object-number.scss
```diff
@@ -0,0 +1,63 @@
...
+$color-states: (
+  "positive": ("color": var(--sapPositiveTextColor)),
+  "critical": ("color": var(--sapCriticalTextColor)),
+  "negative": ("color": var(--sapNegativeTextColor)),
+  "informative": ("color": var(--sapInformativeTextColor))
+);
+
+.#{$block} {
+  // TEXT VARIABLES
+  $fd-object-number-font-family: var(--sapFontFamily) !default;
```
InnaAtanasova, line 18: This is not needed, check my comment bellow

src/object-number.scss
```diff
@@ -0,0 +1,63 @@
...
+  $fd-object-number-font-size-large: 1.5rem;
+
+  @include fd-reset();
+
+  font-family: $fd-object-number-font-family;
+  color: $fd-object-number-neutral;
+  max-width: inherit;
+
+  &__text,
+  &__unit {
```
InnaAtanasova, line 33: missing `fd-reset`

src/object-number.scss
```diff
@@ -0,0 +1,63 @@
...
+    max-width: inherit;
+    display: inline-block;
+    vertical-align: middle;
+    padding-top: 0.125rem;
+
+    @include fd-ellipsis();
+  }
+
+  &__text {
```
InnaAtanasova, line 42: The number is available in both regular and bold weights, so in this case you are missing a modifier class that will set the weight to bold.

src/object-number.scss
```diff
@@ -0,0 +1,63 @@
...
+    font-size: $fd-object-number-text-font-size;
+  }
+
+  &__unit {
+    font-size: $fd-object-number-unit-font-size;
```
InnaAtanasova, line 47: The unit text has different font-family:  `font-family: var(--sapFontLightFamily);`

src/object-number.scss
```diff
@@ -0,0 +1,63 @@
...
+    &--#{$set-name} {
+      color: map_get($color-set, "color");
+    }
+  }
+
+  // LARGE DESIGN
+  &--large {
+    .#{$block}__text,
+    .#{$block}__unit {
+      font-size: $fd-object-number-font-size-large;
```
InnaAtanasova, line 60: In large design the font-family is also light. Use `font-family: var(--sapFontLightFamily);`

stories/object-number/object-number.stories.js
```diff
@@ -0,0 +1,57 @@
...
+</span>
+`;
+
+/**
+ * Object Number can be large (class: `fd-object-number--large`)
+ */
+
+export const large = () => `
+<span class="fd-object-number fd-object-number--large fd-object-number--negative">
+    <span class="fd-object-number__text">-2,000.00</span>&nbsp;<span class="fd-object-number__unit">EUR</span>
```
InnaAtanasova, line 43: The empty space can be rendered with a before attribute of the unit (`<span class="fd-object-number__unit">`). This way the user won't have to manually put it every time.

stories/object-number/object-number.stories.js
```diff
@@ -0,0 +1,57 @@
...
+    tags: ['f3', 'a11y', 'theme']
+};
+
+/**
+ * Object Number can display the semantic status of an object: negative (class: `fd-object-number--negative`), critical
+ * (class: `fd-object-number--critical`), positive (class: `fd-object-number--positive`),
+ * informative (class: `fd-object-number--informative`), or none.
+ */
+
+export const primary = () => `
```
InnaAtanasova, line 19: Missing example with bold text

src/object-number.scss
stefanoScalzo: Screen Shot 2020-08-24 at 9 59 12 AM
Different font families

jbadan:
@droshev Thank you for following up! That answer feels like a little bit of a cop out 😆  

I'm think for the semantic examples, we need to show what an accessible version would actually look like - meaning including the label, icon or attribute. Do the designers have any mockups of what the label/icon/etc would look like? Can we use existing components to make this happen?

src/object-number.scss
JKMarkowski: Honestly I doubt about this. Colors are set below, so shouldn't be overwritten by fd-reset mixin there.

src/object-number.scss
olshansky: @stefanoScalzo thanks! Could you please re-attach pdf-file with correct spec [here](https://app.gitbook.com/@fundamental-styles/s/docs/new-component-object-number)?

InnaAtanasova: @olshansky I updated the pdf in GitBook

src/object-number.scss
InnaAtanasova: These elements need a reset. The colors should be defined after the reset.

src/object-number.scss
```diff
@@ -0,0 +1,71 @@
...
+  color: $fd-object-number-neutral;
+  max-width: 100%;
+  display: inline-block;
+
+  &__text,
+  &__unit {
+    @include fd-reset();
+    @include fd-ellipsis();
+
+    font-size: $fd-object-number-font-size;
```
InnaAtanasova, line 35: fd-reset mixin is bringing `font-size: var(--sapFontSize);` so you don't need this

src/object-number.scss
```diff
@@ -0,0 +1,71 @@
...
+  "positive": ("color": var(--sapPositiveTextColor)),
+  "critical": ("color": var(--sapCriticalTextColor)),
+  "negative": ("color": var(--sapNegativeTextColor)),
+  "informative": ("color": var(--sapInformativeTextColor))
+);
+
+.#{$block} {
+  // TEXT VARIABLES
+  $fd-object-number-font-size: var(--sapFontSize) !default;
+  $fd-object-number-neutral: var(--sapNeutralTextColor);
```
InnaAtanasova, line 19: this doesn't make sense to be a variable since it's used in only one place in the code.

src/object-number.scss
```diff
@@ -0,0 +1,71 @@
...
+
+  // LARGE DESIGN VARIABLES
+  $fd-object-number-font-size-large: 1.5rem;
+
+  @include fd-reset();
+
+  color: $fd-object-number-neutral;
```
InnaAtanasova, line 26: color: var(--sapNeutralTextColor);

stories/object-number/object-number.stories.js
```diff
@@ -0,0 +1,79 @@
+import '../../dist/object-number.css';
+import './custom.scss';
+
+export default {
+    title: 'Components/Object Number',
+    parameters: {
+        description: `The Object Number displays number and number unit properties for an object.
```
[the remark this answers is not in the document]
olshansky: @jessmetivier should this description be in the documentation page instead of mine `The Object Number displays number and number unit properties for an object.......`?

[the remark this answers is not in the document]
olshansky: @droshev @InnaAtanasova @JKMarkowski @prsdthkr how do you think, how I could show semantic and a11y aspects in a correct and understandable way? Maybe by adding a separate demo section with an example as @prsdthkr add [here](https://github.com/SAP/fundamental-styles/pull/1419#discussion_r474139978) _(object number with an icon and description text)_?

JKMarkowski (approved):
LGTM

jbadan:
@JKMarkowski @InnaAtanasova Doesn't look like we've addressed the a11y issues with the component in general. Were you wanting to go forward with just a disclaimer? Don't we want to build out the examples?

stories/object-number/object-number.stories.js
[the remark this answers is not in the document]
jessmetivier: Hi @olshansky, I'm not sure if I understand your question... are you asking me whether the description I wrote should replace the one that's already there?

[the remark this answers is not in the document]
droshev: @olshansky - maybe we can remove the accessibility tag for now and open an issue (for future) to create an example that demonstrate the a11y part.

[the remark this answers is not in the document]
olshansky: @jessmetivier yes, exactly about that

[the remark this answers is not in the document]
olshansky: @droshev ok, thanks. I've removed a11y tag already. Will create an issue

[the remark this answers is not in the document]
jessmetivier: @olshansky If that's the case, then yes. This description should replace it.

jessmetivier (approved):

src/object-number.scss
```diff
@@ -0,0 +1,66 @@
...
+  // LARGE DESIGN VARIABLES
+  $fd-object-number-font-size-large: 1.5rem;
+
+  @include fd-reset();
+
+  color: var(--sapNeutralTextColor);
+  max-width: 100%;
+  display: inline-block;
+
+  &__text,
```
jacobdevera, line 26: I believe it would be appropriate to include the `text-shadow` change in cases of low-contrast as is done for the `Object Status` component in  #1571 

```suggestion
  display: inline-block;
  text-shadow: var(--sapContent_TextShadow);

  &__text,
```

Issue #1384, opened 2020-08-14
Create Component Carousel


droshev:
Create Carousel excluding the Message Page as it is tackled separately

PR #1429, opened 2020-08-24
feat: new Fiori 3 component Carousel

InnaAtanasova:
## Related Issue
Closes SAP/fundamental-styles#1384

## Description
Introduce a new Fiori 3 component - Carousel


stories/carousel/carousel.stories.js
```diff
@@ -0,0 +1,321 @@
...
+                <li 
+                    data-slide-to="6" 
+                    aria-label="Displaying item 6 of 7" 
+                    class="fd-carousel__page-indicator"></li>
+                <li 
+                    data-slide-to="7" 
+                    aria-label="Displaying item 7 of 7" 
+                    class="fd-carousel__page-indicator"></li>
+            </ol>
+            <button
```
meganaconley, line 81: All of the buttons need aria-labels.

stories/carousel/carousel.stories.js
```diff
@@ -0,0 +1,321 @@
...
+        </div>
+        <div class="fd-carousel__content" style="min-height: 15.5rem;"></div>
+    </div>
+    <h4>Navigation Buttons in Content</h4>
+    <div class="fd-carousel" data-ride="carousel" style="margin-bottom: 3rem; max-width: 30rem;">
+        <div class="fd-carousel__page-indicator-container">
+            <ol class="fd-carousel__page-indicators">
+                <li 
+                    data-slide-to="1" 
+                    aria-label="Displaying item 1 of 7"
```
meganaconley, line 226: I don't think the aria-label here is necessary since it isn't focusable. Instead, we might want to include an aria-live container (just one per carousel) that updates as the pages are flipped to say "Displaying item x of y".

src/carousel.scss
```diff
@@ -0,0 +1,155 @@
...
+
+  &__button.#{$button} {
+    @include fd-flex-center();
+    @include set-width($fd-carousel-button-size);
+    @include set-height($fd-carousel-button-size);
+
+    border-radius: 50%;
+    padding: 0;
+    box-shadow: none;
+    margin: 0.25rem;
```
JKMarkowski, line 132: Maybe let's add additional clickable area as additional spacing for buttons. It's done in #1305  button refactor as pseudo element.

InnaAtanasova: Yes, that's why I added the margin now so the spacing is correct. Once we have the refactored button the margin will be removed, the spacing will be correct and we will have the clickable area.

src/carousel.scss
```diff
@@ -0,0 +1,155 @@
...
+
+    flex: 1 1 auto;
+    width: 100%;
+    background: var(--sapBackgroundColor);
+    position: relative;
+
+    .#{$block}__button.#{$button} {
+      box-shadow: var(--sapContent_Shadow1);
+      position: absolute;
+      top: calc(50% - 1.0625rem);
```
salarenko, line 33: What is `1.0625rem`? Is this half of the button size?
We can also position it like that to get rid of this value:
```
      top: calc(50%);
      transform: translateX(-50%);
```

src/carousel.scss
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    width: 9rem;
+    padding: 0.5rem;
+  }
+
+  &__page-indicator {
+    @include fd-reset();
+
+    list-style-type: none;
```
salarenko, line 101: Don't we have some mixin for reseting list styles?

src/carousel.scss
```diff
@@ -0,0 +1,155 @@
...
+        transform: rotate(180deg);
+      }
+    }
+
+    @include fd-focus() {
+      outline-offset: 0;
+    }
+  }
+
+  &--touch-device {
```
salarenko, line 150: Maybe lets name it according to functionality instead of the target (`&--no-nvigation`)? This won't make any difference in current implementation but will be less confusing if some library decides to go against Fiori recommendations and:
- display buttons on mobile device
- hide buttons on desktop so user can't interact with the carousel

salarenko (changes requested):
Looks good, just minor reflections:
- Could you add an example for touch-devices?

src/carousel.scss
```diff
@@ -0,0 +1,155 @@
...
+    background: var(--sapBackgroundColor);
+    position: relative;
+
+    .#{$block}__button.#{$button} {
+      box-shadow: var(--sapContent_Shadow1);
+      position: absolute;
+      top: calc(50% - 1.0625rem);
+      margin: 0;
+
+      &.sap-icon--slim-arrow-left {
```
salarenko, line 36: We should not rely on icon class names. This will prevent users from changing the navigation icons.

src/carousel.scss
InnaAtanasova: I can't find such mixin.

InnaAtanasova:
> Looks good, just minor reflections:
> 
> * Could you add an example for touch-devices?

It makes no sense to add an example for touch devices - would be just a carousel with no buttons. The class that needs to be used is specified in the documentation. In the implementation libraries where we will have the Carousel functionality we will show the example

src/carousel.scss
InnaAtanasova: Screen Shot 2020-08-26 at 9 38 52 AM

Fiori 3 specifies the icons that should be used in the navigation buttons so if someone wants to change them it will be against the guidelines. I am using the icons names to specify the horizontal offset of the buttons when used in content. We can't rely on first-child or last-child because they are using position: absolute. If we are not using the icon names we have to add additional classnames like `fd-carousel__button--left` and `fd-carousel__button--right`. Also these classnames will be used only when the buttons are inside the content, not when they are inside the Page indicator. So adding classes will mean that the buttons will either have different markup based on where they are used or the classnames will be redundant and do nothing when they are used in the Page Indicator. All these problems are solved with using icon names. Let me know what you think about it @salarenko @droshev @stefanoScalzo @katekozlowska @mikerodonnell89

katekozlowska:
All looks good to me and meets the spec.

src/carousel.scss
InnaAtanasova: Changed to `--no-navigation` and added examples in the docs

src/carousel.scss
salarenko: Yeah, but we can just always apply `fd-carousel__button--left` and `fd-carousel__button--right` and "conditionally" apply styles depending whether the buttons are placed inside `fd-carousel__content` or not, exactly at it is happening now. But I'm also curious what the others think.

salarenko (approved):

stories/carousel/carousel.stories.js
```diff
@@ -0,0 +1,480 @@
...
+        Displaying item 1 of 4
+    </div>
+</div>
+`;
+
+carouselBottom.storyName = 'Carousel with Page Indicator on Bottom';
+carouselBottom.parameters = {
+    docs: {
+        iframeHeight: 900,
+        storyDescription: `
```
jessmetivier, line 198: "The bottom page indicator carousel displays a content area, navigation buttons and a page indicator at the bottom. The page indicator wraps onto the next line if the content is too long, and after the pages reach a count of 8, the dots are replaced by numerical representation i.e. _1 of 9_."

stories/carousel/carousel.stories.js
```diff
@@ -0,0 +1,480 @@
...
+            </div>
+        </div>
+    </div>
+    <div style="display: none;" role="region" id="carousel-8" aria-live="polite">
+        Displaying item 1 of 4
+    </div>
+</div>
+`;
+
+carouselTop.storyName = 'Carousel with Page Indicator on Top';
```
jessmetivier, line 364: "The top page indicator carousel displays a content area, navigation buttons and a page indicator at the top."

stories/carousel/carousel.stories.js
```diff
@@ -0,0 +1,480 @@
...
+        </div>
+    </div>
+    <div style="display: none;" role="region" id="carousel-6" aria-live="polite">
+        Displaying item 1 of 4
+    </div>
+
+</div>
+`;
+
+carouselNoNavigation.storyName = 'Hiding the Navigation Buttons';
```
jessmetivier, line 473: "Carousel can also be displayed without navigation buttons. To hide them, add the` fd-carousel--no-navigation` modifier class to the `fd-carousel` class. On touchable devices, the user can navigate with a swipe gesture."

stefanoScalzo:
Why use ordered list and not undordered list?

InnaAtanasova:
> Why use ordered list and not undordered list?

Because it's an ordered list. Items go 1, 2, 3, 4, etc. Please check https://www.w3schools.com/

stories/carousel/carousel.stories.js
```diff
@@ -0,0 +1,480 @@
...
+- **Paging**- the current position inside the set of items is displayed using an optional paging indicator. 
+
+## Empty State
+
+When no pages are loaded, the Carousel displays a MessagePage component with <code class="docs-code">sap-icon--document</code> icon.
+
+
+## Accessibility
+
+The div element with class <code class="docs-code">fd-carousel</code> has a visually hidden sibling div element with <code class="docs-code">role="region"</code> and <code class="docs-code">aria-live="polite"</code>. Its text should be dynamically updated using JavaScript so that content changes of the Carousel are announced to the user.
```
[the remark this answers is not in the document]
InnaAtanasova: Yes, a screen reader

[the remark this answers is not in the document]
jessmetivier: Okay, then this is what I would add at the end: "This way, the text will be dynamically updated using Javascript so that carousel's content changes can be announced to the user by screen readers."

katekozlowska (approved):

PR #1529, opened 2020-08-28
docs: updated info label documentation

jessmetivier:
## Description
{{ _Updating **Info Label** docs_ }}


stories/info-label/info-label.stories.js
```diff
@@ -80,4 +108,11 @@ export const rtl = () => `
     <span class="fd-info-label fd-info-label--accent-color-2 fd-info-label--icon sap-icon--upload-to-cloud"></span>
 </div>
 `;
-rtl.storyName = 'Info Label RTL';
+rtl.storyName = 'RTL';
+rtl.parameters = {
+    docs: {
+        iframeHeight: 500,
+        storyDescription: `The info label can be displayed from right to left on the screen.
```
InnaAtanasova, line 115: @mikerodonnell89 @droshev do you think it's a good idea to add a description here or keep it just RTL in the title?

stories/info-label/info-label.stories.js
```diff
@@ -44,7 +52,14 @@ export const colors = () => `
     <span class="fd-info-label fd-info-label--accent-color-10">Info Label</span>
 </div>
 `;
-colors.storyName = 'Color Flavors';
+colors.storyName = 'Colors';
+colors.parameters = {
+    docs: {
+        iframeHeight: 500,
+        storyDescription: `Info label can be displayed in various colors, by adding the <code>fd-info-label--accent-color</code> modifier class to the element.
+`
```
InnaAtanasova, line 60: I think we need to have this as `<code>fd-info-label--accent-color-*</code>` and specify the options are numbers from 1 to 10. 
Someone can add just the modifier class <code>fd-info-label--accent-color</code> and it won't apply the color since you need to specify the number of the color. The `-*` part usually means a placeholder for a number or letter. @mikerodonnell89 what do you think about it? Does it make sense to add the `-*` part

InnaAtanasova: Maybe use the same naming convention as for {icon-name} ?

jessmetivier: Ah, you're right! I actually meant to put -* there!

salarenko (approved):

stories/info-label/info-label.stories.js
salarenko: For me the description is fine.

jessmetivier: What do you think would be informative to a developer audience?

mikerodonnell89: I say yes to a description but mention that it's typically used for internationalization purposes

jessmetivier: Done.

InnaAtanasova (approved):

PR #1541, opened 2020-08-31
fix: update navigation buttons of Carousel

InnaAtanasova:
## Description
The Button markup has changes so the Navigation buttons of the Carousel have to be updated with the latest markup. Also, additional modifier classes are introduced for the left and right button to place the buttons inside the content according to the specs.
- update the markup of the buttons to use `<i class="sap-icon--slim-arrow-left"></i>` and `<i class="sap-icon--slim-arrow-right"></i>`
- introduce additional modifier classes - `fd-carousel__button--left` and `fd-carousel__button--right`

### DIFF
--- a/src/carousel.scss
+++ b/src/carousel.scss
@@ -7,6 +7,8 @@ $button: #{$fd-namespace}-button;
 .#{$block} {
   $fd-carousel-page-indicator-border: 0.0625rem solid var(--sapPageFooter_BorderColor) !default;
   $fd-carousel-button-size: 2.125rem !default;
+  $fd-carousel-touchable-button-size: 2.625rem !default;
+  $fd-carousel-button-focus-outline-offset: -0.125rem !default;
   $fd-carousel-button-content-offset: 0.5rem !default;
   $fd-carousel-dot-size: 0.25rem !default;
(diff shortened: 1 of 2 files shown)


JKMarkowski (approved):
LGTM

src/carousel.scss
```diff
@@ -132,19 +134,26 @@ $button: #{$fd-namespace}-button;
...
     }
 
     @include fd-focus() {
-      outline-offset: 0;
+      &::after {
+        top: $fd-carousel-button-focus-outline-offset;
+        bottom: $fd-carousel-button-focus-outline-offset;
+        left: $fd-carousel-button-focus-outline-offset;
+        right: $fd-carousel-button-focus-outline-offset;
+      }
```
JKMarkowski, line 156: Could you change offset to `-3px (in rem ofc)`? Because now it looks like that 
image

JKMarkowski (changes requested):

src/carousel.scss
InnaAtanasova: With 0 it looks like this:
Screen Shot 2020-09-01 at 9 16 58 AM

Changed it to -0.125rem

JKMarkowski (approved):
LGTM

PR #1571, opened 2020-09-02
fix: add text-shadow for better contrast

jacobdevera:
## Description
Semantic text and links fails color contrast ratio requirements in Table when a row is selected. This adds a `text-shadow` to table cells to correct this, except in the case of "Critical" semantic text, which is a known deviation. I have also included it for non-inverted links and object statuses in case they are displayed on low-contrast backgrounds.

The change is unnoticeable for most text, and produces a subtle shadow for low-contrast text.

### Before:
text-shadow-before-2
text-shadow-before

### After:
text-shadow-after-2
text-shadow-after

### Reference
https://webaim.org/articles/contrast/#outline


droshev (approved):

droshev:
@jacobdevera can you re-generate the storybook screenshots for those components?

InnaAtanasova (approved):

PR #1578, opened 2020-09-03
feat: new component Action Sheet

JKMarkowski:
## Related Issue
Closes https://github.com/SAP/fundamental-styles/issues/1577

## Description
There is new component called Action Sheet. 

Also to achieve the look I had to add
- new modifier to button, called `--full-width`, to keep width of container.

### After:
Desktop:
image

Tablet:
image

Mobile:
image

### DIFF
--- a/src/action-sheet.scss
+++ b/src/action-sheet.scss
@@ -55,6 +55,7 @@ $block: #{$fd-namespace}-action-sheet;
     @include fd-typography(var(--sapFontHeader6Size));
     @include fd-ellipsis();
 
+    z-index: 1;
     margin: 0 1rem 0.5rem 1rem;
     color: var(--sapContent_ContrastTextColor);
   }
@@ -65,5 +66,17 @@ $block: #{$fd-namespace}-action-sheet;
     .#{$block}__item {
       padding: 0.1875rem 0;
     }
+  }
+
+  &--mobile {
+    background-color: $fd-popover-background-color;
+    z-index: 1;
+    box-shadow: $fd-popover-box-shadow-no-arrow;
+    width: $fd-popover-full-filled-width;
+    position: relative;
+    overflow: auto;
+
+    @include fd-left-popover-offset-placement();
+    @include fd-popover-border-radius();
   }
 }
--- a/src/action-sheet.scss
+++ b/src/action-sheet.scss
@@ -6,7 +6,7 @@ $block: #{$fd-namespace}-action-sheet;
 .#{$block} {
 
   // OVERLAY
-  $fd-action-sheet-overlay-color: rgba(0, 0, 0, 0.6); // Should be fade(var(--sapBlockLayer_Background), 60 or 80 (HCB|HCW))
+  $fd-action-sheet-overlay-color: var(--sapBlockLayer_Background);
 
   // SPACING
   $fd-action-sheet-mobile-padding: 0.1875rem 0.375rem;
@@ -27,7 +27,6 @@ $block: #{$fd-namespace}-action-sheet;
 
   &__wrapper {
     position: absolute;
-    background: $fd-action-sheet-overlay-color;
     width: 100%;
     height: 100%;
     left: 0;
@@ -35,6 +34,13 @@ $block: #{$fd-namespace}-action-sheet;
     display: none;
 
     @include fd-popover-border-radius();
+
+    @include fd-pseudo-expand() {
+      opacity: var(--sapOverlay_Background_Opacity);
+      background-color: $fd-action-sheet-overlay-color;
+
+      @include fd-popover-border-radius();
+    }
 
     &--active {
       @include fd-flex('column') {
--- a/src/action-sheet.scss
+++ b/src/action-sheet.scss
@@ -8,9 +8,6 @@ $block: #{$fd-namespace}-action-sheet;
   // OVERLAY
   $fd-action-sheet-overlay-color: rgba(0, 0, 0, 0.6); // Should be fade(var(--sapBlockLayer_Background), 60 or 80 (HCB|HCW))
 
-  // BACKGROUND
-  $fd-action-sheet-background: var(--sapGroup_ContentBackground) !default;
-
   // SPACING
   $fd-action-sheet-mobile-padding: 0.1875rem 0.375rem;
   $fd-action-sheet-desktop-padding: 0.25rem 0.5rem;
@@ -18,7 +15,6 @@ $block: #{$fd-namespace}-action-sheet;
   @include fd-reset();
 
   padding: $fd-action-sheet-desktop-padding;
-  background-color: $fd-action-sheet-background;
   width: 100%;
 
   &__item {
--- a/src/action-sheet.scss
+++ b/src/action-sheet.scss
@@ -55,8 +55,9 @@ $block: #{$fd-namespace}-action-sheet;
     @include fd-typography(var(--sapFontHeader6Size));
     @include fd-ellipsis();
 
+    max-width: 100%;
     z-index: 1;
-    margin: 0 1rem 0.5rem 1rem;
+    padding: 0 1rem 0.5rem 1rem;
     color: var(--sapContent_ContrastTextColor);
   }
 
--- a/src/action-sheet.scss
+++ b/src/action-sheet.scss
@@ -6,10 +6,10 @@ $block: #{$fd-namespace}-action-sheet;
 .#{$block} {
 
   // OVERLAY
-  $fd-dialog-overlay-color: rgba(0, 0, 0, 0.6); // Should be fade(var(--sapBlockLayer_Background), 60 or 80 (HCB|HCW))
+  $fd-action-sheet-overlay-color: rgba(0, 0, 0, 0.6); // Should be fade(var(--sapBlockLayer_Background), 60 or 80 (HCB|HCW))
 
   // BACKGROUND
-  $fd-action-sheet-background: var(--sapGroup_ContentBackground) !default; // Should be fade(var(--sapBlockLayer_Background), 60 or 80 (HCB|HCW))
+  $fd-action-sheet-background: var(--sapGroup_ContentBackground) !default;
 
   // SPACING
   $fd-action-sheet-mobile-padding: 0.1875rem 0.375rem;
@@ -31,7 +31,7 @@ $block: #{$fd-namespace}-action-sheet;
 
   &__wrapper {
     position: absolute;
-    background: $fd-dialog-overlay-color;
+    background: $fd-action-sheet-overlay-color;
     width: 100%;
     height: 100%;
     left: 0;
@@ -46,11 +46,19 @@ $block: #{$fd-namespace}-action-sheet;
     }
   }
 
+  &__title {
+    @include fd-reset();
+    @include fd-typography(var(--sapFontHeader6Size));
+
+    margin: 0 1rem 0.5rem 1rem;
+    color: var(--sapContent_ContrastTextColor);
+  }
+
   &--compact {
     padding: $fd-action-sheet-mobile-padding;
 
     .#{$block}__item {
-      padding: 0 0.1875rem;
+      padding: 0.1875rem 0;
     }
   }
 }
--- a/src/action-sheet.scss
+++ b/src/action-sheet.scss
@@ -34,6 +34,8 @@ $block: #{$fd-namespace}-action-sheet;
     top: 0;
     display: none;
 
+    @include fd-popover-border-radius();
+
     &--active {
       @include fd-flex('column') {
         justify-content: flex-end;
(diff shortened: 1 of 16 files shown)


InnaAtanasova:
Great job! I have a few visual comments:
1. The corners of the popover body. I think you should add some rounded corners or handle the overflow 
Screen Shot 2020-09-03 at 11 07 58 AM

2. I think `Decide Later` is the same as the buttons above just without an icon.
Fundamentals:
Screen Shot 2020-09-03 at 11 08 19 AM

Fiori:
Screen Shot 2020-09-03 at 11 08 34 AM

3. Could you make the height of the iframes bigger?Thanks
Screen Shot 2020-09-03 at 11 04 09 AM

stories/action-sheet/action-sheet.stories.js
```diff
@@ -0,0 +1,198 @@
...
+            <li class="fd-action-sheet__item">
+                <button class="fd-button fd-button--full-width fd-button--compact fd-button--transparent fd-button--text-alignment-left">
+                    <i class="sap-icon--print"></i>
+                    <span class="fd-button__text">Print</span>
+                </button>
+            </li>
+            <li class="fd-action-sheet__item">
+                <button class="fd-button fd-button--full-width fd-button--compact fd-button--transparent fd-button--text-alignment-left">
+                    <i class="sap-icon--home"></i>
+                    <span class="fd-button__text">Go to landing page</span>
```
prsdthkr, line 67: `Go to landing page` should be a link, not a button. Can action sheets have links? If yes, we should add supporting styles for this:
```suggestion
<li class="fd-action-sheet__item">
    <a href="#" class="fd-link" tabindex="0"><i class="sap-icon--home"></i> Go to home</a>
</li>
```

If no, we should call this out and update the story.

src/action-sheet.scss
```diff
@@ -0,0 +1,60 @@
...
+  $fd-action-sheet-overlay-color: rgba(0, 0, 0, 0.6); // Should be fade(var(--sapBlockLayer_Background), 60 or 80 (HCB|HCW))
+
+  // SPACING
+  $fd-action-sheet-mobile-padding: 0.1875rem 0.375rem;
+  $fd-action-sheet-desktop-padding: 0.25rem 0.5rem;
+
+  @include fd-reset();
+
+  padding: $fd-action-sheet-desktop-padding;
+  width: 100%;
```
prsdthkr, line 18: Should we add max-width and overflow behavior for list items?
image

stories/action-sheet/action-sheet.stories.js
```diff
@@ -0,0 +1,198 @@
...
+<div class="fd-popover">
+    <div class="fd-popover__control">
+        <button class="fd-button fd-button--compact fd-button--transparent" 
+                aria-label="More Options" aria-controls="actionSheetDesktop" aria-expanded="false" 
+                aria-haspopup="true" onclick="onPopoverClick('actionSheetDesktop');">
+            <i class="sap-icon--settings"></i>
+        </button>
+    </div>
+    <div class="fd-popover__body" aria-hidden="true" id="actionSheetDesktop">
+        <ul class="fd-action-sheet fd-action-sheet--compact">
```
prsdthkr, line 45:
```suggestion
        <ul class="fd-action-sheet fd-action-sheet--compact" aria-label="List of contextual options">
```

prsdthkr (commented):
Nice work. Few suggestions.

droshev (changes requested):
@JKMarkowski do not forget the delta theming 
Screen Shot 2020-09-03 at 4 19 54 PM

JKMarkowski:
Thank You @InnaAtanasova , your comments has been addressed :)

src/action-sheet.scss
JKMarkowski: Hello @prsdthkr With current impementation of button it was impossible, I had to add `display: flex` property to `--full-width` modifier on buttons. Buttons allow to keep overflow, because of additional clickable area on `up/down`. So now it looks like that
image

stories/action-sheet/action-sheet.stories.js
JKMarkowski: Following specs only buttons can be added into this list. I will change example. Thanks for noticing

stories/action-sheet/action-sheet.stories.js
JKMarkowski: I completely forgot about arias and roles. Thanks for noticing

salarenko (changes requested):
- Missing ellipsis
image

- I think we should have some scroll behavior in the mobile version (currently list items are overflowing)

- There is a space between popover and screen edge:
image

- Are we sure we want to directly modify popover styling? I think we should rather write a modifier in `Action Sheet` to extend popover behavior.

- Please adjust the height of the iframe for first two examples.

src/action-sheet.scss
salarenko: This issue is rather coming directly from button component implementation and it should be fixed there.
image

JKMarkowski:
> * Missing ellipsis
>   image
> * I think we should have some scroll behavior in the mobile version (currently list items are overflowing)
> * There is a space between popover and screen edge:
>   image
> * Are we sure we want to directly modify popover styling? I think we should rather write a modifier in `Action Sheet` to extend popover behavior.
> * Please adjust the height of the iframe for first two examples.

Thanks for review :) Changes has been applied. So it's ready for second round of review

InnaAtanasova:
Found some minor things:
- Text for Cancel button should be centered:
Screen Shot 2020-09-09 at 11 12 53 AM

- Still missing `...`
Screen Shot 2020-09-09 at 11 35 24 AM

droshev (approved):

JKMarkowski:
Thank You @InnaAtanasova for comments. It has been addressed

InnaAtanasova (approved):

stories/action-sheet/action-sheet.stories.js
```diff
@@ -0,0 +1,200 @@
...
+                    <span class="fd-button__text">Decide Later</span>
+                </button>
+            </li>
+        </ul>
+    </div>
+</div>
+<div style="height: 300px"></div>
+`;
+
+actionSheetTablet.parameters = {
```
jessmetivier, line 136: (If above is makes sense)

For tablet screens, add the `id="actionSheetTablet"` to the `fd-popover__body` component.

stories/action-sheet/action-sheet.stories.js
```diff
@@ -0,0 +1,200 @@
...
+            </li>
+        </ul>
+    </div>
+</div>
+`;
+
+actionSheetPhone.parameters = {
+    docs: {
+        iframeHeight: 800,
+        storyDescription: 'On smartphones the action sheet is wrapped additionally by `fd-action-sheet__wrapper`.'
```
jessmetivier, line 198: To display the action sheet on mobile screens, it is additionally wrapped by `fd-action-sheet__wrapper` together with the `id="actionSheetPhone"`.

Let me know if it makes sense.

The only thing I would suggest about the code is to change "phone" to "mobile" because that's what we've called it so far in the docs.

jessmetivier (changes requested):
Hi Jedrzej!
Here are my comments, let me know if it is the ids that differentiate the components (versus the modifiers in the past) and that everything makes sense technically.

stories/action-sheet/action-sheet.stories.js
```diff
@@ -0,0 +1,200 @@
...
+                </button>
+            </li>
+        </ul>
+    </div>
+</div>
+<div style="height: 300px"></div>
+`;
+
+actionSheetDesktop.parameters = {
+    docs: {
```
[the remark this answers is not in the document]
JKMarkowski: IDs are only to identify popover. It's needed for method `onPopoverClick` to find proper element, that will have changed display mode. Fragment you linked is just to keep height of example's frame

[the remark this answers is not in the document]
JKMarkowski: As you can see above button has linked property `onclick="onPopoverClick('actionSheetTablet');"`. The `onPopoverClick` method takes id argument, which is needed to find element by `document.getElementById(id);` In most of cases we use random 8 chars

stories/action-sheet/action-sheet.stories.js
```diff
@@ -0,0 +1,200 @@
...
+**Do not use the action sheet if:**
+
+- The menu provides only one option. In this case, consider using a button instead.
+- You need to show a hierarchical menu. In this case, use the menu button instead.
+- Your users would benefit more from a split button, which offers an easily-accessible default action, with the option to include additional actions.
+
+## Responsiveness
+
+The action sheet is fully responsive. On smartphones, the actions are displayed as a list inside a dialog.
+On tablets and desktop devices, the actions are displayed in a popover.
```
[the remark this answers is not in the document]
JKMarkowski: I added reponsive tag

JKMarkowski:
Thanks for comments @jessmetivier. Those have been adressed

stories/action-sheet/action-sheet.stories.js
```diff
@@ -0,0 +1,200 @@
...
+
+**Use the action sheet if:**
+
+- You need an option that provides more than one action.
+- It is really important that the user stays in context on a phone.
+- You only have a small number of actions.
+
+**Do not use the action sheet if:**
+
+- The menu provides only one option. In this case, consider using a button instead.
```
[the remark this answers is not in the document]
JKMarkowski: I capitalized it and made it bold. I'm not sure if there is easy way to include links, I would need to investigate it and address this issue in other components as well, because we have a lot of appearance of external components. Maybe it can be skipped for now

stories/action-sheet/action-sheet.stories.js
```diff
@@ -0,0 +1,199 @@
...
+**Use the action sheet if:**
+
+- You need an option that provides more than one action.
+- It is really important that the user stays in context on a phone.
+- You only have a small number of actions.
+
+**Do not use the action sheet if:**
+
+- The menu provides only one option. In this case, consider using a **Button** instead.
+- You need to show a hierarchical menu. In this case, use the **MenuButton** instead.
```
jessmetivier, line 23: Button is the only one that should be emphasized, since it's the main component. However, button and split button are technically components, in the documentation they are considered component "types" / or rather button types. They don't need to be emphasized, you can type them out like you did before :)

"menu button" "split button"

stories/action-sheet/action-sheet.stories.js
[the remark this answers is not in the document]
jessmetivier: Got it. So what differentiates one component type to the next? (difference between Desktop and Tablet for instance?) Is it the `aria-controls`?

The reason why I ask is because we usually reference modifier classes in the docs, to give readers a quick hint on how they can accomplish a certain look/style. What would we tell our users about how to achieve the Desktop mode for this component?

stories/action-sheet/action-sheet.stories.js
jessmetivier: If the ID doesn't really help our readers and is only for us internally, this needs to be rewritten. If we could go over a call to discuss what to include here, feel free to ping me on Slack! Or if it's a short explanation, you can just answer the comment above :)

jessmetivier (commented):
Please see comments, and ping me on Slack for either a review/call. I'll do it first thing in the morning :)

stories/action-sheet/action-sheet.stories.js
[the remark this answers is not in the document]
JKMarkowski: So the only difference between those 2 usages (Desktop and Tablet) is that Desktop uses compact mode for container `fd-action-sheet--compact` and for buttons `fd-button--compact`

salarenko (approved):

stories/action-sheet/action-sheet.stories.js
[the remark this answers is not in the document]
jessmetivier: "The action sheet is displayed in desktop mode by adding the `fd-action-sheet--compact` modifier class to the container, and `fd-button--compact` to each button."

stories/action-sheet/action-sheet.stories.js
```diff
@@ -0,0 +1,199 @@
...
+<div style="height: 300px"></div>
+`;
+
+actionSheetTablet.parameters = {
+    docs: {
+        iframeHeight: 300
+    }
+};
+
+actionSheetTablet.storyName = 'Tablet';
```
jessmetivier, line 139: "Although the tablet mode is visually similar to desktop, it doesn't require any modifier classes. Therefore, the container and buttons appear slightly larger for better visibility."

stories/action-sheet/action-sheet.stories.js
jessmetivier: "To display the action sheet on mobile screens, the container requires a `fd-action-sheet--mobile` modifier class  and is additionally wrapped by `fd-action-sheet__wrapper`."

jessmetivier (changes requested):
Let me know if they make sense now!

JKMarkowski:
@jessmetivier It totally makes sense, thanks for your input, I updated descriptions :)

jessmetivier (changes requested):
Great!

PR #1581, opened 2020-09-03
feat: PDF viewer component

meganaconley:
## Description
Creates the PDF viewer component and fixes the lack of dialog sizes (as the PDF viewer requires a size L dialog).

### Before:

Screen Shot 2020-09-03 at 1 14 53 PM

### After:

Screen Shot 2020-09-03 at 1 13 50 PM

Screen Shot 2020-09-03 at 1 12 45 PM

Screen Shot 2020-09-03 at 1 12 53 PM

Screen Shot 2020-09-03 at 1 13 02 PM


src/dialog.scss
```diff
@@ -31,6 +31,8 @@ $button: #{$fd-namespace}-button;
   $fd-dialog-content-box-shadow: var(--sapContent_Shadow3);
   $fd-dialog-content-border-radius: var(--sapElement_BorderCornerRadius);
   $fd-dialog-content-padding-x: (def: 1rem, s: 1rem, m: 2rem, l: 2rem, xl: 3rem);
+  $fd-dialog-content-min-width-x: (def: 0, s: 0, m: 600px, l: 1024px, xl: 1440px);
+  $fd-dialog-content-max-width-x: (def: 599px, s: 599px, m: 1023px, l: 1439px, xl: 0);
```
meganaconley, line 35: This is causing some display issues in the docs - I tried to adjust them a bit but any further input on how to better display them would be appreciated.

stories/pdf-viewer/pdf-viewer.stories.js
```diff
@@ -0,0 +1,148 @@
...
+        </footer>
+    </div>
+</div>
+`;
+
+dialog.storyName = 'Dialog';
+dialog.parameters = {
+    docs: {
+        iframeHeight: 400,
+        storyDescription: 'When displayed in a dialog, download and close buttons are in the footer. The PDF Viewer dialog does not support resizing. For mobile, ensure the modal takes up the entire screen by using the <code>.fd-dialog__content--mobile</code> class.'
```
mikerodonnell89, line 69: Screen Shot 2020-09-04 at 12 24 01 PM

Do we need to escape the double underscores?  Not sure why it is showing up like this for me

mikerodonnell89 (changes requested):
How is this rule (from the specs) enforced? "If the content height is increased beyond the screen height, the dialog height cannot go beyond 4 rem from the top and bottom of the screen."

Also a tablet/smartphone example would be nice

JKMarkowski (commented):
Hello @meganaconley, I added one question regarding height, could you also change all `px` values to rem`?

stories/pdf-viewer/pdf-viewer.stories.js
meganaconley: I've removed this part of the description

src/pdf-viewer.scss
```diff
@@ -0,0 +1,30 @@
+@import "./new-settings";
+@import "./mixins";
+
+$block: #{$fd-namespace}-pdf-viewer;
+
+.#{$block} {
+  padding: 1rem;
+  min-height: 500px;
```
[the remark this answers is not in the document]
meganaconley: I've updated the `px` values to use `rem`. The reason there is a min-height set is that the iframe does not expand past the dialog's min-height unless I set an explicit value. I have been playing around with it for some time and haven't been able to get it to freely resize/take up the height up to the set maximum.

jacobdevera:
On the docs, the off-white background on the loading error doesn't fill up the entire space. Not sure if this should just be transparent, or use something other than `padding-top` to center the error.
Screen Shot 2020-09-09 at 4 45 09 PM

InnaAtanasova:
Screen Shot 2020-09-10 at 1 39 30 PM

Extra large dialog is broken

meganaconley:
@InnaAtanasova I've fixed the outline issue, however the docs container is just too small to contain the dialogs. I left a comment about this when I opened the PR, but I'm wondering if we need to adjust the dialog examples to include buttons which open them rather than display them within the iframes.

src/dialog.scss
salarenko, line 35: Where do this vaules come from?
I talked to the designer once, and they said there are no predefined sizes for the dialog. I just went through the specs and I also couldn't find any information.

salarenko (changes requested):

src/dialog.scss
meganaconley: Oh I saw it in the fd-styles documentation for dialog and didn't think to cross check it with the specs: https://sap.github.io/fundamental-styles/?path=/docs/components-dialog--default-dialog

If there's no defined sizes, why have size modifiers at all? The spec for the PDF viewer indicates a size `l` dialog as well. I can just set a width for the PDF viewer and revert these changes then.

stefanoScalzo:
The pdf viewers top area should be 2.5rem. Currently it is 2.75 + another 1 rem

stefanoScalzo:
> How is this rule (from the specs) enforced? "If the content height is increased beyond the screen height, the dialog height cannot go beyond 4 rem from the top and bottom of the screen."

This isn't enforced yet? @meganaconley

src/pdf-viewer.scss
```diff
@@ -0,0 +1,34 @@
+@import "./new-settings";
+@import "./mixins";
+
+$block: #{$fd-namespace}-pdf-viewer;
+
+.#{$block} {
+  padding: 1rem;
+
+  &__dialog {
+    max-height: calc(100vh - 4rem);
```
meganaconley, line 10: @stefanoScalzo The max height design spec you're referencing was fixed here.

src/pdf-viewer.scss
```diff
@@ -0,0 +1,34 @@
...
+    box-shadow: none;
+  }
+
+  &__content {
+    width: 100%;
+    min-height: 28rem;
+
+    &--in-page {
+      width: 100%;
+      min-height: 35rem;
```
JKMarkowski, line 26: I got not clue where `min-height` values are coming from, I think it should be better to specify them in inline styling, instead of putting them into library.

src/pdf-viewer.scss
JKMarkowski: `the dialog height cannot go beyond 4 rem from the top and bottom of the screen."` Shouldn't it be `-4rem` per side, so `-8rem`?

src/pdf-viewer.scss
meganaconley: They've been removed.

JKMarkowski (approved):
LGTM

stories/pdf-viewer/pdf-viewer.stories.js
```diff
@@ -0,0 +1,196 @@
...
+        <header class="fd-dialog__header fd-bar fd-bar--header">
+            <div class="fd-bar__left">
+                <div class="fd-bar__element">
+                    <h2 class="fd-title fd-title--h5" id="dialog-title-1">
+                        PDF Viewer title
+                    </h2>
+                </div>
+            </div>
+        </header>
+        <div class="fd-dialog__body">
```
prsdthkr, line 40: Since this can become scrollable, should we add `tabindex="0"` here?
```suggestion
        <div class="fd-dialog__body" tabindex="0">
```

image

prsdthkr (changes requested):

stories/pdf-viewer/pdf-viewer.stories.js
meganaconley: I removed the scrollbar instead and made a note that the heights should not allow the dialog to scroll. We'd like to avoid putting tabindex on non-interactive items as much as necessary and it'd be really annoying to have to tab out, scroll the dialog down, then tab back into the iframe to access the rest.

src/dialog.scss
meganaconley: @salarenko Any further input on this topic? The rest of the PR is ready to go. I can either remove these widths or remove the part of the docs that specifies these widths.
I checked the UX spec and I'm still not clear on what the sizes are supposed to refer to.

prsdthkr:
Is the viewer expected to work with IE 11? IE just asks if the doc should be downloaded!
IE-pdf-viewer

prsdthkr (approved):
The acrobat reader wasn't installed on my machine. It is working in IE 11 now:
IE-pdf-viewer-working

src/dialog.scss
salarenko: @meganaconley 
- I believe the section you are referring to in the Dialog documentation is describing what classes be used depending on the Dialog size, to add proper margins to the Dialog window. There is nothing said about "size modifiers".

- In Fiori Desgin Guideliness: "The PDF viewer control is fully responsive on large-screen devices (size L).". Again there is nothing said about the Dialog size, but rather about the size of the device screen.

I totally agree that Fiori Fundamentals should have something like predefined Dialog sizes, but there is nothing about it in the design, and the designers are not very helpful with this topic... Looking at the Fiori Guideliness I'd rather say that the Dialog should be opened in `width: 100; height: 100%`.
And you see, all of this are only assumptions and because of that I'm not sure if we should even have this component, because it has not been described in "Fiori VIsual Design".

src/_dialog-placeholders.scss
```diff
@@ -26,9 +26,7 @@ $fd-dialog-overlay-color: rgba(0, 0, 0, 0.6); // Should be fade(var(--sapBlockLa
   flex-direction: column;
   outline: none;
   position: absolute;
-  min-width: $fd-dialog-content-min-width;
```
salarenko, line 29: Why this has been removed?
image

stories/pdf-viewer/pdf-viewer.stories.js
```diff
@@ -0,0 +1,198 @@
...
+dialog.parameters = {
+    docs: {
+        iframeHeight: 500,
+        storyDescription: 'When displayed in a dialog, download and close buttons are in the footer. The PDF Viewer dialog does not support resizing. For mobile, ensure the modal takes up the entire screen. Refer to the documentation for Dialog.'
+    }
+};
+
+export const mobileDialog = () => `
+<div class="fd-dialog-docs-static fd-pdf-viewer fd-dialog fd-dialog--active">
+    <div class="fd-pdf-viewer__dialog fd-dialog__content fd-dialog__content--mobile" style="min-height: 28rem;" role="dialog" aria-modal="true" aria-labelledby="dialog-title-1">
```
salarenko, line 78: This is inconsistent naming. Dialog follows structure:
```
- Dialog
    - Dialog Content
       - Dialog Content Elements
```
and PDF Viewer which is build as a composition/extension of the Dialog i using `__dialog` on `__content` level. Please change the naming to avoid confusion.

salarenko (changes requested):
I'm sorry but I think this PR should not be accepted.
Fiori Design Guidelines should not be followed when it comes to styling parameters (paddings, margins etc.), the source for such parameters is Fiori Visual Guideline specification. In this example you can see the inconsistency between Design and Visual Guidelines, by checking the Dialog component:

- Design Guidelines:
image

- Visual Guidelines:
image

And now because PDF Viewer follows values values from Design Guidelines, you are enforcing some rem values to override proper behavior of the Dialog.

I think we can discuss creating this component without Visual Guidelines Specification because this is a very simple one, but current implementation should not be merged.

src/_dialog-placeholders.scss
meganaconley: This was part of my attempt to put sizing in to the Dialog because I thought that the min-width/max-width specified in the Dialog page sizing section were actually supposed to be set by the Dialog. It's unfortunate there's no actual sizing, since the iframe does not have dimensions of its own and simply fills its container.

I reverted any changes to the dialog and set the width inside the story manually so now both height and width of the iframe will be up to the consumer.

src/pdf-viewer.scss
```diff
@@ -0,0 +1,30 @@
...
+@import "./mixins";
+
+$block: #{$fd-namespace}-pdf-viewer;
+
+.#{$block} {
+  &__dialog {
+    max-height: calc(100vh - 8rem);
+  }
+
+  &__title {
```
salarenko, line 11: Dialog is using Bar by default (as specification says) and I think we should still use Bar component to create this layout.
Current Dialog Header/Sub-header implementation provides all necessary styling and blocks for building the PDF Viewer Header.

For sure there is no reason to create PDF Viewer specific styling to make Bar look like Toolbar in transparent mode.

src/pdf-viewer.scss
```diff
@@ -0,0 +1,30 @@
...
+
+  &__content {
+    width: 100%;
+
+    &--in-page {
+      width: 100%;
+    }
+
+    &--error {
+      background: none;
```
salarenko, line 26: Where does this styling come from?

salarenko (changes requested):

salarenko (changes requested):
Could you revert the changes for `src/_dialog-placeholders.scss`? If there is nothing changed, there is no point of creating another change-log for this file.

meganaconley:
Closing until there is a standard design.

Issue #1348, opened 2020-08-11
Update repo from CLA to DCO


droshev:
Update repo from CLA to DCO

PR #1583, opened 2020-09-03
chore: migrate from CLA to DCO

droshev:
Migrate Fundamental Styles from CLA to DCO

fixes: #1348


salarenko (changes requested):
- Could you give a bit more explanation why this change is needed?

- If we use github bot don't we need some script to append our commit messages?
image

salarenko (approved):

PR #1584, opened 2020-09-03
chore: move visual snapshot testing to Chromatic

prsdthkr:
## Related Issue
Closes SAP/fundamental-styles#1586 

## Description
In this change, we 
* remove local image snapshot testing infra (delete images snaps, tests, scripts)
* automate the creation of visual tests for each story
* use GitHub workflow to enable Chromatic snapshot testing and change approval process
* add toolbar add-on for theme switching in story canvas mode
* use formatted HTML story source snippet
* updated all storybook packages to v6.0.26

### Before:
NA

### After:
Theme switch toolbar:
image

Chromatic process:
image

Formatted HTML story source snippet:
image

- [NA] RTL support
3. Testing
- [NA] Updated tests
4. Documentation


jbadan:
@prsdthkr Sorry I didn't see this - I enabled the storyshots for the meantime - hopefully that doesn't mess with you! https://github.com/SAP/fundamental-styles/pull/1585

jbadan:
@stefanoScalzo you would need to `npm install` first. Regardless, we're trying to go down the continuum path at the moment so let's leave this in draft.

prsdthkr:
@jbadan @stefanoScalzo yeah we're in the process of setting-up [Chromatic](https://www.chromatic.com/) to solve this for us. This PR will be a draft until then.

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,15 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - run: |
+        yarn && yarn build
+    - uses: chromaui/action@v1
+      with: 
+        projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
+        token: ${{ secrets.GITHUB_TOKEN }}
+        buildScriptName: storybook:chromatic
```
jbadan, line 15: I know this is a draft so I'll go away but I was curious 🤣  

I'm so excited about this!!! Flee from travis!!!! 🏃‍♂️ 🏃 🏃‍♀️

jbadan: Possibly dumb question - will this work for forks? I can't see why not - the only thing I would be worried about is if the secret gets passed

jbadan:
WOW! Look at this build time:
Screen Shot 2020-10-14 at 10 57 15 AM

versus ~ 7 minutes before 🎉

.storybook/main.js
```diff
@@ -4,7 +4,20 @@ const { exec } = require("child_process");
...
+  const mode = process.env.STORYBOOK_ENV;
+  switch(mode){
+    case 'docs':
+      return '(stories)';
+    case 'chromatic':
+      return '(visual)';
+    default:
+      return '(stories|visual)';
+  }
+}
```
jbadan, line 18: so much cleaner 🧹

.storybook/preview.js
```diff
@@ -50,48 +52,83 @@ export const parameters = {
...
       page: DocsPage,
-      theme: fundamentals
+      theme: fundamentals,
+      transformSource: (src) => {
+        // we strip out the () =>` ` from the story
+        // so that the source can be formatted and
+        // rendered as HTML.
+        const SOURCE_REGEX = /^\(\) => `((.|\n)*)`$/;
+        const match = SOURCE_REGEX.exec(src);
+        return prettify(match ? match[1] : src);
```
jbadan, line 68: 🎉

.storybook/preview.js
```diff
@@ -50,48 +52,83 @@ export const parameters = {
...
+        let stylePath = `${component}-${context?.globals?.theme}.css`;
+        document.head.appendChild(styleLinkTag(stylePath));
+    })
+
+    toRemove.forEach(item => {
+      item.parentNode.removeChild(item);
+    })
+    document.head.append(styleLinkTag(`theming-base-content/content/Base/baseLib/${context?.globals?.theme}/css_variables.css`));
+
+    return storyFn(context);
```
jbadan, line 127: This is great! This is what I was trying to do when I added it to the docs page but couldn't figure it out 💯

.github/workflows/chromatic.yml
prsdthkr: Not at all a dumb question. You're correct. Yikes: https://www.chromatic.com/docs/ci#run-chromatic-on-external-forks-of-open-source-projects

Is it a good solution to expect fork authors to ask repo maintainers to trigger Chromatic 🤷‍♂️ ?

prsdthkr: Also, I think @droshev has set up a free open-source plan for us. If that's the case, we have unlimited snapshots, and sharing the token should be fine.

jbadan (approved):
Huge undertaking 💪  This looks amazing. You can tell how excited I am about this pr because of the 800 or so emojis I've used throughout this review 🤣 

Can we add `storybook-chromatic/` to gitignore? Only other question I had was about the github action with forks, but we can always just add `npm run storybook:chromatic` to travis if we run into problems there - which I hope we don't.

Can you update https://github.com/SAP/fundamental-styles/wiki/Visual-Testing?

PR #1595, opened 2020-09-10
fix: Remove reset from icons

JKMarkowski:
## Description
There is a need form NGX to remove `fd-reset()` mixin from icons. Due to different order of imports styles from reset are more important than other styles, which breaks most of components.

### DIFF
--- a/src/icon.scss
+++ b/src/icon.scss
@@ -3,10 +3,7 @@
 $block: sap-icon;
 
 [class*="#{$block}"] {
-  @include fd-reset();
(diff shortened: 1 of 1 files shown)


jbadan (changes requested):
Can we talk about this during the meeting or would you mind adding an example/additional context of why this is necessary to the description? We've gone down this path a couple times and we really should avoid removing resets when possible.

salarenko:
This solution has been discussed and abandoned.

Issue #1603, opened 2020-09-11
Change strategy of icon usage


JKMarkowski:
So currently our approach to use icons inside components is to put class with component's element.
Ex.
```
<span class="fd-avatar fd-avatar--accent-color-1 fd-avatar--m sap-icon--money-bills" role="presentation"></span>
```
It should be changed to 
```
<span class="fd-avatar fd-avatar--accent-color-1 fd-avatar--m" role="presentation">
    <i class="sap-icon--money-bills"></i>
</span>
```

List of components that are needed to be refactored:
- Avatar
- Generic Tile 
- Info Label
- List
- Menu
- Message Page
- Nested List
- Numeric Content
- Object Marker
- Object Status
- Product Switch
- Switch
- Table
- Tabs
- Tree

If Element contain text:
```
 <li role="listitem" tabindex="0" class="fd-list__item">
    <i role="presentation" class="fd-list__icon sap-icon--cart"></i>
    <span class="fd-list__title">List item 1</span>
</li>
```

If element doesn't have any text add `aria-label`
```
<i aria-label="label" class="fd-list__icon sap-icon--cart"></i>
```

Elements that are meant to be used only with icons, shouldn't have resets

PR #1604, opened 2020-09-11
[WIP] fix: Refactor icons usage

JKMarkowski:
## Related Issue
Closes https://github.com/SAP/fundamental-styles/issues/1603

## Description

So currently our approach to use icons inside components is to put class with component's element.
Ex.
```
<span class="fd-avatar fd-avatar--accent-color-1 fd-avatar--m sap-icon--money-bills" role="presentation"></span>
```
It should be changed to 
```
<span class="fd-avatar fd-avatar--accent-color-1 fd-avatar--m" role="presentation">
    <i class="sap-icon--money-bills"></i>
</span>
```

Components:
- [x] Avatar
- [x] Generic Tile
- [x] Info Label
- [x] List
- [x] Menu
- [x] Message Page
- [x] Nested List
- [x] Numeric Content
- [x] Object Marker
- [x] Object Status
- [x] Product Switch
- [x] Switch
- [x] Table
- [x] Tabs
- [x] Tree

- [ ] all items on the PR Review Checklist are addressed :

### DIFF
--- a/stories/card/card.stories.js
+++ b/stories/card/card.stories.js
@@ -221,7 +221,9 @@ export const analyticalCard = () => `
                             </div>
                             <div class="fd-numeric-content__scale-container">
                                 <div class="fd-numeric-content__scale">
-                                    <span class="fd-numeric-content__scale-arrow sap-icon--down"></span>
+                                    <span class="fd-numeric-content__scale-arrow">
+                                        <i class="sap-icon--down"></i>
+                                    </span>
                                     <span class="fd-numeric-content__scale-text">M</span>
                                 </div>
                             </div>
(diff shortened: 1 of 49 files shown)


stories/card/card.stories.js
```diff
@@ -221,7 +221,9 @@ export const analyticalCard = () => `
                             </div>
                             <div class="fd-numeric-content__scale-container">
                                 <div class="fd-numeric-content__scale">
-                                    <span class="fd-numeric-content__scale-arrow sap-icon--down"></span>
+                                    <span class="fd-numeric-content__scale-arrow">
+                                        <i class="sap-icon--down"></i>
```
jbadan, line 225: I know this is a draft but just wanted to catch it early before you do too much work - all of these icons will need either aria-hidden or aria-label attributes. aria-hidden if there is already text, aria-label if not (like this example)

Issue #747, opened 2020-01-31
Wizard component


andresfelipe66:
#### Is this a bug, enhancement, or feature request?
Feature Request

#### Briefly describe your proposal.
We are trying to implement other Wizards available on the Web (Bootstrap), but I'm trying to validate if it's possible to have something related to Fiori.

#### Which versions of Angular and Fundamental NGX are affected? (If this is a feature request, use current version.)
Current Version

#### Is there anything else we should know?
I'd like to have something similar to openui5 Wizard. https://openui5.hana.ondemand.com/api/sap.m.Wizard#overview

We are implementing a solution with different Wizards (Process step by step) (In our architecture we are using Ui5 web components and Fiori fundamental-ngx). And, off course we like to have the SAP Fiori User Experience.

droshev:
@estelaxu3 can you validate the status of the design?

salarenko:
By now "Wizard Functionality" can be achieved using `Tabs` component in process mode. Its slightly different visual experience, but I think it covers the desired functionality.

[NGX Core](https://sap.github.io/fundamental-ngx/#/core/tabs#processMode)
[Fundamental Styles](https://sap.github.io/fundamental-styles/components/tabs.html)

widra:
Is it possible to implement the 'real' wizard function from ui5?

droshev:
@andresfelipe66 @widra please follow the Wizard epic #1366

PR #1621, opened 2020-09-15
feat: new Fiori 3 Wizard component

InnaAtanasova:
## Related Issue
Closes SAP/fundamental-styles#747

## Description
Introduce new Fiori 3 component to Fundamental-styles - Wizard component, which is used to divide into sections a long or unfamiliar task and guides the user through it.


src/bar.scss
```diff
@@ -23,8 +23,8 @@ $block: #{$fd-namespace}-bar;
     padding: 0 0.5rem;
     margin-top: 0.5rem;
     margin-bottom: 0.5rem;
-    margin-left: $margin-x;
-    margin-right: $margin-x;
+    left: $margin-x;
+    right: $margin-x;
```
JKMarkowski, line 27: Floating bar examples are not working well cause of changes in bar.scss
image
Let's add some let's add left: 0 / right: 0 by default

src/wizard.scss
```diff
@@ -0,0 +1,300 @@
...
+    @include fd-reset();
+    @include fd-flex-vertical-center();
+
+    height: 2rem;
+  }
+
+  &__step {
+    @include fd-reset();
+
+    width: 100%;
```
JKMarkowski, line 65: image
Last one without any line has same width as rest of them. I'm not sure if it's correct behaviour

src/wizard.scss
```diff
@@ -0,0 +1,300 @@
...
+        @include line-clamp(1);
+      }
+    }
+  }
+
+  &__label {
+    @include fd-reset();
+    @include line-clamp();
+
+    line-height: 1;
```
JKMarkowski, line 237: image
Let's change line-height to 1rem

src/wizard.scss
```diff
@@ -0,0 +1,300 @@
...
+          @include fd-rtl() {
+            padding-right: 0;
+            padding-left: $fd-wizard-stacked-upcoming-step-offset;
+          }
+        }
+
+        // Fiori 3 recommends a maximum of 8 steps
+        @for $i from 1 through 8 {
+          &:nth-child(#{$i}) {
+            z-index: calc(10 - #{$i});
```
JKMarkowski, line 169: Great Job with it, but could you add more z-index for focused upcoming elements?
wizard

src/wizard.scss
JKMarkowski, line 237: IE11
image

JKMarkowski (changes requested):
Hi @InnaAtanasova I added some comments. Great Job with this component, I was wondering how you will deal with sizing. I think your approach there is simplest&best.

src/wizard.scss
```diff
@@ -0,0 +1,300 @@
...
+
+  position: relative;
+
+  &__navigation {
+    @include fd-reset();
+  }
+
+  &__progress-bar {
+    @include fd-reset();
+    @include fd-wizard-responsive-paddings();
```
JKMarkowski, line 42: What do you think about moving `--s/m/l/xl` to block and applying padding on progress bar same as for content

JKMarkowski:
image

What about the situation, when there is optional text + the title too long for 1 line?

src/wizard.scss
InnaAtanasova: I don't understand :) The line-heigh is 1rem. The height of the label is supposed to be the same as the circle - 2 rem

src/wizard.scss
InnaAtanasova: I don't understand how more z-index will solve the problem. Could you elaborate, please? The upcoming elements should not be focused. I disabled the pointer events but for tabs I need to use JavaScript so we will handle this in the implementation libraries.

InnaAtanasova:
> image
> 
> What about the situation, when there is optional text + the title too long for 1 line?

Hi @JKMarkowski , it's described in the documentation part and also in example:
Screen Shot 2020-09-15 at 9 26 13 AM
Screen Shot 2020-09-15 at 9 26 05 AM

src/wizard.scss
InnaAtanasova: This is ok :) we are using line-clamp and it's not working in IE11. I can check the line-height, but the rest can't be fixed (as we agreed and discussed multiple times when we decided to go with line-clamp)

src/wizard.scss
InnaAtanasova: Could you be more specific? I don't understand your idea

src/wizard.scss
InnaAtanasova: Screen Shot 2020-09-15 at 9 31 22 AM

That's because it's following the specs and has a min-width. In this specific case, since there are multiple steps, their width is equal and equal to the min-width.

JKMarkowski:
Hi @InnaAtanasova, It doesn't work on IE11 then
Ie11
image


Chrome:
image

InnaAtanasova:
> Hi @InnaAtanasova, It doesn't work on IE11 then
> Ie11
> image
> 
> Chrome:
> image



> Hi @InnaAtanasova, It doesn't work on IE11 then
> Ie11
> image
> 
> Chrome:
> image

yes, it doesn't work :) it's using line-clamp and it's not supported in IE11. Other solutions are more than welcome

src/bar.scss
InnaAtanasova: Fixed! Thanks for catching it

src/wizard.scss
InnaAtanasova: Done

InnaAtanasova:
hey @JKMarkowski can you take a look at it again? I addressed the IE11 issue.

mikerodonnell89:
@InnaAtanasova why do we need the "fd-wizard__label-container" div? Seems like an extra element when we could just have the fd-wizard__label

InnaAtanasova:
> @InnaAtanasova why do we need the "fd-wizard__label-container" div? Seems like an extra element when we could just have the fd-wizard__label

Multiple reasons for the container div. First, because of the truncation rules specified for the label and the optional text. Inside the container you can have either 1 element - the label itself or 2 elements - label and optional text. When you have only label, the label goes to the second line and then truncates. If you have a label+optional text, each of the elements truncates on the first line. For this reason you need a modifier class `fd-wizard__label-container fd-wizard__label-container--optional`. In css we can target a sibling that comes after a specific element, but not the sibling that comes before a specific element. For this reason you need to apply this modifier class on a wrapper element.

The second reason is because you have an <a> tag holding 3 elements - a span (step-indicator), a label and an optional text. They need to be placed like this:
Screen Shot 2020-09-16 at 9 28 07 AM

Using flex I wrap the label and the optional text in 1 div and this way I get 2 divs (step-indicator and label container) that I can manipulate.

stories/wizard/wizard.stories.js
```diff
@@ -0,0 +1,868 @@
...
+
+|  &nbsp;&nbsp;**rem** &nbsp;&nbsp; |  &nbsp;&nbsp;**min-width** &nbsp;&nbsp; |  &nbsp;&nbsp;**max width** &nbsp;&nbsp; |  &nbsp;&nbsp;**modifier class** &nbsp;&nbsp; |
+| :----: | :----------: | :----------: | :----------------------------------: |
+| 1rem | _n/a_ | 599px | <code>fd-wizard\\_\\_progress-bar--s</code> / <code>fd-wizard\\_\\_content--s</code> |
+| 2rem | 600px | 1023px | <code>fd-wizard\\_\\_progress-bar--m</code> / <code>fd-wizard\\_\\_content--m</code> |
+| 2rem | 1024px | 1439px | <code>fd-wizard\\_\\_progress-bar--l</code> / <code>fd-wizard\\_\\_content--l</code> |
+| 3rem | 1440px | _n/a_ | <code>fd-wizard\\_\\_progress-bar--xl</code> / <code>fd-wizard\\_\\_content--xl</code> |
+
+
+## Step types
```
jessmetivier, line 41: I think we need to refer to it as ##Modifiers here

and add a ##Types section under the Usage section, with this:

"There are two types of wizard that offer different functionality: _**standard**_ and _**branching**_.

**Use the standard type if:**

- The total number of steps is known in advance.
- The number of steps does not change during usage.
- There is linear progression from one step to the next.

**Use the branching type if:**

- The total number of steps is not known.
- The number of steps may change during usage.
- There is non-linear progression. In other words, the user’s choice during one step determines which step comes next.
- In both types of wizard you can let users skip steps. Label these steps as “Optional”.

stories/wizard/wizard.stories.js
```diff
@@ -0,0 +1,868 @@
...
+`;
+
+customized.parameters = {
+    docs: {
+        iframeHeight: 950,
+        storyDescription: `Modifier classes can be applied to customize the Wizard's elements. In the example below responsive paddings are added to the progress bar (<code class="docs-code">fd-wizard\\_\\_progress-bar--m</code>), the wizard content (<code class="docs-code">fd-wizard\\_\\_content--m</code>) and the footer (<code class="docs-code">fd-bar--page-m_l</code>). The background of the wizard content is set to list (<code class="docs-code">fd-wizard\\_\\_content--list</code>).
+
+In certain use cases, the steps in the wizard depend on the choices the user makes along the way. The user’s entries for one step determine the follow-on steps (“branching”). Use the <code class="docs-code">fd-wizard\\_\\_connector--branching</code> applied on the step connector to indicate a branching step.
+`
+    }
```
jessmetivier, line 233: The wizard component can be customized with modifier classes. For instance, the example below has several modified elements:

| Modified | Modification |
| :------------- | :------------|
| <code class="docs-code">fd-wizard\\_\\_progress-bar--m</code> | Added responsive padding |
| <code class="docs-code">fd-wizard\\_\\_content--m</code> | Added responsive padding |
| <code class="docs-code">fd-bar--page-m_l</code> | Added responsive padding |
| <code class="docs-code">fd-wizard\\_\\_content--list</code> | The background of the wizard content is set to list |

###Branching
To display a branching step, add the <code class="docs-code">fd-wizard\\_\\_connector--branching</code> to the step connector element.

stories/wizard/wizard.stories.js
```diff
@@ -0,0 +1,868 @@
...
+| current       |  &nbsp;&nbsp; &nbsp;&nbsp; <code>fd-wizard\\_\\_step--current</code>       |
+| upcoming      |  &nbsp;&nbsp; &nbsp;&nbsp; <code>fd-wizard\\_\\_step--upcoming</code>      |
+| no-label      |  &nbsp;&nbsp; &nbsp;&nbsp; <code>fd-wizard\\_\\_step--no-label</code>      |
+| stacked       |  &nbsp;&nbsp; &nbsp;&nbsp; <code>fd-wizard\\_\\_step--stacked</code>       |
+| stacked-top   |  &nbsp;&nbsp; &nbsp;&nbsp; <code>fd-wizard\\_\\_step--stacked-top</code>   |
+| active        |  &nbsp;&nbsp; &nbsp;&nbsp; <code>fd-wizard\\_\\_step--active</code>        |
+
+## Connector types
+
+When connecting Inactive steps, or Active and Inactive steps, use the default <code>fd-wizard\\_\\_connector</code> class. For connecting Active steps the <code>fd-wizard\\_\\_connector--active</code> modifier class should be applied together with <code>fd-wizard\\_\\_connector</code>. For branching step connector use <code>fd-wizard\\_\\_connector--branching</code> modifier class.
```
jessmetivier, line 58: Just want to make sure here: Inactive steps are the greyed out steps that have yet to be selected right?

stories/wizard/wizard.stories.js
```diff
@@ -0,0 +1,868 @@
...
+| 2rem | 1024px | 1439px | <code>fd-wizard\\_\\_progress-bar--l</code> / <code>fd-wizard\\_\\_content--l</code> |
+| 3rem | 1440px | _n/a_ | <code>fd-wizard\\_\\_progress-bar--xl</code> / <code>fd-wizard\\_\\_content--xl</code> |
+
+
+## Step types
+
+Use the following modifier classes to change the type of the wizard step:
+
+
+| **Type** | **Modifier class** |
```
jessmetivier, line 46: Instead of "Type" I would put "Steps"

stories/wizard/wizard.stories.js
```diff
@@ -0,0 +1,868 @@
...
+            </div>
+        </div>
+    </footer>
+</div>
+`;
+
+responsive.parameters = {
+    docs: {
+        iframeHeight: 950,
+        storyDescription: `If there is enough horizontal space, all labels are shown.
```
jessmetivier, line 567: If there is insufficient horizontal space, all step labels are displayed. However, when the width decreases, the unselected step labels disappear. To hide the labels, add the <code class="docs-code">fd-wizard\\_\\_step--no-label</code> to the **(?) element/container**.

The unselected steps should stack on top of each other when there is limited screen space. To stack them, add the <code class="docs-code">fd-wizard\\_\\_step--stacked</code> modifier class. The last completed step should display on top, and requires a <code>fd-wizard__step--stacked-top</code> modifier class.

**I don't understand this sentence: No such modifier class is required for the upcoming stacked steps.** Could you clarify this?

stories/wizard/wizard.stories.js
```diff
@@ -0,0 +1,868 @@
...
+            </footer>
+        </div>
+    </div>
+</div>
+`;
+
+mobile.parameters = {
+    docs: {
+        iframeHeight: 950,
+        storyDescription: `In mobile mode use the <code class="docs-code">fd-bar--floating-footer</code> modifier class to achieve a floating footer. The Next Step button is placed after the content when possible. It can be placed above the floating footer by applying the <code class="docs-code">fd-wizard\\_\\_next-step--floating</code> modifier class.
```
jessmetivier, line 865: For mobile devices, add the <code class="docs-code">fd-bar--floating-footer</code> modifier class to display a floating footer. The _Next Step_ button is placed after the content whenever possible. However, it can be placed above the floating footer by adding the <code class="docs-code">fd-wizard\\_\\_next-step--floating</code> modifier class.

jessmetivier (changes requested):
This was a big one, and you did an amazing job with the docs! I have a few questions, and actually need your help to finish some of my descriptions. Let me know on Slack when you've reviewed it! 😄

stories/wizard/wizard.stories.js
```diff
@@ -0,0 +1,868 @@
...
+
+These modifier classes will add horizontal paddings to the content and can be applied on the <code>fd-wizard\\_\\_progress-bar</code> level and/or on the <code>fd-wizard\\_\\_content</code> level.
+
+
+|  &nbsp;&nbsp;**rem** &nbsp;&nbsp; |  &nbsp;&nbsp;**min-width** &nbsp;&nbsp; |  &nbsp;&nbsp;**max width** &nbsp;&nbsp; |  &nbsp;&nbsp;**modifier class** &nbsp;&nbsp; |
+| :----: | :----------: | :----------: | :----------------------------------: |
+| 1rem | _n/a_ | 599px | <code>fd-wizard\\_\\_progress-bar--s</code> / <code>fd-wizard\\_\\_content--s</code> |
+| 2rem | 600px | 1023px | <code>fd-wizard\\_\\_progress-bar--m</code> / <code>fd-wizard\\_\\_content--m</code> |
+| 2rem | 1024px | 1439px | <code>fd-wizard\\_\\_progress-bar--l</code> / <code>fd-wizard\\_\\_content--l</code> |
+| 3rem | 1440px | _n/a_ | <code>fd-wizard\\_\\_progress-bar--xl</code> / <code>fd-wizard\\_\\_content--xl</code> |
```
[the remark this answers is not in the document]
jessmetivier: If it looks all jagged and weird, follow what we've done with the calendar! Where one column is right aligned, and the other is left. Just so it's easier to scan.

stories/wizard/wizard.stories.js
InnaAtanasova: The stacked items on the left-hand side are completed items, and the ones on the right-hand side are upcoming. You don't need to apply `fd-wizard__step--stacked-top`modifier class on the right-hand (upcoming steps) stack, just for the left-hand (completed steps)

stories/wizard/wizard.stories.js
jessmetivier: There are multiple connector types that can be displayed depending on the steps involved.

| Connector type | Connection | Modifier/Class |
| :--------------- | :----------- | :------------- |
| Default | Inactive step (or Active step) to Inactive step | <code>fd-wizard\\_\\_connector</code> |
| Active | Active step to Active step | <code>fd-wizard\\_\\_connector--active</code> |
| Branching | Branching step to Inactive step | <code>fd-wizard\\_\\_connector--branching</code> |

jessmetivier: Does this make sense? @InnaAtanasova

stories/wizard/wizard.stories.js
```diff
@@ -0,0 +1,868 @@
...
+        </div>
+    </footer>
+</div>
+`;
+
+defaultExample.storyName = 'Default';
+defaultExample.parameters = {
+    docs: {
+        iframeHeight: 950,
+        storyDescription: `When the wizard is triggered, the user is presented a walkthrough screen, which shows only a part of the long form. After all necessary fields are filled out, a button appears in the Wizard content that allows the user to move to the next step. A Cancel button situated in the footer can exit the wizard. If the form fields have been modified, a data loss message can be prompted to the user.
```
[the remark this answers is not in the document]
jessmetivier: The default wizard displays a walkthrough screen with forms split into sections. Once the user fills all of the necessary input fields, a button is displayed to go to the next section. A _Cancel_ button is displayed in the footer, so the user can exit the wizard at any point during the process. However, if a user exits the wizard after adding their input, they will receive a warning that their information will be lost.

stories/wizard/wizard.stories.js
InnaAtanasova: This I changed a bit. These are modifier classes, so I added a column for the element and a column for the modifier class

stories/wizard/wizard.stories.js
jessmetivier: Ok so here goes:

"If there is insufficient horizontal space, all step labels are displayed. However, when the width decreases, the unselected step labels disappear. To hide the labels, add the `fd-wizard\_\_step--no-label` to the **(?) element/container**. ------ Add your input here @InnaAtanasova 

The unselected steps should stack on top of each other when there is limited screen space. To stack them, add the `fd-wizard\_\_step--stacked` modifier class. The last completed step should display on top, and requires a `fd-wizard__step--stacked-top` modifier class. Although they are stacked, upcoming steps do not require the stacked-top modifier class. "

jessmetivier (changes requested):
Let me know if you see the new comments

stories/wizard/wizard.stories.js
```diff
@@ -0,0 +1,897 @@
...
+    </nav>
+    <section class="fd-wizard__content" id="wizard-section-1" style="min-height: 300px;">
+        <div>
+            Sed fermentum, mi et tristique ullamcorper, sapien sapien faucibus sem, quis pretium nibh lorem malesuada diam. Nulla quis arcu aliquet, feugiat massa semper, volutpat diam. Nam vitae ante posuere, molestie neque sit amet, dapibus velit. Maecenas eleifend tempor lorem. Mauris vitae elementum mi, sed eleifend ligula. Nulla tempor vulputate dolor, nec dignissim quam convallis ut. Praesent vitae commodo felis, ut iaculis felis. Fusce quis eleifend sapien, eget facilisis nibh. Suspendisse est velit, scelerisque ut commodo eget, dignissim quis metus. Cras faucibus consequat gravida. Curabitur vitae quam felis. Phasellus ac leo eleifend, commodo tortor et, varius quam. Aliquam erat volutpat
+        </div>
+        <div class="fd-wizard__next-step">
+            <button class="fd-button fd-button--compact fd-button--emphasized">Step 3</button>
+        </div>
+    </section>
+    <footer>
```
prsdthkr, line 166: Semantically all `<footer>` elements should be enclosed in sectioning content like `<section>`. See https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Sectioning_content

prsdthkr (changes requested):

stories/wizard/wizard.stories.js
```diff
@@ -0,0 +1,897 @@
...
+| Active | &nbsp;&nbsp; &nbsp;&nbsp; Active step to Active step | &nbsp;&nbsp; &nbsp;&nbsp; <code>fd-wizard\\_\\_connector--active</code> |
+| Branching | &nbsp;&nbsp; &nbsp;&nbsp; Branching step to Inactive step or no step | &nbsp;&nbsp; &nbsp;&nbsp; <code>fd-wizard\\_\\_connector--branching</code> |
+
+## Truncation rules
+
+| **Type** | &nbsp;&nbsp; &nbsp;&nbsp; **Max lines** | &nbsp;&nbsp; &nbsp;&nbsp; **Modifier class** |
+| :---- | :-------------- | :-------------- |
+| Label | &nbsp;&nbsp; &nbsp;&nbsp; 2 |  &nbsp;&nbsp; &nbsp;&nbsp; Text truncates when it exceeds the maximum space available. When a step has optional text, the label can only accommodate 1 line. |
+| Label with optional text | &nbsp;&nbsp; &nbsp;&nbsp; 1 | &nbsp;&nbsp; &nbsp;&nbsp; Text truncates when it exceeds the maximum space available. The <code>fd-wizard\\_\\_label-container--optional</code> modifier class is applied together with <code>fd-wizard\\_\\_label-container</code>. |
+| Optional text | &nbsp;&nbsp; &nbsp;&nbsp; 1 | &nbsp;&nbsp; &nbsp;&nbsp; Text truncates when it exceeds the maximum space available. |
```
jessmetivier, line 89: So far, all of the tables look fantastic! I tried making the window smaller to see how they all looked, and I think here it's the only place where the text wraps too soon (the **Max lines** column). I'm not sure if more spacing would help here or if it would look more jarring. But I figured I would point this out, I can get to it later and test out some things. :)

stories/wizard/wizard.stories.js
```diff
@@ -0,0 +1,897 @@
...
+## Layout
+
+The wizard can be used both in a full-screen layout and in the flexible column layout. There is no navigation from the wizard to a next page. After completion or cancellation, the user is always navigated back to the page the wizard was triggered from.
+
+## Responsive paddings
+
+These modifier classes will add horizontal paddings to the content and can be applied on the <code>fd-wizard\\_\\_progress-bar</code> level and/or on the <code>fd-wizard\\_\\_content</code> level.
+
+
+|  **rem** |  &nbsp;&nbsp;&nbsp;&nbsp;**min-width** |  &nbsp;&nbsp;&nbsp;&nbsp;**max width** |  &nbsp;&nbsp;&nbsp;&nbsp;**modifier class** |
```
jessmetivier, line 52: I missed this, let's capitalize: Min-width, Max-width, Modifier class
To stay aligned with the other tables.

jessmetivier (approved):
I left a couple comments for minor changes, but overall great job!!

JKMarkowski (approved):
LGTM. Great Job!

prsdthkr (approved):

PR #1638, opened 2020-09-15
fix: update file uploader accessibility

meganaconley:
## Description
Update File Uploader accessibility:
- Add aria-live region with file name
- Make readonly input tabbable
- Fix incorrect aria attributes on input
- Add example of post file selection

## Screenshots
No visual changes

### DIFF
--- a/stories/file-uploader/file-uploader.stories.js
+++ b/stories/file-uploader/file-uploader.stories.js
@@ -16,12 +16,6 @@ Use the File Uploader if:
 
 * The user needs to select the files to upload by browsing the system folders.
 * The user needs to select the files based on specific file formats to upload.
-
-Note that the input includes an autocomplete attribute with a value of "off". 
-Historically, setting autocomplete=“off” has prevented the browser from predicting this value.
-Chrome, however, has frequently changed its stance on support of removing the autocomplete.
-With some versions, autocomplete=“off” will work, while in others, autocomplete=“nope” or a semantic autocomplete=“specific example here” will work.
-Make sure to test your implementation to ensure the correct autocomplete value is used to prevent predictions.
 `,
         tags: ['f3', 'a11y', 'theme']
     }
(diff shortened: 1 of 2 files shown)


InnaAtanasova:
Hey @meganaconley I see the markup is changed. Could you update the Breaking changes in the wiki and also mention them in the PR template. Thanks!

meganaconley:
@InnaAtanasova I've updated the wiki.

stories/file-uploader/file-uploader.stories.js
```diff
@@ -23,32 +23,30 @@ Use the File Uploader if:
...
-      onchange="selectFile(this,'browse_input1')">
+  <div class="fd-file-uploader">
+    <label for="browse_input1" class="fd-file-uploader__hidden">File upload</label>
+    <input 
+      class="fd-input fd-file-uploader__input" 
+      onclick="browseFile('input1');" 
+      title="Choose a file for upload" 
+      type="text" 
+      id="browse_input1" 
+      autocomplete="off"
```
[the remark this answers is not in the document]
meganaconley: I'm not entirely sure it's necessary here since this input should only ever be `readonly`, but it was already here so I wasn't sure if there was a reason. I don't think even Chrome would add predictions to a readonly field.

stories/file-uploader/file-uploader.stories.js
```diff
@@ -23,32 +23,30 @@ Use the File Uploader if:
...
-          aria-label="Select a file for uploading">Browse...
-        </button>
-    </div>
-    <input
-      id="input1"
-      class="fd-file-uploader__hidden"
-      type="file"
-      onchange="selectFile(this,'browse_input1')">
+  <div class="fd-file-uploader">
+    <label for="browse_input1" class="fd-file-uploader__hidden">File upload</label>
```
prsdthkr, line 27: For some reason, the label value is not being computed as an accessible name for the input field so not announced by VioceOver too. This also causes a storybook a11y violation. I think it is interfering with the title attribute. Should we revert back to using aria-label?
image

prsdthkr (changes requested):

stories/file-uploader/file-uploader.stories.js
meganaconley: It turns out it's not necessarily the `title` that's the problem but the fact that the `label` is visually hidden. In these cases `aria-label` is preferred over a visually hidden `label`. The More You Know ✫彡

prsdthkr (approved):
LGTM after snapshot update ⛴

jacobdevera (approved):

JKMarkowski (approved):
LGTM. There are some errors in console upon clicking on file uploader, but it's not scope of this PR
image

Issue #1639, opened 2020-09-16
Fundamental Styles 0.12.0 defect hunting


stefanoScalzo:
- [ ] ![Screen Shot 2020-09-16 at 9 08 03 AM](https://user-images.githubusercontent.com/50607147/93342009-d2486300-f7fc-11ea-8a71-1afb13eda22e.png)
Missing url

PR #1657, opened 2020-09-17
fix: Refactor icons and text in object marker

JKMarkowski:
## Related Issue
Part of: #1603 
and #1639

## Description
There is added 1 new element:
`fd-object-status__text`
And added A11Y support in markup.

Breaking Change:
Before:
```
<span class="fd-object-marker">Locked</span>
```

After:
```
<div class="fd-object-marker">
    <span class="fd-object-marker__text">Draft</span>
</div>
```

### DIFF
--- a/src/object-marker.scss
+++ b/src/object-marker.scss
@@ -48,8 +48,6 @@ $block: #{$fd-namespace}-object-marker;
   }
 
   &__icon {
-    @include fd-reset();
-
     @include fd-icon-selector() {
       @include fd-flex-center();
 
--- a/src/object-marker.scss
+++ b/src/object-marker.scss
@@ -15,98 +15,70 @@ $block: #{$fd-namespace}-object-marker;
   $fd-object-marker-padding-right: 0.25rem !default;
 
   @include fd-reset();
-  @include fd-ellipsis();
 
   max-width: 100%;
-  color: $fd-object-marker-text-color;
-  word-break: break-word;
-  align-items: center;
   line-height: 1;
-  display: inline-block;
   font-size: 1rem;
-  white-space: nowrap;
-  overflow: hidden;
-  text-overflow: ellipsis;
-
-  &::before {
-    font-size: 1rem;
-    color: $fd-object-marker-color;
-    padding: 0 $fd-object-marker-padding-right 0 0;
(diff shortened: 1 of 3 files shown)


src/object-marker.scss
```diff
@@ -15,98 +15,70 @@ $block: #{$fd-namespace}-object-marker;
...
-    display: inline-block;
+    cursor: pointer;
     text-decoration: none;
     color: $fd-object-marker-color;
-    cursor: pointer;
 
-    &::before {
-      position: relative;
-      display: inline-block;
+    .#{$block}__text {
```
InnaAtanasova, line 31: missing reset?

src/object-marker.scss
```diff
@@ -15,98 +15,70 @@ $block: #{$fd-namespace}-object-marker;
...
-      text-decoration: none;
-    }
   }
 
   &__icon {
-    color: $fd-object-marker-color;
-    font-style: normal;
-    line-height: normal;
-    font-size: 1rem;
+    @include fd-reset();
```
InnaAtanasova, line 51: do you need a reset? it's a class always applied with `sap-icon--*` which already brings a reset

InnaAtanasova (changes requested):
please document the Breaking changes in the PR templte

src/object-marker.scss
```diff
@@ -15,98 +15,68 @@ $block: #{$fd-namespace}-object-marker;
...
-  @include fd-empty() {
-    justify-content: center;
-    padding-right: 0;
-    padding-left: 0;
 
-    &::before {
-      padding-right: 0;
-      padding-left: 0;
+      @include fd-rtl() {
+        padding-left: $fd-object-marker-padding-right;
```
mikerodonnell89, line 65: So there's some padding in RTL mode but not in LTR?

mikerodonnell89 (changes requested):

src/object-marker.scss
JKMarkowski: There is `padding-right` set in line 57 for LTR

salarenko (approved):

InnaAtanasova (approved):

mikerodonnell89 (approved):

PR #1659, opened 2020-09-17
fix: Numeric Content - Refactor icons usage

salarenko:
## Related Issue
Part of: SAP/fundamental-styles#1603

## Description
This PR:
- Separates numeric content icons into separate HTML elements
- Adds aria attributes

### Before:
```
<span class="fd-numeric-content__scale-arrow sap-icon--down"></span>
```

### After:
```
<span class="fd-numeric-content__scale-arrow" aria-label="decrease">
    <i class="sap-icon--down"></i>
</span>
```

### DIFF
--- a/stories/card/card.stories.js
+++ b/stories/card/card.stories.js
@@ -338,27 +338,27 @@ export const listCard = () => `
             <div class="fd-card__content" role="group" aria-label="Card Content">
                 <ul class="fd-list fd-list--no-border" role="list">
                     <li role="listitem" tabindex="0" class="fd-list__item">
-                        <span aria-hidden="true" class="fd-list__icon sap-icon--cart"></span>
+                        <span class="fd-list__icon sap-icon--cart"></span>
                         <span class="fd-list__title">List item 1</span>
                     </li>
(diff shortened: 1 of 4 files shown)


stories/card/card.stories.js
```diff
@@ -338,27 +342,27 @@ export const listCard = () => `
             <div class="fd-card__content" role="group" aria-label="Card Content">
                 <ul class="fd-list fd-list--no-border" role="list">
                     <li role="listitem" tabindex="0" class="fd-list__item">
-                        <span class="fd-list__icon sap-icon--cart"></span>
+                        <span aria-hidden="true" aria-hidden="true" class="fd-list__icon sap-icon--cart"></span>
                         <span class="fd-list__title">List item 1</span>
```
InnaAtanasova, line 346: Why is aria-hidden added here? Shouldn't we put the icon in a separate <i> tag for consistency?

InnaAtanasova (changes requested):
Since we are separating the update for each component I don't think this PR should contain Cards and Numeric content at the same time. At least rename the PR

PR #1660, opened 2020-09-17
chore: rename the delta theming parameters with fd prefix

droshev:
## Related Issue
rename the delta theming parameters with fd prefix


jbadan (approved):
Looks good to me - I think we should mark this as a breaking change since we technically export these

JKMarkowski (approved):

PR #1678, opened 2020-09-21
feat: Object list item Component styles

### LINKED ISSUE #1165: Object list Styles required as per Fiori3
Object list styles have to be created as per Fiori3 standards.

Screen Shot 2020-06-19 at 4 17 41 PM

### LINKED ISSUE #1512: Object List Item - implement styles

sKudum:
## Related Issue
Closes #1165 
Closes #1512 

## Description
Introducing  Object list item component, one of the variants of the list as pere Fiori3.

## Screenshots


src/_listObject.scss
```diff
@@ -0,0 +1,228 @@
+
+@import "./listDefinitions.scss";
```
Betrozov, line 2: plz remove `.scss` it is not needed

stories/list/object/object-list.stories.js
```diff
@@ -0,0 +1,496 @@
...
+To create a object list add the \`fd-list__object\` modifier class to ul and add (\`fd-list__item--object\`) 
+modifier class to the element.
+(\`fd-list-optional\`) and (\`fd-list-optional__text\`) modifier class will help to add optional text information.
+(\`fd-list__object--top-row\`) and (\`fd-list__object--bottom-row\`) class will help to devide the list item into two parts.
+(\`fd-list__object--top-right\`) and (\`fd-list__object--top-left\`) modifiers will help to divide the top row into 2 column with ratio of 60% and 40%.
+Similary (\`fd-list__object--bottom-right\`) and (\`fd-list__object--bottom-left\`) modifiers will help to divide the bottom rows into 2 column with ratio of 50% and 50%.
+and to represent semantics for respective element append element-- \`neutral\`, \`positive\`, \`critical\`, \`negative\`
+and \`informative\` status.
+`,
+        tags: ['f3', 'a11y', 'theme', 'development']
```
Betrozov, line 20: is it correct tags?

Betrozov (changes requested):

Betrozov:
Screenshot 2020-09-22 at 11 22 28
letters cut off

Betrozov:
plz fix travis

src/_listObject.scss
```diff
@@ -0,0 +1,228 @@
...
+      }
+    }
+
+    &--top-right {
+      width: 40%;
+      flex: 2 2 10%;
+      text-align: right;
+      padding-left: $fd-list-object-with-padding-value-one !important;
+      vertical-align: $fd-list-object-with-vertical-align-top !important;
+      margin-right: $fd-list-object-with-margin-value-zero !important;
```
dimamarksman, line 65: Can we avoid `!important` usage?

src/_listObject.scss
```diff
@@ -0,0 +1,228 @@
...
+      white-space: normal;
+      text-overflow: ellipsis;
+
+      @include fd-rtl() {
+        float: right;
+      }
+    }
+  }
+
+  .fd-object-status {
```
dimamarksman, line 171: Is it allowed to use external selectors for styling?
Should not we create our ourn selector for `fd-object-status`?

dimamarksman (commented):
Just questions

stories/list/object/object-list.stories.js
sKudum: included, as per other list variants

src/_listObject.scss
sKudum: styles are not getting loaded sometimes when I am removing of .scss.

sKudum:
> Screenshot 2020-09-22 at 11 22 28
> 
> letters cut off

fixed

src/_listObject.scss
sKudum: we have default styles at the list item level, whereas object list-style values need to override those values and ! import help here to achieve the required styles.

src/_listObject.scss
sKudum: we need all the styles of fd-object-status, only a few values in it are changing in the object list, so instead of rewriting whole styles again. I reusing the class and overriding the required values. The same is done where ever it is applicable.

sKudum:
> plz fix travis

Travis build will fail, because we had [WIP] in the title, which was added to mention that it has not completed internal review. and the second reason is it needs at least one approval from fundamental team members. When we click on Details it gives us detailed information on the failure.

stories/list/object/object-list.stories.js
```diff
@@ -0,0 +1,798 @@
...
+For a object list with navigation links and navigation indicator add \`fd-list--navigation\` \`fd-list--navigation-indication\` along with \`fd-list\` and \`fd-list--object\` in list and \`fd-list__link\`, \`fd-list__link--navigation-indicator\`, \`fd-list--navigation\` modifier class to the link(anchor) and the
+\`fd-list__item--link\` modifier class to the list element. that contain links.This will help all items to be navigable with navigation indicator.
+`
+    }
+};
+
+export const selection = () => `
+<h4>Object List Item With Row Selection</h4>
+<div style="max-width: 450px">
+<ul class="fd-list fd-list--object fd-list--selection fd-list__row-selection" role="list">
```
prsdthkr, line 405: Selection lists should have listbox role:
```suggestion
<ul class="fd-list fd-list--object fd-list--selection fd-list__row-selection" role="listbox">
```

stories/list/object/object-list.stories.js
```diff
@@ -0,0 +1,798 @@
...
+  <li role="listitem" tabindex="-1" class="fd-list__item">
```
prsdthkr, line 406: Selection list item should have option role:
```suggestion
  <li role="option" tabindex="0" class="fd-list__item">
```

stories/list/object/object-list.stories.js
```diff
@@ -0,0 +1,798 @@
...
+    docs: {
+        iframeHeight: 305,
+        storyDescription: 'To create selection list use `fd-list--selection ` and `fd-list--row-selection` along with `fd-list and fd-list--object` modifier.'
+    }
+};
+
+export const multiSelection = () => `
+<h4>Object List Item With Multi Selection</h4>
+<div style="max-width: 450px">
+<ul class="fd-list fd-list--object fd-list--selection" role="list">
```
prsdthkr, line 500:
```suggestion
<ul class="fd-list fd-list--object fd-list--selection" role="listbox">
```

stories/list/object/object-list.stories.js
```diff
@@ -0,0 +1,798 @@
...
+  <li role="listitem" tabindex="-1" class="fd-list__item is-selected">
```
prsdthkr, line 501:
```suggestion
  <li role="option" aria-selected="true" tabindex="0" class="fd-list__item is-selected">
```

stories/list/object/object-list.stories.js
```diff
@@ -0,0 +1,798 @@
...
+   </div>
+   <div class="fd-list__object fd-list__object--bottom-row">
+   <div class="fd-list__object fd-list__object--bottom-left">
+     <span class="fd-list-attribute">Third Attribute</span>
+   </div>
+   </div>
+   </div>
+   </div> 
+  </li>
+  <li role="listitem" tabindex="-1" class="fd-list__item">
```
prsdthkr, line 544:
```suggestion
  <li role="option" aria-selected="true" tabindex="0" class="fd-list__item">
```

stories/list/object/object-list.stories.js
```diff
@@ -0,0 +1,798 @@
...
+  <div class="fd-form-item fd-list__form-item">
+          <input type="checkbox" class="fd-checkbox" id="Ai4ez6111Z" checked="" aria-labelledby="O09lk1">
```
prsdthkr, line 546:
```suggestion
          <input type="checkbox" class="fd-checkbox" id="Ai4ez6111Z" checked="">
```

stories/list/object/object-list.stories.js
```diff
@@ -0,0 +1,798 @@
...
+    }
+};
+
+export const multiSelection = () => `
+<h4>Object List Item With Multi Selection</h4>
+<div style="max-width: 450px">
+<ul class="fd-list fd-list--object fd-list--selection" role="list">
+  <li role="listitem" tabindex="-1" class="fd-list__item is-selected">
+  <div class="fd-form-item fd-list__form-item">
+          <input type="checkbox" class="fd-checkbox" id="Ai4ez6111Z" checked="" aria-labelledby="O09lk1">
```
prsdthkr, line 503: Use unique ids.
```suggestion
          <input type="checkbox" class="fd-checkbox" id="Ai4ez6111Y" checked="">
```

stories/list/object/object-list.stories.js
```diff
@@ -0,0 +1,798 @@
...
+          <label tabindex="-1" class="fd-checkbox__label" for="Ai4ez6111Z"></label>
```
prsdthkr, line 504:
```suggestion
          <label aria-label="select/de-select option 1" tabindex="-1" class="fd-checkbox__label" for="Ai4ez6111Y"></label>
```

stories/list/object/object-list.stories.js
```diff
@@ -0,0 +1,798 @@
...
+     <span class="fd-list-attribute">Third Attribute</span>
+   </div>
+   </div>
+   </div>
+   </div> 
+  </li>
+  <li role="listitem" tabindex="-1" class="fd-list__item">
+  <div class="fd-form-item fd-list__form-item">
+          <input type="checkbox" class="fd-checkbox" id="Ai4ez6111Z" checked="" aria-labelledby="O09lk1">
+          <label tabindex="-1" class="fd-checkbox__label" for="Ai4ez6111Z"></label>
```
prsdthkr, line 547:
```suggestion
          <label aria-label="select/de-select option 2" tabindex="-1" class="fd-checkbox__label" for="Ai4ez6111Z"></label>
```

stories/list/object/object-list.stories.js
```diff
@@ -0,0 +1,798 @@
...
+Similary (\`fd-list__object--bottom-right\`) and (\`fd-list__object--bottom-left\`) modifiers will help to divide the bottom rows into 2 column with ratio of 50% and 50%.
+and to represent semantics for respective element append element-- \`neutral\`, \`positive\`, \`critical\`, \`negative\`
+and \`informative\` status.
+`,
+        tags: ['a11y', 'theme', 'development']
+    }
+};
+
+export const object = () => `
+<h4>Object List Item</h4>
```
prsdthkr, line 24:
```suggestion
<h4 id="objectListItemHeader" >Object List Item</h4>
```

stories/list/object/object-list.stories.js
```diff
@@ -0,0 +1,798 @@
...
+<ul class="fd-list fd-list--object" role="list">
```
prsdthkr, line 25: Would be a good idea to label all the unordered lists by something like the header.
```suggestion
<ul class="fd-list fd-list--object" role="list" aria-labelledby="objectListItemHeader">
```

prsdthkr (changes requested):
Some accessibility improvement:

stories/list/object/object-list.stories.js
sKudum: updated

stories/list/object/object-list.stories.js
sKudum: updated

stories/list/object/object-list.stories.js
sKudum: updated

stories/list/object/object-list.stories.js
sKudum: updated

stories/list/object/object-list.stories.js
sKudum: updated role and tab index values, as it is not a selected row did not update aria-selected.

stories/list/object/object-list.stories.js
sKudum: updated

stories/list/object/object-list.stories.js
sKudum: updated

stories/list/object/object-list.stories.js
sKudum: updated

stories/list/object/object-list.stories.js
sKudum: updated

stories/list/object/object-list.stories.js
sKudum: updated

stories/list/object/object-list.stories.js
sKudum: updated

src/_listObject.scss
dimamarksman: @InnaAtanasova could you please tell if such approach is acceptable here?

src/_listDefinitions.scss
```diff
@@ -40,6 +40,27 @@ $fd-list-navigation-navigator-bar-right: inset -0.1875rem 0 0 0 var(--sapSelecte
 $fd-list-navigation-navigator-bar-left: inset 0.1875rem 0 0 0 var(--sapSelectedColor);
 $fd-list-navigation-indicator-width: 2.75rem !default;
 
+$fd-list-optional-text-font-family: var(--sapUiFontFamily);
```
jbadan, line 43: Is this the right variable? I don't believe `sapUiFontFamily` is something that is a part of @sap-theming repo

https://github.com/SAP/theming-base-content/blob/master/content/Base/baseLib/sap_fiori_3/css_variables.css

Additionally, things like fontFamily should be set by the reset and should be unecessary

src/_listDefinitions.scss
```diff
@@ -40,6 +40,27 @@ $fd-list-navigation-navigator-bar-right: inset -0.1875rem 0 0 0 var(--sapSelecte
...
+$fd-list-optional-text-font-size: var(--sapMFontMediumSize);
```
jbadan, line 44: `sapMFontMediumSize` is not from the theming repo - https://github.com/SAP/theming-base-content/blob/master/content/Base/baseLib/sap_fiori_3/css_variables.css

src/_listDefinitions.scss
```diff
@@ -40,6 +40,27 @@ $fd-list-navigation-navigator-bar-right: inset -0.1875rem 0 0 0 var(--sapSelecte
...
+$fd-list-optional-text-font-weight: normal;
+$fd-list-optional-text-font-color: var(--sapUiContentLabelColor);
```
jbadan, line 46: `sapUiContentLabelColor` is not the correct variable

src/_listDefinitions.scss
```diff
@@ -40,6 +40,27 @@ $fd-list-navigation-navigator-bar-right: inset -0.1875rem 0 0 0 var(--sapSelecte
...
+$fd-list-object-with-vertical-align-bottom: bottom;
+$fd-list-object-attribute-font-family: var(--sapFontFamily);
+$fd-list-object-font-size: var(--sapFontSize);
+$fd-list-object-font-weight: normal;
+$fd-list-object-attribute-font-color: var(--sapContent_LabelColor);
+$fd-list-object-number-font-family: var(--sapFontLightFamily);
+$fd-object-identifier-title-font-family: var(--sapFontHeaderFamily);
+$fd-object-identifier-title-font-size: var(--sapFontHeader4Size);
+$fd-object-identifier-title-font-weight: var(--sapFontHeaderWeight);
+$fd-object-identifier-title-color: var(--sapGroup_TitleTextColor);
```
jbadan, line 62: A lot of these variables seem unnecessary - can you do an audit that they are 1) the correct variables and 2) that they are actually needed and not already set by `fd-reset()`

jbadan (changes requested):
See comments

src/_listDefinitions.scss
sKudum: updated sapUiFontFamily, where ever usage of fd-reset is not causing issues updated it as well.

src/_listDefinitions.scss
sKudum: updated

src/_listDefinitions.scss
sKudum: updated

src/_listDefinitions.scss
sKudum: where ever usage of fd-reset and fd-typography is not causing issues updated it as well and removed un required variables.

stories/list/object/object-list.stories.js
```diff
@@ -0,0 +1,900 @@
...
+        iframeHeight: 785
+    }
+};
+
+export const navigation = () => `
+
+<h4 id="objectListItemNavigation">Object List Item With Navigation</h4>
+
+<div style="max-width: 450px">
+<ul class="fd-list fd-list--object fd-list--navigation" role="listbox" aria-labelledby="object List Item With Navigation">
```
prsdthkr, line 218: `aria-labelledby` value should be the `id` of the header above for all `<ul>`s
```suggestion
<ul class="fd-list fd-list--object fd-list--navigation" role="listbox" aria-labelledby="objectListItemNavigation">
```

stories/list/object/object-list.stories.js
```diff
@@ -0,0 +1,900 @@
...
+    docs: {
+        iframeHeight: 305,
+        storyDescription: 'To create selection list use `fd-list--selection ` and `fd-list--row-selection` along with `fd-list and fd-list--object` modifier.'
+    }
+};
+
+export const multiSelection = () => `
+<h4 id="objectListItemMultiSelection">Object List Item With Multi Selection</h4>
+<div style="max-width: 450px">
+<ul class="fd-list fd-list--object fd-list--selection" role="listbox" aria-labelledby="objectListItem with multi selection">
```
prsdthkr, line 537: `aria-multiselectable="true"` should be added to multi selectable object list

```suggestion
<ul class="fd-list fd-list--object fd-list--selection" role="listbox" aria-multiselectable="true" aria-labelledby="objectListItemMultiSelection">
```

prsdthkr (changes requested):
Some a11y changes.

stories/list/object/object-list.stories.js
sKudum: updated

stories/list/object/object-list.stories.js
sKudum: updated

stories/list/object/object-list.stories.js
```diff
@@ -0,0 +1,900 @@
+import '../../../dist/list.css';
+import '../../../dist/icon.css';
+import '../../../dist/checkbox.css';
+
+export default {
+    title: 'Components/List/Object',
+    parameters: {
+        description: `
+        Object list provides a structure for Object list item content.
```
prsdthkr, line 9:
```suggestion
Object list provides a structure for Object list item content.
```

prsdthkr (changes requested):
Styles are broken on IE 11.
object-list-IE-styles

src/_listObject.scss
```diff
@@ -0,0 +1,226 @@
...
+    .#{$block}__object {
+      width: 100%;
+
+      &--top-row {
+        display: flex;
+        vertical-align: top;
+      }
+
+      &--top-left {
+        flex: 3 3 10%;
```
InnaAtanasova, line 48: This will most probably cause problems in IE11

src/_listObject.scss
```diff
@@ -0,0 +1,226 @@
...
+          padding: 0.5rem 0.5rem 0.5rem 0;
+        }
+      }
+
+      &--bottom {
+        margin-top: 0.25rem;
+      }
+    }
+
+    .fd-object-number {
```
jbadan, line 116: I would prefer to avoid this pattern. We should add an additional class to `fd-object-number` and not use the same classes.

src/_listObject.scss
sKudum: updated it as per Inna's commets now

stories/list/object/object-list.stories.js
sKudum: updated

sKudum:
> Styles are broken on IE 11.
> object-list-IE-styles
updated the details whiich may cause issues in IE, as I don't have thr browser can't test the fix, please let me know in case issue still presists

src/_listObject.scss
```diff
@@ -0,0 +1,226 @@
...
+      }
+    }
+
+    .fd-avatar {
+      width: $fd-list-object-with-padding-value-three;
+      height: $fd-list-object-with-padding-value-three;
+      min-width: $fd-list-object-with-padding-value-three;
+      max-width: $fd-list-object-with-padding-value-three;
+      min-height: $fd-list-object-with-padding-value-three;
+      max-height: $fd-list-object-with-padding-value-three;
```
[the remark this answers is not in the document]
sKudum: updated

src/_listObject.scss
```diff
@@ -0,0 +1,226 @@
...
+      margin-right: 0.75rem;
+
+      @include fd-rtl() {
+        margin-left: 0.75rem;
```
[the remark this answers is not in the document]
sKudum: updated

src/_listObject.scss
```diff
@@ -0,0 +1,226 @@
...
+        white-space: normal;
+        text-overflow: ellipsis;
+
+        @include fd-rtl() {
+          float: right;
+        }
+      }
+    }
+
+    .fd-object-status {
```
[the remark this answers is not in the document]
sKudum: updated

src/_listObject.scss
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    .fd-avatar {
+      width: $fd-list-object-with-padding-value-three;
+      height: $fd-list-object-with-padding-value-three;
+      min-width: $fd-list-object-with-padding-value-three;
+      max-width: $fd-list-object-with-padding-value-three;
+      min-height: $fd-list-object-with-padding-value-three;
+      max-height: $fd-list-object-with-padding-value-three;
+      margin-right: 0.75rem;
```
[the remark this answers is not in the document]
sKudum: updated

src/_listObject.scss
```diff
@@ -0,0 +1,226 @@
...
+        display: inline-block;
+      }
+    }
+
+    .#{$block}-attribute {
+      @include fd-ellipsis();
+      @include fd-reset();
+
+      width: 100%;
+      text-align: left;
```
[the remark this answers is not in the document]
sKudum: updated

src/_listObject.scss
```diff
@@ -0,0 +1,226 @@
...
+        }
+      }
+    }
+
+    .fd-object-status {
+      @include fd-ellipsis();
+
+      text-align: right;
+      width: 100%;
+      display: initial;
```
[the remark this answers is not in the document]
sKudum: updated

src/_listObject.scss
```diff
@@ -0,0 +1,226 @@
...
+
+        font-weight: $fd-object-identifier-title-font-weight;
+        line-height: 1.5;
+        vertical-align: top;
+        display: -webkit-box;
+        -webkit-line-clamp: 2;
+        -webkit-box-orient: vertical;
+        overflow: hidden;
+        white-space: normal;
+        text-overflow: ellipsis;
```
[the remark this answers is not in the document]
sKudum: updated

src/_listObject.scss
```diff
@@ -0,0 +1,226 @@
...
+
+        > div:only-child {
+          padding: 0;
+        }
+      }
+
+      &--bottom-left {
+        flex: 3 3 10%;
+        text-align: left;
+        padding: 0.5rem 0.5rem 0.5rem 0;
```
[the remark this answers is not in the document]
sKudum: updated

src/_listObject.scss
```diff
@@ -0,0 +1,226 @@
...
+        display: inline-grid;
```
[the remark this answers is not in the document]
sKudum: updated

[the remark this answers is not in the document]
sKudum: As (parent classes)fd-list_item, _content has display:flex by which truncation was not happening, when I tried a different approach then display inline grid and initial helped to achieve it so I added them.
if you can share the suggested approach for truncation that will help. I will follow the same.
Thank you.

src/_listObject.scss
sKudum: updated, as suggested by Inna in below comments.

src/_listObject.scss
sKudum: updated.

dimamarksman (approved):

jbadan:
> > updated the details whiich may cause issues in IE, as I don't have thr browser can't test the fix, please let me know in case issue still presists

@sKudum How was it possible to complete the pr review checklist without having a working virtual machine with IE11? 
Screen Shot 2020-09-25 at 8 33 19 AM


The checklist is not there by accident. You should download a windows machine from VirtualBox or vmware. Storybook is setup to be easily tested in that environment.

sKudum:
> > > updated the details whiich may cause issues in IE, as I don't have thr browser can't test the fix, please let me know in case issue still presists
> 
> @sKudum How was it possible to complete the pr review checklist without having a working virtual machine with IE11?
> Screen Shot 2020-09-25 at 8 33 19 AM
> 
> The checklist is not there by accident. You should download a windows machine from VirtualBox or VMware. The storybook is setup to be easily tested in that environment.

Hi @jbadan , I use Citrix in general to test, but since last 2 days am not able to open it due to network issues so (meanwhile the issue get resolved), I requested prasad in case he comes across the issue again I can look into it by trying other ways. 
because when I tested before raising the PR did not came across this issue.

stories/list/object/object-list.stories.js
```diff
@@ -0,0 +1,1003 @@
+import '../../../dist/list.css';
+import '../../../dist/icon.css';
+import '../../../dist/checkbox.css';
+
+export default {
+    title: 'Components/List/Object',
```
jbadan, line 6: @jessmetivier Should this be under "List" or grouped with the other "Object" components. This looks a little weird to me:

Screen Shot 2020-09-25 at 9 59 52 AM

src/_listDefinitions.scss
```diff
@@ -40,6 +40,22 @@ $fd-list-navigation-navigator-bar-right: inset -0.1875rem 0 0 0 var(--sapSelecte
...
+$fd-list-object-with-value-zero: 0;
+$fd-list-object-with-padding-value-half: 0.5rem;
+$fd-list-object-with-padding-value-one: 1rem;
+$fd-list-object-with-padding-value-three: 3rem;
+$fd-list-object-with-value-half: 0.75rem;
+$fd-list-object-align-left: left;
+$fd-list-object-align-right: right;
+$fd-list-object-with-vertical-align-top: top;
+$fd-list-object-with-vertical-align-bottom: bottom;
+$fd-list-object-with-no-text-shadow: 0 0 black;
```
jbadan, line 53: Is this correct? I can't find it in the spec. If it should be here, it should definitely be using a variable. It looks pretty strange:
With:
Screen Shot 2020-09-25 at 10 01 09 AM
Screen Shot 2020-09-25 at 10 01 17 AM

src/_listObject.scss
```diff
@@ -0,0 +1,220 @@
...
+
+      &-text {
+        font-weight: normal;
+        font-size: 1.375rem;
+        font-family: $fd-list-object-number-font-family;
+        text-shadow: $fd-list-object-with-no-text-shadow;
+      }
+
+      &-unit {
+        font-family: $fd-list-object-number-font-family;
```
jbadan, line 131: You're overriding the same value here coming from `fd-object-number__unit`. This should be removed:

Screen Shot 2020-09-25 at 10 08 28 AM

src/_listObject.scss
```diff
@@ -0,0 +1,220 @@
...
+        font-size: $fd-list-object-font-size;
```
jbadan, line 132: Same with font-size - it's already set to this. The design spec says it should be `1.375rem;`

src/_listObject.scss
```diff
@@ -0,0 +1,220 @@
...
+        text-shadow: $fd-list-object-with-no-text-shadow;
```
jbadan, line 133: This isn't in the spec. Please remove

src/_listObject.scss
```diff
@@ -0,0 +1,220 @@
...
+
+    .#{$block}__object-number {
+      vertical-align: $fd-list-object-with-vertical-align-top !important;
+      margin-right: $fd-list-object-with-value-zero;
+
+      &-text {
+        font-weight: normal;
+        font-size: 1.375rem;
+        font-family: $fd-list-object-number-font-family;
+        text-shadow: $fd-list-object-with-no-text-shadow;
```
jbadan, line 127: Same as unit - the only thing that should stay in this block is the font-size.

src/_listObject.scss
```diff
@@ -0,0 +1,220 @@
...
+      }
+
+      @include fd-rtl() {
+        margin-left: $fd-list-object-with-value-zero;
+      }
+    }
+
+    .#{$block}__object-identifier {
+      &-title {
+        @include fd-reset();
```
jbadan, line 153: Why does this need `fd-reset`? It's not a new element block. Please remove

src/_listObject.scss
```diff
@@ -0,0 +1,220 @@
...
+        @include fd-typography(var(--sapFontHeader4Size));
+        @include line-clamp($lines: 2);
+
+        line-height: 1.5;
+        vertical-align: $fd-list-object-with-vertical-align-top;
+        font-weight: $fd-object-identifier-title-font-weight;
```
jbadan, line 159: Please remove font-weight, it's now set to the same value 4 times on this element.

src/_listObject.scss
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    .#{$block}__object-status {
+      @include fd-ellipsis();
+
+      text-align: $fd-list-object-align-right;
+      margin-right: $fd-list-object-with-value-zero;
+
+      &-text {
+        text-shadow: $fd-list-object-with-no-text-shadow;
```
jbadan, line 174: Remove text-shadow

src/_listObject.scss
```diff
@@ -0,0 +1,220 @@
...
+      }
+
+      &-icon[class*=sap-icon] {
+        display: inline-block;
+      }
+    }
+
+    .#{$block}__object-attribute {
+      @include fd-ellipsis();
+      @include fd-reset();
```
jbadan, line 189: Why is fd-ellipsis before fd-reset?

jbadan (changes requested):
In addition to my comments about the css, something feels wrong about this whole component. Should it really be part of "List" and not it's own standalone "ObjectList" component?

stories/list/object/object-list.stories.js
sKudum: As per Fiori3 design guidelines, the object list item is one of the list variants and it will be reusing most the list and standard list item styles. 
Following the above points, it is placed under the list. Please let me know if that is not correct understanding, so I can place it accordingly.

src/_listDefinitions.scss
sKudum: Object variants such as status, number by default have a shadow effect on text when it is active but in the case of object lists item, these shadow effects should not be present as per the images shared in Fiori guidelines. To stay in sync with object list item styles look and feel this property is updated here.

As per Fiori.





Without text-shadow update look in local.

src/_listObject.scss
sKudum: updated

src/_listObject.scss
sKudum: added too stay in object list item look as per Fiori3 specs.(as mentioned in the above comments.)

src/_listObject.scss
sKudum: The number needs a front-light as per fiori3, 
if we remove this it takes font family bold. that's, why we added, have this here.

src/_listObject.scss
sKudum: updated

src/_listObject.scss
sKudum: updated

src/_listObject.scss
sKudum: updated

src/_listObject.scss
sKudum: text-shawdow -added to stay in object list item look as per Fiori3 specs.(as mentioned in the above comments.)

src/_listObject.scss
sKudum: added to stay in object list item look as per Fiori3 specs.(as mentioned in the above comments.)

sKudum:
Hi @jbadan, I have updated the changes as suggested but for a few such as text-shadow and font family, size removal is causing a mismatch with the object list item look as mentioned in Fiori.

 If you feel having a mismatch in look is ok or any other approach I need to follow, please do let me know.

 I will update them accordingly.

Thank You.

src/_listDefinitions.scss
jbadan: Can you show me where this is defined in the design spec? It's not mentioned in ObjectList or ObjectNumber:
Screen Shot 2020-09-25 at 12 43 26 PM
Screen Shot 2020-09-25 at 12 43 52 PM

sKudum: Yes, that's right, inspite of this the current styles of object number and object status are showing shadow effect on text which in turn effecting object list item look as mentioned in the above images.

jbadan: Sorry, I don't think I'm following. Why are you adding text shadow? Are you doing that based on the image you added with a blue background? If so, where is that coming from?

jbadan:
@sKudum I think this should be rebuilt as `fd-object-status` component. In my opinion, mixing it in with `fd-list` is confusing.  That will help this issue with following BEM naming conventions. Open to other opinions of course though! 😃

sKudum:
> @sKudum I think this should be rebuilt as `fd-object-status` component. In my opinion, mixing it in with `fd-list` is confusing. That will help this issue with following BEM naming conventions. Open to other opinions of course though! 😃

Object number and object status.scss has defined text-shadow (though it is not mentioned is the spec) which is resulting in updated text-shadow to match the per-object list item look as mentioned in the screenshot above.

I would like to take other reviewers' opinions on this, as they also suggested to not using component class names directly to override.

src/_listDefinitions.scss
sKudum: Object number and object status.scss has defined text-shadow (though it is not mentioned is the spec) which is resulting in updated text-shadow to match as per-object list item look as mentioned in the screenshot above.

jbadan: I'm sorry, I'm still not following. How does it have a defined text-shadow if it's not mentioned in the spec? Are there incorrect styles in `object-status.scss` that should be removed instead of doing hacks to override them?

jbadan: Doing some research, I see now where it's added. That contrast in objectStatus was added for accessibility reasons - https://github.com/SAP/fundamental-styles/pull/1571. We shouldn't need to override it here.

sKudum: As per my understanding from the spec, yes we should not have text-shadow in object number and object status styles. 

As am not ware of the requirement behind adding the text-shadow there, I prefer to override it in object list item (As a temporary fix).

Until it is fixed (if it is considered as bug) in object number and object status.

jbadan: The current text shadow was added as an accessibility fix. @jacobdevera can you provide additional context for the text shadow added in  https://github.com/SAP/fundamental-styles/pull/1571?

A possible override (if @jacobdevera can confirm that his changes were meant only for when the components are in a table and not in general), would be to set `text-shadow: none`.

jbadan:
@sKudum I meant `fd-object-list`, it is correct that we shouldn't be using other component names in the other scss files.

InnaAtanasova:
Hi @sKudum , I share @jbadan opinion on having `fd-object-list` as a separate component. Right now we are violating BEM rules in multiple places. For example, `<div class="fd-list__object fd-list__object--top-row"></div>` is completely against BEM guidelines. The `fd-list__object--top-row` is not a modifier class but a container (a.k.a element) and should not be using `--`. The correct markup would be `<div class="fd-object-list fd-object-list__top-row"></div>` , indicating clearly that the top row is an element of `fd-object-list`.

Element D from the specs - Number: it's specified that this component inherits the styling and behaviour from the "Large Design" ObjectNumber (Fiori 3) with some deviation on the font-size and Element E is following the styling and behaviour from the ObjectNumber with no deviations. In this case you need to have a markup like this:
```
<span class="fd-object-number fd-object-number--large fd-object-number--critical fd-object-list__object-marker">
    <span class="fd-object-number__text">0.00</span><span class="fd-object-number__unit">EUR</span>
</span>
```
and then in css you only need to overwrite the font-size to be 1.375rem. You shouldn't be modifying the font-weight, the text-shadow or the font-family. For the unit (element E) you shouldn't have an additional class and overwrite anything. 

For element C - you need to follow the Avatar component. Avatar with size S has a width and height of 3rem so you don't need any additional classes on this element. The right spacing specified in the specs can be added as a left-spacing on the Title container and it can be removed if no avatar is added (simply checking if the title container is a first child and if it is, remove the left spacing). This way the user won't have to add an additional class to the Avatar just for the spacing and also it will be easier to handle the RTL.

InnaAtanasova:
For the class names I'd suggest a simplified naming. 
You currently have:
```
<div class="fd-list__object fd-list__object--bottom-row">
    <div class="fd-list__object fd-list__object--bottom-left">
        <div class="fd-list__object-attribute">
                  <span class="fd-list__object-attribute-text">First Attribute</span>
        </div>
    </div>
</div>
```

You can name each row in the component `fd-object-list__row` and either use a modifier classes like `fd-object-list__row--top` and `fd-object-list__row--bottom` or rely on the specificity and have `.fd-object-list__top .fd-object-list__row` and `.fd-object-list__bottom .fd-object-list__row`. 
Then your markup will be very simple:
```
<div class="fd-object-list__bottom">
     <div class="fd-object-list__row">
         <div class="fd-object-list__attribute">
                <span class="fd-object-list__attribute-text"> First Attribute</span>
         </div>
        <div class="fd-object-list__status">
                <span class="fd-object-status--inverted fd-object-status--critical">Inverted Warning</span>
        </div>
    </div>
</div>
```

InnaAtanasova:
You can also avoid names like `top` and `bottom` by using something like:
```
<li class="fd-object-list__item">
    <div class="fd-object-list__intro">...</div>
    <div class="fd-object-list__header">...</div>
    <div class="fd-object-list__content">...</div>
</li>
```

src/_listDefinitions.scss
jacobdevera: The `text-shadow` was added primarily for Table when rows are selected as the selected color causes contrast issues. If it needs to be overridden, it makes sense to use `text-shadow: none`, as the black `text-shadow` currently being set seems off.

The reason it looks off in your screenshot is due to the inverted styling (white text against dark background) making it appear to have a glow effect. I also set the `text-shadow` to `none` in those cases as there doesn't seem to be an appropriate `text-shadow` variable to use for inverted text.

jacobdevera: That said, the Object List does appear to have a selection mode with semantic text, which would run into the same problem as Table as it uses the same colors as Table. If it must be added to the spec to be incorporated here, the text-shadow should be added or be accepted as a known deviation, as the Positive text does _narrowly_ go below a 4.5:1 contrast without the text shadow.

src/_listObject.scss
```diff
@@ -0,0 +1,218 @@
...
+      &-text {
+        @include fd-reset();
+        @include fd-ellipsis();
+
+        color: $fd-list-object-font-color;
+      }
+    }
+
+    .#{$block}__content {
+      flex: auto;
```
JKMarkowski, line 29: @include fd-reset(); missing

src/_listObject.scss
```diff
@@ -0,0 +1,218 @@
...
+          text-align: $fd-list-object-align-right;
+        }
+      }
+
+      &--top-right {
+        width: 40%;
+        text-align: $fd-list-object-align-right;
+        padding-left: $fd-list-object-with-padding-value-one !important;
+        vertical-align: $fd-list-object-with-vertical-align-top !important;
+        margin-right: $fd-list-object-with-value-zero !important;
```
JKMarkowski, line 54: Please remove `!important`

src/_listObject.scss
```diff
@@ -0,0 +1,218 @@
...
+
+        @include fd-rtl() {
+          text-align: $fd-list-object-align-left;
+          padding-right: $fd-list-object-with-padding-value-one !important;
+          padding-left: $fd-list-object-with-value-zero !important;
+          margin-left: $fd-list-object-with-value-zero !important;
```
JKMarkowski, line 60: Let' remove important everywhere. It's forbidden

src/_listObject.scss
```diff
@@ -0,0 +1,218 @@
...
+        color: $fd-list-object-font-color;
+      }
+    }
+
+    .#{$block}__content {
+      flex: auto;
+    }
+
+    .#{$block}__object {
+      width: 100%;
```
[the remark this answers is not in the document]
JKMarkowski: You should change it to `flex: 1 1 0`, also `fd-list__object-attribute` whouldn't have width: 50%;
Width with flex containers works quite poor, with `flex: 1 1 0` you would have:
- `100%` width, when there is only one object
- `50%` width (50/50) when there are two

src/_listObject.scss
```diff
@@ -0,0 +1,218 @@
...
+          margin-left: $fd-list-object-with-value-zero !important;
+        }
+      }
+
+      &--bottom-row {
+        display: flex;
+        align-items: flex-end;
+        justify-content: flex-end;
+
+        > div:only-child {
```
JKMarkowski, line 69: That selector can't look like that, relying on pure elements (div) is forbidden

JKMarkowski (changes requested):
I added few comments regarding code, please take a look at Jenna's and Inna's sugestion, I totally agree with their comments. Let me know if it has been applied, so I will continue with review

sKudum:
> @sKudum I meant `fd-object-list`, it is correct that we shouldn't be using other component names in the other scss files.

Hi @jbadan, @InnaAtanasova , @JKMarkowski ,
Below are the changes done.

- Created object list as a new component.
- blocks, elements, modifier names are created as suggested.
- updated text-shadow to none.
- removed !important and used strong selectors instead of it.
- Width and reset details are updated as suggested.
-  Kept only relevant code as per the spec in the respective element classes such as object-identifier,object-attribute. 
- Only object number has display: block and content: none by which unit will be placed under number as shown in spec.
- For places I added comments, to understand why that piece of code is present will be removed once the review is done.

src/object-list.scss
```diff
@@ -0,0 +1,210 @@
...
+    @include fd-reset();
+    @include fd-flex(column);
+
+    height: 100%;
+    width: 100%;
+  }
+
+  &__header {
+    display: flex;
+    vertical-align: $fd-object-list-with-vertical-align-top;
```
InnaAtanasova, line 68: This is used only in 1 place, no need to be a variable. We use variables if something is used in multiple places so in case something change in future it will be easier to update only one value

src/object-list.scss
```diff
@@ -0,0 +1,210 @@
...
+
+    .#{$block}__header-left:not(:first-child) {
+      .#{$block}__object-identifier {
+        margin-left: $fd-object-list-with-value-half;
+
+        @include fd-rtl() {
+          margin-right: $fd-object-list-with-value-half;
+        }
+      }
```
InnaAtanasova, line 77: This is not necessary. You can apply the margin-left on the container. You will have:
```
.#{$block}__header-left:not(:first-child) {
    margin-left: $fd-object-list-with-value-half;
    
    @include fd-rtl() {
           margin-left:0;
           margin-right: $fd-object-list-with-value-half;
    }
}
```

src/object-list.scss
```diff
@@ -0,0 +1,210 @@
...
+  &__container {
+    @include fd-reset();
+    @include fd-flex(column);
+
+    height: 100%;
+    width: 100%;
+  }
+
+  &__header {
+    display: flex;
```
InnaAtanasova, line 67: missing reset

src/object-list.scss
```diff
@@ -0,0 +1,210 @@
...
+        margin-left: $fd-object-list-with-value-half;
+
+        @include fd-rtl() {
+          margin-right: $fd-object-list-with-value-half;
+        }
+      }
+    }
+  }
+
+  &__header-left {
```
InnaAtanasova, line 81: missing reset

src/object-list.scss
```diff
@@ -0,0 +1,210 @@
...
+    width: 60%;
+
+    @include fd-rtl() {
+      text-align: $fd-object-list-align-right;
+    }
+  }
+
+  &__header-right {
+    width: 40%;
```
InnaAtanasova, line 90: missing reset

src/object-list.scss
```diff
@@ -0,0 +1,210 @@
...
+    text-align: $fd-object-list-align-right;
+    padding-left: $fd-object-list-with-padding-value-one;
+
+    // object number and object status extra margin values,
+    // by which all right items are not aligned same
+    //  comments will be removed after review
+    >[class*=number] {
```
InnaAtanasova, line 97: not the way to select an element

src/object-list.scss
```diff
@@ -0,0 +1,210 @@
...
+        $fd-object-list-with-padding-value-half
+        $fd-object-list-with-value-zero;
+
+      >[class*=status] {
+        margin-left: $fd-object-list-with-value-zero;
+      }
+    }
+  }
+
+  &__content {
```
InnaAtanasova, line 178: missing reset

src/object-list.scss
```diff
@@ -0,0 +1,210 @@
...
+      padding-right: $fd-object-list-with-padding-value-one;
+      padding-left: $fd-object-list-with-value-zero;
+
+      >[class*=number] {
+        margin-left: $fd-object-list-with-value-zero;
+      }
+    }
+  }
+
+  &__row {
```
InnaAtanasova, line 112: missing reset

src/object-list.scss
```diff
@@ -0,0 +1,210 @@
...
+    align-items: flex-end;
+    justify-content: flex-end;
+
+    > *:only-child {
+      padding: $fd-object-list-with-value-zero;
+      width: 100%;
+    }
+  }
+
+  &__row-left {
```
InnaAtanasova, line 123: missing reset

src/object-list.scss
```diff
@@ -0,0 +1,210 @@
...
+      text-align: $fd-object-list-align-right;
+      padding:
+        $fd-object-list-with-padding-value-half
+        $fd-object-list-with-value-zero
+        $fd-object-list-with-padding-value-half
+        $fd-object-list-with-padding-value-half;
+    }
+  }
+
+  &__row-right {
```
InnaAtanasova, line 144: missing reset

src/object-list.scss
```diff
@@ -0,0 +1,210 @@
...
+      $fd-object-list-with-padding-value-half
+      $fd-object-list-with-value-zero
+      $fd-object-list-with-padding-value-half
+      $fd-object-list-with-padding-value-half;
+    // added for object status as it parent has display:flex,
+    //  comments will be removed after review
+    >[class*=status] {
+      > * {
+        @include fd-ellipsis();
+      }
```
InnaAtanasova, line 159: please re-write this

src/object-list.scss
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  &__object-attribute {
+    @include fd-reset();
+    @include fd-ellipsis();
+
+    &-text {
+      color: $fd-object-list-font-color;
+    }
+  }
```
InnaAtanasova, line 209: You don't need a container that holds a simple text.

src/object-list.scss
```diff
@@ -0,0 +1,210 @@
...
+
+      &::after {
+        content: none;
+      }
+    }
+  }
+
+  &__object-identifier {
+    >[class*=identifier__title] {
+      @include fd-typography(var(--sapFontHeader4Size));
```
InnaAtanasova, line 195: Please use just `font-size`

InnaAtanasova (changes requested):
Hi @sKudum , please change all selectors that are of the format  `>[class*=.....]` . This is not the way to target an element.

sKudum:
> Hi @sKudum , please change all selectors that are of the format `>[class*=.....]` . This is not the way to target an element.

Hi @InnaAtanasova , addressed all the comments .

- updated fd-reset added for requested elements.
- >[class*=.....] replaced with class names.
- used classname for span,tag,div ,i
- removed styles which are present in fd-list__item

src/object-list.scss
```diff
@@ -0,0 +1,212 @@
...
+        .#{$fd-namespace}-avatar,
+        .#{$fd-namespace}-object-identifier__title,
+        .#{$fd-namespace}-object-marker__text,
+        .#{$fd-namespace}-object-marker__icon,
+        .#{$fd-namespace}-object-number__text,
+        .#{$fd-namespace}-object-number__unit,
+        .#{$fd-namespace}-object-status__text,
+        .#{$fd-namespace}-object-status__icon {
+          color: var(--sapList_Active_TextColor);
+          text-shadow: none;
```
jbadan, line 35: @jacobdevera Is it ok to remove the text shadow here? I thought we decided it should stay for a11y reasons.

jbadan (commented):
Pending a sign off from the accessibility team (@jacobdevera , @meganaconley , @Mike-Diaz ), I think these changes look good! Thanks for taking the time to redo this component. This iteration is more sustainable long term.

src/object-list.scss
sKudum: Testers have raised concerns to handle text shadow that's why I made it as none.

 I removed it for now and will inform them to post here in case of concerns on this hope that will help.

InnaAtanasova:
@sKudum just a question, not really a comment :) 
I see object list items have 1rem top and bottom paddings. In list with Byline, in borderless design, the "inner" items have 0.5rem top and bottom paddings, the first and last items have padding-bottom 0.5rem and padding-top 0.5rem respectively. I attach an image with what I'm trying to explain :D
Screen Shot 2020-10-02 at 11 47 48 AM

Do you know if this is also the case with Object list?

sKudum:
> @sKudum just a question, not really a comment :)
> I see object list items have 1rem top and bottom paddings. In list with Byline, in borderless design, the "inner" items have 0.5rem top and bottom paddings, the first and last items have padding-bottom 0.5rem and padding-top 0.5rem respectively. I attach an image with what I'm trying to explain :D
> Screen Shot 2020-10-02 at 11 47 48 AM
> 
> Do you know if this is also the case with Object list?

Not sure @InnaAtanasova on this did not come across this, as per the discussion with spec designers I came to know object list item still styles changes will happen as of now only PH1 ready. I remember the fundamentals team is also in that mail thread. if you want I can ask them again on this or else wait for some time as they suggested already.

InnaAtanasova:
The rest LGTM

InnaAtanasova:
> > @sKudum just a question, not really a comment :)
> > I see object list items have 1rem top and bottom paddings. In list with Byline, in borderless design, the "inner" items have 0.5rem top and bottom paddings, the first and last items have padding-bottom 0.5rem and padding-top 0.5rem respectively. I attach an image with what I'm trying to explain :D
> > Screen Shot 2020-10-02 at 11 47 48 AM
> > Do you know if this is also the case with Object list?
> 
> Not sure @InnaAtanasova on this did not come across this, as per the discussion with spec designers I came to know object list item still styles changes will happen as of now only PH1 ready. I remember the fundamentals team is also in that mail thread. if you want I can ask them again on this or else wait for some time as they suggested already.

@jessmetivier could you include this in the next email to the designers?
@sKudum it's not a blocker for the PR so you can merge it once it's approved, no need to wait for the designers answer. If something needs to be done it will be logged as a separate issue. Thanks for answering my question.

InnaAtanasova (approved):

jacobdevera (approved):
LGTM 🚢

jbadan (approved):
🚀

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1193 @@
...
+      </div>
+    </div>
+    </div>
+    </a>
+  </li>  
+ </ul>
+</div>
+`;
+
+selectionWithSelectionAndNavigation.storyName = 'selectionWithSelectionAndNavigation';
```
prsdthkr, line 735:
```suggestion
selectionWithSelectionAndNavigation.storyName = 'Selection with Selection and Navigation';
```

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1193 @@
...
+           - <code class="docs-code">fd-object-list\\_\\_object-attribute</code> object attribute element
+        
+`,
+        tags: ['f3', 'a11y', 'theme', 'development']
+    }
+};
+
+export const object = () => `
+<h4 id="objectListItemHeader">Object List Item</h4>
+<ul class="fd-list fd-object-list" role="listbox" aria-labelledby="objectListItemHeader">
```
prsdthkr, line 57: Only a selectable list should have a role `listbox`. Non-selectable lists should have the role `list`, which is the default.
```suggestion
<ul class="fd-list fd-object-list" aria-labelledby="objectListItemHeader">
```

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1193 @@
...
+  <li role="option" tabindex="0" class="fd-list__item fd-object-list__item">
```
prsdthkr, line 58: Only a selectable list item should have a role `option`. Non-selectable list item should have the role `listitem`, which is the default.
```suggestion
  <li tabindex="0" class="fd-list__item fd-object-list__item">
```

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1193 @@
...
+`
+    }
+};
+
+export const navigation = () => `
+
+<h4 id="objectListItemNavigation">Object List Item With Navigation</h4>
+
+<div style="max-width: 450px">
+<ul class="fd-list fd-object-list fd-list--navigation-object fd-list--navigation" role="listbox" aria-labelledby="objectListItemNavigation">
```
prsdthkr, line 321: Navigation items need to be wrapped in the navigation role with a valid label. See https://www.w3.org/TR/wai-aria-practices/examples/landmarks/navigation.html 

```suggestion
<div role="navigation"  aria-labelledby="objectListItemNavigation">
<h4 id="objectListItemNavigation">Object List Item With Navigation</h4>
<ul class="fd-list fd-object-list fd-list--navigation-object fd-list--navigation">
```

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1193 @@
...
+  <li role="option" tabindex="0" class="fd-list__item fd-object-list__item fd-list__item--link">
```
prsdthkr, line 322: option role is only required in selectable list items.
```suggestion
  <li tabindex="0" class="fd-list__item fd-object-list__item fd-list__item--link">
```

prsdthkr (changes requested):

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1193 @@
...
+ </ul>
+</div>
+`;
+
+selection.storyName = 'Selection';
+
+selection.parameters = {
+    docs: {
+        iframeHeight: 305,
+        storyDescription: 'To create row selection list: Create a object list with `fd-list--selection` `fd-list--selection-row` modifiers.'
```
jessmetivier, line 596: Object list items can be selectable by adding the `fd-list--selection` and `fd-list--selection-row` modifier classes to the main element.

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1193 @@
...
+    </div>
+    </a>
+  </li>  
+ </ul>
+</div>
+`;
+
+selectionWithSelectionAndNavigation.storyName = 'selectionWithSelectionAndNavigation';
+
+selectionWithSelectionAndNavigation.parameters = {
```
jessmetivier, line 737: Honestly, this is too convoluted... I think we can call this a "Complex" object list item.

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1193 @@
...
+    docs: {
+        iframeHeight: 305,
+        storyDescription: ' To create row selection list with navigation add `fd-list--selection` `fd-list--navigation` `fd-list--navigation-object` modifier classes to the list and the `fd-list__item--link` modifier class to the list elements that contain links. All items should be navigable.'
```
jessmetivier, line 740: The complex object list is both selectable and navigable. To display selectable object list items with navigation links, add the `fd-list--selection`, `fd-list--navigation` and `fd-list--navigation-object` modifier classes to the list and the `fd-list__item--link` modifier class to the list elements that contain links. All items should be navigable.

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1193 @@
...
+</ul>
+</div>
+`;
+
+borderless.storyName = 'Borderless';
+
+borderless.parameters = {
+    docs: {
+        iframeHeight: 305,
+        storyDescription: 'To create borderless list use `fd-list--no-border` modifier to list.'
```
jessmetivier, line 937: Object list items can be displayed without borders. To display a borderless list, add the `fd-list--no-border` modifier class to the list element.

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1193 @@
...
+        </div>
+      </div>
+    </div>
+    </div>
+  </li>
+</ul>
+</div>
+`;
+
+rtl.storyName = 'Rtl';
```
jessmetivier, line 1186: Object list items can be displayed from right to left on the screen, so that they may be used internationally.

jessmetivier, line 1186: Capitalize RTL

jessmetivier (changes requested):
Well done! Please see my documentation updates, and let me know if you have any questions!

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1193 @@
...
+            </div>
+        </div>
+      </div>
+    </div>
+    </div>
+  </li>
+</ul>
+`;
+
+object.storyName = 'Object';
```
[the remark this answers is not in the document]
sKudum: > Modifiers are an overarching concept, so they should be mentioned with the main description.
> 
> What would you say about the examples here? Are they default object list items? How can we describe them?

No, these do not default, I placed them in main as suggested in another review.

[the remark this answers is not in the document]
sKudum: No, these do not default, I placed them in main as suggested in another review.

stories/object-list/object-list.stories.js
sKudum: renamed it to Selection with Navigation, few more features are expected in the next release of the version of the style from the Fiori team which have complex relations too. Let me know if this name does not suite.

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1193 @@
...
+- Introduction text
+- Avatar
+- Object Identfier
+- Object Number
+- Object Attribute
+- Object Marker
+- Object Status
+
+## Structure
+**Object List consists of the following elements along with standard list:**
```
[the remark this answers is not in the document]
sKudum: I indented to say, in the below structure of object list item classes such as fd-list__item etc are used the same as standard list item do. 
But now I feel it may not be required to mention because it is common to all list variants.

sKudum:
Hi @jessmetivier , @prsdthkr addressed your comments, please do let me know in case anything needs to be done further.

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1194 @@
...
+        storyDescription: 'An object list can display links that the users can navigate to. To display a navigation object list, add the `fd-list--navigation-object` modifier class together with `fd-list--navigation` and the `fd-list__item--link` modifier class to the list elements that contain links. All items should be navigable.'
+    }
+};
+
+
+export const selection = () => `
+<h4 id="objectListItemRowSelection">Object List Item With Row Selection</h4>
+<div style="max-width: 450px">
+<ul class="fd-list fd-object-list fd-list--selection fd-list--selection-row" role="listbox" aria-labelledby="objectListItemRowSelection">
+  <li role="option" tabindex="0" class="fd-list__item fd-object-list__item is-selected">
```
prsdthkr, line 468:
```suggestion
  <li role="option" aria-selected="true" tabindex="0" class="fd-list__item fd-object-list__item is-selected">
```

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1194 @@
...
+          <div class="fd-object-list__row-left">
+            <div class="fd-object-list__object-attribute">
+              Third Attribute
+            </div>
+        </div>
+      </div>
+    </div>
+    </div>
+  </li>
+  <li role="option" tabindex="0" class="fd-list__item fd-object-list__item">
```
prsdthkr, line 528:
```suggestion
  <li role="option" aria-selected="false" tabindex="0" class="fd-list__item fd-object-list__item">
```

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1194 @@
...
+        iframeHeight: 305,
+        storyDescription: 'Object list items can be selectable by adding the `fd-list--selection` and `fd-list--selection-row` modifier classes to the main element.'
+    }
+};
+
+export const selectionWithNavigation = () => `
+<h4 id="objectListItemRowSelectionAndNavigation">Object List Item With Row Selection And Navigation</h4>
+<div role="navigation" style="max-width: 450px">
+<ul class="fd-list fd-object-list fd-list--navigation-object fd-list--navigation fd-list--selection" role="listbox" aria-labelledby="objectListItemRowSelectionAndNavigation">
+  <li role="option" tabindex="0" class="fd-list__item fd-object-list__item fd-list__item--link is-selected">
```
prsdthkr, line 605:
```suggestion
  <li role="option" aria-selected="true" tabindex="0" class="fd-list__item fd-object-list__item fd-list__item--link is-selected">
```

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1194 @@
...
+            <div class="fd-object-list__object-attribute">
+              Third Attribute
+            </div>
+        </div>
+      </div>
+    </div>
+    </div>
+    </a>
+  </li>
+  <li role="option" tabindex="0" class="fd-list__item fd-object-list__item fd-list__item--link">
```
prsdthkr, line 670:
```suggestion
  <li role="option" aria-selected="false" tabindex="0" class="fd-list__item fd-object-list__item fd-list__item--link">
```

prsdthkr (changes requested):
Sorry did not notice this earlier. Here are some final suggestions for accessibility.

prsdthkr (approved):
LGTM

JKMarkowski (approved):
LGTM

Betrozov (approved):

stories/object-list/object-list.stories.js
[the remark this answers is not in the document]
jessmetivier: How about 'Standard'? Would this be the basic way to style this component?

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1194 @@
...
+An object list is another type of list that consists of object list items, and each item provides a quick 
+overview of an object. Object list items should only contain essential information for the user to identify what 
+objects to prioritize. Avoid long descriptive texts as the text space is limited and will truncate.
+         
+**Object list items can display the following elements:**
+        
+- Title of the object
+- Avatar
+- Object display components such as **Object Identifier**, **Object Number**, **Object Attribute**, **Object Marker** and **Object Status**
+- Introductory text indicating the origin of the object, such as -Forwarded by...- or -On Behalf of...-
```
jessmetivier, line 18: Oops, I think I meant to put underscores for "-Forwarded by...- or -On Behalf of...-". They should be italicized like this: _Forwarded by..._ or _On Behalf of..._

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1194 @@
...
+| **Status** | <span style="margin-left: 2rem;">**Object Number**</span> | <span style="margin-left: 2rem;">**Object Status**</span> |
+| :---- | :-------------- | :-------------- |
+| Positive|<code class="docs-code" style="margin-left: 2rem;">fd-object-number--positive</code> |<code class="docs-code" style="margin-left: 2rem;">fd-object-status--positive</code> |
+| Negative|<code class="docs-code" style="margin-left: 2rem;">fd-object-number--negative</code> |<code class="docs-code" style="margin-left: 2rem;">fd-object-status--negative</code> |
+| Critical|<code class="docs-code" style="margin-left: 2rem;">fd-object-number--critical</code> |<code class="docs-code" style="margin-left: 2rem;">fd-object-status--critical</code> |
+| Informative|<code class="docs-code" style="margin-left: 2rem;">fd-object-number--informative</code> |<code class="docs-code" style="margin-left: 2rem;">fd-object-status--informative</code> |
+
+
+## Structure
+**Object List consists of the following elements:**
```
jessmetivier, line 46: Decapitalize the first letter in "List", so it should be "Object list"

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1194 @@
...
+- Title of the object
+- Avatar
+- Object display components such as **Object Identifier**, **Object Number**, **Object Attribute**, **Object Marker** and **Object Status**
+- Introductory text indicating the origin of the object, such as -Forwarded by...- or -On Behalf of...-
+- Icon that identifies the object
+- The first status line can contain indicator icons for locked items, favorites, or items that have been
+  flagged for follow-up
+
+##Usage
+**Use the object List if:**
```
jessmetivier, line 24: Decapitalize L in "List"

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1194 @@
...
+
+- You want to display a brief details to user.
+
+**Do not use the object List if:**
```
jessmetivier, line 28: Decapitalize L in "List"

jessmetivier (changes requested):
This looks great! Just a few minor adjustments, and need to establish what the first component description is.

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1194 @@
...
+            </div>
+        </div>
+      </div>
+    </div>
+    </div>
+  </li>
+</ul>
+`;
+
+object.storyName = 'Object';
```
jessmetivier: **Standard**
The standard object list item displays a title, introductory text, an avatar and object display components. It is recommended to display avatar in size S, by adding the `fd-avatar--s` modifier class to the `fd-avatar` element.

jessmetivier (commented):
Hi Sushma,

I've added a description for the standard object list item! Just title it as "Standard", like we've done in other docs :)

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1193 @@
...
+import '../../dist/icon.css';
+import '../../dist/object-list.css';
+
+export default {
+    title: 'Components/Object List',
+    parameters: {
+        description: `
+An object list is another type of list that consists of object list items, and each item provides a quick 
+overview of an object. Object list items should only contain essential information for the user to identify what 
+objects to prioritize. Avoid long descriptive texts as the text space is limited and will truncate.
```
jessmetivier, line 11: Just a slight correction of my previous intro:

An object list is a type of list that consists of object list items, where each item provides a quick 
overview of an object. Object list items should only contain essential information for the user to identify what objects to prioritize. Avoid long descriptive texts as the text space is limited and will truncate.

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1193 @@
...
+  flagged for follow-up
+
+##Usage
+**Use the object list if:**
+
+- You want to display a brief details to user.
+
+**Do not use the object list if:**
+
+- You want to display a detailed information to user.
```
jessmetivier, line 30: **Use the object list if:**

- You want to display brief details to the user.

**Do not use the object list if:**

- You want to display detailed information to the user.
- You want to trigger more than one event. Instead, consider using **Action List Item**.

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1193 @@
...
+ </ul>
+</div>
+`;
+
+selectionWithNavigation.storyName = 'Selection With Navigation';
+
+selectionWithNavigation.parameters = {
+    docs: {
+        iframeHeight: 305,
+        storyDescription: 'This object list is both selectable and navigable. To display selectable object list items with navigation links, add the `fd-list--selection`, `fd-list--navigation` and `fd-list--navigation-object` modifier classes to the list and the `fd-list__item--link` modifier class to the list elements that contain links. All items should be navigable.'
```
jessmetivier, line 740: Let's change this to:

This object list is both selectable and navigable. To display selectable object list items with navigation, add the `fd-list--selection`, `fd-list--navigation` and `fd-list--navigation-object` modifier classes to the list and the `fd-list__item--link` modifier class to the list elements that contain links. All items should be navigable.

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1193 @@
...
+</ul>
+</div>
+`;
+
+borderless.storyName = 'Borderless';
+
+borderless.parameters = {
+    docs: {
+        iframeHeight: 305,
+        storyDescription: 'Object list items can be displayed without borders. To display a borderless list, add the `fd-list--no-border` modifier class to the list element'
```
jessmetivier, line 937: Missing period at the end of the sentence

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1193 @@
...
+  </li>
+</ul>
+`;
+
+object.storyName = 'Standard';
+
+object.parameters = {
+    docs: {
+        iframeHeight: 785,
+        storyDescription: 'The standard object list item displays a title, introductory text, an avatar and object display components. It is recommended to display avatar in size S, by adding the `fd-avatar--s` modifier class to the `fd-avatar` element.'
```
jessmetivier, line 312: Let's change this to:

The standard object list item displays a title, introductory text, an avatar and object display components. It is recommended to display avatar in size S, therefore add the `fd-avatar--s` modifier class to the `fd-avatar` element.

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1193 @@
...
+  </ul>
+</div>
+`;
+
+navigation.storyName = 'Navigation';
+
+navigation.parameters = {
+    docs: {
+        iframeHeight: 355,
+        storyDescription: 'An object list can display links that the users can navigate to. To display a navigation object list, add the `fd-list--navigation-object` modifier class together with `fd-list--navigation` and the `fd-list__item--link` modifier class to the list elements that contain links. All items should be navigable.'
```
jessmetivier, line 458: Let's change this to:

An object list can display links that the users can navigate to. To display a navigation object list, add the `fd-list--navigation-object` modifier class together with `fd-list--navigation` and the `fd-list__item--link` modifier class to the list elements that contain a link. All items should be navigable.

jessmetivier (changes requested):
This should be the last one ;)

stories/object-list/object-list.stories.js
```diff
@@ -0,0 +1,1194 @@
...
+
+##Usage
+**Use the object list if:**
+
+- You want to display brief details to the user.
+
+**Do not use the object list if:**
+
+- You want to display detailed information to the user.
+- You want to trigger more than one event. Instead, consider using Action List Item.
```
jessmetivier, line 31: Oh right, we don't have Action List Item... so this should instead be:

You want to trigger more than one event, as each object list item should contain one link.

jessmetivier (changes requested):
Hi Sushma,

I've added two comments, please address :)

I'll approve anyway, because they are so minor and I trust you're on the ball with making the changes!

jessmetivier (approved):

PR #1713, opened 2020-09-25
chore: update pr template

jbadan:
## Description
It has become pretty clear that the pr review checklist is being ignored, leading to reviews being more difficult. This pr moves the checklist to the template directly. I know this is way more wordy, but it seems necessary to prevent blindly checking "all items on the PR Review Checklist are addressed :" without having checked all of them.


InnaAtanasova (approved):

greg-a-smith (approved):
Looks good. 🚢

PR #1714, opened 2020-09-25
fix: add fix for Notifications examples markup and for the max-width of the footer content

InnaAtanasova:
## Description
The last item of the footer (7 minutes ago) was wrapped with a span and `fd-notification__footer-content` class in all examples. The footer items were always the same size, changed the max-width to be 50%, positioning the second item close to the first one.

### Before:
Screen Shot 2020-09-25 at 3 44 36 PM

### After:
Screen Shot 2020-09-25 at 3 38 50 PM
Screen Shot 2020-09-25 at 3 38 41 PM
Screen Shot 2020-09-25 at 3 38 33 PM

2. Documentation

### DIFF
--- a/src/notification.scss
+++ b/src/notification.scss
@@ -128,7 +128,11 @@ $block: #{$fd-namespace}-notification;
     &-content {
       @include fd-ellipsis();
 
-      flex: 1 1 auto;
+      max-width: 50%;
+
+      &:only-child {
+        max-width: 100%;
+      }
     }
   }
 
--- a/src/notification.scss
+++ b/src/notification.scss
@@ -1,9 +1,8 @@
-@import "./mixins";
+@import './mixins';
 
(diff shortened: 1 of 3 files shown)


droshev (approved):

prsdthkr (changes requested):
Probably unrelated but It'd be great if it's fixed here:
* popover in mobile is skewed down
image
* could we also make `fd-tabs` focusable and add focus indicator

JKMarkowski (approved):
Let's add `align-items: flex-start` to `fd-notification__actions`, to prevent from height expansion. So Popover won't behave that weird. Rest LGTM

InnaAtanasova:
Hey @prsdthkr your comments have been addressed. Could you check again? Thanks

prsdthkr (approved):
QA LGTM 🛫

PR #1730, opened 2020-09-29
fix: Truncation on menu button

JKMarkowski:
## Description
Previously there was no way to make truncation work correctly on menu buttons. With `__text` element, there is a way to fix it, adding flex/center property to button.

### Before:
image

### After:
image

3. Testing

### DIFF
--- a/src/button.scss
+++ b/src/button.scss
@@ -104,6 +104,8 @@ $block: #{$fd-namespace}-button;
   overflow: hidden;
   white-space: nowrap;
   text-overflow: ellipsis;
+  display: inline-flex;
+  align-items: center;
 }
 
 @mixin menu-icon() {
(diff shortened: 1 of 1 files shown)


prsdthkr (approved):
QA LGTM 🚢 
I'm wondering if we should add `fd-button__text` to regular/all the buttons so that they get this truncation behavior too?

JKMarkowski:
@prsdthkr We had some talks about it. Buttons usage should be for sure unified and this will be included before `0.13.0` version

katekozlowska (approved):

InnaAtanasova:
@JKMarkowski please remove the checkboxes from the template that are not relevant to the PR and check the rest, o.w. the Mergeable check fails

Issue #1734, opened 2020-09-30
Add modifier on grid column so that it takes the remaining width


stefanoScalzo:
**Is your feature request related to a problem? Please describe.**
It would be nice to have a modifier class that allows a column to take the remaining space left on the row

> Is this issue related to a specific component?
Layout grids
> What browser are you working in?
All
> What version of Fundamental Library Styles are you using?
Latest

PR #1736, opened 2020-09-30
feat: add full class modifier

stefanoScalzo:
## Related Issue
Closes SAP/fundamental-styles#1734

## Description
New modifier class called `--full` to make layout grid column take the remaining space left
Due to the grid being min 1/12 it still does have a min-width of 1/12 columns
### Before:
none

### After:
Screen Shot 2020-09-30 at 9 31 07 AM

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/layout-grid.scss
+++ b/src/layout-grid.scss
@@ -13,6 +13,12 @@ $l: 1024;
 $xl: 1440;
 $cols: 12;
 
+@mixin full {
+  min-width: ( ( 1 / 12 ) *100%);
+  max-width: none;
+  flex: 1;
+}
+
 .#{$blockContainer} {
   @include fd-reset();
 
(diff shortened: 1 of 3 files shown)


src/layout-grid.scss
```diff
@@ -73,6 +73,12 @@ $cols: 12;
   min-width: 100%;
   max-width: 100%;
 
+  &--full {
+    min-width: ( ( 1 / 12 ) *100%);
+    max-width: none;
+    flex: 1;
+  }
```
InnaAtanasova, line 80: You have the same block of code repeated 4 times. Move it to a mixin.

InnaAtanasova (changes requested):
Why is this a `feature`?

src/layout-grid.scss
stefanoScalzo: Fixed!

InnaAtanasova:
@stefanoScalzo why us this a `feat`?

JKMarkowski (approved):
LGTM, that's the thing we were missing

InnaAtanasova (approved):

PR #1744, opened 2020-10-01
docs: update docs to reference previous versions wiki page

### LINKED ISSUE #1670: display older doc versions

mikerodonnell89:
fixes #1670 

Added a "previous versions" page to the wiki and a note about it in the readme.  Clicking the version number in the top right will no longer take you to the github page of that tag, but instead to the previous versions page

### DIFF
--- a/.storybook/custom/components/Header.js
+++ b/.storybook/custom/components/Header.js
@@ -18,7 +18,7 @@ const Header = () => {
                 </a>
                 <a
                 className='fddocs-header__anchor'
-                href={`https://github.com/SAP/fundamental-styles/tree/v${packageJson.version}`}
+                href={`https://github.com/SAP/fundamental-styles/wiki/Previous-Versions`}
                 target='_blank'>
                         {`v${packageJson.version}`}
                 </a>
--- a/README.md
+++ b/README.md
@@ -85,6 +85,8 @@ If you want to contribute, please check the [Contribution Guidelines](https://gi
 
 ## Versioning
 
+Documentation for previous versions can be found [here](https://github.com/SAP/fundamental-styles/wiki/Previous-Versions).
+
 The `fundamental-styles` library follows [Semantic Versioning](https://semver.org/). These components strictly adhere to the `[MAJOR].[MINOR].[PATCH]` numbering system (also known as `[BREAKING].[FEATURE].[FIX]`).
 
 Merges to the `master` branch will be published as a prerelease. Pre-releases will include an **rc** version (_e.g._ `[MAJOR].[MINOR].[PATCH]-rc.[RC]`).


.storybook/custom/components/Header.js
```diff
@@ -18,7 +18,7 @@ const Header = () => {
                 </a>
                 <a
                 className='fddocs-header__anchor'
-                href={`https://github.com/SAP/fundamental-styles/tree/v${packageJson.version}`}
```
jbadan, line 21: The whole point of this is to take you to the version that you're seeing docs for though - could we create a separate link for previous versions?

jbadan: Or better yet, add a dropdown that has previous version links in it?

mikerodonnell89: I don't see why linking someone to the tag for this specific version would be very useful, figured this would be more helpful.

Not sure what dropdown we could use here that would fit stylisticly so I just went with the link

jbadan: The logic behind it is: say you're looking at docs for version `4.0.0`, when you click the link, you're going to want to see the code from `4.0.0`. If master is on `4.0.0-rc.20`, you don't want that because the code on master may no longer match the documentation you were looking at. 

If we want to support multiple versions, we should just set that up in general. The whole point of being 0th version is to not have to support old versions and be able to move quickly. Checkout out the Concur docs for how we supported multiple versions - https://hig.concur.com/pages/develop/web/components/nui-widgets/v13/card.html

jbadan (changes requested):
I'm not a huge fan of editing the github link. If we want to support multiple versions, we should create a programmatic solution

prsdthkr (changes requested):
I agree with @jbadan. It would be a better dev experience if we have a dropdown menu for previous versions. We then try to switch the versions in Storybook itself or maybe link to the netifly previews for previous versions. Then there would be a single link to our Github repo master branch. Kinda like how bootstrap does it. https://getbootstrap.com/docs/4.0/getting-started/introduction/

But maybe that's too ambitious 🤷 

image

.storybook/custom/components/Header.js
greg-a-smith: If we do anything, I would prefer to see a docs site version switcher.  Ultimately, it would be nice if this was automated as part of the build/release process, but is there a way we could make a dropdown list of the older versions (even if it's just the hard-coded links from the wiki page)?  Either build the current version tag directly into it or leave this version tag and put an "Other Releases" dropdown next to it.  Although a wiki page is nice and quick, it's also kind of kludgy.

JKMarkowski: I like Greg's Idea of having `Other Releases` dropdown - fd-select, or native select, similar to NGX.

mikerodonnell89:
Closing for now, as we want to research how this can be automated with netlify

PR #1745, opened 2020-10-01
docs: updated card documentation

jessmetivier:
## Description
Updated **Card** docs.

### DIFF
--- a/stories/card/card.stories.js
+++ b/stories/card/card.stories.js
@@ -19,7 +19,7 @@ A card is a container that consists of two main components: a selectable header 
 
 Components | Description
 :--------- |:-----------
-Header (main) | The header displays a mandatory title, indicating what the card is about and functions as a navigation control that opens the parent app.
+Header (main) | The header displays a mandatory title, indicating what the card is about and functions as a navigation control that directs the user the parent app.
 Content (main) | The content area is reserved for application content.
 Title | A title is mandatory to explain what content is being displayed to the user.
 Avatar (optional) | An avatar can be displayed in a size S (3rem).
@@ -285,10 +285,10 @@ analyticalCard.storyName = 'Analytical card';
 analyticalCard.parameters = {
     docs: {
         iframeHeight: 450,
-        storyDescription: `The analytical card is used for data visualization. It can display a KPI header, and various chart types in the content. The only difference between a KPI header and a default header is that the former requires a subtitle, and displays a KPI area and optional second subtitle. To display an analytical card, add the <code>fd-card--analytical</code> modifier class to the main element.
+        storyDescription: `The analytical card is used for data visualization. It can display a KPI header and various chart types in the content. The only difference between a KPI header and a default header is that the former requires a subtitle, a KPI area and displays an optional second subtitle. To display an analytical card, add the <code>fd-card--analytical</code> modifier class to the main element.
 
 ####Chart types
-**Analytical cards content area can display 8 different chart types:**
+**The content area of analytical card can display 8 different chart types:**
 
 - Line
 - Bubble
--- a/stories/card/card.stories.js
+++ b/stories/card/card.stories.js
@@ -12,7 +12,7 @@ export default {
 - informative text
 - a combination of two elements
 
-Card sizes vary depending on the layout, and they are not editable. A card can focus on either a single object or topic, or on a group of objects. Several cards can reference the same application, but the information should be distinct from one card to another.
+Card sizes vary depending on the layout, and they are not editable. A card can focus on either a single object or topic, or on a group of objects. Several cards can reference the same application, but the information should be distinct from one another.
 
 ##Card anatomy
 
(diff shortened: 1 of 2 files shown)


stories/card/card.stories.js
```diff
@@ -411,10 +397,11 @@ export const listCard = () => `
 </div>
 `;
 
+listCard.storyName = 'List card';
 listCard.parameters = {
     docs: {
         iframeHeight: 400,
-        storyDescription: 'The Header and the Content sections follow the specification described in the Card Anatomy. The Content can accommodate different list types. All controls placed inside should behave natively.'
+        storyDescription: 'A card can display various types of lists. To display a list, add <code>fd-list</code> to the content element. All components placed inside should behave natively.'
```
InnaAtanasova, line 404: "To display a list, add <code>fd-list</code> to the content element." shouldn't be there as it might confuse the user. Adding <code>fd-list</code> will not be enough to render a list so it's better without this sentence.

stories/card/card.stories.js
```diff
@@ -574,10 +561,11 @@ export const tableCard = () => `
 </div>
 `;
 
+tableCard.storyName = 'Table card';
 tableCard.parameters = {
     docs: {
         iframeHeight: 400,
-        storyDescription: 'The Header and the Content sections follow the specification described in the Card Anatomy. The Table component belongs to the Content area. The <code>.fd-card--table</code> modifier class needs to be applied with <code>.fd-card</code>.'
+        storyDescription: 'Cards can display tables within the content area. To add a table, add the <code>fd-card--table</code> modifier class to the main element.'
```
InnaAtanasova, line 568: "To add a table, add the <code>fd-card--table</code> modifier class to the main element." 

The way it's written I'm afraid that the user might think that this modifier class will add a table to the card. This modifier class is needed to turn a standard card to a Table card and apply some additional styling. Something like:
"To render a Table card, add the <code>fd-card--table</code> modifier class to the main element." I think will make more sense (feel free to modify it, it's just to explain what I mean :) )

stories/card/card.stories.js
```diff
@@ -724,19 +712,23 @@ export const objectCard = () => `
     </div>
 </div>
 `;
+
+objectCard.storyName = 'Object card';
 objectCard.parameters = {
     docs: {
         iframeHeight: 600,
-        storyDescription: `The Object Card Header is the same specified in the Card Anatomy. The Content has a 1rem padding applied on all sides. The <code>.fd-card--object</code> modifier class needs to be applied with <code>.fd-card</code>.
+        storyDescription: `The object card can display a single object or a group of objects. Contrary to other cards, object card has 1rem padding on all sides of the content area. To display an object card, add the <code>fd-card--object</code> modifier class to the main element.
```
InnaAtanasova, line 720: Ah! I saw this after my comment on Table card. This is perfect :) Could you make Table card last sentence similar to this one?

InnaAtanasova (changes requested):

stories/card/card.stories.js
jessmetivier: Oops, you're totally right about this one!

stories/card/card.stories.js
```diff
@@ -289,35 +286,24 @@ export const analyticalCard = () => `
...
-
-- Standard Header
-  - Title (mandatory) - The title provides the most important information. It is recommend to use a single-line text, but it can also wrap to two lines.
-  - Subtitle (optional) - The subtitle can wrap to two lines, and gets truncated at the end of the second line.
-- KPI Header
-  - Title (mandatory) - same as for Standard Header
-  - Subtitle (mandatory) - The subtitle can wrap to two lines, and gets truncated at the end of the second line. The unit of measure is shown at the end of the subtitle.
-  - KPI area (mandatory)
-  - Second subtitle (optional)
+        storyDescription: `The analytical card is used for data visualization. It can display a KPI header and various chart types in the content. The only difference between a KPI header and a default header is that the former requires a subtitle, a KPI area and can display an optional second subtitle. To display an analytical card, add the <code>fd-card--analytical</code> modifier class to the main element.
```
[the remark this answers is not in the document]
jessmetivier: You can use two header types for the analytical card, depending on the use case:

**Standard Header**
Title (mandatory): The title provides the most important information. We recommend using a single-line text, but you can also wrap the title to two lines.
Subtitle (optional): The subtitle can wrap to two lines, and gets truncated at the end of the second line. If the subtitle contains multiple qualifiers, separate them with comma. Do not repeat the chart title.

**KPI Header**
Title (mandatory): The title provides the most important information. We recommend using a single-line text, but you can also wrap the title to two lines.
Subtitle (mandatory): The subtitle can wrap to two lines, and gets truncated at the end of the second line. The unit of measure is shown at the end of the subtitle. We therefore recommend keeping the subtitle short and within one line. If the subtitle contains multiple qualifiers, separate them with comma. Do not repeat the chart title.

stories/card/card.stories.js
jessmetivier: Is there anyway to make this description more useful to the user? It seems flat to me now. What about referencing the List component page? Like...

"To display a list in the content area, see **Standard List**." (and then link to the page)

stories/card/card.stories.js
jessmetivier: Done.

stories/card/card.stories.js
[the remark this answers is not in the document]
InnaAtanasova: oh true! It was the second subtitle that was optional...

stories/card/card.stories.js
InnaAtanasova: You can add the link. In my opinion with and without will be good enough. I let you decide if you want to add it :)

InnaAtanasova (approved):

PR #1752, opened 2020-10-06
feat: support fallbacks in ie11 for all themes

jbadan:
## Description

Previously, only fiori-light was supported in ie11. We were relying on css variables for other themes - css variables are not supported in ie11.

This pr changes our setup. Instead of outputting 1 file (ex: `action-bar.css`), we output 1 default files `action-bar.css` and 1 file per theme - `action-bar-sap_fiori_3.css`, `action-bar-sap_fiori_3_dark.css`, `action-bar-sap_fiori_3_hcb.css`, etc.

The default file has `sap_fiori_3` (light) ie11 fallbacks to reduce breaking changes.

Each themed file has fallbacks derived from the corresponding sap-theming file and delta theming file.


config/postcss-add-fallback.js
```diff
@@ -1,9 +1,16 @@
...
 const postcss = require('postcss');
 const fs = require('fs');
 const arrayUniq = require('array-uniq');
 
+const supportedThemes = [
+    'sap_fiori_3',
+    'sap_fiori_3_dark',
+    'sap_fiori_3_hcb',
+    'sap_fiori_3_hcw',
+    'sap_fiori_3_light_dark'
```
greg-a-smith, line 11: Are we doing "Light Dark"?  Is this a real thing? 😆

scripts/theme-css.js
```diff
@@ -0,0 +1,22 @@
+const fs = require('fs-extra');
+const path = require('path');
+
+const supportedThemes = [
+    'sap_fiori_3',
+    'sap_fiori_3_dark',
+    'sap_fiori_3_hcb',
+    'sap_fiori_3_hcw',
+    'sap_fiori_3_light_dark'
+];
```
greg-a-smith, line 10: I see this array used in two places.  Should we make a constants file that both can import?

greg-a-smith (commented):
Not that this isn't the correct path to go, but removing fallbacks from the base CSS file (_e.g._ `alert-bar.css`) technically makes this a breaking change since that file will be missing numerous styles in IE11.

config/postcss-add-fallback.js
jbadan: I know 😆  but yes

scripts/theme-css.js
jbadan: Yeah that's the plan! This is the _very_ rough version

package.json
```diff
@@ -20,15 +20,15 @@
     "std-version": "standard-version --releaseCommitMessageFormat \"chore(release): version {{currentTag}} build ${TRAVIS_BUILD_NUMBER} [ci skip]\"",
     "style:assets": "./scripts/copy-assets.sh",
     "style:compile": "node-sass -q --output-style expanded --precision 5 src/ --output dist/",
-    "style:theming:postCSS": "postcss --config config/postcss.config.js  --replace dist/theming/*.css",
```
jbadan, line 23: I'm not sure when `style:theming:postCSS` was added, but it's not necessary

stories/action-bar/action-bar.stories.js
```diff
@@ -12,7 +7,8 @@ export default {
 - Main Actions for the page
 
 For the page title, a Title component is utilized that is an H1 but visually styled as an H3. As the first heading on the page, for accessibility purposes it must be an H1. See the Title component page for further information on semantic versus visual heading levels.`,
-        tags: ['non-f3', 'a11y', 'theme']
+        tags: ['non-f3', 'a11y', 'theme'],
+        components: ['button', 'icon', 'title', 'action-bar', 'menu', 'popover']
```
jbadan, line 11: All css files needed should be added to this array in each file.

🚨 🚨  ** If a component is used in the markup and not present in this list, styling may look correct at first, but once you navigate from a different page using a different theme, you'll see that this page doesn't have it's own unique instance tied ot the theme provider**. I had to audit all the components and I think I've got most of them, but probably missed a bunch.

stories/generictile/generic-tile.css
```diff
@@ -1,3 +0,0 @@
-.fd-tile {
```
jbadan, line 1: We should never be overriding any `fd-*` classes for documentation. Add a custom `fddocs-` class instead.

stories/generictile/generic-tile.stories.js
```diff
@@ -1,150 +1,44 @@
...
-                    <a href="#line-tile" class="toc-link node-name--H3 " target="_self">
-                        Line Tile
-                    </a>
-                </li>
-            </ol>
-        </li>
-    </ol>
-</nav>
-
+export default {
```
jbadan, line 1: This file had to be rewritten, theming is not supported in mdx files.

stories/object-status/custom.scss
```diff
@@ -1,4 +0,0 @@
-.fd-object-status,
-.fd-inverted-status {
-  margin-right: 20px;
-}
```
jbadan, line 4: Same here - no `fd-` custom styles for documentation should be added.

stories/rating-indicator/rating-indicator.stories.js
```diff
@@ -13,9 +11,10 @@ Although the maximum amount is 7 for the icons or images, it is highly recommend
 
 ##Usage
 Use the rating indicator in forms, tables, or in a dialog box.
-        `
-    },
-    tags: ['f3', 'a11y', 'theme']
+        `,
+        tags: ['f3', 'a11y', 'theme'],
+        components: ['rating-indicator']
+    }
```
jbadan, line 17: Note: `tags` and `components` belongs _inside_ `parameter` block or else it won't work

stories/rating-indicator/custom.scss
```diff
@@ -1,10 +0,0 @@
-.example-container {
```
jbadan, line 1: These should be reused in `custom.scss` inside `.storybook` folder instead of per component.

stories/object-number/object-number.stories.js
```diff
@@ -43,63 +41,69 @@ Make sure that the object number is properly described and semantically understa
...
-<span class="fd-object-number fd-object-number--large fd-object-number--positive">
-    <span class="fd-object-number__text">1,000.00</span><span class="fd-object-number__unit">EUR</span>
-</span>
-<span class="fd-object-number fd-object-number--large fd-object-number--informative">
-    <span class="fd-object-number__text">100.00</span><span class="fd-object-number__unit">EUR</span>
-</span>
-<span class="fd-object-number fd-object-number--large">
-    <span class="fd-object-number__text">999.00</span><span class="fd-object-number__unit">EUR</span>
-</span>
+<div class="fddocs-container">
```
jbadan, line 68: Most of these diffs are due to adding `<div class="fddocs-container">` because I got rid of the docs override styling for these components.

stories/object-status/object-status.stories.js
```diff
@@ -349,64 +279,89 @@ export const inverted = () => `
...
+    </span>
+    <span class="fd-object-status fd-object-status--inverted fd-object-status--link fd-object-status--indication-6">
+        <span class="fd-object-status__text">Indication6</span>
+    </span>
+    <span class="fd-object-status fd-object-status--inverted fd-object-status--link fd-object-status--indication-7">
+        <span class="fd-object-status__text">Indication7</span>
+    </span>
+    <span class="fd-object-status fd-object-status--inverted fd-object-status--link fd-object-status--indication-8">
+        <span class="fd-object-status__text">Indication8</span>
+    </span>
```
jbadan, line 365: I removed some of the `rtl` examples in here - once they were using layouts instead of `br` and override styling, rtl doesn't really mean anything with this component.

.storybook/static/HCB_css_variables.css
```diff
@@ -1,494 +0,0 @@
-/**
```
jbadan, line 1: I've deleted all these storybook/static files a couple times now 🤣  They keep sneaking back in. We reference all of these from `sap-theming` now.

droshev (approved):

.storybook/main.js
```diff
@@ -23,6 +23,17 @@ module.exports = {
...
     });
+    config.module.rules.push({
+      test: /\.css?$/,
+      use: [
+        {
+            loader: 'style-loader',
+            options: {
+                injectType: 'lazyStyleTag'
+            }
+        }
```
greg-a-smith, line 34: Does the webpack rule also need a `css-loader`?  🤷‍♂️

README.md
```diff
@@ -34,6 +34,12 @@ npm install fundamental-styles --save
 
 > NOTE: We only distribute compiled CSS for each component, not the full project or HTML for specific components. 
 
+### Distributed CSS Files
+
+The distrubtion folder contains several css files per component: a default file and one file per supported theme (`sap_fiori_3`, `sap_fiori_3_dark`, `sap_fori_3_light_dark`, `sap_fiori_3_hcb`, `sap_fori_3_hcw`). Each themed file includes fallback support for css variables in IE11 for the corresponding theme.
```
greg-a-smith, line 39: Should we note that the default version of each file does have fallbacks for IE11, but they are hard-coded to Fiori Light?

README.md
```diff
@@ -34,6 +34,12 @@ npm install fundamental-styles --save
 
 > NOTE: We only distribute compiled CSS for each component, not the full project or HTML for specific components. 
 
+### Distributed CSS Files
+
+The distrubtion folder contains several css files per component: a default file and one file per supported theme (`sap_fiori_3`, `sap_fiori_3_dark`, `sap_fori_3_light_dark`, `sap_fiori_3_hcb`, `sap_fori_3_hcw`). Each themed file includes fallback support for css variables in IE11 for the corresponding theme. 
+
+If you project does not need to support theming in IE11, use the default files.
```
greg-a-smith, line 41: Nit: "If your project..."

greg-a-smith (changes requested):
Aside from a few comments, I am unable to get the story pages to load properly.  When storybook starts, I see the Overview page just fine:
Screen Shot 2020-10-07 at 11 19 38 PM

However, when I click on any of the other story links, I see the page with no styling (and no header where the theme dropdown is located):
Screen Shot 2020-10-07 at 11 19 42 PM

Am I missing something?  I pulled the latest code, did `npm install`, then `npm start`.

config/postcss-add-fallback.js
bcullman: taking over under  💵 on when `dark_light` appears...

jbadan:
@greg-a-smith 
> Aside from a few comments, I am unable to get the story pages to load properly. When storybook starts, I see the Overview page just fine:

> However, when I click on any of the other story links, I see the page with no styling (and no header where the theme 
> Am I missing something? I pulled the latest code, did `npm install`, then `npm start`.

Yes, you need to click on "Docs" , the canvas tab is meant to just be a playground for each story. 
Screen Shot 2020-10-08 at 6 18 47 AM

But it is definitely a 🐛  that the styles are not loaded by default in the canvas tab as well. It's because I'm loading the styles in the DocsPage wrapper, which doesn't exist in Canvas. For React, this will be easier because we can wrap each story in ta ThemeProvider, but with storybook html it's more challenging.

.storybook/main.js
jbadan: It hasn't had any problems without it 🤷

greg-a-smith (approved):
Looks good. 🚢

PR #1756, opened 2020-10-07
fix: remove deprecated localization editor

droshev:
## Description

BREAKING CHANGE:
remove deprecated localization editor

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation


jbadan (approved):
🔥

jbadan:
🎗️ remember to mark as a breaking change

Issue #1709, opened 2020-09-25
Slider - implement styles


manjunathanagaraj:

manjunathanagaraj:
To be assigned to Ilya

PR #1761, opened 2020-10-08
feat: introduce Slider

IlyaSurmay:
## Related Issue
Closes SAP/fundamental-styles#1709

## Description
Introduce Slider styles

## Screenshots

image
image
image

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation


droshev:
the numbers don't look centered with the slider tracks
Screen Shot 2020-10-08 at 3 34 23 PM

src/slider.scss
```diff
@@ -0,0 +1,132 @@
...
+    @include fd-reset();
+
+    height: var(--fdSlider_Track_Height);
+    width: 100%;
+    background-color: var(--fdSlider_Track_Background);
+    border-radius: var(--fdSlider_Track_BorderRadius);
+    position: relative;
+    z-index: 1;
+
+    &-range {
```
InnaAtanasova, line 34: this block can't be here in the way you use it. If it's an element (like you used it) it should be outside of the `&__track` block since it's a different element.
If you want to keep it here you need to make it a modifier class and rename it to `--range`. In this case you will have to use both classes on your div: 
```
<div class="fd-slider__track fd-slider__track--range"></div>
```

In this case inside the `&--track` block you don't need the reset, the position: relative, border-radius, etc., as they will come from `fd-slider__track ` class. That's why you make a modifier class, to modify stuff from the base class. Which approach you will choose is up to you :)

src/slider.scss
```diff
@@ -0,0 +1,132 @@
...
+    position: absolute;
+    width: 100%;
+    top: -0.4rem;
+  }
+
+  &__ticks {
+    @include fd-reset();
+
+    display: flex;
+    justify-content: space-between;
```
InnaAtanasova, line 57: there are mixins for the flex. Check `fd-flex()`

src/slider.scss
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  &__labels {
+    @include fd-reset();
+
+    display: flex;
+    justify-content: space-between;
```
InnaAtanasova, line 64: You can check the flex mixins in the mixins file

src/slider.scss
```diff
@@ -0,0 +1,132 @@
...
+    position: relative;
+    font-size: var(--sapFontSmallSize);
+    color: var(--sapContent_LabelColor);
+  }
+
+  &__handle {
+    @include fd-reset();
+
+    width: 1.625rem;
+    height: 1.625rem;
```
InnaAtanasova, line 93: can you put these in variables?

src/slider.scss
```diff
@@ -0,0 +1,132 @@
...
+    cursor: pointer;
+    border-radius: 50%;
+    z-index: 1;
+    background-color: var(--fdSlider_Handle_Background);
+    border: 0.125rem solid var(--fdSlider_Handle_Border_Color);
+
+    &::before {
+      content: '';
+      width: 2.75rem;
+      height: 2.75rem;
```
InnaAtanasova, line 106: can you put these in variables?

src/slider.scss
```diff
@@ -0,0 +1,132 @@
...
+
+    @include fd-active() {
+      background: var(--fdSlider_Handle_Background_Active);
+      border-color: var(--fdSlider_Handle_Border_Color_Active);
+    }
+
+    @include fd-screen(l) {
+      top: -0.5rem;
+      width: 1.25rem;
+      height: 1.25rem;
```
InnaAtanasova, line 124: these should be in variables

src/slider.scss
```diff
@@ -0,0 +1,132 @@
...
+
+      &::before {
+        width: 2rem;
+        height: 2rem;
```
InnaAtanasova, line 128: same as above

InnaAtanasova (changes requested):

stories/slider/slider.stories.js
```diff
@@ -0,0 +1,127 @@
...
+ */
+
+export const range = () => `
+<div style="width: 50%">
+    <div class="fd-slider" role="slider">
+        <div class="fd-slider__wrapper">
+            <div class="fd-slider__track">
+                <div class="fd-slider__track-range" style="width: 40%;left: 40%"></div>
+                <div class="fd-slider__handle" style="left: 40%;"></div>
+                <div class="fd-slider__handle" style="left: 80%;"></div>
```
InnaAtanasova, line 40: why do you add inline styling here? Our components should have inline styling only for better representation in the documentation (usually it's applied on the container div).

src/theming/sap_fiori_3.scss
```diff
@@ -31,4 +31,19 @@
...
+  --fdSlider_Tick_Height: 1rem;
+  --fdSlider_Track_Background: var(--sapField_BorderColor);
+  --fdSlider_Track_Active_Background: var(--sapActiveColor);
+  --fdSlider_Track_Active_Border: var(--sapActiveColor);
+  --fdSlider_Handle_Background: var(--sapButton_Background);
+  --fdSlider_Handle_Border_Color: var(--sapField_BorderColor);
+  --fdSlider_Handle_Background_Hover: var(--sapButton_Hover_Background);
+  --fdSlider_Handle_Border_Color_Hover: var(--sapButton_Hover_BorderColor);
+  --fdSlider_Handle_Background_Active: var(--sapButton_Active_Background);
+  --fdSlider_Handle_Border_Color_Active: var(--sapButton_Active_BorderColor);
```
jbadan, line 48: Custom css variables are meant for situations where high contrast has a different value than light/dark and sap-theming repo has not defined a variable for it. A lot of these seem to be the same in all themes? If so, just use the `var(--sap*)` variable in `slider.scss`

For example, ` fdSlider_Track_Background`, `fdSlider_Handle_Background`, `fdSlider_Handle_Border_Color`, `fdSlider_Handle_Background_Hover`, `fdSlider_Handle_Border_Color_Hover`, `fdSlider_Handle_Background_Active`, `fdSlider_Handle_Border_Color_Active` should be removed and just use the sap variable in their place in the code. 

The goal is to keep these theming files as small as possible as we slowly work these values back into sap theming repo - we ideally want to have 0 custom css variables in this repository.

stories/slider/slider.stories.js
jbadan: I feel like the inline css might be ok _for this component only_. It's being used to display the different states of the slider without having to go through the trouble of setting up the interaction to work.

We should take the `<div style="width: 50%">` off the wrapper and use one of the docs classes instead though. You can find some containers here - https://github.com/SAP/fundamental-styles/blob/master/.storybook/custom/custom.scss or add another if none of those fit the needs of this component.

stories/slider/slider.stories.js
```diff
@@ -0,0 +1,127 @@
...
+</div>
+`;
+
+/**
+ * Range slider
+ */
+
+export const range = () => `
+<div style="width: 50%">
+    <div class="fd-slider" role="slider">
```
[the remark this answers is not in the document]
meganaconley: The slider in the example is also an input. It's possible to have it be a div, but we'll need a manual tab-index as well. Here's an example of a slider made up of divs: https://www.w3.org/TR/wai-aria-practices/examples/slider/slider-1.html

Specifically we need the aria-valuemin/valuemax/valuenow attributes, the tabindex, and an aria-label.

I have not been able to find an example of a range slider that uses a div, however; they all use an input.

InnaAtanasova (commented):
Removing the requested changes as I am going on vacation

stories/slider/slider.stories.js
IlyaSurmay: @jbadan that's right, inline styling on slider handles was used to give them different positions so that way user sees a more or less real-world example of how range slider would look like. Of course, in component implementation these values are going to be set dynamically.

I'll use helper class from custom scss file to set wrapper's width, thanks

IlyaSurmay:
> the numbers don't look centered with the slider tracks
> Screen Shot 2020-10-08 at 3 34 23 PM

that's right, they aren't aligned perfectly but I don't think this needs (and can be) done with CSS only because labels position depends on many factors like what text is used, label's width, how many labels there are
I believe that each label position should be calculated on the component level in JS implementation as CSS doesn't provide enough possibilities to do that, please correct me if I'm wrong. Also, Fiori design guidelines mentions that labels should have a responsive mechanism which will be activated if there is not enough horizontal space to display all the labels.

For demo purposes I think I can tweak CSS or add inline styling to align labels with ticks if that's necessary.

jbadan:
@droshev We might need to take a look at the CLA bot - did it really take _4 days_ to check this pr?

jbadan (changes requested):
Requesting changes until a11y feedback is resolved

src/theming/sap_fiori_3.scss
IlyaSurmay: thanks! I removed variables that are the same for all themes

src/slider.scss
```diff
@@ -0,0 +1,139 @@
...
+    background-color: var(--fdSlider_Track_Active_Background);
+    border: 0.0625rem solid var(--fdSlider_Track_Active_Border);
+  }
+
+  &__tick-wrapper {
+    @include fd-reset();
+
+    position: absolute;
+    width: 100%;
+    top: -0.4rem;
```
JKMarkowski, line 54: Let's change it to `top: 50%` and `transform: translateY(-50%);` for proper center

src/slider.scss
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  &__handle {
+    @include fd-reset();
+
+    width: $handle-mobile-dimensions;
+    height: $handle-mobile-dimensions;
+    position: absolute;
+    top: -0.7rem;
+    margin-left: -0.5rem;
```
JKMarkowski, line 103: You don't need those values. center it by `top: 50%` and `transform: translate(-50%, -50%)`;

src/slider.scss
```diff
@@ -0,0 +1,139 @@
...
+      background: var(--sapButton_Hover_Background);
+      border-color: var(--sapButton_Hover_BorderColor);
+    }
+
+    @include fd-active() {
+      background: var(--sapButton_Active_Background);
+      border-color: var(--sapButton_Active_BorderColor);
+    }
+
+    @include fd-screen(l) {
```
JKMarkowski, line 128: Please change `fd-screen` to `--lg` modifier

src/slider.scss
```diff
@@ -0,0 +1,139 @@
...
+      top: -0.5rem;
```
JKMarkowski, line 129: Also remove it, because it's going to be centered anyway

src/slider.scss
```diff
@@ -0,0 +1,139 @@
...
+    height: var(--fdSlider_Tick_Height);
+    background-color: var(--sapField_BorderColor);
+
+    &--in-range {
+      background-color: var(--fdSlider_Track_Active_Border);
+    }
+  }
+
+  &__label {
+    position: relative;
```
JKMarkowski, line 91: Let's add fd-reset there

JKMarkowski (changes requested):
Hello @IlyaSurmay Great Job! I left some comment regarding mostly centering elements. Use `left/top` + `translate` to make sure that element will be always centered.
Also mobile mode example is missing there.

src/slider.scss
IlyaSurmay: this will only work when tick labels aren't displayed, that's why I added just `top` here

stories/slider/slider.stories.js
```diff
@@ -0,0 +1,146 @@
...
+            </div>
+        </div>
+    </div>
+</div>
+`;
+
+export const ticks = () => `
+<div class="slider-container">
+    <div class="fd-slider" tabindex="0" role="slider" aria-label="slider with ticks" aria-valuemin="1" aria-valuemax="100" aria-valuenow="50">
+        <div class="fd-slider__wrapper">
```
dimamarksman, line 49: Maybe it should be `fd-slider__inner` since it lays in `fd-slider` and doesn't wrap it?

src/slider.scss
```diff
@@ -0,0 +1,144 @@
...
+    }
+  }
+
+  &__label {
+    @include fd-reset();
+
+    position: relative;
+    font-size: var(--sapFontSmallSize);
+    color: var(--sapContent_LabelColor);
+  }
```
dimamarksman, line 100: Can we center labels doing this? 
image

dimamarksman (commented):
Looks good!
Couple of questions.

src/slider.scss
IlyaSurmay: awesome solution, thank you!

stories/slider/slider.stories.js
```diff
@@ -0,0 +1,146 @@
...
+
+export const range = () => `
+<div class="slider-container">
+    <div class="fd-slider" tabindex="0" role="slider" aria-label="range slider" aria-valuemin="1" aria-valuemax="100" aria-valuenow="40">
+        <div class="fd-slider__inner">
+            <div class="fd-slider__track">
+                <div class="fd-slider__track-range" style="width: 40%;left: 40%"></div>
+                <div class="fd-slider__handle" style="left: 40%;"></div>
+                <div class="fd-slider__handle" style="left: 80%;"></div>
+            </div>
```
jacobdevera, line 40: Similar to the single-handle examples, the handles themselves should be the ones with role `slider` with the associated ARIA attributes as they would be treated as two separate tabbable elements. [Example](https://www.w3.org/TR/wai-aria-practices/examples/slider/multithumb-slider.html)

```suggestion
    <div class="fd-slider">
        <div class="fd-slider__inner">
            <div class="fd-slider__track">
                <div class="fd-slider__track-range" style="width: 40%;left: 40%"></div>
                <div class="fd-slider__handle" tabindex="0" role="slider" aria-label="min slider" aria-valuemin="1" aria-valuemax="80" aria-valuenow="40" style="left: 40%;"></div>
                <div class="fd-slider__handle" tabindex="0" role="slider" aria-label="max slider" aria-valuemin="40" aria-valuemax="100" aria-valuenow="80" style="left: 80%;"></div>
            </div>
```

stories/slider/slider.stories.js
```diff
@@ -0,0 +1,146 @@
...
+export const basic = () => `
+<div class="slider-container">
+    <div class="fd-slider" tabindex="0" role="slider" aria-label="slider" aria-valuemin="1" aria-valuemax="100" aria-valuenow="50">
+        <div class="fd-slider__inner">
+            <div class="fd-slider__track">
+                <div class="fd-slider__track-range" style="width: 50%;"></div>
+                <div class="fd-slider__handle" style="left: 50%;"></div>
+            </div>
+        </div>
+    </div>
```
jacobdevera, line 24: The handle itself should contain the `slider` attributes as it will be the element to receive focus.

```suggestion
    <div class="fd-slider">
        <div class="fd-slider__inner">
            <div class="fd-slider__track">
                <div class="fd-slider__track-range" style="width: 50%;"></div>
                <div class="fd-slider__handle" tabindex="0" role="slider" aria-label="slider" aria-valuemin="1" aria-valuemax="100" aria-valuenow="50" style="left: 50%;"></div>
            </div>
        </div>
    </div>
```

It's fine that the role is not on the main `fd-slider` element; notice in [this example](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_slider_role#Example_2_Text_Values) how the "slider" class is used as an outer container but the handle receives the actual slider role.

jacobdevera (changes requested):
Few adjustments to where slider attributes should be placed. Thanks for taking accessibility into consideration!

stories/slider/slider.stories.js
IlyaSurmay: Thanks! I moved ARIA attributes to handle elements.

However, it seems to me that it's different for other libraries, for instance, Angular Material Slider has ARIA attributes on the host container and ng5-slider on handle. Anyway, let's stick to official guidelines from MDN.

JKMarkowski (approved):
LGTM

jbadan (approved):
Looks great to me (pending sign off from a11y team member)! 🚢

jacobdevera (approved):
LGTM 👍

stories/slider/slider.stories.js
jacobdevera: It's definitely possible to keep the attributes on the host container given appropriate CSS and proper focus management, though it's a bit harder, especially with multiple handles.

prsdthkr (changes requested):
I'm unable to see the focus indicator when tabbing through the stories.

stories/slider/slider.visual.js
```diff
@@ -0,0 +1,17 @@
+import * as stories from './slider.stories.js';
```
jbadan, line 1: We just moved to automated visual testing via chromatic 🎉 , so you can remove this file

IlyaSurmay: Cool :) 
removed

prsdthkr:
Should focus align to this slider example seen in the spec here: https://wiki.wdf.sap.corp/wiki/pages/viewpage.action?pageId=2046198993 ?

image

jacobdevera:
> Should focus align to this slider example seen in the spec here: https://wiki.wdf.sap.corp/wiki/pages/viewpage.action?pageId=2046198993 ?
If we do it this way, wouldn't it be impossible to manage focus between multi-handle sliders?

prsdthkr:
> 
> > Should focus align to this slider example seen in the spec here: https://wiki.wdf.sap.corp/wiki/pages/viewpage.action?pageId=2046198993 ?
> 
> If we do it this way, wouldn't it be impossible to manage focus between multi-handle sliders?

Good point.

On the handle, it should align to the standard Fiori focus outline on all zoom levels.
image

I'm seeing a different focus indicator on zoom level 125 in Chrome 🤔 :
image

IlyaSurmay:
> > > Should focus align to this slider example seen in the spec here: https://wiki.wdf.sap.corp/wiki/pages/viewpage.action?pageId=2046198993 ?
> > 
> > 
> > If we do it this way, wouldn't it be impossible to manage focus between multi-handle sliders?
> 
> Good point.
> 
> On the handle, it should align to the standard Fiori focus outline on all zoom levels.
> image
> 
> I'm seeing a different focus indicator on zoom level 125 in Chrome 🤔 :
> image

I couldn't reproduce this in latest Chrome with zoom 125%, outline looks exactly the same
Снимок экрана 2020-10-16 в 12 53 49

prsdthkr:
@IlyaSurmay Cool, I think it is just me.

Also, don't we need an example for Element F and G as per spec here: https://wiki.wdf.sap.corp/wiki/pages/viewpage.action?pageId=2033852394 ?
image

IlyaSurmay:
> @IlyaSurmay Cool, I think it is just me.
> 
> Also, don't we need an example for Element F and G as per spec here: https://wiki.wdf.sap.corp/wiki/pages/viewpage.action?pageId=2033852394 ?
> image

@prsdthkr 
I think examples with popover and input aren't necessary in this case because they aren't related to slider's styles, they're more like an integration of different Fiori components. Also, It'd mess with slider's markup in examples and make it harder to understand because each implementation (ng, vue, react) will use its own way of including popover/input in slider. 

And one more thing is that including markup of other components in slider is adding additional complexity because if there are any breaking changes in popover/input markup we'll have to change it in sliders' examples as well.

prsdthkr:
I think it is okay to compose styles from various components to achieve the spec. We do it with many fd-styles components. It reduces ambiguity and standardizes the implementations. However, I'll defer to @droshev and team to take the call on this.

prsdthkr:
Also, Range (Right to Left) seems to be misstyled: https://5f85e3904b19d600221872af-lfemkaaqsw.chromatic.com/?path=/story/visual-light-slider--slider

image

IlyaSurmay:
> Also, Range (Right to Left) seems to be misstyled: https://5f85e3904b19d600221872af-lfemkaaqsw.chromatic.com/?path=/story/visual-light-slider--slider
> 
> image

@prsdthkr 
That's an interesting one. Selected range element is positioned with using `left` property and for basic slider it works fine in both LTR/RTL modes but for range slider the value of `left` needs to be changed, I'm not sure this can be achieved with CSS only because the offset of selected range element has to be recalculated based on the position of handles. Of course this won't be a big issue in JS implementations but in CSS it is.

Any ideas on this, am I missing something?

JKMarkowski (approved):
LGTM, great job

IlyaSurmay:
@prsdthkr 
I fixed this issue with wrong offset of selected range element (for this demo only), please take a look

JKMarkowski:
> > Also, Range (Right to Left) seems to be misstyled: https://5f85e3904b19d600221872af-lfemkaaqsw.chromatic.com/?path=/story/visual-light-slider--slider
> > image
> 
> @prsdthkr
> That's an interesting one. Selected range element is positioned using `left` property and for basic slider it works fine in both LTR/RTL modes but for range slider the value of `left` needs to be changed, I'm not sure this can be achieved with CSS only because the offset of selected range element has to be recalculated based on the position of handles. Of course this won't be a big issue in JS implementations but in CSS it is.
> 
> Any ideas on this, am I missing something?

The thing you can do there is to add class with left: 40% and width: 40%. And change `left` to `-20%`, when there is RTL

stories/slider/slider.stories.js
```diff
@@ -0,0 +1,146 @@
...
+ * Basic slider
+ */
+
+export const basic = () => `
+<div class="slider-container">
+    <div class="fd-slider">
+        <div class="fd-slider__inner">
+            <div class="fd-slider__track">
+                <div class="fd-slider__track-range" style="width: 50%;"></div>
+                <div class="fd-slider__handle" tabindex="0" role="slider" aria-label="slider" aria-valuemin="1" aria-valuemax="100" aria-valuenow="50" style="left: 50%;"></div>
```
meganaconley, line 21: After looking into this some more, this should really be an `input`. Using `div`s will put a large burden on the implementation libraries to handle keyboard and screen reader support. Mobile screen readers in particular seem to struggle with this. Using an `input` would make the experience consistent across all platforms and assistive technologies.

See the following examples:
https://butterpep.com/slider-accessibility.html
https://thepaciellogroup.github.io/AT-browser-tests/test-files/input-range.html
https://www.w3.org/TR/wai-aria-1.1/#slider

meganaconley (changes requested):

stories/slider/slider.stories.js
IlyaSurmay: > After looking into this some more, this should really be an `input`. Using `div`s will put a large burden on the implementation libraries to handle keyboard and screen reader support. Mobile screen readers in particular seem to struggle with this. Using an `input` would make the experience consistent across all platforms and assistive technologies.
> 
> See the following examples:
> https://butterpep.com/slider-accessibility.html
> https://thepaciellogroup.github.io/AT-browser-tests/test-files/input-range.html
> https://www.w3.org/TR/wai-aria-1.1/#slider

I got your point. However, I think that using `div` is OK in this case for a variety of reasons:
- poor styling abilities for `input[type="range"]` and inability to style in according to Fiori spec without both CSS/JS hacks
- implementing a range slider on native `input[type="range"]` will be way more complicated compared to `div`s
- handling keyboard/screen reader support and adding required `ARIA` attributes doesn't seem to be a large burden to me because frameworks provide in-built support of such functionality (Angular/React/Vue)
- existing implementation of an old Fiori slider is using `div`s (https://sapui5.hana.ondemand.com/1.82.4/#/entity/sap.m.RangeSlider/sample/sap.m.sample.RangeSlider)
- in addition, I couldn't find a popular 3rd party slider lib for Angular/React/Vue that uses `input[type="range"]`, IMO this is because of the first three reasons

> - https://material.angular.io/components/slider/overview
> - https://angular-slider.github.io/ng5-slider/
> - https://www.telerik.com/kendo-angular-ui/components/inputs/slider/
> - https://www.primefaces.org/primeng/showcase/#/slider
> - https://material-ui.com/components/slider/
> - https://zillow.github.io/react-slider/

meganaconley: I'm not sure that other 3rd party libraries not doing this is necessarily a good reason. We had accessibility issues with a popular React datepicker that lead to us making improvements to the one here, porting it to fundamental-react, and using that instead, for example. Accessibility isn't always their number one concern. There are several UI5 components with accessibility issues as well.

And yes, React has tools that can help add better keyboard functionality but it is still a manual process to implement it; it doesn't come for free. It's unfortunate that the spec didn't take into consideration the native html markup that is built to support this and this design can't easily be implemented with the correct markup. 

If it's not possible to use inputs, the second best alternative is to ensure the documentation explicitly states the expected behavior for voice readouts and keyboard handling.

IlyaSurmay: Implementation libraries would still have to implement keyboard functionality regardless of what's used,`div` or `input` because spec describes such functionality as `jump` (incrementing/decrementing slider's value by a bigger step when `shift + arrow key` is pressed) which is something that native `input` doesn't provide.

Slider is an easier component to implement than a datepicker and I don't think adding a11y functionality will be a big problem (because of the fact that a lot of other libraries is using this approach and their implementation of a11y is pretty much the same which makes me think a11y isn't that hard to implement as it may seems)

Also I found W3 ARIA examples for sliders (both single and range) and they're using `div` with no mention that `input[type="range"]` is more preferable. 
https://www.w3.org/TR/wai-aria-practices-1.1/examples/slider/slider-1.html
https://www.w3.org/TR/wai-aria-practices-1.1/examples/slider/multithumb-slider.html

I added a section in the docs that describes what attributes need to be added in order to make the slider accessible.

image

prsdthkr (approved):
QA LGTM after accessibility callouts.

meganaconley (approved):

Issue #1005, opened 2020-05-12
'High Contrast Black' theme Form (Radio buttons and default Checkbox) checked state color contrast


prsdthkr:
**Describe the bug**
The Form (Radio buttons and default Checkbox) checked state color contrast ratio does not meet the minimum criteria i.e. `4.5:1`

> What version of the Fundamental Library Styles are you using?
v0.9.1-rc.0

> What front-end framework are you implementing in, e.g., Angular, React, etc.?
React

**To Reproduce**
1. Go to https://deploy-preview-928--fundamental-styles.netlify.app/components/form.html
2. Check the Radio buttons and Checkbox checked state icon color contrast with the background

**Expected behavior**
The Form (Radio buttons and default Checkbox) checked state color contrast ratio should be 4.5:1 or above. Ideally 7:1 or above for WCAG AAA compliance. `#fff` would be the best choice.

**Screenshots**
![image](https://user-images.githubusercontent.com/43764832/81745843-6161fa80-945a-11ea-8c18-4cef9580582a.png)
![image](https://user-images.githubusercontent.com/43764832/81746549-8acf5600-945b-11ea-9079-1b2709e3623c.png)
![image](https://user-images.githubusercontent.com/43764832/81745901-75a5f780-945a-11ea-8346-3117ecccfc68.png)

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: Chrome 
 - Version: 81.0.4044.129 (Official Build) (64-bit)

prsdthkr:
![image](https://user-images.githubusercontent.com/43764832/95624460-c3d01e80-0a2b-11eb-832d-9c65622a9420.png)

PR #1763, opened 2020-10-09
docs: add additional selector for hcm radio color

Mike-Diaz:
## Related Issue
Closes SAP/fundamental-styles#1005

## Description
Radio button checked state icon does not have HCM styles applied due to a lack of specificity coming from the selectors. Additional style rules have been added to all HCM styles where needed.

### Before Chrome/Mac:
Screen Shot 2020-10-08 at 5 10 24 PM
Screen Shot 2020-10-08 at 5 10 33 PM
Screen Shot 2020-10-08 at 5 10 41 PM
Screen Shot 2020-10-08 at 5 10 48 PM

### After Chrome/Mac:
Screen Shot 2020-10-08 at 5 14 35 PM
Screen Shot 2020-10-08 at 5 14 41 PM
Screen Shot 2020-10-08 at 5 14 48 PM
Screen Shot 2020-10-08 at 5 14 53 PM

### Before IE11/Windows:
Capture
Capture2
Capture3
Capture4

### After IE11/Windows
Capture
Capture2
Capture3
Capture4

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/.storybook/static/windows-hcm-1.css
+++ b/.storybook/static/windows-hcm-1.css
@@ -56,3 +56,7 @@ textarea:disabled {
 :focus {
   outline: 1px solid white !important;
 }
+
+input[type=radio]:checked + label::after{
+  background:  yellow !important;
+}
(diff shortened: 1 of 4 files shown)


jbadan (approved):
Should this just be "docs" because it's a change only for storybook?

InnaAtanasova (approved):

Issue #1764, opened 2020-10-09
Object Attribute - implement styles


sKudum:
Implement styles for Object Attribute Component as per Fiori3.
<img width="311" alt="2020-10-09_10-07-20" src="https://user-images.githubusercontent.com/53487468/95543855-3a165600-0a17-11eb-8897-a3ed16f19456.png">

PR #1765, opened 2020-10-09
feat: introduce Object Attribute

sKudum:
## Related Issue

Closes SAP/fundamental-styles#1764

## Description
Object Attribute styles file is created as per fiori3 standards.

- [x] All values are in `rem`
- [x] RTL support
3. Testing


jbadan:
Hi @sKudum! Is there a reason several items were removed from the pr template checklist? Are you planning on adding examples/documentation? If so, can we convert this to a "Draft" pull request.

sKudum:
> Hi @sKudum! Is there a reason several items were removed from the pr template checklist? Are you planning on adding examples/documentation? If so, can we convert this to a "Draft" pull request.

Hi @jbadan , Forwarded mail regarding the same discussion as per the instructions given in mail, created only styles for it

jbadan:
@sKudum Sounds good - I didn't know that :)

droshev (changes requested):
@sKudum we have another component that is not documented but we have testing page 
https://github.com/SAP/fundamental-styles/blob/master/stories/numeric-content/numeric-content.stories.js

src/object-attribute.scss
```diff
@@ -0,0 +1,47 @@
...
+
+    &:hover {
+      @include fd-hover() {
+        text-decoration: $fd-object-attribute-text-decoration-underline;
+        color: var(--sapLink_Hover_Color);
+      }
+    }
+
+    &[disabled],
+    &[aria-disabled="true"] {
```
InnaAtanasova, line 35: there's fd-disabled mixin you canuse

src/object-attribute.scss
```diff
@@ -0,0 +1,47 @@
...
+      opacity: var(--sapContent_DisabledOpacity);
+      cursor: not-allowed;
+    }
+
+    &:focus {
+      @include fd-focus();
+      @include fd-fiori-focus($offset: -0.0625rem);
```
InnaAtanasova, line 42: are you sure this is working in IE11?

src/object-attribute.scss
sKudum: fd-disabled is causing build failures, so I removed it.

sKudum:
> @sKudum we have another component that is not documented but we have testing page
> https://github.com/SAP/fundamental-styles/blob/master/stories/numeric-content/numeric-content.stories.js
> 
> Hi @droshev , Sure will do it as well, as It was mentioned in the mail to a similar way as badge and nested-list so did the same.

Hi @droshev , updated with examples as suggested.

JKMarkowski:
Hello @sKudum I can't see anything in dev page of object attribute 
https://deploy-preview-1765--fundamental-styles.netlify.app/?path=/docs/dev-object-attribute--text
Can you recheck it?

JKMarkowski:
@sKudum Yes I think you should enable it, but on dev section

jbadan:
@JKMarkowski @sKudum the items under dev are supposed to be disabled for the documentation website -that's why they're under a "Dev" umbrella. If we want them publicly exposed, they should be a Component. I think it was decided this component should live under "Dev", so we should just be reviewing this pr by running it locally and not relying on the netlify preview.

stories/object-attribute/object-attribute.visual.js
```diff
@@ -0,0 +1,17 @@
+import * as stories from './object-attribute.stories.js';
```
jbadan, line 1: We just moved to automated visual testing via chromatic 🎉 , so you can remove this file

sKudum:
Hi @jbadan , @JKMarkowski , @droshev , 

Please do let me know in case anything else needs to be done in this.So that I can proceed on this further.

stories/object-attribute/object-attribute.stories.js
```diff
@@ -0,0 +1,33 @@
...
+
+export const TextWithLink = () => `
+<div id="firsttext" class="fd-object-attribute" aria-label="Object Attribute">Object Attribute :
+<span id="textwithlink" class="fd-object-attribute--link" aria-label="Text With Link" role="link" tabindex="0">Text With Link</span></div>
+`;
+
+TextWithLink.parameters = { docs: { disable: true } };
+
+export const DisabledLink = () => `
+    <span id="disabledlink" class="fd-object-attribute fd-object-attribute--link" role="link" tabindex="0" disabled="true" aria-disabled="true" aria-label="Disabled Object Attribute linked Text">Disabled Object Attribute linked Text</span>
```
jacobdevera, line 22: The `disabled` attribute isn't supported for `span` nor an `a`. You can keep `aria-disabled` though either way.

```suggestion
    <span id="disabledlink" class="fd-object-attribute fd-object-attribute--link" role="link" tabindex="0" aria-disabled="true">Disabled Object Attribute linked Text</span>
```

stories/object-attribute/object-attribute.stories.js
```diff
@@ -0,0 +1,33 @@
...
+`;
+
+DisabledLink.parameters = { docs: { disable: true } };
+
+export const TruncatedText = () => `
+<div style="width: 120px;">
+    <span id="truncated" class="fd-object-attribute" tabindex="0" title="Truncate Very long Object Attribute Text" aria-label="Truncate Very long Object Attribute Text">Truncate Very long Object Attribute Text</span>
```
jacobdevera, line 29: `aria-label` is redundant if the text is fully in the DOM; screen readers can see the full text even if it's truncated for sighted users.
```suggestion
    <span id="truncated" class="fd-object-attribute" tabindex="0" title="Truncate Very long Object Attribute Text">Truncate Very long Object Attribute Text</span>
```

stories/object-attribute/object-attribute.stories.js
```diff
@@ -0,0 +1,33 @@
...
+
+<div class="fd-object-attribute" id="standardtext" aria-label="Standard Object Attribute Text">Standard Object Attribute Text</div>
+`;
+
+Text.parameters = { docs: { disable: true } };
+
+
+export const TextWithLink = () => `
+<div id="firsttext" class="fd-object-attribute" aria-label="Object Attribute">Object Attribute :
+<span id="textwithlink" class="fd-object-attribute--link" aria-label="Text With Link" role="link" tabindex="0">Text With Link</span></div>
```
jacobdevera, line 16: 1. Is there a reason this cannot instead be an `a`?
2. `aria-label` is redundant

```suggestion
<a href="https://link" id="textwithlink" class="fd-object-attribute--link">Text With Link</a></div>
```

jacobdevera (changes requested):

stories/object-attribute/object-attribute.stories.js
sKudum: Hi @jacobdevera , this should be in a position to trigger any action on user demand, such as click open a dialog or popup, alert message, or navigation etc. because of this requirement, we are not restricting it as <a here.

stories/object-attribute/object-attribute.stories.js
```diff
@@ -0,0 +1,33 @@
...
+
+export const TextWithLink = () => `
+<div id="firsttext" class="fd-object-attribute">Object Attribute :
+<span id="textwithlink" class="fd-object-attribute--link" role="link" tabindex="0">Text With Link</span></div>
+`;
+
+TextWithLink.parameters = { docs: { disable: true } };
+
+export const DisabledLink = () => `
+    <span id="disabledlink" class="fd-object-attribute fd-object-attribute--link" role="link" tabindex="0" aria-disabled=true>Disabled Object Attribute linked Text</span>
```
jacobdevera, line 22: `aria-disabled` needs quotes around `"true"` (even though it seems to properly resolve in the browser), otherwise this looks fine.

src/object-attribute.scss
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    &[disabled],
+    &[aria-disabled="true"] {
+      @include fd-disabled() {
+        opacity: var(--sapContent_DisabledOpacity);
+        pointer-events: none;
+      }
+    }
```
jacobdevera, line 41: You can omit these selectors, as they are already provided in the mixin.

```suggestion
    @include fd-hover() {
      text-decoration: underline;
      color: var(--sapLink_Hover_Color);
    }

    @include fd-focus();
    @include fd-fiori-focus($offset: -0.0625rem);

    @include fd-disabled() {
      opacity: var(--sapContent_DisabledOpacity);
      pointer-events: none;
    }
```

JKMarkowski (approved):
LGTM

jacobdevera (approved):
Awesome 👍

stories/object-attribute/object-attribute.stories.js
```diff
@@ -0,0 +1,33 @@
+export default {
+    title: 'Dev/Object Attribute',
+    components: ['object-attribute']
+};
+
+export const Text = () => `
+
+<div class="fd-object-attribute" id="standardtext">Standard Object Attribute Text</div>
+`;
```
prsdthkr, line 9: Remove first newline from each story to fix empty jest storyshots.
```suggestion
export const Text = () => `<div class="fd-object-attribute" id="standardtext">Standard Object Attribute Text</div>
`;
```

prsdthkr (changes requested):

stories/object-attribute/object-attribute.stories.js
prsdthkr: @sKudum the first new line needs to be removed too.

stories/object-attribute/object-attribute.stories.js
```diff
@@ -0,0 +1,27 @@
...
+    components: ['object-attribute']
+};
+
+export const Text = () => `<div class="fd-object-attribute" id="standardtext">Standard Object Attribute Text</div>
+`;
+
+Text.parameters = { docs: { disable: true } };
+
+export const TextWithLink = () => `<div id="firsttext" class="fd-object-attribute">Object Attribute :
+<span id="textwithlink" class="fd-object-attribute--link" role="link" tabindex="0">Text With Link</span></div>
```
prsdthkr, line 12: For the sake of semantics, all links should be an `<a>` with proper href. If we want generic actions an example for a button should be added.

stories/object-attribute/object-attribute.stories.js
```diff
@@ -0,0 +1,27 @@
...
+`;
+
+TextWithLink.parameters = { docs: { disable: true } };
+
+export const DisabledLink = () => `<span id="disabledlink" class="fd-object-attribute fd-object-attribute--link" role="link" tabindex="0" aria-disabled="true">Disabled Object Attribute linked Text</span>
```
prsdthkr, line 17: Not sure if disabled links should be focusable.

stories/object-attribute/object-attribute.stories.js
```diff
@@ -0,0 +1,27 @@
...
+`;
+
+DisabledLink.parameters = { docs: { disable: true } };
+
+export const TruncatedText = () => `<div style="width: 120px;">
+<span id="truncated" class="fd-object-attribute" tabindex="0" title="Truncate Very long Object Attribute Text">Truncate Very long Object Attribute Text</span>
```
prsdthkr, line 23: Not sure if this text should be focusable.

prsdthkr (changes requested):

katekozlowska (approved):
Looks good!

stories/object-attribute/object-attribute.stories.js
```diff
@@ -0,0 +1,26 @@
...
+    title: 'Dev/Object Attribute',
+    components: ['object-attribute', 'button']
+};
+
+export const Text = () => `<div class="fd-object-attribute" id="standardtext">Standard Object Attribute Text</div>
+`;
+
+Text.parameters = { docs: { disable: true } };
+
+export const TextWithLink = () => `<a id="textwithlink" href="#" class="fd-object-attribute--link" role="link" tabindex="0">Text With Link</a>
```
prsdthkr, line 11: `role="link"` is redundant on `<a>`

```suggestion
export const TextWithLink = () => `<a id="textwithlink" href="#" class="fd-object-attribute--link" tabindex="0">Text With Link</a>
```

prsdthkr:
@sKudum w.r.t. using buttons... I would argue that the spec is not correct if it specifies that a thing that looks like a link should perform actions like a button. That is fundamentally wrong IMO. However, in the interest of time, you can style this by using a `div` with a `role="button"`  and `onClick` as a separate example.

InnaAtanasova (approved):

prsdthkr (approved):
QA LGTM ✅

droshev (approved):

PR #1774, opened 2020-10-10
docs: introduce RTL select and remove previous RTL examples

droshev:
Introduce Directionality select on the doc website


.storybook/custom/components/DocsPage.js
```diff
@@ -41,7 +41,17 @@ const DocsPage = () => {
...
+    const previousDirectionality = useRef();
+
     useEffect(() => {
+        if(!previousDirectionality.current || previousDirectionality.current !== directionalityState){
+            document.getElementsByClassName('sbdocs-preview').forEach(item => {
+                item.childNodes[1].setAttribute('dir', directionalityState);
+            });
+            previousDirectionality.current = directionalityState;
+        }
+
```
jbadan, line 54: While you're here, would you mind removing the `console.log(item)` I accidentally left on line 50

.storybook/custom/components/Header.js
```diff
@@ -1,13 +1,17 @@
...
+const Header = ({onThemeChange, onDirectionalityChange}) => {
     return (
         <header className='fddocs-header'>
             <section className='fddocs-header--left'>
                 <span>Fundamental Styles</span>
             </section>
             <section className='fddocs-header--right'>
+                <select className='fddocs-select' onChange={(e) => onDirectionalityChange(e)}>
+                    <option value='ltr'>Left-to-Right</option>
+                    <option value='rtl'>Right-to-Left</option>
```
jbadan, line 13: Should these just be `LTR` and `RTL` or do we need to explicitly call out "Left-to-Right"? Maybe @jessmetivier knows the best way to communicate this

jbadan (approved):
Looks good to me 🚢  So much cleaner! Left a question for Jessie about word choice but other than that this looks great to me!

.storybook/custom/components/Header.js
droshev: @jbadan either way - whatever @jessmetivier say :)

.storybook/custom/components/DocsPage.js
droshev: done

JKMarkowski (approved):
Great Job! Works like a miracle, this is the thing we needed

.storybook/custom/components/DocsPage.js
jbadan: Thank you!

.storybook/custom/components/Header.js
jessmetivier: Since this is only a label and our users are both developers and designers, it needs to be explicitly stated.

jbadan:
@droshev Can you bring in master and update the visual tests - we just moved to chromatic

Issue #1710, opened 2020-09-25
Icon - Implement styles


manjunathanagaraj:
To be assigned to Oleksii Aleksandrov

skocharian:
https://app.gitbook.com/@fundamental-styles/s/docs/icon

manjunathanagaraj:
PR 1778 merging is paused. Need more discussions to proceed further. Therefore removing the milestone assignment.

droshev:
let's close this for now. If needed we can either reopen or create a new one

PR #1778, opened 2020-10-13
[WIP] feat: Icon component default styles and new class

EvilAlexei:
## Related Issue
Closes SAP/fundamental-styles#1710

## Description

BREAKING CHANGE:
Icon class changed from `.sap-icon` to `.fd-icon`

Changes:
- Renamed main class of Icon component from `sap-icon` to `fd-icon`
- Renamed all cases of `sap-icon` class to `fd-icon`
- Default styles according to a https://app.gitbook.com/@fundamental-styles/s/docs/icon#styling.
  - font-size: 1rem;
  - color: var(--sapContent_NonInteractiveIconColor)
  - background-color: var(--sapBackgroundColor);
- Added `Styles` section to stories
#### Note:
by documentation background-color should have different value `--sapContent_IconColor`, but this breaks design,
after discussion with @JKMarkowski it was decided to use `--sapBackgroundColor` as a temporary solution

### Before:
prev-sizes
prev-icons

### After:
new-sizes
new-styles
new-icons

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

BREAKING CHANGE:
Icon class changed from `.sap-icon` to `.fd-icon`


src/fd-icon.scss
```diff
@@ -0,0 +1,22 @@
+@import "./mixins";
+
+$block: #{$fd-namespace}-icon;
+
+[class*="#{$block}"] {
```
jbadan, line 5: We're trying to move away from `[class*="#{$block}"] `, can this just reference the block like on line 16?

jbadan (changes requested):
Does this need to be a new icon component? Could we not just replace https://github.com/SAP/fundamental-styles/blob/master/src/icon.scss + documentation so that we don't have 2 icons at once?

Other than those small nits this looks good to me!

EvilAlexei:
@jbadan at this moment I think we can't just replace it because class `.sap-icon` already in use in many places,
for example, the button component uses it as well as the icon component himself.
So to replace the class name there should be added changes in related styles and related components in fundamentals-ngx.
In my opinion, these changes are breaking changes and should be done in this way, like creating a similar component, and then replacing it.

src/fd-icon.scss
EvilAlexei: Changed

EvilAlexei:
@jbadan I always for the easier way :laughing: In my opinion, this way is the easiest, but I'm open to an alternative opinion too as well.

I see it this way: 
All libraries such as fundamentals-ngx are using this old class `sap-icon`.
If we change class in styles library and we do not change it in related libraries all cases with icons will be broken.
So we need to create separate PRs for each related library, and it can't be done in one PR, 
and they should be merged at the same time as PR for the styles library, otherwise all cases with the sap-icon will be broken.
I'm not really seeing it's possible to merge all PR's at the same time.

If I'm missing anything, please correct me.

jbadan:
@EvilAlexei The implementation libraries should be keeping their dependency on fundamental-styles up to date and keeping up with breaking changes.

In fundamental-react, we bump the fundamental-styles version every 2 weeks or so (and I hope ngx is doing something more or less similar). Since this change would be marked as a breaking change, the libraries shouldn't get this change unless they explicitly bump their version to include this breaking change - which will be called out in the release notes. The pr to bump fundamental-styles in fundamental-react or fundamental-ngx would also include all the changes that need to be made in the implementation library to bring this change in. In fundamental-react's case, this is a pretty simple change because we're using a component for icons and very rarely referring directly to `sap-icon--*`. It's going to be a find and replace for us (and hopefully for ngx unless something else is going on).

By making a breaking change in fundamental styles, you're not on the hook for making the changes in the implementation libraries - while it's nice when someone does do them all, it's the maintainers responsibility there (unless you are also a maintainer of ngx).

TLDR: We can do this asynchronously, there's no need for it to be done at the exact same time in all the libraries. Granted, we will be blocked from bringing in new features or fixes that are merged after this pr, but that's the nature of it and why we should be keeping the fundamental-styles dependency up to date as much as possible.

src/fd-icon.scss
EvilAlexei: Reverted. 
We can't remove this part, because most of the icons doesn't have common unique class,
all of them looks like `.sap-icon--washing-machine` or `.sap-icon--wrench`.

It could be a good idea to add this class in future, for example as in the font-awesome:
```
<i class="fas fa-arrow-right"></i>
<i class="fas fa-arrow-left"></i>
```
I think this update shouldn't be a part of this PR.

EvilAlexei:
@jbadan updated PR according to the last comments, removed new component `fd-icon` and transferred all new changes to the `icon` component.

Also reverted changes you mentioned earlier: `[class*="#{$block}"]`, answered in the original comment why I reverted this.

IlyaSurmay (approved):

droshev (changes requested):
@EvilAlexei thank you for your hard work! we need to discuss how we want to continue with the icons before moving forward with this PR

src/fd-icon.scss
jbadan: That sounds reasonable! Can we make a followup issue for that? All places using an icon should probably be something like: `<i class='fd-icon fd-icon--iconName" />`

jbadan:
> @jbadan updated PR according to the last comments, removed new component `fd-icon` and transferred all new changes to the `icon` component.
> 
Thank you for doing that - I know it was a lot of extra work :) Other than needing to have a more global group conversation about the direction of icons, I think this work looks good.

src/fd-icon.scss
EvilAlexei: @jbadan could you review please this issue :)
https://github.com/SAP/fundamental-styles/issues/1794

InnaAtanasova:
Let's park this for now and discuss it later.

Issue #1753, opened 2020-10-06
Rating Indicator: remove icon import and local css variables


jbadan:
We should never be importing all of icons into a component - https://github.com/SAP/fundamental-styles/blob/master/src/rating-indicator.scss#L3

The implementation library can also import icon styles into the component, but rating indicator shouldn't have this built in.

Additionally, adding locally scoped css variables makes adding fallbacks for ie11 difficult: https://github.com/SAP/fundamental-styles/blob/master/src/rating-indicator.scss#L7

PR #1782, opened 2020-10-14
[WIP] fix: Fix compile issue for rating indicator styles

mykolavorobiov:
## Description
image

The rating indicator can't be compiled when is used in the ngx-fundamental component.

### Before:
image

### After:
image

3. Testing

### DIFF
--- a/src/rating-indicator.scss
+++ b/src/rating-indicator.scss
@@ -5,8 +5,8 @@
 $block: #{$fd-namespace}-rating-indicator;
 
 :root {
-  --sapRating-indicator-icon-rated: url();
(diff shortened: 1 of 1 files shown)


IlyaSurmay (approved):

EvilAlexei (approved):

katekozlowska (approved):

jbadan (changes requested):
We should just remove these variables - please see https://github.com/SAP/fundamental-styles/issues/1753

If we new css variables that aren't defined by sap-theming, they should go in the theming/sap*.scss files.

mykolavorobiov:
@jbadan Got it

PR #1784, opened 2020-10-14
chore: update README visual testing link.

prsdthkr:
## Description
* Update README link to point to updated documentation, after this change: https://github.com/SAP/fundamental-styles/pull/1584
* Added a Storybook badge pointing to the storybook built by Chromatic for the master branch

- [NA] RTL support
3. Testing
- [NA] Updated tests
4. Documentation


jbadan (approved):
Do we need to add a chromatic open source badge?

prsdthkr:
> Do we need to add a chromatic open source badge?

Nice idea. There isn't a badge for chromatic, but I added one for Storybook pointing to the storybook built by Chromatic fro the master branch:

[Storybook](https://master--5f85e3904b19d600221872af.chromatic.com)

Issue #1038, opened 2020-05-13
'High Contrast White' theme Tabs selected text color contrast


prsdthkr:
**Describe the bug**
The Tabs selected text color contrast ratio is less than 4.5:1
Also for selected/active tabs, fd-tabs__icon text color should be in line with the theme i.e white.

> What version of the Fundamental Library Styles are you using?
v0.10.0-rc.1

**To Reproduce**
1. Go to https://deploy-preview-1025--fundamental-styles.netlify.app/components/tabs.html
2. Check the Tabs selected text color contrast ratio

**Expected behavior**
The Tabs selected text color contrast ratio should be 4.5:1 or above. Ideally 7:1 or above for WCAG AAA compliance. #0049DB would be the best choice.

Text color for icon based tabs that are active should be white.

**Screenshots**
![image](https://user-images.githubusercontent.com/43764832/81863841-ab151880-9520-11ea-95f8-4ac38d376de5.png)
Fails all requirements: https://webaim.org/resources/contrastchecker/?fcolor=8FB5FF&bcolor=FFFFFF

![image](https://user-images.githubusercontent.com/43764832/81864886-16131f00-9522-11ea-8587-6fcc6125d6f3.png)

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: Chrome 
 - Version: 81.0.4044.129 (Official Build) (64-bit)

Mike-Diaz:
Fixed as of version 0.13.0 - Contrast ratio is now 7.5:1

<img width="238" alt="Screen Shot 2020-10-08 at 5 53 39 PM" src="https://user-images.githubusercontent.com/21978807/95529218-3808b480-098f-11eb-9a95-0d2bad0c1767.png">

prsdthkr:
![image](https://user-images.githubusercontent.com/43764832/96192563-1feae500-0efb-11eb-84f4-74cba722c896.png)

Issue #1033, opened 2020-05-13
'High Contrast White' theme Form (Radio buttons and default Checkbox) checked state color contrast


prsdthkr:
**Describe the bug**
The Form (Radio buttons and default Checkbox) checked state color contrast ratio does not meet the minimum criteria i.e. `4.5:1`

> What version of the Fundamental Library Styles are you using?
v0.10.0-rc.1

**To Reproduce**
1. Go to https://deploy-preview-1025--fundamental-styles.netlify.app/components/form.html
2. Check the Radio buttons and Checkbox checked state icon color contrast with the background

**Expected behavior**
The Form (Radio buttons and default Checkbox) checked state color contrast ratio should be 4.5:1 or above. Ideally 7:1 or above for WCAG AAA compliance. `#000` would be the best choice.

**Screenshots**
![image](https://user-images.githubusercontent.com/43764832/81858224-376f0d80-9518-11ea-97ab-336d65126609.png)
![image](https://user-images.githubusercontent.com/43764832/81858388-874dd480-9518-11ea-8279-945973883a19.png)
![image](https://user-images.githubusercontent.com/43764832/81858260-4b1a7400-9518-11ea-9a0f-d3d981037f2b.png)

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: Chrome 
 - Version: 81.0.4044.129 (Official Build) (64-bit)

Mike-Diaz:
Fixed as of version 0.13.0 - Contrast ratio now 7.5:1
<img width="178" alt="Screen Shot 2020-10-08 at 3 18 27 PM" src="https://user-images.githubusercontent.com/21978807/95529720-866a8300-0990-11eb-9c82-4c7f2795131f.png">

prsdthkr:
![image](https://user-images.githubusercontent.com/43764832/96193827-eff11100-0efd-11eb-93a8-52cef29cff0b.png)

Issue #1030, opened 2020-05-13
'High Contrast White' theme Avatar (TileIcon Background) color


prsdthkr:
**Describe the bug**
TileIcon Background component in the Avatar component has the same text and background-color: black

> What version of the Fundamental Library Styles are you using?
v0.10.0-rc.1

**To Reproduce**
1. Go to https://deploy-preview-1025--fundamental-styles.netlify.app/components/avatar.html
2. Check TileIcon Background component

**Expected behavior**
TileIcon Background component in the Avatar component should have a `color: #fff`

**Screenshots**
![image](https://user-images.githubusercontent.com/43764832/81855237-ec52fb80-9513-11ea-93e4-135ddfba392f.png)

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: Chrome 
 - Version: 81.0.4044.129 (Official Build) (64-bit)

Mike-Diaz:
Fixed as of version 0.13.0 - Contrast ratio now 5:1
<img width="377" alt="Screen Shot 2020-10-07 at 5 32 42 PM" src="https://user-images.githubusercontent.com/21978807/95529822-e19c7580-0990-11eb-92cd-1136fb0d03da.png">

prsdthkr:
@Mike-Diaz I just checked this component in the 'High Contrast White' and its not fixed:
![image](https://user-images.githubusercontent.com/43764832/95619008-bf533800-0a22-11eb-9fb3-4d0965b5167b.png)

To reproduce:
* go to https://deploy-preview-1752--fundamental-styles.netlify.app/?path=/docs/components-avatar--icon
* change the theme to 'High Contrast White' from the top right dropdown
* see 'Tile Icon Background' story

Issue #1029, opened 2020-05-13
'High Contrast White' theme Avatar color contrast


prsdthkr:
**Describe the bug**
Icons in the Avatar component do not have a contrast ratio of 3:1 or above.
Icons that can be used to understand the content need a minimum contrast ratio of 3:1 according to this WCAG 2.1 specification: https://www.w3.org/WAI/WCAG21/Techniques/general/G207.html 

> What version of the Fundamental Library Styles are you using?
v0.10.0-rc.1

**To Reproduce**
1. Go to https://deploy-preview-1025--fundamental-styles.netlify.app/components/avatar.html
2. Check Icon color contrast

**Expected behavior**
Icon in the Avatar component should have a color contrast ratio of 3:1 or above.

**Screenshots**
![image](https://user-images.githubusercontent.com/43764832/81855034-9a11da80-9513-11ea-80bd-75adb1d09bf7.png)
![image](https://user-images.githubusercontent.com/43764832/81855063-a433d900-9513-11ea-8f12-b2601071c760.png)

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: Chrome 
 - Version: 81.0.4044.129 (Official Build) (64-bit)

Mike-Diaz:
Fixed as of version 0.13.0 - Contrast ratio now 5.3:1
<img width="377" alt="Screen Shot 2020-10-07 at 5 38 27 PM" src="https://user-images.githubusercontent.com/21978807/95529883-15779b00-0991-11eb-893c-fd5d49d24f01.png">

prsdthkr:
See this page in HCW Theme:
https://deploy-preview-1752--fundamental-styles.netlify.app/?path=/docs/components-avatar--icon

PR #1789, opened 2020-10-15
fix: fiori 3 high contrast mode contrast ratios

Mike-Diaz:
## Related Issue
Closes SAP/fundamental-styles#1030
Closes SAP/fundamental-styles#1029
Closes SAP/fundamental-styles#1018
Closes SAP/fundamental-styles#1005
Closes SAP/fundamental-styles#998
Closes SAP/fundamental-styles#1038
Closes SAP/fundamental-styles#1033

## Description
This PR is focused on fixing Fiori High Contrast White and High Contrast Black issues with several components. Components currently broken, or violating WCAGAA contrast ratio guidelines have been addressed and now pass ratio recommendations.

#### New ratios:
Tested with PCG Colour Contrast Analyser

- `Avatar` Tile Icon Background Fiori HCW: 21:1
- `Avatar` Tile Icon Background Fiori HCB: 21:1
- `Avatar` Icon Fiori HCW: 7.3:1
- `Tabs` Default Fiori HCB: 11.7:1
- `Tabs` Icon Only Fiori HCB: 11.7:1
- `Tabs` Default Fiori HCW: 7.1:1
- `Tabs` Icon Only HCW: 7.1:1
- `Radio` Responsiveness Fiori HCB: 21:1
- `Radio` Responsiveness Fiori HCW: 7.3:1
- `Checkbox` Desktop Fiori HCB: 21:1
- `Checkbox` Desktop Fiori HCW: 7.3:1

#### Note:
`Checkbox` and `Radio` screenshots in IE11 are broken. Using CSS Addons to switch styles in Canvas does not work (in IE11). Using the Theme selector in Docs works partially. Theme selector only shows some of the styling changes, but it will not change the background color of the frame they are housed in, therefore; any theme using white colors for text or other elements will not be visible since the background is also white.

When testing these changes, please use Canvas instead of Docs to prevent the issues mentioned above with other components.

### Before:
Avatar_Before
Avatar_Icon_Before
Avatar_TileIcon_Before
Radio_Before
Radio_Before_HCW
Checkbox_Before_HCW
Tabs_Before

#### Chrome
Avatar_Icon_After
Avatar_Placeholder_HCB
Avatar_Placeholder_HCW
Avatar_TileIcon_HCB
Avatar_TileIcon_HCW
Checkbox_Desktop_After
Checkbox_Desktop_HCW
Radio_Responsive_After
Radio_Responsive_HCW
Tabs_Counter_After
Tabs_Default_AFter
Tabs_Icon_After
Tabs_Default_HCW
Tabs_Icon_HCW

#### IE11
IE11_Avatar_TileIcon_HCB
IE11_Avatar_TileIcon_HCW
IE11_Avatar_Placeholder_HCB
IE11_Avatar_Placeholder_HCW
IE11_Icon_After
IE11_Checkbox_After
IE11_Checkbox_HCW
IE11_Radio_After
IE11_Radio_HCW
IE11_Tabs_After
IE11_Tabs_Icon_After
IE11_Tabs_Default_HCW
IE11_Tabs_Icon_HCW

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation


prsdthkr:
Hi @mike, could you merge the master branch into this so that the new visual snapshots be generated in Chromatic?

prsdthkr (changes requested):
Avatar > Tile Icon Background in High Contrast Light:
image

prsdthkr:
Can we also fix https://github.com/SAP/fundamental-styles/issues/1038 as part of this change?

jbadan:
Did any of these changes deviate from the published spec on the wiki? If so, we should email the designer responsible and let them know. If not, carry on!

prsdthkr:
Contrast for Avatar with a white icon/text on cyan(ish) bg for HCD theme isn't enough:
image

****

prsdthkr:
Avatar with borders has no borders in the Dark theme
image

Mike-Diaz:
@prsdthkr 

> Contrast for Avatar with a white icon/text on cyan(ish) bg for HCD theme isn't enough:

Changed the inner icon color to black.
Screen Shot 2020-10-18 at 2 44 06 PM

> Avatar with borders has no borders in the Dark theme

Oops, I messed this one up when I was adding borders to HCB elements. Added a new variable for the optional border element to use.
Screen Shot 2020-10-18 at 2 43 32 PM

src/avatar.scss
```diff
@@ -40,7 +40,7 @@ $block: #{$fd-namespace}-avatar;
     [class*="#{$block}"] {
       @include fd-reset();
 
-      color: var(--sapContent_ContrastTextColor);
+      color: var(--fdAvatar_DefaultColor);
```
jbadan, line 43: I think this should stay ` --sapContent_ContrastTextColor`. It's specified for all themes for the default color - this should be white in hcb and black in hcw. We've got the opposite happening.

src/avatar.scss
```diff
@@ -40,7 +40,7 @@ $block: #{$fd-namespace}-avatar;
...
       background-color: $fd-avatar-background-color;
```
jbadan, line 44: `$fd-avatar-background-color` needs a new variable - it should be `--sapAccentColor6` in light + dark, `--sapBackgroundColor` in hcw/hcb.

src/avatar.scss
```diff
@@ -97,17 +97,21 @@ $block: #{$fd-namespace}-avatar;
...
-    background-color: var(--sapTile_IconColor);
+    background-color: var(--fdAvatar_Tile_BackgroundColor);
+    border: var(--fdAvatar_Border);
+    color: var(--fdAvatar_Tile_DefaultColor);
   }
 
   // optional border
   &--border {
-    border: 0.0625rem solid var(--sapGroup_ContentBorderColor);
+    border: var(--fdAvatar_Optional_Border);
```
jbadan, line 114: This one can just be `0.0625rem solid var(--sapGroup_ContentBorderColor);` because it's the same in all themes. We can delete `--fdAvatar_Optional_Border`

src/avatar.scss
```diff
@@ -97,17 +97,21 @@ $block: #{$fd-namespace}-avatar;
 
   // placeholder background
   &--placeholder {
-    background-color: var(--sapContent_ImagePlaceholderBackground);
+    background-color: var(--fdAvatar_Placeholder_BackgroundColor);
```
jbadan, line 100: This one can just be `var(--sapContent_ImagePlaceholderBackground);` and delete `--fdAvatar_Placeholder_BackgroundColor`

src/avatar.scss
```diff
@@ -97,17 +97,21 @@ $block: #{$fd-namespace}-avatar;
...
+    border: var(--fdAvatar_Border);
```
jbadan, line 101: I think we can get rid of this border here and apply it to the root as it's the same for default.

src/avatar.scss
```diff
@@ -97,17 +97,21 @@ $block: #{$fd-namespace}-avatar;
...
+    color: var(--fdAvatar_Placeholder_DefaultColor);
   }
 
   // tile background
   &--tile {
-    background-color: var(--sapTile_IconColor);
+    background-color: var(--fdAvatar_Tile_BackgroundColor);
+    border: var(--fdAvatar_Border);
```
jbadan, line 108: I think we can get rid of this border here and apply it to the root as it's the same for default.

src/avatar.scss
```diff
@@ -97,17 +97,21 @@ $block: #{$fd-namespace}-avatar;
 
   // placeholder background
   &--placeholder {
-    background-color: var(--sapContent_ImagePlaceholderBackground);
+    background-color: var(--fdAvatar_Placeholder_BackgroundColor);
+    border: var(--fdAvatar_Border);
+    color: var(--fdAvatar_Placeholder_DefaultColor);
```
jbadan, line 102: We can also remove this color - I think sorting the default background out will actually solve most of the color contrast issues.

jbadan:
I pushed a couple changes for Avatar - feel free to revert if you don't like it :) 

Screen Shot 2020-10-19 at 3 53 45 PM
Screen Shot 2020-10-19 at 3 53 51 PM
Screen Shot 2020-10-19 at 3 53 57 PM
Screen Shot 2020-10-19 at 3 54 01 PM
Screen Shot 2020-10-19 at 2 58 22 PM

JKMarkowski (commented):
Hello @Mike-Diaz Is this PR only about colors? If so I will approve it. I see checkbox should have different border style on HCB on some states.
image

jbadan:
@JKMarkowski I can't speak for @Mike-Diaz, but I was ignoring that one for the moment. I think we need to do a massive audit and use 1 variable for border style. That same `solid` versus `dashed` appears in the majority of components. I'd hate to do it piecemeal and miss components or end up with 20 variables that are the same thing. 

Maybe in a followup pr we create some shared component variables `--fd-border` or whatever naming convention we want to use for these.

Mike-Diaz:
> Hello @Mike-Diaz Is this PR only about colors? If so I will approve it. I see checkbox should have different border style on HCB on some states.
> image

@JKMarkowski Thanks for taking a look at the PR! Yes, this PR was mainly to deal with WCAG AA high contrast color issues. There is certainly more work that can be done, but I would like to keep the scope of the PR to only the issues mentioned in the description, if possible, so that it can eventually be merged in haha!

prsdthkr:
Tabs, Checkbox, and Radio LGTM.

For accented Avatars, 
Seems like an icon with a black color would offer better contrast in the HCD theme:
image

Seems like an icon with a white color would offer better contrast in the HCL theme:
image

prsdthkr:
Also, could you merge master into this again for snapshot consistency?

Mike-Diaz:
> For accented Avatars,
> Seems like an icon with a black color would offer better contrast in the HCD theme:

> Seems like an icon with a white color would offer better contrast in the HCL theme:

Ah yes, I did notice this. I didn't pay much attention since this was not part of the issues addressed in this PR, but I went ahead and changed the inner icon color as requested :)

Also merged master in so that the snapshots are consistent.

prsdthkr (approved):
QA LGTM 🚢 
Thanks for the extra accent fix 👍

PR #1792, opened 2020-10-15
chore: take themed + RTL visual snapshots for each component

prsdthkr:
## Description
In this change, we:
* update visual story builder script to create visual stories, with RTL version, for each story in themes:
  * sap_fiori_3
  * sap_fiori_3_dark
  * sap_fiori_3_hcw
  * sap_fiori_3_hcb
* add 'Thanks' section in README for Chromatic
* fix an issue with jest storyshot that resulted in empty storyshots for single element stories, like wrapped in `<div>`, that had newline in the first line.
* update all empty storyshots to proper storyshots after the above was fixed.

For example:
```jsx
export const types = () => `
<div class="fddocs-container"> 
   ...
   ...
</div>`;
```
resulted in an empty storyshot.

Removing the newline and making it
```jsx
export const types = () => `<div class="fddocs-container"> 
   ...
   ...
</div>`;
```
creates the correct jest storyshot.

- [NA] RTL support
3. Testing
- [NA] Updated tests
4. Documentation


droshev (approved):
That's an amazing step forward!

PR #1793, opened 2020-10-18
docs: add story dependencies

droshev:
## Description
update component dependencies for the storybook and replace the images URL as the previous weren't working

- [N/A] All values are in `rem`
- [N/A] RTL support
3. Testing
- [N/A] Updated tests
4. Documentation


prsdthkr:
Hey @droshev could you merge master into this? I have made some changes to generic-tile.

InnaAtanasova (approved):

PR #1804, opened 2020-10-21
feat: introduce Text component

EvilAlexei:
## Related Issue
No related issue

## Description
Introduce text component styles

#### Default
deafult
#### Whitespaces
whitespaces
#### Max lines
maxlines
#### Hyphenation
hyphenation

- [x] All values are in `rem`
- [x] RTL support
- [x] Mixins are used for repeatable code (`fd-rtl`, `fd-ellipsis`, `fd-flex`, `fd-selected`, `fd-focus`, etc.)
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/.storybook/custom/custom.scss
+++ b/.storybook/custom/custom.scss
@@ -459,6 +459,11 @@ body.sb-show-main {
     margin-right: 10px;
     min-width: 75px;
   }
+
+  .mw-300 {
+    max-width: 300px;
+    padding: 0 15px;
+  }
 }
 
 .fddocs-select {
--- a/.storybook/custom/custom.scss
+++ b/.storybook/custom/custom.scss
@@ -460,7 +460,7 @@ body.sb-show-main {
     min-width: 75px;
   }
 
-  .mw-300 {
+  .fddocs-text {
     max-width: 300px;
     padding: 0 15px;
   }
--- a/src/text.scss
+++ b/src/text.scss
@@ -1,0 +1,18 @@
+@import "./new-settings";
+@import './mixins';
+
+$block: #{$fd-namespace}-text;
+
+.#{$block} {
+  @include fd-reset();
+
+  font-weight: normal;
+  font-family: var(--sapFontFamily);
+  font-size: var(--sapFontSize);
+  color: var(--sapTextColor);
+
+  &::selection {
+    color: var(--sapContent_ContrastTextColor);
+    background-color: var(--sapSelectedColor);
+  }
+}
--- a/src/text.scss
+++ b/src/text.scss
@@ -6,13 +6,22 @@ $block: #{$fd-namespace}-text;
 .#{$block} {
   @include fd-reset();
 
-  font-weight: normal;
-  font-family: var(--sapFontFamily);
-  font-size: var(--sapFontSize);
-  color: var(--sapTextColor);
-
   &::selection {
     color: var(--sapContent_ContrastTextColor);
     background-color: var(--sapSelectedColor);
+  }
+
+  &--max-lines {
+    display: -webkit-box;
+    -webkit-box-orient: vertical;
+    overflow: hidden;
+  }
+
+  &--pre-wrap {
+    white-space: pre-wrap;
+  }
+
+  &--hyphenation {
+    hyphens: auto;
   }
 }
--- a/stories/text/__snapshots__/text.stories.storyshot
+++ b/stories/text/__snapshots__/text.stories.storyshot
@@ -25,7 +25,7 @@ exports[`Storyshots Components/Text Hyphenation 1`] = `
     
             
     <h3>
-      Without hyphens
+      Without hyphenation
     </h3>
     
             
@@ -185,7 +185,7 @@ exports[`Storyshots Components/Text Wrapping 1`] = `
     
     
   <h3>
-    Whitespace pre-wrap
+    Pre-wrapped whitespace
   </h3>
   
     
--- a/stories/text/__snapshots__/text.stories.storyshot
+++ b/stories/text/__snapshots__/text.stories.storyshot
@@ -1,0 +1,248 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Storyshots Components/Text Default 1`] = `
+
+    
+`;
+
+exports[`Storyshots Components/Text Hyphenation 1`] = `
+<section>
+  
+
+    
+  <h3>
+    Without hyphens
+  </h3>
+  
+    
+  <div
+    style="width: 300px"
+  >
+    
+        
+    <p
+      class="fd-text"
+    >
+      Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt
+        ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut
+        aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore
+        eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia
+        deserunt mollit anim id est laborum.
+    </p>
+    
+    
+  </div>
+  
+    
+    
+  <h3>
+    Hyphen
+  </h3>
+  
+    
+  <div
+    style="width: 300px"
+  >
+    
+        
+    <p
+      class="fd-text"
+    >
+      Lorem ipsum dolor sit amet, consectetur adip-iscing elit, sed do eiusmod tempor
+        incididunt ut labore et dolore magna aliqua. Ut enim ad mi-nim veniam, quis nostrud exercitation ullamco laboris nisi ut
+        aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore
+        eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia
+        deserunt mollit anim id est laborum.
+    </p>
+    
+    
+  </div>
+  
+    
+    
+  <h3>
+    Shy
+  </h3>
+  
+    
+  <div
+    style="width: 300px"
+  >
+    
+        
+    <p
+      class="fd-text"
+    >
+      Lorem ipsum dolor sit amet, con­sectetur adip­iscing elit, sed do eiusmod tempor
+        incididunt ut labore et dolore magna aliqua. Ut enim ad mi­nim veniam, quis nostrud exer­citation ullamco laboris nisi ut
+        aliquip ex ea commodo conse­quat. Duis aute irure dolor in reprehenderit in vo­luptate velit esse cillum dolore
+        eu fu­giat nulla pari­atur. Excepteur sint occaecat cupidatat non pro­ident, sunt in culpa qui officia
+        deserunt mollit anim id est la­bo­rum.
+    </p>
+    
+    
+  </div>
+  
+
+</section>
+`;
+
+exports[`Storyshots Components/Text Max lines 1`] = `
+<section>
+  
+    
+  <h3>
+    No max lines rule
+  </h3>
+  
+    
+  <p
+    classname="fd-text"
+    style="white-space: normal;"
+  >
+    Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
+    eiusmod tempor incididunt ut labore et dolore magna aliqua. Malesuada bibendum arcu vitae elementum. Ut etiam sit
+    amet nisl. Laoreet suspendisse interdum consectetur libero id faucibus nisl. Tellus in metus vulputate eu
+    scelerisque felis imperdiet proin fermentum. Ullamcorper morbi tincidunt ornare massa eget egestas purus viverra
+    accumsan. Mauris vitae ultricies leo integer malesuada nunc vel risus. Nisi est sit amet facilisis. Eu turpis
+    egestas pretium aenean. Nunc id cursus metus aliquam eleifend. Placerat in egestas erat imperdiet. Etiam dignissim
+    diam quis enim lobortis scelerisque. A erat nam at lectus. Amet nisl purus in mollis nunc sed id. Purus semper eget
+    duis at tellus at urna.
+    Aliquet eget sit amet tellus cras adipiscing enim. Risus ultricies tristique nulla aliquet enim tortor. Varius quam
+    quisque id diam vel quam elementum pulvinar. Eget aliquet nibh praesent tristique magna. Id interdum velit laoreet
+    id donec ultrices tincidunt arcu. Suspendisse ultrices gravida dictum fusce ut placerat orci nulla. Montes nascetur
+    ridiculus mus mauris vitae ultricies. In nisl nisi scelerisque eu. Suspendisse interdum consectetur libero id
+    faucibus. Ut lectus arcu bibendum at varius. Placerat vestibulum lectus mauris ultrices eros in cursus turpis massa.
+    Dolor magna eget est lorem ipsum dolor sit. Aliquam nulla facilisi cras fermentum odio eu feugiat pretium nibh.
+    Augue neque gravida in fermentum et. Proin fermentum leo vel orci porta non pulvinar.
+  </p>
+  
+    
+    
+  <h3>
+    2 max lines
+  </h3>
+  
+    
+  <p
+    class="fd-text"
+    style="display: -webkit-box; -webkit-box-orient: vertical; -webkit-line-clamp: 2; overflow: hidden;"
+  >
+    
+    Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
+    eiusmod tempor incididunt ut labore et dolore magna aliqua. Malesuada bibendum arcu vitae elementum. Ut etiam sit
+    amet nisl. Laoreet suspendisse interdum consectetur libero id faucibus nisl. Tellus in metus vulputate eu
+    scelerisque felis imperdiet proin fermentum. Ullamcorper morbi tincidunt ornare massa eget egestas purus viverra
+    accumsan. Mauris vitae ultricies leo integer malesuada nunc vel risus. Nisi est sit amet facilisis. Eu turpis
+    egestas pretium aenean. Nunc id cursus metus aliquam eleifend. Placerat in egestas erat imperdiet. Etiam dignissim
+    diam quis enim lobortis scelerisque. A erat nam at lectus. Amet nisl purus in mollis nunc sed id. Purus semper eget
+    duis at tellus at urna.
+    Aliquet eget sit amet tellus cras adipiscing enim. Risus ultricies tristique nulla aliquet enim tortor. Varius quam
+    quisque id diam vel quam elementum pulvinar. Eget aliquet nibh praesent tristique magna. Id interdum velit laoreet
+    id donec ultrices tincidunt arcu. Suspendisse ultrices gravida dictum fusce ut placerat orci nulla. Montes nascetur
+    ridiculus mus mauris vitae ultricies. In nisl nisi scelerisque eu. Suspendisse interdum consectetur libero id
+    faucibus. Ut lectus arcu bibendum at varius. Placerat vestibulum lectus mauris ultrices eros in cursus turpis massa.
+    Dolor magna eget est lorem ipsum dolor sit. Aliquam nulla facilisi cras fermentum odio eu feugiat pretium nibh.
+    Augue neque gravida in fermentum et. Proin fermentum leo vel orci porta non pulvinar.
+  </p>
+  
+
+    
+  <h3>
+    3 max lines
+  </h3>
+  
+    
+  <p
+    class="fd-text"
+    style="display: -webkit-box; -webkit-box-orient: vertical; -webkit-line-clamp: 3; overflow: hidden;"
+  >
+    
+    Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
+    eiusmod tempor incididunt ut labore et dolore magna aliqua. Malesuada bibendum arcu vitae elementum. Ut etiam sit
+    amet nisl. Laoreet suspendisse interdum consectetur libero id faucibus nisl. Tellus in metus vulputate eu
+    scelerisque felis imperdiet proin fermentum. Ullamcorper morbi tincidunt ornare massa eget egestas purus viverra
+    accumsan. Mauris vitae ultricies leo integer malesuada nunc vel risus. Nisi est sit amet facilisis. Eu turpis
+    egestas pretium aenean. Nunc id cursus metus aliquam eleifend. Placerat in egestas erat imperdiet. Etiam dignissim
+    diam quis enim lobortis scelerisque. A erat nam at lectus. Amet nisl purus in mollis nunc sed id. Purus semper eget
+    duis at tellus at urna.
+    Aliquet eget sit amet tellus cras adipiscing enim. Risus ultricies tristique nulla aliquet enim tortor. Varius quam
+    quisque id diam vel quam elementum pulvinar. Eget aliquet nibh praesent tristique magna. Id interdum velit laoreet
+    id donec ultrices tincidunt arcu. Suspendisse ultrices gravida dictum fusce ut placerat orci nulla. Montes nascetur
+    ridiculus mus mauris vitae ultricies. In nisl nisi scelerisque eu. Suspendisse interdum consectetur libero id
+    faucibus. Ut lectus arcu bibendum at varius. Placerat vestibulum lectus mauris ultrices eros in cursus turpis massa.
+    Dolor magna eget est lorem ipsum dolor sit. Aliquam nulla facilisi cras fermentum odio eu feugiat pretium nibh.
+    Augue neque gravida in fermentum et. Proin fermentum leo vel orci porta non pulvinar.
+  </p>
+  
+
+</section>
+`;
+
+exports[`Storyshots Components/Text Wrapping 1`] = `
+<section>
+  
+    
+  <h3>
+    Whitespace normal
+  </h3>
+  
+    
+  <p
+    class="fd-text"
+    style="white-space: normal;"
+  >
+    Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt
+    ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut
+    aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore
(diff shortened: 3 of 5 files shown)


src/text.scss
```diff
@@ -0,0 +1,18 @@
...
+
+$block: #{$fd-namespace}-text;
+
+.#{$block} {
+  @include fd-reset();
+
+  font-weight: normal;
+  font-family: var(--sapFontFamily);
+  font-size: var(--sapFontSize);
+  color: var(--sapTextColor);
```
jbadan, line 12:
```
  font-weight: normal;
  font-family: var(--sapFontFamily);
  color: var(--sapTextColor);
```
are already included in `fd-reset()`

stories/text/text.stories.js
```diff
@@ -0,0 +1,172 @@
...
+    
+    <h3>Whitespace pre-wrap</h3>
+    <p class="fd-text" style="white-space: pre-wrap;">Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt
+ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut
+aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore
+eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia
+deserunt mollit anim id est laborum.</p>
+
+    <h3>Whitespace nowrap</h3>
+    <div style="overflow: auto;">
```
jbadan, line 43: Please remove all inline styles - uses classes instead https://github.com/SAP/fundamental-styles/blob/master/.storybook/custom/custom.scss

src/text.scss
EvilAlexei: @jbadan Yeap, I saw it, but my idea was that these styles are specific for text component, from design guidelines.
And styles from `fd-reset()` are styles from `fd-reset()` they have different scopes, I mean if something will change in `fd-reset` It will affect the text component, but it shouldn't.

jbadan: We shouldn't repeat css - it needs to be removed.

EvilAlexei: On another side, we should make components independent.

And a few other points:
- It's a common practice to overwrite styles from reset styles, we just got lucky that these style values are the same.
- It's a common practice inside an already created components.

stories/text/text.stories.js
```diff
@@ -0,0 +1,172 @@
...
+    title: 'Components/Text',
+    parameters: {
+        description: `The text control is used to display text. It generally contains the text
+         that developers want apps to display (property: text).`,
+        tags: ['f3', 'a11y', 'theme'],
+        components: ['text']
+    }
+};
+
+export const defaultExample = () => `
```
prsdthkr, line 11: Remove newline at the end of this file to fix [empty jest snapshot](https://github.com/SAP/fundamental-styles/pull/1804/files#diff-4f34f45b3d2cc609fb28216e10a082a914ee15c29538f2d39a630b8b54af9994R3-R6).

prsdthkr (changes requested):

stories/text/text.stories.js
EvilAlexei: If I got your idea correct, you propose to remove all inline styles from examples and replace them inside `custom.scss`.
In this way, all custom, specific, and not reusable styles for unique examples will be stored in one place.
A huge file with a lot of unique styles, and a lot of classes.

But with the current implementation, we have all data for example in one place, independent, easy to modify, or to expand.
And it's already a common practice in these examples. And yes, I don't like inline styles either, but I think the advantages  of this approach is stronger than disadvantage on the inline styles

src/text.scss
jbadan: It is repeating styles, you're overwriting a value with the same value. It is only common to overwrite reset styles when the styles are different from the ones in the reset. The reset serves a purpose to not allow outside global styles to leak into our components.

stories/text/text.stories.js
jbadan: One of the coding style guidelines we have for fundamental-styles is to avoid inline styles whenever possible. We've made an exception for `fd-slider` as that is a special case. If we start doing this here, it's going to creep in all over the place. Please remove them.

EvilAlexei: I found 424 cases of the inline styles in the stories files, and it's definitely not only the `fd-slider`.
If we start collecting all stuff in custom.scss it will become one huge unsupportable file

InnaAtanasova: @EvilAlexei, 95% of these 426 cases the inline styling is used to make the documentation more presentable or if we want to showcase an overflow, the other 5% go to adding a background image for Avatar and slider.

jbadan: It sounds like we have a lot cases to fix then 😃  

There are some legitimate uses of inline styles... I just don't think this qualifies. You're showing off instances of a component that don't exist without inline styles - that doesn't seem right. 
Why are we telling consumers to set inline styles? For this wrapping example, shouldn't we have `<span class='fd-text fd-text--wrap'`, where `--wrap` has `white-space: nowrap`?

src/text.scss
EvilAlexei: As I mentioned earlier, it's just a coincidence that this component and `fd-reset` styles have the same values.

We have an independent component with its styles, and we have a set of reset styles, it is two different scopes.
Reset styles - is a global set of rules and global scope.
And the text component has its own set of rules, with its own scop.

The last one has the main priority, and if something will be changed in `fd-reset`, by mistake or on purpose this shouldn't affect the main text component styles.

In my opinion, this is a more bulletproof approach, without unclear relations.

stories/text/text.stories.js
```diff
@@ -0,0 +1,172 @@
...
+
+maxLines.storyName = 'Max lines';
+maxLines.parameters = {
+    docs: {
+        iframeHeight: 300,
+        storyDescription: `You can define maximum number of lines for text, after reaching the limit text will be
+        truncated and ellipsis will be added. To set maximum numbers you should use these styles:
+        <code class="docs-code">
+        display: -webkit-box; -webkit-box-orient: vertical; -webkit-line-clamp: [number of lines]; overflow: hidden;
+        </code>`
```
jbadan, line 124: This should be in a `fd-text--modifier` class.

stories/text/text.stories.js
EvilAlexei: @InnaAtanasova maybe I get something wrong, but in my opinion that inline styles in this component serve the same purpose, only to show examples how users could use it, they are not the part component itself.

EvilAlexei: @jbadan in this way we should have classes for all styles, for example, `fd-text-color-red`. In my opinion, it's a little overkill.

stories/text/text.stories.js
EvilAlexei: @jbadan why? It's a common recipe for line clamp, it's not a unique mechanic for the text component.

stories/text/text.stories.js
EvilAlexei: I think each approach has its own pros and cons.
I see really strong pros for an isolated approach, where each example has its own styles,
like incapsulation, easy to support and easy to modify.
And cons - inline styles, but in the case of isolated examples, in my opinion, it's not a big deal.

And pros for `custom.scss` - not inline styles.
cons - all styles in one place, it will be difficult to support this file and isolate styles, and in time really difficult.

If I missed something else - I will be glad for the additional info, I'm still learning about this library and approaches here.

stories/text/text.stories.js
jbadan: Then maybe this isn't a component, but a pattern?

EvilAlexei: Hey, @jbadan, 
I think maybe there is a misunderstanding about these examples and why we have so different thoughts about those styles?

Just in case, I'll try to clarify my reason to add them.
I had a task to create a Text component styles, in the documentation described only base 4 style rules and selected state.

Also, in fiori designs were sections about wrapping, maxlines, hyphenation, but this functionality is the default for any text in html+css, you can control it as easily as setting color or font-size, but because it was mentioned in the design doc I added these examples, to show how it could be achieved. I saw similar examples here as well, for example showing icon size change by changing font-size via inline styles.

The reason to not create specific classes for those cases, that's this is not a component functionality, those are things that could be done with this component, the same as color change, font-size change and etc.

src/text.scss
InnaAtanasova: Hi @EvillAlexei, the current approach of Fundamental-styles is to use the fd-reset and not to repeat the css. This set of rules is not specific only for the Text component, it's used in many other components dealing with text. Could you add the reset here and we can go on with the review?

stories/text/text.stories.js
InnaAtanasova: > @InnaAtanasova maybe I get something wrong, but in my opinion that inline styles in this component serve the same purpose, only to show examples how users could use it, they are not the part component itself.

Imagine you have 2 fundamental buttons in a div in the documentation. 
```
<div>
    <button class="fd-button">Button 1</button>
    <button class="fd-button">Button 2</button>
</div>
```

In this case your 2 buttons will be rendered next to each other with no spacing, which is ugly. In this case we add inline styling on the container div and you get something like:
```
<div style="display:flex; justify-content: space-between;">
    <button class="fd-button">Button 1</button>
    <button class="fd-button">Button 2</button>
</div>
```

This type of inline styling is ok because it only concerns the docs part and does not affect the components (the user doesn't have to do anything). Your inline styling IS affecting the component directly. For a specific case the user will have to write `<p class="fd-text" style="white-space: nowrap;">` or `<p class="fd-text" style="white-space: pre-wrap;">` or `<p class="fd-text" style="white-space: normal;">`. That's exactly why we have modifier classes in the library. We have 1 default behaviour and we use the modifier classes to overwrite it. 
Also, if you take a look from the implementation library perspective it will be much easier to manipulate this with classes and not with inline styling. The user can pass to the input property the desired value for white-space and internally we will add the required class. 
Anyways, inline styling on the component level is not something we want to have in the library, so you need to change it to modifier classes.

stories/text/text.stories.js
InnaAtanasova: @EvilAlexei, I understand your point of view but I'd rather remove these examples and stick with the example from the specs than providing all these examples with inline styling. You only need the styling that comes from the reset function, the text selection rules and the truncation (word by word). If the user wants to modify something let them figure out how to do it. We only show them how to achieve Fiori 3 case using classes.

src/text.scss
EvilAlexei: Hi @InnaAtanasova , I wasn't against `fd-reset` and it's already presented in the component.
I'm against removing scoped styles, named as repeated. I checked and found such behaviour in many existed components:
alerts.scss, menu.scss, pagination.scss and else. So as I understand it is or maybe it was ok to use styles in this way.

And I see a lot of sense in this approach, `fd-reset` - it's a set of reset styles, it's a common mixin, and it's could be changed easily, but this shouldn't affect components main styles, from design specification. But in this case, If I remove the main independent styles all possible changes in `fd-reset` will affect main component representation.

Anyway, If you think that I should remove it anyway - I will remove it, but I think this could case issues in the future.

stories/text/text.stories.js
EvilAlexei: @InnaAtanasova @jbadan My bad, I found the documentation with a description about classes for text-component that should have such functionality, I will add them in next commit, and I will try to remove all inline styles

stories/text/text.stories.js
```diff
@@ -0,0 +1,159 @@
...
+    docs: {
+        iframeHeight: 300,
+        storyDescription: `The Text component has a property allowing browsers to render whitespace and tabs. 
+To enable whitespace rendering add a <code class="docs-code">fd-text--pre-wrap</code> class`
+    }
+};
+
+export const maxLines = () => `
+    <h3>No max lines rule</h3>
+    <p className="fd-text">Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
```
prsdthkr, line 57:
```suggestion
    <p class="fd-text">Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
```

JKMarkowski (approved):
LGTM, It's also worth to mention that `-webkit-line-clamp:` doesn't work on IE11 and in this case user should add limit height to `X*height of line`

EvilAlexei:
@JKMarkowski good point about support IE11, thank you
What do you think, should I create an example or add a note in the description about this one?

InnaAtanasova:
> @JKMarkowski good point about support IE11, thank you
> What do you think, should I create an example or add a note in the description about this one?

You can add a note in the description above the example. I will also tag our documentation guru to check the PR.
@jessmetivier could you take a look if you have time? Thanks

InnaAtanasova (approved):

.storybook/custom/custom.scss
```diff
@@ -435,6 +435,11 @@ body.sb-show-main {
     margin-right: 10px;
     min-width: 75px;
   }
+
+  .mw-300 {
```
InnaAtanasova, line 439: could you rename this to `fddocs-text`?

stories/text/text.stories.js
```diff
@@ -0,0 +1,159 @@
...
+            deserunt mollit anim id est la&shy;bo&shy;rum.</p>
+        </div>
+    </div>
+`;
+
+hyphenation.storyName = 'Hyphenation';
+hyphenation.parameters = {
+    docs: {
+        iframeHeight: 200,
+        storyDescription: `To allow browser to automatically break words at appropriate hyphenation points,
```
jessmetivier, line 147: The text component can display words that are broken at appropriate hyphenation points in a text block. To display hyphens, add the <code class="docs-code">fd-text--hyphenation</code> to the main element. 

It is also possible to suggest line break opportunities with two Unicode characters that manually specify potential line breakpoints:
- Hyphen: The "hard" hyphen character indicates a visible line break opportunity. Even if the line is not actually broken at that point, the hyphen is still displayed.
- Shy: An invisible, "soft" hyphen. Although this character is not visible, it marks a place where the browser should break the word if hyphenation is necessary.

In HTML, add the `&shy;` Unicode to insert a soft hyphen.

jessmetivier (changes requested):
Great job on the docs! Please see my comments.

stories/text/text.stories.js
```diff
@@ -0,0 +1,159 @@
+export default {
+    title: 'Components/Text',
+    parameters: {
+        description: `The text control is used to display text. It generally contains the text
```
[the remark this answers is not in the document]
EvilAlexei: @jeannewalters changed. I have a small note, that in design documentation there is a note about using text components inside forms:
```
Do not use the text control if you need a label, or vice versa.
```

Should I add the same note?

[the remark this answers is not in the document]
EvilAlexei: changed

.storybook/custom/custom.scss
EvilAlexei: changed

stories/text/text.stories.js
```diff
@@ -0,0 +1,159 @@
...
+    aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore
+    eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia
+    deserunt mollit anim id est laborum.</p>
+`;
+
+defaultExample.storyName = 'Default';
+defaultExample.parameters = {
+    docs: {
+        iframeHeight: 100,
+        storyDescription: 'Default example of the text component'
```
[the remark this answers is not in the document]
EvilAlexei: changed

stories/text/text.stories.js
```diff
@@ -0,0 +1,159 @@
...
+    Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore
+eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia
+deserunt mollit anim id est laborum.</p>
+`;
+
+wrapping.storyName = 'Wrapping';
+wrapping.parameters = {
+    docs: {
+        iframeHeight: 300,
+        storyDescription: `The Text component has a property allowing browsers to render whitespace and tabs.
```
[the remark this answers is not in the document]
EvilAlexei: changed

stories/text/text.stories.js
```diff
@@ -0,0 +1,159 @@
...
+    faucibus. Ut lectus arcu bibendum at varius. Placerat vestibulum lectus mauris ultrices eros in cursus turpis massa.
+    Dolor magna eget est lorem ipsum dolor sit. Aliquam nulla facilisi cras fermentum odio eu feugiat pretium nibh.
+    Augue neque gravida in fermentum et. Proin fermentum leo vel orci porta non pulvinar.</p>
+`;
+
+maxLines.storyName = 'Max lines';
+maxLines.parameters = {
+    docs: {
+        iframeHeight: 300,
+        storyDescription: `You can define maximum number of lines for text, after reaching the limit text will be
```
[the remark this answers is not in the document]
EvilAlexei: changed

stories/text/text.stories.js
```diff
@@ -0,0 +1,159 @@
...
+defaultExample.storyName = 'Default';
+defaultExample.parameters = {
+    docs: {
+        iframeHeight: 100,
+        storyDescription: 'Default example of the text component'
+    }
+};
+
+export const wrapping = () => `
+    <h3>Whitespace normal</h3>
```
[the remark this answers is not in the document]
EvilAlexei: changed

stories/text/text.stories.js
EvilAlexei: changed

stories/text/text.stories.js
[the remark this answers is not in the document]
jessmetivier: I saw that too, I made a note to add it when we have the [Label](https://experience.sap.com/fiori-design-web/label/) component in Fundamental Styles, that way we can add a **Usage** section like we have in all the other docs.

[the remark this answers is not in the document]
jessmetivier: Then we can put something like...

Do not use text if:
- You need to display text to define controls or form fields. Instead, use a **Label**.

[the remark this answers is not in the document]
EvilAlexei: Got it, sounds fine to me :)
Thank you!

stories/text/text.stories.js
```diff
@@ -0,0 +1,167 @@
...
+maxLines.parameters = {
+    docs: {
+        iframeHeight: 300,
+        storyDescription: `The text component can be displayed with a maximum number of lines.
+When the maximum is reached, the text truncates and displays an ellipsis. To display text with a maximum line count,
+add the <code class="docs-code">fd-text--max-lines</code> modifier class and an inline style rule with the number of
+lines to the main element. For example, add <code class="docs-code">style="-webkit-line-clamp: 3;"</code> to display
+three lines of text.
+
+**Note**: Property <code class="docs-code">-webkit-line-clamp</code> doesn't work in IE11 and should be changed
```
jessmetivier, line 122: I would just add "The" at the beginning of this sentence, the rest is great.

stories/text/text.stories.js
```diff
@@ -0,0 +1,167 @@
...
+`;
+
+hyphenation.storyName = 'Hyphenation';
+hyphenation.parameters = {
+    docs: {
+        iframeHeight: 200,
+        storyDescription: `The text component can display words that are broken at appropriate hyphenation
+points in a text block. To display hyphens, add the <code class="docs-code">fd-text--hyphenation</code> to the main element.
+        
+It is also possible to suggest line break opportunities with two Unicode characters that manually specify
```
jessmetivier, line 156: Oops my bad here, all sentences with a colon ":" should be bolded. So just bold "It is also possible to suggest line break opportunities with two Unicode characters that manually specify potential line breakpoints:"

stories/text/__snapshots__/text.stories.storyshot
```diff
@@ -0,0 +1,207 @@
...
+    ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut
+    aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore
+    eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia
+    deserunt mollit anim id est laborum.
+  </p>
+  
+    
+    
+  <h3>
+    Whitespace pre-wrap
```
jessmetivier, line 188: At first glance, this seems incomprehensible. I know it's in the code, but how can we make it more conversational?

Pre-wrapped whitespace maybe? (for the designer audience that visits the site).

stories/text/text.stories.js
```diff
@@ -0,0 +1,167 @@
...
+
+**Note**: Property <code class="docs-code">-webkit-line-clamp</code> doesn't work in IE11 and should be changed
+to <code class="docs-code">height</code>. For example, <code class="docs-code">style="height: 200px;"</code> 
+`
+    }
+};
+
+export const hyphenation = () => `<div class="example-container">
+        <div class="fddocs-text">
+            <h3>Without hyphens</h3>
```
jessmetivier, line 130: Without hyphenation

(To stay consistent with the next example)

jessmetivier (changes requested):
I just have a question about the "whitespace pre-wrap" title, but should be good to go after that's cleared up :)

stories/text/__snapshots__/text.stories.storyshot
EvilAlexei: So basically it renders whitespaces, maybe `Render whitespaces`?

EvilAlexei: @jessmetivier just in case, I uploaded PR with your variant, but I can change it in case of need.

jessmetivier: Ah, okay. In that case, I would clarify it by writing:

"Wrapped text with indents and whitespace"

jessmetivier: And in the example above say instead of "No whitespace":

No indents and whitespace

jessmetivier: Maybe "Wrapping" shouldn't be the title to this example section... since this is more about "Whitespace" than anything else, and indentation is a white space in itself...

jessmetivier: Not just visually, but literally — whitespace is a design principle that helps guide the users' eyes. In this case, makes the text more readable! 😉

EvilAlexei: What do you think about the plural form? I mean use "Whitespace`s`"

stories/text/text.stories.js
```diff
@@ -0,0 +1,167 @@
...
+    docs: {
+        iframeHeight: 300,
+        storyDescription: `The text component can be displayed with a maximum number of lines.
+When the maximum is reached, the text truncates and displays an ellipsis. To display text with a maximum line count,
+add the <code class="docs-code">fd-text--max-lines</code> modifier class and an inline style rule with the number of
+lines to the main element. For example, add <code class="docs-code">style="-webkit-line-clamp: 3;"</code> to display
+three lines of text.
+
+**Note**: The property <code class="docs-code">-webkit-line-clamp</code> doesn't work in IE11 and should be changed
+to <code class="docs-code">height</code>. For example, <code class="docs-code">style="height: 200px;"</code>
```
jessmetivier, line 123: Just missing a period at the end of this sentence!

jessmetivier (approved):
I'll go ahead and approve this one since I trust you'll add the finishing touches. Again, great job on this one! Was a pleasure to review.

stories/text/__snapshots__/text.stories.storyshot
jessmetivier: Whitespace is the proper term, and it is always singular. If you want to talk about white spaces individually, it works — but you can always replace it with "whitespace" when referring to design/writing.

"There is so much whitespace on this website!"
"There are so many white spaces on this website!" (Awkward, but still technically correct)

EvilAlexei:
@jessmetivier thank you for help :)

prsdthkr:
The selection background-color is not completely opaque. This is best seen with an image in the background:
image

Since this might affect foreground/background color contrast calculations we should set selection background opacity to 0.99:
image

EvilAlexei:
@prsdthkr could you please provide more info on this?
I wasn't able to find info or requirements about background opacity in the design documentation,
and the default browser selection color has this opacity too.
Could you also add more details about opacity value, why `0.99` and not `1`?

And in my opinion, this is a very specific case.
If there will be such background, despite selection color opacity there will be a contrast calculation issue.

And anyway, I think it's up to user to set proper background for text.
For example, if the background will be the same color as the selection color, there will be the same contrast calculation issue,
and you can't avoid this.

prsdthkr:
@EvilAlexei, I don't know the exact reason for this but I  think the browser agent sets some opacity to the selection background when opacity is `1` but not when opacity is `0.99` so basically is 0.99 ends up being more opaque than 1. This will not only be an issue when the consumer sets a custom background, but also in our own example:

opacity: 0.99 has a better contrast ratio:
image


than opacity: 1
image

stories/text/text.stories.js
```diff
@@ -0,0 +1,167 @@
...
+defaultExample.parameters = {
+    docs: {
+        iframeHeight: 100,
+        storyDescription: `The default text component can display lines of text that wrap to the next line
+        once they reach the end of the content container width.`
+    }
+};
+
+export const whitespace = () => `
+    <h3>No indents and whitespace</h3>
```
jessmetivier, line 29: Hey Alexei, last thing I promise! Because it's a negative sentence it's "No indents or whitespace"

jessmetivier (commented):
Last fix 🙃

EvilAlexei:
@prsdthkr thank you for the update!

Sounds reasonable, but I have issues with these changes:
To add opacity to the background I should use rgba function, rgba function from scss
able to work with hex values, but not with css variables, so I can't use color variables from themes,
to get this work all themes should be updated with rgb equivalent of this selection color,
and in my opinion, it's out of the scope of this task.

stories/text/text.stories.js
EvilAlexei: It's ok, no worries :)

EvilAlexei: Updated PR

prsdthkr (approved):

stefanoScalzo (approved):

EvilAlexei:
Hello @jbadan and @InnaAtanasova , could you please re-review this PR :)

JKMarkowski:
Hello @EvilAlexei, can you push to this branch, just to retrigger UI Test pipeline?

EvilAlexei:
@JKMarkowski I updated PR with the latest master. This will be ok?

JKMarkowski:
@EvilAlexei Everything is fine, we will merge this after realeasing `0.13.0`. You don't have to worry about this PR anymore

EvilAlexei:
@JKMarkowski awesome, thank you!

Issue #1760, opened 2020-10-08
Defect Hunting Sprint 48


InnaAtanasova:
Before releasing we need to perform a Defect Hunting and address all the issues that are found. Document all the issues found under this issue.

jbadan:
I probably should have just put #1772 , #1771, #1770, #1769, #1768, #1767 and #1766 under here!

jbadan:
Generic Tile is missing some imports for the docs:
<img width="1429" alt="Screen Shot 2020-10-12 at 1 42 15 PM" src="https://user-images.githubusercontent.com/29607818/95788756-fb8bd000-0c90-11eb-8d85-dc5768c74e3d.png">

InnaAtanasova:
- [ ] the overflow button does not open the contextual menu in Action Bar
- [ ] images are not loaded in the Avatar component. The provided urls return 404
- [ ] Busy Indicator, the Standard example is not visible in High Contrast Dark mode (needs a dark background)
- [ ] Dialog - wrong position of the close button in mobile mode
<img width="779" alt="Screen Shot 2020-10-21 at 2 35 17 PM" src="https://user-images.githubusercontent.com/39598672/96767601-be1ef500-13aa-11eb-8e48-09aeae780237.png">
<img width="674" alt="Screen Shot 2020-10-21 at 2 35 09 PM" src="https://user-images.githubusercontent.com/39598672/96767614-c1b27c00-13aa-11eb-8b2d-6cd7aa154bf2.png">
- [ ] Card - missing avatar image (same problem as for Avatar)
- [ ] Checkbox - missing the third state (TriState example)
- [ ] Should be changing the background of the example container div in High Contrast because the text is not visible.
- [ ] Dialog issues - I am able to click Show Code and other buttons when a dialog is open. Clicking on the buttons is not detected by the backdrop. Broken example for Selectable dialog:
<img width="2535" alt="Screen Shot 2020-10-21 at 2 44 42 PM" src="https://user-images.githubusercontent.com/39598672/96768824-4c47ab00-13ac-11eb-9120-d96cb77360df.png">
<img width="2295" alt="Screen Shot 2020-10-21 at 2 45 24 PM" src="https://user-images.githubusercontent.com/39598672/96768875-59fd3080-13ac-11eb-9a21-e2cd1a6e3f4f.png">
- Tiles - HCD button issue:
[...]

PR #1810, opened 2020-10-22
docs: bring different image into examples

JKMarkowski:
## Related Issue
part of https://github.com/SAP/fundamental-styles/issues/1760


prsdthkr:
Instead of depending on these image services, would it make sense to just pull some images from https://www.sapbrandtools.com/ and save locally in storybook static? We don't have to hop from one service to another that way. Also, image snapshots will give consistent results.

InnaAtanasova:
> https://www.sapbrandtools.com/

I completely agree with you!

InnaAtanasova:
> Instead of depending on these image services, would it make sense to just pull some images from https://www.sapbrandtools.com/ and save locally in storybook static? We don't have to hop from one service to another that way. Also, image snapshots will give consistent results.

I 100%  share @prsdthkr  opinion on this.

JKMarkowski:
Hello @prsdthkr, that's great idea, I added some SAP image as a asset

InnaAtanasova (approved):

salarenko (approved):

JKMarkowski:
Thanks @prsdthkr for awesome work there :)

PR #1815, opened 2020-10-23
fix: Remove focus on pressed button

JKMarkowski:
## Related Issue
Part of https://github.com/SAP/fundamental-ngx/issues/3641
Part of https://github.com/SAP/fundamental-styles/issues/1760

## Description
Focus outline is not needed on selected buttons, when button is not focused.

### Before:
image

### After:
image


mikerodonnell89 (approved):
Good catch thanks!

salarenko (approved):

PR #1814, opened 2020-10-23
fix: wcag aaa text color contrast

Mike-Diaz:
## Related Issue
Closes SAP/fundamental-styles#1020
Closes SAP/fundamental-styles#1019
Closes SAP/fundamental-styles#1016
Closes SAP/fundamental-styles#1015
Closes SAP/fundamental-styles#1014
Closes SAP/fundamental-styles#1010
Closes SAP/fundamental-styles#1009
Closes SAP/fundamental-styles#1006
Closes SAP/fundamental-styles#1004
Closes SAP/fundamental-styles#1003
Closes SAP/fundamental-styles#996

## Description
This PR is addressing WCAG AAA Color contrast issue above which mainly pertain to the text contrast of these components against their backgrounds in Fiori **High Contrast Black** mode. The two colors used throughout are `#FFFFFF` for text, `#0F5D94` and `#7A5100` for background. This is currently failing color contrast WCAG AAA minimums.

`#FFFFFF` text color against `#7A5100` background results in 6.98:1 contrast ratio which fails WCAG AAA
https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=7A5100
 
`#FFFFFF` text color against `#0F5D94` results in 6.96:1 contrast ratio which fails WCAG AAA
https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=0F5D94 

Additionally for the `Calendar` component, there were no surrounding borders for items containing these colors further failing `#0F5D94` and `#7A5100` vs. `#000000` background. Borders have been added to active calendar days.
https://webaim.org/resources/contrastchecker/?fcolor=000000&bcolor=0F5D94
https://webaim.org/resources/contrastchecker/?fcolor=000000&bcolor=7A5100

In order to comply with WCAG AAA guidelines, a text shadow has been added to offending components in order to have a passing foreground/background contrast ratio.

### Contrast Ratios
- Token
   - `#FFFFFF` vs. `#0A3A5B` 11.9:1
   - `#FFFFF` vs. `#5B3C03` 10:1
- Side Nav
   - `#FFFFFF` vs. `#0A3A5B` 11.9:1
   - `#FFFFF` vs. `#5B3C03` 10:1
- Time
   - `#FFFFFF` vs. `#0A3A5B` 11.9:1
   - `#FFFFF` vs. `#5B3C03` 10:1
- Select
   - `#FFFFFF` vs. `#0A3A5B` 11.9:1
[...]

PR #1825, opened 2020-10-27
fix: high contrast black side navigation nested-list/icon text shadow

Mike-Diaz:
## Related Issue
Closes SAP/fundamental-styles#1016

## Description
This PR focuses on a small change with the `SideNav` component. While a partial fix was implemented to add text shadow [here](https://github.com/SAP/fundamental-styles/pull/1814), some of the examples, particularly nested lists, did not receive this change which caused a failure in WCAG AAA contrast ratio minimums. Text shadow has now been added to the remaining examples which did not originally receive it.

- Side Nav Icon/Nested
   - `#FFFFFF` vs. `#0A3A5B` 11.9:1
   - `#FFFFFF` vs. `#5B3C03` 10:1

### Before:
image
image

#### Chrome:
image
image

#### IE11:
image
image

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation


jacobdevera (approved):
LGTM

prsdthkr:
Hi @Mike-Diaz, change looks good. Could you sync with `master`.

InnaAtanasova:
Hi @Mike-Diaz looks great. Why are so many tests updated? Could you sync with master as @prsdthkr suggested?

Mike-Diaz:
@prsdthkr Thanks for checking it out! Merged!
@InnaAtanasova Thank you! Got it merged!

prsdthkr (approved):
QA LGTM ✅

PR #1827, opened 2020-10-27
docs: show storybook examples with themed background

### LINKED ISSUE #1767: Docs: Form components: update to use container with themed background
We should update the Form component documentation to have each example wrapped in another component or a playground example div that is themed. When looking at the examples in "Dark" theme, you can't really see the text.

Screen Shot 2020-10-09 at 6 35 44 AM
Screen Shot 2020-10-09 at 6 34 31 AM

prsdthkr:
## Related Issue
Closes #1767

## Description
In this change, we update Storybook so that: 
* Changing the theme should now change the background for story containers
* Directionality and theme `<select>`s have labels

### Before:
image

### After:
image

- [N.A.] All values are in `rem`
- [N.A. ] Text elements follow the truncation rules
- [N.A.] hover state of the element follow design spec
- [N.A.] focus state of the element follow design spec
- [N.A.] active state of the element follow design spec
- [N.A.] selected state of the element follow design spec
- [N.A.] selected hover state of the element follow design spec
- [N.A.] pressed state of the element follow design spec
- [N.A.] Responsiveness rules - the component has modifier classes for all breakpoints
- [N.A.] Includes Compact/Cosy/Tablet design
- [N.A.] RTL support
- [N.A.] BEM naming convention is used
- [N.A.] Mixins are used for repeatable code (`fd-rtl`, `fd-ellipsis`, `fd-flex`, `fd-selected`, `fd-focus`, ect.)
- [N.A.] A11y support - keyboard support, screenreader support, proper ARIA attributes, etc.
- [N.A.] `fd-reset()` mixin is applied to all elements
- [N.A.] Variables are used, if some value is used more than twice.
- [N.A.] Checked if current components can be reused, instead of having new code.
3. Testing
- [N.A.] tested Storybook examples with "CSS Resources" `normalize` option 
- [N.A.] tested Storybook examples with "CSS Resources" `unnormalize` option 
- [N.A.] Verified all styles in IE11
- [N.A.] Updated tests
4. Documentation
- [N.A.] Storybook documentation has been created/updated
- [N.A.] Breaking Changes [wiki](https://github.com/SAP/fundamental-styles/wiki/Breaking-Changes) has been updated in case of breaking changes.


InnaAtanasova (approved):

InnaAtanasova:
Awesome! Thanks @prsdthkr I was just about to start working on this issue :)
I will close https://github.com/SAP/fundamental-styles/pull/1830 as it doesn't make sense anymore. 
Thank again!

PR #1836, opened 2020-10-30
feat: add Form Header

InnaAtanasova:
## Description
Introduce Form Header which is needed for Wizard Summary page

Note:
--sapFontHeaderWeight is not present in sap-theming, the designers have been contacted about this.

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/form-header.scss
+++ b/src/form-header.scss
@@ -1,0 +1,31 @@
+@import './new-settings';
+@import './mixins';
+
+/*!
+.fd-form-header
+   .fd-form-header__text
+*/
+$block: #{$fd-namespace}-form-header;
+$fd-form-header-padding-x: 1rem !default;
+
+.#{$block} {
+  @include fd-reset();
+  @include fd-flex-vertical-center();
+
+  width: 100%;
+  height: 2.75rem;
+  background-color: var(--sapGroup_TitleBackground);
+  padding-left: $fd-form-header-padding-x;
+  padding-right: $fd-form-header-padding-x;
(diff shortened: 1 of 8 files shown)


salarenko (approved):
Looks good. I'm just wondering if we shouldn't have an example how to use form with form-header.

InnaAtanasova:
> Looks good. I'm just wondering if we shouldn't have an example how to use form with form-header.

There will be one in the Wizard Summary page

src/form-header.scss
```diff
@@ -0,0 +1,31 @@
...
+.#{$block} {
+  @include fd-reset();
+  @include fd-flex-vertical-center();
+
+  width: 100%;
+  height: 2.75rem;
+  background-color: var(--sapGroup_TitleBackground);
+  padding-left: $fd-form-header-padding-x;
+  padding-right: $fd-form-header-padding-x;
+  box-shadow: 0 var(--sapList_BorderWidth) var(--sapList_BorderWidth) var(--sapGroup_TitleBorderColor);
```
JKMarkowski, line 20: image
Can you add new parameter to theming, so for high contrast there will be `3 * --sapList_BorderWidth` and for rest `--sapList_BorderWidth`?

JKMarkowski (commented):
Great Job! I left one comment regarding theming, rest LGTM

jbadan (approved):
What a hilariously specific component 🤣

JKMarkowski (approved):
LGTM

stefanoScalzo (approved):

PR #1843, opened 2020-11-02
fix: Prevent token from growing

salarenko:
## Related Issue
Part of: #1760

## Description
This PR:
- removes growing behavior from Token
- add proper max-width for tokens inside the tokenizer

### Before:
image
image

### After:
image
image

### DIFF
--- a/src/token.scss
+++ b/src/token.scss
@@ -44,6 +44,7 @@ $block: #{$fd-namespace}-token;
   color: $fd-token-color;
   font-size: var(--sapFontSize);
   display: inline-flex;
+  flex-shrink: 0;
   align-items: center;
   text-align: center;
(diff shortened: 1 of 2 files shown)


InnaAtanasova (changes requested):
I think this change is causing all the tokens to have `...` even if they don't need.

Screen Shot 2020-11-03 at 8 56 11 AM

salarenko:
@InnaAtanasova Thanks, should be good now.
I've also updated the `max-width` of the token inside tokenizer according to Fiori3 docs.

JKMarkowski (approved):
Looks good for me, tested also on IE11

salarenko:
During the review we found some different issue (not related to this PR) with controls maximum width. The issue has been described in [this ticket](https://github.com/SAP/fundamental-styles/issues/1883).

InnaAtanasova (approved):

Issue #867, opened 2020-04-16
Grid List component styles creation as per Fiori3


sKudum:
<img width="1286" alt="Screen Shot 2020-04-16 at 2 51 42 PM" src="https://user-images.githubusercontent.com/53487468/79438920-cceab080-7ff1-11ea-95c6-20e671c8fda8.png">

Create styles for Grid list component as per Fiori3 standards.

sKudum:
Before starting with the coding need to discuss with styles team for the preferred approach. Should we have it as a new component or as a modifier in the list.scss file

PR #1845, opened 2020-11-03
feat: introduce Grid List component

Betrozov:
## Related Issue
Closes https://github.com/SAP/fundamental-styles/issues/867

## Description
Introducing the new Grid List component

## Screenshots
grid-list

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/.storybook/custom/custom.scss
+++ b/.storybook/custom/custom.scss
@@ -504,3 +504,25 @@ body.sb-show-main {
   border: 1px dashed var(--sapTextColor) !important;
   border-radius: 0.25rem;
 }
+
+.grid-list-item-body-example {
+  display: flex;
+
+  .grid-list-item-body-example-avatar {
+    background-image: url('http://lorempixel.com/400/400/nature/4/');
+    margin-right: 1rem;
+  }
+
+  p {
+    margin: 0;
+
+    + div {
+      margin: 1rem 0;
+    }
+  }
+}
+
+[dir="rtl"] .grid-list-item-body-example .grid-list-item-body-example-avatar {
+  margin-right: 0;
+  margin-left: 1rem;
+}
--- a/src/list-grid.scss
+++ b/src/list-grid.scss
@@ -52,18 +52,45 @@ $block: #{$fd-namespace}-grid-list;
       color: var(--sapContent_MarkerTextColor);
     }
 
+    .#{$block}__radio--input {
+      @include fd-focus() {
+        + .#{$block}__radio--label {
+          outline-offset: 0;
+        }
+      }
+
+      + .#{$block}__radio--label {
+        @include fd-empty() {
+          padding: 0;
+        }
+
+        &::after {
+          top: 0;
+          right: 0;
+          bottom: 0;
+          left: 0;
+          margin: auto !important;
+        }
+      }
+    }
+
+    .#{$block}__checkbox--input {
+      @include fd-focus() {
+        + .#{$block}__checkbox--label {
+          outline-offset: 0;
+        }
+      }
+
+      + .#{$block}__checkbox--label {
+        @include fd-empty() {
+          padding: 0;
+        }
+      }
+    }
+
     .#{$block}__checkbox--label,
     .#{$block}__radio--label {
       display: none;
-      padding: 0 !important;
-    }
-
-    .#{$block}__radio--label::after {
-      top: 0 !important;
-      right: 0;
-      bottom: 0;
-      left: 0 !important;
-      margin: auto !important;
     }
 
     &--state-unread {
@@ -108,7 +135,7 @@ $block: #{$fd-namespace}-grid-list;
     }
   }
 
-  &__toolbar {
+  .#{$block}__toolbar {
     @include fd-flex-vertical-center();
 
     padding: 0 1rem;
--- a/src/list-grid.scss
+++ b/src/list-grid.scss
@@ -10,17 +10,19 @@ $block: #{$fd-namespace}-grid-list;
   $item: #{$block}__item;
   $more: #{$block}__more;
 
+  $fd-grid-list-selection-border-color: var(--sapList_SelectionBorderColor);
   $fd-grid-list-border-radius: var(--sapElement_BorderCornerRadius);
   $fd-grid-list-active-text-color: var(--sapList_Active_TextColor);
   $fd-grid-list-item-box-shadow: var(--sapContent_Shadow0);
-  $fd-grid-list-mode-box-shadow: 0 0 0 0.125rem var(--sapList_SelectionBorderColor);
+  $fd-grid-list-mode-box-shadow: 0 0 0 0.125rem $fd-grid-list-selection-border-color;
   $fd-grid-list-item-selected-background-color: var(--sapList_SelectionBackgroundColor);
   $fd-grid-list-item-active-background-color: var(--sapList_Active_Background);
   $fd-grid-list-item-hover-background-color: var(--sapList_Hover_Background);
+  $fd-grid-list-item-hover-box-shadow: 0 0 0 0.0625rem var(--fdGrid_List_Item_Hover_Border_Color);
   $fd-grid-list-highlight: (
-    'success': ('backgroundColor': var(--sapSuccessBorderColor)),
-    'error': ('backgroundColor': var(--sapErrorBorderColor)),
-    'warning': ('backgroundColor': var(--sapWarningBorderColor)),
+    'positive': ('backgroundColor': var(--sapSuccessBorderColor)),
+    'negative': ('backgroundColor': var(--sapErrorBorderColor)),
+    'critical': ('backgroundColor': var(--sapWarningBorderColor)),
     'neutral': ('backgroundColor': var(--sapNeutralBorderColor))
   );
 
@@ -28,6 +30,7 @@ $block: #{$fd-namespace}-grid-list;
   @include fd-flex(column);
 
   &__item {
+    @include fd-reset();
     @include fd-flex(column);
     @include fd-fiori-focus();
 
@@ -37,36 +40,21 @@ $block: #{$fd-namespace}-grid-list;
     height: 100%;
     position: relative;
 
-    &--body {
-      @include fd-flex(column);
-
-      height: 100%;
-      justify-content: space-between;
-
-      &--padding {
-        padding: 1rem;
-      }
-    }
-
-    &--counter {
-      color: var(--sapContent_MarkerTextColor);
-    }
-
-    .#{$block}__radio--input {
+    .#{$block}__radio-input {
       @include fd-focus() {
-        + .#{$block}__radio--label {
+        + .#{$block}__radio-label {
           outline-offset: 0;
         }
       }
 
-      + .#{$block}__radio--label {
+      + .#{$block}__radio-label {
         @include fd-empty() {
           padding: 0;
         }
 
         &::after {
           top: 0;
-          right: 0;
+          right: 0 !important;
           bottom: 0;
           left: 0;
           margin: auto !important;
@@ -74,48 +62,44 @@ $block: #{$fd-namespace}-grid-list;
       }
     }
 
-    .#{$block}__checkbox--input {
+    .#{$block}__checkbox-input {
       @include fd-focus() {
-        + .#{$block}__checkbox--label {
+        + .#{$block}__checkbox-label {
           outline-offset: 0;
         }
       }
 
-      + .#{$block}__checkbox--label {
+      + .#{$block}__checkbox-label {
         @include fd-empty() {
           padding: 0;
         }
       }
     }
 
-    .#{$block}__checkbox--label,
-    .#{$block}__radio--label {
+    .#{$block}__checkbox-label,
+    .#{$block}__radio-label {
       display: none;
     }
 
-    &--state-unread {
-      .#{$item}--body * {
+    &--unread {
+      .#{$item}-body * {
         font-weight: bold;
       }
-    }
-
-    &--state-actions {
-      margin-top: 1rem;
     }
 
     &--link {
       cursor: pointer;
 
       @include fd-hover() {
-        box-shadow: var(--fdGrid_List_Item_Hover_BoxShadow);
+        box-shadow: $fd-grid-list-item-hover-box-shadow;
         background-color: $fd-grid-list-item-hover-background-color;
       }
 
       @include fd-active() {
-        background-color: $fd-grid-list-item-active-background-color;
+        background-color: $fd-grid-list-item-active-background-color !important;
 
         * {
-          color: $fd-grid-list-active-text-color;
+          color: $fd-grid-list-active-text-color !important;
         }
       }
 
@@ -129,10 +113,36 @@ $block: #{$fd-namespace}-grid-list;
           width: 0.1875rem;
           display: block;
           border-radius: 0 $fd-grid-list-border-radius $fd-grid-list-border-radius 0;
-          background-color: var(--sapList_SelectionBorderColor);
+          background-color: $fd-grid-list-selection-border-color;
         }
       }
     }
+  }
+
+  &__item-body {
+    @include fd-reset();
+
+    @include fd-flex(column) {
+      justify-content: space-between;
+    }
+
+    height: 100%;
+    padding: 1rem;
+
+    &---no-padding {
+      padding: 0;
+    }
+  }
+
+  &__item-body-actions {
+    margin-top: 1rem;
+    order: 1;
+  }
+
+  &__item-counter {
+    @include fd-fiori-focus(0);
+
+    color: var(--sapContent_MarkerTextColor);
   }
 
   .#{$block}__toolbar {
@@ -142,28 +152,43 @@ $block: #{$fd-namespace}-grid-list;
   }
 
   .#{$block}__group-header {
+    @include fd-reset();
     @include fd-flex-vertical-center();
     @include fd-ellipsis();
+    @include fd-fiori-focus();
 
     font-size: var(--sapFontHeader5Size);
-    padding-top: 0.75rem;
-    margin-top: 1rem;
-    height: 2.75rem;
+    margin-top: 0.75rem;
+    height: 2rem;
     color: var(--sapList_GroupHeaderTextColo);
     font-weight: bold;
     line-height: 2rem;
 
-    &:first-child {
+    &--no-margin {
       margin-top: 0;
     }
   }
 
-  &__mode {
-    &--single-select-left,
-    &--multi-select {
+  &--mode {
+    &-single-select {
       .#{$item} {
-        .#{$block}__checkbox--label,
-        .#{$block}__radio--label {
+        cursor: pointer;
+
+        @include fd-hover() {
+          &:not(.#{$item}--link, [aria-selected="true"], .is-selected) {
+            box-shadow: $fd-grid-list-item-hover-box-shadow;
+            background-color: $fd-grid-list-item-hover-background-color;
+          }
+        };
+      }
+    }
+
+    &-single-select-right,
+    &-single-select-left,
+    &-multi-select {
+      .#{$item} {
+        .#{$block}__checkbox-label,
+        .#{$block}__radio-label {
           @include fd-flex(column);
 
           align-self: inherit;
@@ -172,12 +197,11 @@ $block: #{$fd-namespace}-grid-list;
       }
     }
 
-    &--single-select-master,
-    &--multi-select,
-    &--single-select-left {
+    &-single-select,
+    &-multi-select,
+    &-single-select-left,
+    &-single-select-right {
       .#{$item} {
-        cursor: default;
-
         @include fd-selected() {
           box-shadow: $fd-grid-list-mode-box-shadow;
           background-color: $fd-grid-list-item-selected-background-color;
@@ -196,18 +220,22 @@ $block: #{$fd-namespace}-grid-list;
       }
     }
 
-    &--single-select-master {
-      .#{$item} {
-        cursor: pointer;
-
-        @include fd-selected() {
-          cursor: default;
-        }
+    &-single-select-right {
+      .#{$block}__radio-label {
+        order: 1;
+        margin-left: 0.5rem !important;
       }
     }
   }
 
+  &__filter {
+    @include fd-fiori-focus();
+  }
+
   &__more {
+    @include fd-reset();
+    @include fd-fiori-focus();
+
     cursor: pointer;
     padding: 0.813rem;
     background-color: var(--sapGroup_ContentBackground);
@@ -215,16 +243,20 @@ $block: #{$fd-namespace}-grid-list;
     box-shadow: $fd-grid-list-item-box-shadow;
     text-align: center;
 
-    .#{$more}--title {
+    .#{$more}-title {
+      @include fd-reset();
+
       font-weight: bold;
     }
 
-    &--description {
-      margin: 0.813rem 0 0 0;
+    &-description {
+      @include fd-reset();
+
+      margin: 0.813rem 0 0;
     }
 
-    .#{$more}--title,
-    &--description {
+    .#{$more}-title,
+    &-description {
       color: var(--sapLink_Visited_Color);
     }
 
@@ -235,19 +267,24 @@ $block: #{$fd-namespace}-grid-list;
     &:active {
       background-color: $fd-grid-list-item-active-background-color;
 
-      .#{$more}--title,
-      .#{$more}--description {
+      .#{$more}-title,
+      .#{$more}-description {
         color: $fd-grid-list-active-text-color;
       }
     }
   }
 
   &__footer {
+    @include fd-reset();
+    @include fd-fiori-focus(0);
+
     color: var(--sapList_FooterTextColor);
   }
 
   @each $set-name, $colors in $fd-grid-list-highlight {
-    &__highlight--#{$set-name} {
+    &__highlight {
+      @include fd-reset();
+
       position: absolute;
       top: 0;
       left: 0;
@@ -255,7 +292,10 @@ $block: #{$fd-namespace}-grid-list;
       width: 0.375rem;
       display: block;
       border-radius: $fd-grid-list-border-radius 0 0 $fd-grid-list-border-radius;
-      background-color: map_get($colors, 'backgroundColor');
+
+      &--#{$set-name} {
+        background-color: map_get($colors, 'backgroundColor');
+      }
     }
   }
 }
--- a/src/list-grid.scss
+++ b/src/list-grid.scss
@@ -253,6 +253,17 @@ $block: #{$fd-namespace}-grid-list;
     @include fd-fiori-focus();
   }
 
+  &__filter,
+  &__filter-button {
+    color: #fff !important;
+  }
+
+  &__filter-button {
+    @include fd-hover() {
+      color: var(--sapButton_TextColor) !important;
+    }
+  }
+
   &__more {
     @include fd-reset();
     @include fd-fiori-focus();
--- a/src/list-grid.scss
+++ b/src/list-grid.scss
@@ -41,6 +41,9 @@ $block: #{$fd-namespace}-grid-list;
     position: relative;
 
     .#{$block}__radio-input {
+      height: 1rem;
+      width: 1rem;
+
       @include fd-focus() {
         + .#{$block}__radio-label {
           outline-offset: 0;
@@ -63,6 +66,9 @@ $block: #{$fd-namespace}-grid-list;
     }
 
     .#{$block}__checkbox-input {
+      height: 1rem;
+      width: 1rem;
+
       @include fd-focus() {
         + .#{$block}__checkbox-label {
           outline-offset: 0;
@@ -180,6 +186,12 @@ $block: #{$fd-namespace}-grid-list;
             background-color: $fd-grid-list-item-hover-background-color;
           }
         };
+
+        .#{$block}__radio-input {
+          width: 100%;
+          height: 100%;
+          z-index: -1;
+        }
       }
     }
 
@@ -221,6 +233,10 @@ $block: #{$fd-namespace}-grid-list;
     }
 
     &-single-select-right {
+      .#{$block}__radio-input {
+        right: 1rem;
+      }
+
       .#{$block}__radio-label {
         order: 1;
         margin-left: 0.5rem !important;
--- a/src/list-grid.scss
+++ b/src/list-grid.scss
@@ -117,6 +117,12 @@ $block: #{$fd-namespace}-grid-list;
         }
       }
     }
+
+    .#{$item}-toolbar {
+      @include fd-flex-vertical-center();
+
+      padding: 0 1rem;
+    }
   }
 
   &__item-body {
@@ -129,7 +135,7 @@ $block: #{$fd-namespace}-grid-list;
     height: 100%;
     padding: 1rem;
 
-    &---no-padding {
+    &--no-padding {
       padding: 0;
     }
   }
@@ -143,12 +149,6 @@ $block: #{$fd-namespace}-grid-list;
     @include fd-fiori-focus(0);
 
     color: var(--sapContent_MarkerTextColor);
-  }
-
-  .#{$block}__toolbar {
-    @include fd-flex-vertical-center();
-
-    padding: 0 1rem;
   }
 
   .#{$block}__group-header {
--- a/src/list-grid.scss
+++ b/src/list-grid.scss
@@ -1,0 +1,234 @@
+@import './new-settings';
+@import './mixins';
+
+/*!
+.fd-grid-list
+*/
+$block: #{$fd-namespace}-grid-list;
+
+.#{$block} {
+  $item: #{$block}__item;
+  $more: #{$block}__more;
+
+  $fd-grid-list-border-radius: var(--sapElement_BorderCornerRadius);
+  $fd-grid-list-active-text-color: var(--sapList_Active_TextColor);
+  $fd-grid-list-item-box-shadow: var(--sapContent_Shadow0);
+  $fd-grid-list-mode-box-shadow: 0 0 0 0.125rem var(--sapList_SelectionBorderColor);
+  $fd-grid-list-item-selected-background-color: var(--sapList_SelectionBackgroundColor);
+  $fd-grid-list-item-active-background-color: var(--sapList_Active_Background);
+  $fd-grid-list-item-hover-background-color: var(--sapList_Hover_Background);
+  $fd-grid-list-highlight: (
+    'success': ('backgroundColor': var(--sapSuccessBorderColor)),
+    'error': ('backgroundColor': var(--sapErrorBorderColor)),
+    'warning': ('backgroundColor': var(--sapWarningBorderColor)),
+    'neutral': ('backgroundColor': var(--sapNeutralBorderColor))
+  );
+
+  @include fd-reset();
+  @include fd-flex(column);
+
+  &__item {
+    @include fd-flex(column);
+    @include fd-fiori-focus();
+
+    background-color: var(--sapList_Background);
+    border-radius: $fd-grid-list-border-radius;
+    box-shadow: $fd-grid-list-item-box-shadow;
+    height: 100%;
+    position: relative;
+
+    &--body {
+      @include fd-flex(column);
+
+      height: 100%;
+      justify-content: space-between;
+
+      &--padding {
+        padding: 1rem;
+      }
+    }
+
+    &--counter {
+      color: var(--sapContent_MarkerTextColor);
+    }
+
+    .#{$block}__checkbox--label,
+    .#{$block}__radio--label {
+      display: none;
+      padding: 0 !important;
+    }
+
+    .#{$block}__radio--label::after {
+      top: 0 !important;
+      right: 0;
+      bottom: 0;
+      left: 0 !important;
+      margin: auto !important;
+    }
+
+    &--state-unread {
+      .#{$item}--body * {
+        font-weight: bold;
+      }
+    }
+
+    &--state-actions {
+      margin-top: 1rem;
+    }
+
+    &--link {
+      cursor: pointer;
+
+      @include fd-hover() {
+        box-shadow: var(--fdGrid_List_Item_Hover_BoxShadow);
+        background-color: $fd-grid-list-item-hover-background-color;
+      }
+
+      @include fd-active() {
+        background-color: $fd-grid-list-item-active-background-color;
+
+        * {
+          color: $fd-grid-list-active-text-color;
+        }
+      }
+
+      @include fd-navigated() {
+        &::after {
+          content: "";
+          position: absolute;
+          top: 0;
+          right: 0;
+          height: 100%;
+          width: 0.1875rem;
+          display: block;
+          border-radius: 0 $fd-grid-list-border-radius $fd-grid-list-border-radius 0;
+          background-color: var(--sapList_SelectionBorderColor);
+        }
+      }
+    }
+  }
+
+  &__toolbar {
+    @include fd-flex-vertical-center();
+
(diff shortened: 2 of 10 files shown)


src/list-grid.scss
```diff
@@ -0,0 +1,232 @@
...
+
+  &__more {
+    cursor: pointer;
+    padding: 0.813rem;
+    background-color: var(--sapGroup_ContentBackground);
+    border-radius: $fd-grid-list-border-radius;
+    box-shadow: $fd-grid-list-item-box-shadow;
+    text-align: center;
+
+    h6 {
```
jbadan, line 189: Can you remove all the element selectors?

src/theming/sap_fiori_3_dark.scss
```diff
@@ -102,4 +102,7 @@
   /* Calendar */
   --fdCalendar_Text_Shadow: none;
   --fdCalendar_Active_Item_Border: none;
+
+  /* Grid List */
+  --fdGrid_List_SingleSelect_Hover_BoxShadow: 0 0.125rem 0.5rem 0 fade(var(--sapContent_ShadowColor), 30);
```
jbadan, line 107: `fade()` is a less function right? If it is, let's get rid of it - we can make a new variable with the output of `fade(var(--sapContent_ShadowColor), 30);`. The designers are supposed to be removing all less functions from the design specs.

src/list-grid.scss
```diff
@@ -0,0 +1,232 @@
...
+    'success': ('backgroundColor': var(--sapSuccessBorderColor)),
+    'error': ('backgroundColor': var(--sapErrorBorderColor)),
+    'warning': ('backgroundColor': var(--sapWarningBorderColor)),
+    'neutral': ('backgroundColor': var(--sapNeutralBorderColor))
+  );
+
+  @include fd-reset();
+
+  display: flex;
+  flex-direction: column;
```
JKMarkowski, line 24: You can use `@include flex(column(` there

src/list-grid.scss
```diff
@@ -0,0 +1,232 @@
...
+      .fd-toolbar {
+        padding: 0 1rem;
+        align-items: center;
+      }
+
+      &-body {
+        padding: 1rem;
+        height: 100%;
+        display: flex;
+        flex-direction: column;
```
JKMarkowski, line 45: You can use `@include flex(column)` there

JKMarkowski (changes requested):
These styles don't match to newest specifications.

src/list-grid.scss
```diff
@@ -0,0 +1,232 @@
...
+    'warning': ('backgroundColor': var(--sapWarningBorderColor)),
+    'neutral': ('backgroundColor': var(--sapNeutralBorderColor))
+  );
+
+  @include fd-reset();
+
+  display: flex;
+  flex-direction: column;
+
+  .fd-list {
```
InnaAtanasova, line 26: Should be `#{$fd-namespace}` instead of `fd`

src/list-grid.scss
```diff
@@ -0,0 +1,232 @@
...
+    &__item {
```
InnaAtanasova, line 27: You are not supposed to modify the properties of the List directly here. You can create a new class and rely on the css specificity to overwrite the properties of the list-item. For example:
`div class="fd-list__item fd-grid-list__item"....</div>`

Then in `.fd-grid-list__item` you overwrite the properties coming from `fd-list__item`. This will also eliminate the need of using !important, which is not recommended.

src/list-grid.scss
```diff
@@ -0,0 +1,232 @@
...
+      flex-direction: column;
+      align-items: normal !important;
+      height: 100% !important;
+
+      .fd-toolbar {
+        padding: 0 1rem;
+        align-items: center;
+      }
+
+      &-body {
```
InnaAtanasova, line 41: same for these 2 classes. Create additional classes that will overwrite the "default" ones

src/list-grid.scss
```diff
@@ -0,0 +1,232 @@
...
+      }
+
+      .fd-checkbox + .fd-checkbox__label,
+      .fd-radio + .fd-radio__label {
+        display: none;
+        padding: 0 !important;
+        margin: 0 !important;
+      }
+
+      .fd-radio + .fd-radio__label::after {
```
InnaAtanasova, line 60: Same as above. And for future reference, if you need to bring `.fd-xxxx` you need to use `#{$fd-namespace}`. If in future we decide to change the prefix we can do it from 1 place.

src/list-grid.scss
```diff
@@ -0,0 +1,232 @@
...
+
+    &__link {
+      flex-direction: column;
+      align-items: normal;
+      padding: 0;
+
+      @include fd-active() {
+        * {
+          color: $fd-grid-list-active-text-color !important;
+        }
```
InnaAtanasova, line 103: That's not a good idea :)

src/list-grid.scss
InnaAtanasova: +1

src/theming/sap_fiori_3_dark.scss
InnaAtanasova: +1 
You can contact @jessmetivier and she will handle the communication with the designers.

InnaAtanasova (changes requested):

src/list-grid.scss
JKMarkowski, line 26: I think you should consider getting rid of `fd-list` usage. The only things that are taken from list are interractive states (hover/active/select). I know that it's named grid-**list**. But now you overwrite almost every element, which will force users to put 2 classes for 1 element (`fd-list__x` and `fd-grid-list__x`), For me you can export interractive states to some mixin(just to keep same for list and grid list) and treat grid list as a standalone component. Let me know what you think about it, or for some more clarifications

InnaAtanasova:
"Visual" issues:
1. On hover and selected states the inside content covers the rounded borders:
Screen Shot 2020-11-04 at 9 07 28 AM
Screen Shot 2020-11-04 at 9 09 46 AM

2. RTL issues:
Screen Shot 2020-11-04 at 9 05 04 AM

src/list-grid.scss
InnaAtanasova: +1

Betrozov:
> RTL issues:
> 
> Screen Shot 2020-11-04 at 9 05 04 AM

Hi @InnaAtanasova 
In the fd-grid-list__item-body can be whatever developer wants and I think for styles in the body will responsible developer
What do you think?

src/list-grid.scss
Betrozov: Do you know another way to avoid using * if we need to apply color for all inside an item and namely inside the body where we do not know what will be inside?
I will be glad for your help @InnaAtanasova :slightly_smiling_face:

InnaAtanasova:
> > RTL issues:
> > Screen Shot 2020-11-04 at 9 05 04 AM
> 
> Hi @InnaAtanasova
> In the fd-grid-list__item-body can be whatever developer wants and I think for styles in the body will responsible developer
> What do you think?

True, but the documentation (examples) should not look "broken".

katekozlowska (changes requested):
with`.fd-list__group-header` comes `padding: 0 1rem,` and it breaks the layout - header should be aligned with a box (first` <li> `element) below

current view:
Screenshot from 2020-11-04 20-02-03

Screenshot from 2020-11-04 20-05-13

design:
Screenshot from 2020-11-04 20-02-03 (1)

src/list-grid.scss
Betrozov: fixed

src/theming/sap_fiori_3_dark.scss
Betrozov: Asked @jessmetivier to help
thanks for the comment :slightly_smiling_face:

src/list-grid.scss
Betrozov: fixed

src/list-grid.scss
Betrozov: fixed

src/list-grid.scss
Betrozov: fixed

src/list-grid.scss
Betrozov: already not use `fd-list`

src/list-grid.scss
Betrozov: fixed

src/list-grid.scss
Betrozov: fixed

Betrozov:
> "Visual" issues:
> 
> 1. On hover and selected states the inside content covers the rounded borders:
> 
> Screen Shot 2020-11-04 at 9 07 28 AM
> 
> No rounded borders + focus line issue
> Screen Shot 2020-11-04 at 9 09 46 AM

fixed

> 1. RTL issues:
> 
> Screen Shot 2020-11-04 at 9 05 04 AM

fixed

Betrozov:
> with`.fd-list__group-header` comes `padding: 0 1rem,` and it breaks the layout - header should be aligned with a box (first`<li>`element) below
> 
> current view:
> Screenshot from 2020-11-04 20-02-03
> 
> (see padding)
> Screenshot from 2020-11-04 20-05-13
> 
> design:
> Screenshot from 2020-11-04 20-02-03 (1)

fixed
thanks for the comment

src/list-grid.scss
Betrozov: agree with you
but about to move interactive states to mixin I think it needs to be done by separate PR
1 PR = 1 component

.storybook/custom/custom.scss
```diff
@@ -491,3 +491,27 @@ body.sb-show-main {
   border: 1px dashed var(--sapTextColor) !important;
   border-radius: 0.25rem;
 }
+
+.fd-grid-list__item--body {
```
jbadan, line 495: There shouldn't be any `fd-` classes in this file - just playground classes. That way the styles don't get confusing whether it's part of the actual component or just for the documentation.

src/list-grid.scss
```diff
@@ -0,0 +1,234 @@
...
+    @include fd-flex(column);
+    @include fd-fiori-focus();
+
+    background-color: var(--sapList_Background);
+    border-radius: $fd-grid-list-border-radius;
+    box-shadow: $fd-grid-list-item-box-shadow;
+    height: 100%;
+    position: relative;
+
+    .#{$fd-namespace}-toolbar {
```
jbadan, line 40: We should not be using `#{$fd-namespace}` in any file. There should only be 1 `#{$fd-namespace}` class per file. The reason for this is that it 1) violates self contained styles 2) it messes up css modules in fundamental-react

You can change this to `&__toolbar {` and add `fd-grid-list__toolbar` to toolbars in the examples.

src/list-grid.scss
```diff
@@ -0,0 +1,234 @@
...
+      color: var(--sapContent_MarkerTextColor);
+    }
+
+    .#{$block}__checkbox--label,
+    .#{$block}__radio--label {
+      display: none;
+      padding: 0 !important;
+    }
+
+    .#{$block}__radio--label::after {
```
jbadan, line 67: This should work find as `&__radio--label::after {` without the `#{$block}`

src/list-grid.scss
```diff
@@ -0,0 +1,234 @@
...
+
+  &__more {
+    cursor: pointer;
+    padding: 0.813rem;
+    background-color: var(--sapGroup_ContentBackground);
+    border-radius: $fd-grid-list-border-radius;
+    box-shadow: $fd-grid-list-item-box-shadow;
+    text-align: center;
+
+    .#{$more}--title {
```
jbadan, line 191: Can be `&--title`

src/list-grid.scss
```diff
@@ -0,0 +1,234 @@
...
+      font-weight: bold;
+    }
+
+    .#{$more}--description {
+      margin: 0.813rem 0 0 0;
+    }
+
+    .#{$more}--title,
+    .#{$more}--description {
```
jbadan, line 200: Can be `&--title` and `&--description`

jbadan (changes requested):
See comments

src/list-grid.scss
Betrozov: Hi @jbadan 
if I make like you want `.#{$block}__radio--label` then I need to use `!important` to avoid that I just made a higher priority 
selector

src/list-grid.scss
Betrozov: the same answer

src/list-grid.scss
```diff
@@ -0,0 +1,234 @@
...
+    background-color: var(--sapGroup_ContentBackground);
+    border-radius: $fd-grid-list-border-radius;
+    box-shadow: $fd-grid-list-item-box-shadow;
+    text-align: center;
+
+    .#{$more}--title {
+      font-weight: bold;
+    }
+
+    .#{$more}--description {
```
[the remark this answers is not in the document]
Betrozov: fixed

src/list-grid.scss
Betrozov: the same answer `&--description`

src/list-grid.scss
Betrozov: fixed

.storybook/custom/custom.scss
Betrozov: fixed

src/list-grid.scss
jbadan: oh, I didn't realize! Thank you for explaining 😃

src/list-grid.scss
```diff
@@ -0,0 +1,234 @@
...
+      &--padding {
+        padding: 1rem;
+      }
+    }
+
+    &--counter {
+      color: var(--sapContent_MarkerTextColor);
+    }
+
+    .#{$block}__checkbox--label,
```
JKMarkowski, line 55: image
Focus in checkbox inside grid list looks weird

JKMarkowski (changes requested):
Great Job! This PR looks much better, after getting rid of `list` reference. I added 2 minor things, let me know when it's addressed, so I will re-review

src/list-grid.scss
```diff
@@ -0,0 +1,234 @@
...
+        }
+      }
+    }
+  }
+
+  &__toolbar {
+    @include fd-flex-vertical-center();
+
+    padding: 0 1rem;
+  }
```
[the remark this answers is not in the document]
Betrozov: fixed

src/list-grid.scss
Betrozov: fixed

src/list-grid.scss
```diff
@@ -0,0 +1,261 @@
...
+        @include fd-empty() {
+          padding: 0;
+        }
+
+        &::after {
+          top: 0;
+          right: 0;
+          bottom: 0;
+          left: 0;
+          margin: auto !important;
```
JKMarkowski, line 72: I'm not fan of importants in library, but for this case this is for sure better, than creating additional 2-3 more nesting levels

JKMarkowski (approved):

JKMarkowski:
@Betrozov 
Can you change `fd-col-lg--3` in all examples to `fd-col-lg--4`? 
image

Betrozov:
> @Betrozov
> Can you change `fd-col-lg--3` in all examples to `fd-col-lg--4`?
> image

changed

InnaAtanasova:
Screen Shot 2020-11-12 at 11 56 34 AM

In this example the cursor is of type pointer on hover over all items. Shouldn't it be pointer only when it's over a clickable item?

InnaAtanasova:
Screen Shot 2020-11-12 at 11 58 48 AM

RTL issue with the Radio button. Please check if it's in the scope of this PR and if not, please log an issue. Thanks

src/list-grid.scss
```diff
@@ -0,0 +1,261 @@
...
+    'error': ('backgroundColor': var(--sapErrorBorderColor)),
+    'warning': ('backgroundColor': var(--sapWarningBorderColor)),
+    'neutral': ('backgroundColor': var(--sapNeutralBorderColor))
+  );
+
+  @include fd-reset();
+  @include fd-flex(column);
+
+  &__item {
+    @include fd-flex(column);
```
InnaAtanasova, line 31: I think it's missing a reset

src/list-grid.scss
```diff
@@ -0,0 +1,261 @@
...
+    @include fd-fiori-focus();
+
+    background-color: var(--sapList_Background);
+    border-radius: $fd-grid-list-border-radius;
+    box-shadow: $fd-grid-list-item-box-shadow;
+    height: 100%;
+    position: relative;
+
+    &--body {
```
InnaAtanasova, line 40: This is not a modifier class. You are using it like this:
Screen Shot 2020-11-12 at 12 07 32 PM

It's a separate div element, it's not modifying the properties of the `fd-grid-list__item--body`. You can name it `fd-grid-list__item-body` and move it out of the `fd-grid-list__item` block. It will also need a reset.

src/list-grid.scss
```diff
@@ -0,0 +1,261 @@
...
+      @include fd-flex(column);
```
InnaAtanasova, line 41:
```
@include fd-flex(column) {
}
```

You can use it like this. Keeps everything related to flex in one place.

src/list-grid.scss
```diff
@@ -0,0 +1,261 @@
...
+
+      height: 100%;
+      justify-content: space-between;
+
+      &--padding {
+        padding: 1rem;
+      }
+    }
+
+    &--counter {
```
InnaAtanasova, line 51: Screen Shot 2020-11-12 at 12 11 39 PM

This is also not a modifier class. It would be a modifier class if you had `fd-grid-list__item fd-grid-list__item--counter`, then `fd-grid-list__item--counter` will modify the properties coming from `fd-grid-list__item`.
In your case it's a separate element and can be named `fd-grid-list__item-counter` or `fd-grid-list__counter` for simplicity. And you will have to add a reset mixin as it's a standalone element. Only modifier classes don't have reset, as the main class that they are modifying is bringing the reset.

src/list-grid.scss
```diff
@@ -0,0 +1,261 @@
...
+      color: var(--sapContent_MarkerTextColor);
+    }
+
+    .#{$block}__radio--input {
```
InnaAtanasova, line 55: `.#{$block}__radio-input` or `.#{$block}__radio`

src/list-grid.scss
```diff
@@ -0,0 +1,261 @@
...
+      @include fd-focus() {
+        + .#{$block}__radio--label {
+          outline-offset: 0;
+        }
+      }
+
+      + .#{$block}__radio--label {
```
InnaAtanasova, line 62: `.#{$block}__radio-label`

src/list-grid.scss
```diff
@@ -0,0 +1,261 @@
...
+          top: 0;
+          right: 0;
+          bottom: 0;
+          left: 0;
+          margin: auto !important;
+        }
+      }
+    }
+
+    .#{$block}__checkbox--input {
```
InnaAtanasova, line 77: `.#{$block}__checkbox-input` or `.#{$block}__checkbox`

src/list-grid.scss
```diff
@@ -0,0 +1,261 @@
...
+      @include fd-focus() {
+        + .#{$block}__checkbox--label {
```
InnaAtanasova, line 79: `.#{$block}__checkbox-label`

src/list-grid.scss
```diff
@@ -0,0 +1,261 @@
...
+          outline-offset: 0;
+        }
+      }
+
+      + .#{$block}__checkbox--label {
```
InnaAtanasova, line 84: `.#{$block}__checkbox-label`

src/list-grid.scss
```diff
@@ -0,0 +1,261 @@
...
+        @include fd-empty() {
+          padding: 0;
+        }
+      }
+    }
+
+    .#{$block}__checkbox--label,
+    .#{$block}__radio--label {
```
InnaAtanasova, line 92: `.#{$block}__radio-label`

src/list-grid.scss
```diff
@@ -0,0 +1,261 @@
...
+      display: none;
+    }
+
+    &--state-unread {
```
InnaAtanasova, line 96: can be `&--unread`

src/list-grid.scss
```diff
@@ -0,0 +1,261 @@
...
+      .#{$item}--body * {
+        font-weight: bold;
+      }
+    }
+
+    &--state-actions {
```
InnaAtanasova, line 102: can be `&--actions`

src/list-grid.scss
```diff
@@ -0,0 +1,261 @@
...
+    }
+  }
+
+  .#{$block}__toolbar {
+    @include fd-flex-vertical-center();
+
+    padding: 0 1rem;
+  }
+
+  .#{$block}__group-header {
```
InnaAtanasova, line 144: needs a reset?

InnaAtanasova: although the way it is the rest might break the styling coming from `fd-col`. Are you sure `fd-col fd-col--12 fd-grid-list__group-header` has to be together? In other cases you add the `fd-col` classes on a separate wrapper div.

src/list-grid.scss
```diff
@@ -0,0 +1,261 @@
...
+    color: var(--sapList_GroupHeaderTextColo);
+    font-weight: bold;
+    line-height: 2rem;
+
+    &:first-child {
+      margin-top: 0;
+    }
+  }
+
+  &__mode {
```
InnaAtanasova, line 161: Screen Shot 2020-11-12 at 12 22 42 PM

You can't have a structure like this. Following BEM:
`fd-grid-list` is the **Block**. You can't have a **Block** AND an **Element** `fd-grid-list__mode` on the same level, meaning you can't have `fd-grid-list fd-grid-list__mode`. The **Element** (the one that you add `__` in the name) is a child of the Block. The modifier (`--` in the name) is the one you can use with Blocks and Elements. So you can have `fd-grid-list fd-grid-list--dark-mode` for example, but you can't have `fd-grid-list fd-grid-list__mode`.

InnaAtanasova: And you will need a reset if it's an **Element**

src/list-grid.scss
```diff
@@ -0,0 +1,261 @@
...
+        .#{$block}__radio--label {
+          @include fd-flex(column);
+
+          align-self: inherit;
+          color: var(--sapContent_LabelColor);
+        }
+      }
+    }
+
+    &--single-select-master,
```
InnaAtanasova, line 175: Starting Jan 1st we need to get abandon the usage of words like `master`, `slave`, etc. Could you find another name for this class?

src/list-grid.scss
```diff
@@ -0,0 +1,261 @@
...
+        cursor: pointer;
+
+        @include fd-selected() {
+          cursor: default;
+        }
+      }
+    }
+  }
+
+  &__more {
```
InnaAtanasova, line 210: This needs a reset

src/list-grid.scss
```diff
@@ -0,0 +1,261 @@
...
+    cursor: pointer;
+    padding: 0.813rem;
+    background-color: var(--sapGroup_ContentBackground);
+    border-radius: $fd-grid-list-border-radius;
+    box-shadow: $fd-grid-list-item-box-shadow;
+    text-align: center;
+
+    .#{$more}--title {
```
InnaAtanasova, line 218: This is not a modifier class. It should be named `fd-grid-list__title` or `fd-grid-list__more-title`

src/list-grid.scss
```diff
@@ -0,0 +1,261 @@
...
+      font-weight: bold;
+    }
+
+    &--description {
```
InnaAtanasova, line 222: same here. It should be names `fd-grid-list__more-description` and since it's a standalone it needs a reset

src/list-grid.scss
```diff
@@ -0,0 +1,261 @@
...
+      background-color: $fd-grid-list-item-active-background-color;
+
+      .#{$more}--title,
+      .#{$more}--description {
+        color: $fd-grid-list-active-text-color;
+      }
+    }
+  }
+
+  &__footer {
```
InnaAtanasova, line 245: missing reset

src/list-grid.scss
```diff
@@ -0,0 +1,261 @@
...
+    color: var(--sapList_FooterTextColor);
+  }
+
+  @each $set-name, $colors in $fd-grid-list-highlight {
+    &__highlight--#{$set-name} {
```
InnaAtanasova, line 250: You are missing a `fd-grid-list__highlight` class that will bring the reset. You put there all the rules that are shared amongst all states (error, succes, etc). You then create a modifier class `fd-grid-list__highlight--error` that will overwrite the property specific for the base case. In this case it will be the color. The rest of the properties (position, top, left, etc. will go in the base class as they are shared). You need to have something like:
`<span class="fd-grid-list__highlight fd-grid-list__highlight--error">`

InnaAtanasova (changes requested):

stories/list-grid/list-grid.stories.js
```diff
@@ -0,0 +1,1880 @@
...
+                    </button>
+                </div>
+            </div>
+        </div>
+
+        <div class="fd-row">
+            <div class="fd-col fd-col--12 fd-col-md--6 fd-col-lg--4 fd-col-xl--3">
+                <div tabindex="0" class="fd-grid-list__item fd-grid-list__item--link">
+                    <input type="radio" class="fd-radio fd-radio--compact fd-grid-list__radio--input" id="pDidh761" name="singleSelectMasterRadio">
+                    <label class="fd-radio__label fd-grid-list__radio--label" for="pDidh761"></label>
```
meganaconley, line 289: This should not be an empty label. At the moment when you focus the radio button, the entire card reads out. There should be some sort of label to summarize what the item is (title, etc)

stories/list-grid/list-grid.stories.js
```diff
@@ -0,0 +1,1880 @@
...
+                </div>
+            </div>
+        </div>
+
+        <div class="fd-row">
+            <div  class="fd-col fd-col--12 fd-col-md--6 fd-col-lg--4 fd-col-xl--3">
+                <div tabindex="0" class="fd-grid-list__item fd-grid-list__item--link is-selected">
+                    <div class="fd-toolbar fd-grid-list__toolbar">
+                        <input type="checkbox" class="fd-checkbox fd-checkbox--compact fd-grid-list__checkbox--input" id="Ai4ez611c" checked>
+                        <label class="fd-checkbox__label fd-checkbox__label--compact fd-grid-list__checkbox--label" for="Ai4ez611c"></label>
```
meganaconley, line 614: Same feedback as for the radio buttons, except for some reason the content of the grid item doesn't read out when you access this checkbox, just 'checkbox'. Every input needs some sort of label.

stories/list-grid/list-grid.stories.js
```diff
@@ -0,0 +1,1880 @@
...
+                    <button class="fd-button fd-button--compact fd-button--transparent" aria-label="Group">
+                        <i class="sap-icon--group-2"></i>
+                    </button>
+                </div>
+            </div>
+        </div>
+
+        <div class="fd-row">
+            <div  class="fd-col fd-col--12 fd-grid-list__group-header">
+                State: Unread
```
meganaconley, line 1104: This group header is not announced when you tab into the group. It should be read upon accessing the first grid item. The same applies for all header examples.

stories/list-grid/list-grid.stories.js
```diff
@@ -0,0 +1,1880 @@
...
+                                </div>
+                                
+
+                                <a href="#" class="fd-link" tabindex="0">john_li@example.com</a>
+                            </div>
+                        </div>
+
+                        <div class="fd-grid-list__item--state-actions">
+                            <span class="fd-object-status fd-object-status--negative">
+                                <span class="fd-object-status__text">(Contains errors)</span>
```
meganaconley, line 1175: The only problem with this layout is that the error state is not read out until the end of the content. So a screen reader user would not know unless they sat there to listen to the entire card content until it got to this line. Ideally the validation state should be announced first.

stories/list-grid/list-grid.stories.js
```diff
@@ -0,0 +1,1880 @@
...
+                        <i class="sap-icon--group-2"></i>
+                    </button>
+                </div>
+            </div>
+        </div>
+
+        <div class="fd-row">
+            <div  class="fd-col fd-col--12 fd-col-md--6 fd-col-lg--4 fd-col-xl--3">
+                <div tabindex="0" class="fd-grid-list__item">
+                    <span class="fd-grid-list__highlight--success"></span>
```
meganaconley, line 1288: There needs to be some screen reader support for the highlight states so screen reader users are aware of a) that it's highlighted at all and b) what the color means. 

This should honestly go back to UX since this is using only color to indicate a state and that is not accessible for several user groups, not just the visually impaired.

stories/list-grid/list-grid.stories.js
```diff
@@ -0,0 +1,1880 @@
...
+                    <button class="fd-button fd-button--compact fd-button--transparent" aria-label="Group">
+                        <i class="sap-icon--group-2"></i>
+                    </button>
+                </div>
+            </div>
+        </div>
+
+        <div class="fd-row">
+            <div class="fd-col fd-col--12">
+                <div class="fd-toolbar fd-toolbar--info fd-toolbar--active">
```
meganaconley, line 1441: The content of this div is not announced to screen readers. The button is accessible, but simply announces 'Cancel' without context.

stories/list-grid/list-grid.stories.js
```diff
@@ -0,0 +1,1880 @@
...
+                            </div>
+                        </div>
+                    </div>
+                </div>
+            </div>
+        </div>
+
+        <div class="fd-row">
+            <div class="fd-col fd-col--12">
+                <div class="fd-grid-list__more">
```
meganaconley, line 1707: This section is not accessible by keyboard.

stories/list-grid/list-grid.stories.js
```diff
@@ -0,0 +1,1880 @@
...
+                        </div>
+                    </div>
+                </div>
+            </div>
+        </div>
+
+        <div class="fd-row">
+            <div class="fd-col fd-col--12">
+                <div class="fd-grid-list__footer">
+                    This is footer text
```
meganaconley, line 1863: This is not accessible to screen reader users.

meganaconley (changes requested):

Betrozov:
> Screen Shot 2020-11-12 at 11 56 34 AM
> 
> In this example the cursor is of type pointer on hover over all items. Shouldn't it be pointer only when it's over a clickable item?

@InnaAtanasova It is a single select mode we can select only one item from all items in the list that's why we need cursor is of type pointer here

InnaAtanasova:
> > Screen Shot 2020-11-12 at 11 56 34 AM
> > In this example the cursor is of type pointer on hover over all items. Shouldn't it be pointer only when it's over a clickable item?
> 
> @InnaAtanasova It is a single select mode we can select only one item from all items in the list that's why we need cursor is of type pointer here

In this case why the cursor is not of type pointer when you hover over the selected item?

InnaAtanasova:
Shouldn't the `delete` button be placed before the `navigation indicator`?
How it's now:
Screen Shot 2020-11-13 at 7 22 19 AM

In specs:
Screen Shot 2020-11-13 at 7 22 25 AM

InnaAtanasova:
I think you are missing an example in the documentation for `GridList with mode SingleSelect` where the radio buttons are displayed on the right hand side:
Screen Shot 2020-11-13 at 7 25 40 AM

Betrozov:
> Screen Shot 2020-11-12 at 11 58 48 AM
> 
> RTL issue with the Radio button. Please check if it's in the scope of this PR and if not, please log an issue. Thanks

fixed

Betrozov:
> > > Screen Shot 2020-11-12 at 11 56 34 AM
> > > In this example the cursor is of type pointer on hover over all items. Shouldn't it be pointer only when it's over a clickable item?
> > 
> > 
> > @InnaAtanasova It is a single select mode we can select only one item from all items in the list that's why we need cursor is of type pointer here
> 
> In this case why the cursor is not of type pointer when you hover over the selected item?

fixed

Betrozov:
> Shouldn't the `delete` button be placed before the `navigation indicator`?
> How it's now:
> Screen Shot 2020-11-13 at 7 22 19 AM
> 
> In specs:
> Screen Shot 2020-11-13 at 7 22 25 AM

fixed

Betrozov:
> I think you are missing an example in the documentation for `GridList with mode SingleSelect` where the radio buttons are displayed on the right hand side:
> Screen Shot 2020-11-13 at 7 25 40 AM

added

src/list-grid.scss
Betrozov: fixed

src/list-grid.scss
Betrozov: fixed

src/list-grid.scss
Betrozov: fixed

src/list-grid.scss
Betrozov: fixed

src/list-grid.scss
Betrozov: fixed

src/list-grid.scss
Betrozov: fixed

src/list-grid.scss
Betrozov: fixed

src/list-grid.scss
Betrozov: fixed

src/list-grid.scss
Betrozov: fixed

src/list-grid.scss
Betrozov: fixed

src/list-grid.scss
Betrozov: fixed

src/list-grid.scss
Betrozov: fixed

src/list-grid.scss
Betrozov: fixed

src/list-grid.scss
Betrozov: fixed

Betrozov: thank you for the comment :slightly_smiling_face: 
@InnaAtanasova

src/list-grid.scss
Betrozov: fixed

src/list-grid.scss
Betrozov: fixed

src/list-grid.scss
Betrozov: fixed

src/list-grid.scss
Betrozov: fixed

src/list-grid.scss
Betrozov: fixed

src/list-grid.scss
Betrozov: fixed

stories/list-grid/list-grid.stories.js
Betrozov: fixed

stories/list-grid/list-grid.stories.js
Betrozov: fixed

stories/list-grid/list-grid.stories.js
Betrozov: fixed

stories/list-grid/list-grid.stories.js
Betrozov: fixed

stories/list-grid/list-grid.stories.js
Betrozov: fixed

stories/list-grid/list-grid.stories.js
Betrozov: fixed

stories/list-grid/list-grid.stories.js
Betrozov: fixed

stories/list-grid/list-grid.stories.js
Betrozov: fixed
thank you for the comments @meganaconley

src/list-grid.scss
```diff
@@ -0,0 +1,301 @@
...
+    @include fd-reset();
+
+    @include fd-flex(column) {
+      justify-content: space-between;
+    }
+
+    height: 100%;
+    padding: 1rem;
+
+    &---no-padding {
```
InnaAtanasova, line 132: Should be 2 `--` not 3

src/list-grid.scss
```diff
@@ -0,0 +1,301 @@
...
+    order: 1;
+  }
+
+  &__item-counter {
+    @include fd-fiori-focus(0);
+
+    color: var(--sapContent_MarkerTextColor);
+  }
+
+  .#{$block}__toolbar {
```
InnaAtanasova, line 148: should be renamed to `__item-toolbar` as it's part of the item

src/list-grid.scss
Betrozov: fixed

src/list-grid.scss
Betrozov: fixed

InnaAtanasova:
Sorry, one last thing for the RTL:
Screen Shot 2020-11-18 at 8 43 21 AM
Screen Shot 2020-11-18 at 8 43 13 AM

You just need to move the margin to the other side on RTL
Screen Shot 2020-11-18 at 8 43 47 AM

Betrozov:
> Sorry, one last thing for the RTL:
> Screen Shot 2020-11-18 at 8 43 21 AM
> Screen Shot 2020-11-18 at 8 43 13 AM
> 
> You just need to move the margin to the other side on RTL
> Screen Shot 2020-11-18 at 8 43 47 AM

fixed

InnaAtanasova (approved):

stories/list-grid/list-grid.stories.js
```diff
@@ -0,0 +1,2069 @@
...
+                            </div>
+                        </div>
+                    </div>
+                </div>
+            </div>
+        </div>
+
+        <div class="fd-row">
+            <div class="fd-col fd-col--12">
+                <div tabindex="0" class="fd-grid-list__more">
```
meganaconley, line 1896: Sorry I missed this initially, but this should have `role="button"` as well if it can't be changed to be a true button.

meganaconley (changes requested):
Great work on the accessibility feedback. I really like the callouts for the highlighted items and error state.

I think there were some regressions though. The focus ring for the checkbox/radio buttons is off center. There is an invisible focusable radio button on the single select cards as well. The focus for that one should be just around the card, and using arrow keys should navigate between them. I believed these looked correct last time I reviewed.

Screen Shot 2020-11-18 at 11 52 28 AM

Screen Shot 2020-11-18 at 11 53 10 AM

Screen Shot 2020-11-18 at 11 53 36 AM

stories/list-grid/list-grid.stories.js
Betrozov: fixed))

Betrozov:
> Great work on the accessibility feedback. I really like the callouts for the highlighted items and error state.
> 
> I think there were some regressions though. The focus ring for the checkbox/radio buttons is off center. There is an invisible focusable radio button on the single select cards as well. The focus for that one should be just around the card, and using arrow keys should navigate between them. I believed these looked correct last time I reviewed.
> 
> Screen Shot 2020-11-18 at 11 52 28 AM
> 
> Screen Shot 2020-11-18 at 11 53 10 AM
> 
> Screen Shot 2020-11-18 at 11 53 36 AM

fixed

meganaconley (approved):

PR #1856, opened 2020-11-05
fix: Make stronger selector for message strip

JKMarkowski:
## Description
There is stronger selector for message-strip close element.

## Screenshots

No visual/usage changes predicted

### DIFF
--- a/src/message-strip.scss
+++ b/src/message-strip.scss
@@ -67,7 +67,7 @@ $block: #{$fd-namespace}-message-strip;
   min-height: 2rem;
 
   // Elements
-  &__close {
+  &__close.#{$fd-namespace}-button {
     @include fd-message-strip-close-btn-container();
 
     @include fd-rtl() {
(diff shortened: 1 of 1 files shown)


JKMarkowski:
Chromatic detected  changes in checkbox/radio buttons. Which i guess are leftovers

src/message-strip.scss
```diff
@@ -67,7 +67,7 @@ $block: #{$fd-namespace}-message-strip;
   min-height: 2rem;
 
   // Elements
-  &__close {
+  &__close.#{$fd-namespace}-button {
```
jbadan, line 70: I thought we agreed not to do this in the last few meetings?

jbadan (changes requested):
We should not mix `fd-*` classes in the same file. It breaks css modules in fundamental-react.

JKMarkowski:
> We should not mix `fd-*` classes in the same file. It breaks css modules in fundamental-react.

I just added second level of nesting, to make it stronger

jbadan (approved):

droshev (approved):

PR #1858, opened 2020-11-05
chore: update pr template

jbadan:
Update pr template to include using only one `fd-` class per file


Mike-Diaz (approved):
Unfortunate the list keeps getting bigger :( but a necessary addition.

Issue #1779, opened 2020-10-13
Bug: Panel shows double arrow in RTL mode


KevinOkamoto:
**Describe the bug**

In "RTL" mode the Panel component shows double arrows.

> Is this issue related to a specific component?

Panel.

> What version of the Fundamental Library Styles are you using?

v0.12.0

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

Fundamental-NGX

> What front-end framework are you implementing in, e.g., Angular, React, etc.?

Angular

**To Reproduce**
1. Go to https://sap.github.io/fundamental-styles/iframe.html?id=components-panel--expandable&viewMode=story
2. Open developer tools
3. Using developer tool add following attribute to `<body>`:  `dir="rtl"`

**Expected behavior**
In RTL mode should only show single arros

**Screenshots**
![Screen Shot 2020-10-13 at 9 21 08 AM](https://user-images.githubusercontent.com/48103491/95888519-09dff780-0d36-11eb-871d-2854616013c7.png)

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Chrome
 - Version: 85.0.4183.121

PR #1861, opened 2020-11-06
fix: RTL mode arrow button direction

olshansky:
## Related Issue
Closes https://github.com/SAP/fundamental-styles/issues/1779
Related https://github.com/SAP/fundamental-ngx/issues/3674 

## Description
- Rotate arrow icon

### Before:
image

### After:
image

### DIFF
--- a/src/panel.scss
+++ b/src/panel.scss
@@ -55,7 +55,9 @@ $block: #{$fd-namespace}-panel;
     font-size: 1rem;
 
     @include fd-rtl() {
-      @include fd-icon('slim-arrow-left');
+      .sap-icon--slim-arrow-right {
+        transform: rotate(180deg);
+      }
     }
   }
 


InnaAtanasova (changes requested):
Default:
Screen Shot 2020-11-06 at 11 06 01 AM

RTL:
Screen Shot 2020-11-06 at 11 06 10 AM

The Panel button does not "rotate" when we switch to RTL. It only changes in the RTL example but since we have a class for this button we can rotate it with styling

src/panel.scss
```diff
@@ -53,10 +53,6 @@ $block: #{$fd-namespace}-panel;
 
   &__button {
     font-size: 1rem;
-
-    @include fd-rtl() {
-      @include fd-icon('slim-arrow-left');
```
InnaAtanasova, line 58: You can keep the `fd-rtl` mixin and just rotate the icon, not add a new one.

stories/panel/panel.stories.js
```diff
@@ -197,3 +197,64 @@ fixedHeightContent.parameters = {
         iframeHeight: 300
     }
 };
+
+export const rtl = () => `
```
InnaAtanasova, line 201: Screen Shot 2020-11-06 at 11 14 23 AM

You don't have to add this example, we have a "switcher" on the top of the doc page. All examples should apply the RTL rules when you switch to this mode.

src/panel.scss
olshansky: @InnaAtanasova thanks, fixed

InnaAtanasova (approved):

mikerodonnell89 (approved):

mikerodonnell89 (approved):

JKMarkowski (approved):

JKMarkowski:
@olshansky Can you push some empty commit to this branch, just to trigger chromatic pipeline?

PR #1865, opened 2020-11-06
fix: accessibility improvements + integrate automated accessibility testing with jest-axe

prsdthkr:
## Description
* Introduce [jest-axe](https://www.npmjs.com/package/jest-axe) automated accessibility testing. This will help us automatically catch some accessibility issues. Run `npm run test:accessibility` to run accessibility tests. These tests are also part of the build now.
* Accessibility improvements for multiple components to pass all jest-axe testing.

BREAKING CHANGE: DOM updated for multiple components for accessibility purposes. Notably, tab > tab items need the tab role and have aria-selected set to true, instead of the link within.

- [n/a] RTL support
- [x] A11y support - keyboard support, screen reader support, proper ARIA attributes, etc.
3. Testing
- [x] Updated tests
4. Documentation


jbadan:
This looks so nice! 🎉  I'm going to hold off approving until other get a chance to play around with this/understand how it works

InnaAtanasova:
This is super cool! Thanks for introducing it to the library. For sure will be very extremely helpful!

InnaAtanasova (approved):

src/tabs.scss
```diff
@@ -179,6 +160,25 @@ $block: #{$fd-namespace}-tabs;
...
+
+      .#{$block}__icon,
+      .#{$block}__tag,
+      & {
+        &::after {
+          background-color: var(--fdTabs_Selected_Color);
+        }
+      }
+    }
+
```
JKMarkowski, line 181: I think it will be better to mention about this change separately. I saw you added annotation, that many component has been changed, but in this case it requires serious action from user.

JKMarkowski (approved):
Looks great for me. It will save a lot of time on a11y validation. I left one comment about tabs breaking change.

meganaconley (approved):
Awesome work!

Issue #1880, opened 2020-11-12
Introduce Form Group Header to Form Group


kavya-b:
**Is your feature request related to a problem? Please describe.**
Group Header styles need to be implemented as per Fiori3.

> Is this issue related to a specific component?

Form Group

PR #1881, opened 2020-11-12
feat: introduce Form Group Header to Form Group

kavya-b:
## Related Issue
Closes SAP/fundamental-styles#1880

## Description
PR introduces Form Group Header to Form Group

### Before:
N/A

### After:
Screenshot 2020-11-12 at 10 04 42 PM
Screenshot 2020-11-12 at 10 04 56 PM

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/form-layout-grid.scss
+++ b/src/form-layout-grid.scss
@@ -15,7 +15,6 @@ $cols: 11;
 
 @media (min-width: 0) {
   .#{$blockContainer}.#{$block}-container {
-    padding: 1rem;
     margin: 0 -0.25rem;
 
     & > .fd-row {
@@ -53,7 +52,6 @@ $cols: 11;
 
         .fd-form-label {
           display: inline-block;
-          margin-right: 0;
           padding-bottom: 0.125rem;
         }
 
@@ -84,6 +82,10 @@ $cols: 11;
         .#{$blockRow} {
           margin: 0 -0.25rem;
         }
+      }
+
+      .#{$fd-namespace}-form-group.#{$blockCol} {
+        padding: 1rem;
       }
     }
 
--- a/src/form-layout-grid.scss
+++ b/src/form-layout-grid.scss
@@ -92,10 +92,10 @@ $cols: 11;
         }
       }
 
-      .#{$fd-namespace}-form-group.#{$blockCol} {
+      .#{$blockCol}__form-group {
         padding: 1rem;
       }
-      .#{$fd-namespace}-form-item.#{$blockRow} {
+      .#{$blockRow}__form-item {
         flex-direction: row;
       }
     }
--- a/src/form-layout-grid.scss
+++ b/src/form-layout-grid.scss
@@ -52,7 +52,15 @@ $cols: 11;
 
         .fd-form-label {
           display: inline-block;
+          margin-right: 0;
+          padding-right: 0.5rem;
           padding-bottom: 0.125rem;
+
+          @include fd-rtl() {
+            padding-left: 0.5rem;
+            margin-left: 0;
+            padding-right: 0;
+          }
         }
 
         @for $n from 0 through $cols {
@@ -86,6 +94,9 @@ $cols: 11;
 
       .#{$fd-namespace}-form-group.#{$blockCol} {
         padding: 1rem;
+      }
+      .#{$fd-namespace}-form-item.#{$blockRow} {
+        flex-direction: row;
       }
     }
 
--- a/stories/form-group/form-group.stories.js
+++ b/stories/form-group/form-group.stories.js
@@ -7,7 +7,7 @@ export default {
         These components can be used alone. For example, the 'form__item' element with the label and control could  
         be used without the full form group since not every form field will need error messages.`,
         tags: ['f3', 'a11y', 'theme'],
-        components: ['form-group', 'form-item', 'form-label', 'input']
+        components: ['form-layout-grid', 'form-group', 'form-item', 'form-label', 'input']
     }
 };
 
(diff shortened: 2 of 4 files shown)


kavya-b:
I have tried running `npm run test` and `npm run test:update` but strangely the storyshots don't update with the newly added code. I get only empty blocks like this:
image
Therefore I haven't added the storyshot file here and travis will fail until I fix this.
Am I doing something wrong, @InnaAtanasova @prsdthkr ?

stories/form-group/form-group.stories.js
```diff
@@ -36,3 +36,125 @@ export const required = () => `<div class="fd-form-group">
         </div>
     </div>
 `;
+
+export const groupHeader = () => `
+    <div class="fd-form-group">
+        <div class="fd-form-group__header">
+            <span class="fd-form-group__header-text">Group Header</span>
```
meganaconley, line 43: I see you checked off the a11y requirements on the PR template, what did you test this with? I'm not able to get this read out when accessing the form with VoiceOver. I think it needs to be associated using `aria-labelledby`. See Approach 2 here for an example:

https://www.w3.org/WAI/tutorials/forms/grouping/

meganaconley (changes requested):

stories/form-group/form-group.stories.js
kavya-b: pressing Ctrl+Option+Shift+Right Arrow read out the header text for me in VoiceOver.

stories/form-group/form-group.stories.js
```diff
@@ -36,3 +36,125 @@ export const required = () => `<div class="fd-form-group">
...
+                <label class="fd-form-label" for="input-14c-name">Name:</label>
+                </div>
+                <div class="fd-col fd-col-md--10 fd-col-lg--8 fd-col-xl--12">
+                <input class="fd-input" type="text" id="input-14c-name" placeholder="Enter First and Last Name" value="Amelia Perry">
+                </div>
+            </div>
+        </div>
+        <div class="fd-form-group fd-col fd-col-md--6 fd-col-lg--6 fd-col-xl--6 fd-col--wrap">
+            <div class="fd-form-group__header">
+                <span class="fd-form-group__header-text">Group Header 3</span>
```
salarenko, line 135: Shouldn't he a `h` tag?

stories/form-group/form-group.stories.js
```diff
@@ -36,3 +36,125 @@ export const required = () => `<div class="fd-form-group">
...
+            </div>
+            <div class="fd-row">
+                <div class="fd-col fd-col-md--2 fd-col-lg--4 fd-col-xl--12">
+                <label class="fd-form-label" for="input-14j-name">Name:</label>
+                </div>
+                <div class="fd-col fd-col-md--10 fd-col-lg--8 fd-col-xl--12">
+                <input class="fd-input" type="text" id="input-14j-name" placeholder="Enter First and Last Name" value="Amelia Perry">
+                </div>
+                <div class="fd-col fd-col-md--2 fd-col-lg--4 fd-col-xl--12">
+                <label class="fd-form-label" for="input-14h-name">Name:</label>
```
salarenko, line 145: I'm not sure if keeping label and input in separate grid columns is a good idea. You need to make double work to maintain responsive grid. In my opinion this is quite unusual approach bringing to value but more code into the HTML.

stories/form-group/form-group.stories.js
```diff
@@ -36,3 +36,125 @@ export const required = () => `<div class="fd-form-group">
...
+                </div>
+            </div>
+        </div>
+
+        <div class="fd-form-group fd-col fd-col-md--6 fd-col-lg--6 fd-col-xl--6 fd-col--wrap">
+            <div class="fd-form-group__header">
+                <span class="fd-form-group__header-text">Group Header 2</span>
+            </div>
+            <div class="fd-row">
+                <div class="fd-col fd-col-md--2 fd-col-lg--4 fd-col-xl--12">
```
salarenko, line 113: This examples should use `form-item`.

salarenko (commented):
- [x] Check A11y violations (mostly non-unique id issues)
- [ ] We don't need separate RTL examples, user can switch the direction using documentation built in control.

salarenko (changes requested):
Needs changes

stories/form-group/form-group.stories.js
```diff
@@ -36,3 +36,125 @@ export const required = () => `<div class="fd-form-group">
...
+                </div>
+            </div>
+        </div>
+        <div class="fd-form-group fd-col fd-col-md--6 fd-col-lg--6 fd-col-xl--6 fd-col--wrap">
+            <div class="fd-form-group__header">
+                <span class="fd-form-group__header-text">Group Header 3</span>
+            </div>
+            <div class="fd-row">
+                <div class="fd-col fd-col-md--2 fd-col-lg--4 fd-col-xl--12">
+                <label class="fd-form-label" for="input-14j-name">Name:</label>
```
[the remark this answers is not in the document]
kavya-b: fixed

stories/form-group/form-group.stories.js
kavya-b: replaced with `h1` tags

stories/form-group/form-group.stories.js
kavya-b: I am just following the existing Form Grid example here and have not modified any underlying logic for responsive grid column usage.

stories/form-group/form-group.stories.js
kavya-b: using `fd-form-item` now

stories/form-group/form-group.stories.js
kavya-b: added `aria-labelledby` with `id` and `role=group` to get VoiceOver to read the group name on entering the group.

src/form-layout-grid.scss
```diff
@@ -85,6 +91,13 @@ $cols: 11;
           margin: 0 -0.25rem;
         }
       }
+
+      .#{$fd-namespace}-form-group.#{$blockCol} {
+        padding: 1rem;
+      }
+      .#{$fd-namespace}-form-item.#{$blockRow} {
+        flex-direction: row;
+      }
```
JKMarkowski, line 100: Some time ago we decided to not mix blocks in files. Is there any way. Could you create other, stronger element to overwrite `.#{$fd-namespace}-form-group` / `.#{$fd-namespace}-form-item`. Or use them in different elements?

JKMarkowski: image

JKMarkowski (changes requested):
Check my previous comment, let me know if something is not clear

src/form-layout-grid.scss
kavya-b: so would it be okay to have separate element modifiers such as this?:
image

kavya-b:
@JKMarkowski  using `fd-row__form-item` and `fd-col__form-group` elements now, please check.

meganaconley (approved):

src/form-layout-grid.scss
JKMarkowski: Yes, it's much better there

JKMarkowski (approved):
LGTM

InnaAtanasova (approved):

kavya-b:
@salarenko could you please check this again, your comments were addressed; thanks!

salarenko (approved):

Issue #1878, opened 2020-11-12
Introduce Flexible Column Layout


droshev:
Interaction : https://experience.sap.com/fiori-design-web/flexible-column-layout/

2d

PR #1887, opened 2020-11-16
feat: flexible column layout

JKMarkowski:
## Related Issue
Closes SAP/fundamental-styles#1878

## Description
There is added `flexible column layout` component with delta theming and examples of usage.

- [x] All values are in `rem`
- [NA] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/flexible-column-layout.scss
+++ b/src/flexible-column-layout.scss
@@ -5,6 +5,8 @@ $block: #{$fd-namespace}-flexible-column-layout;
 
 .#{$block} {
 
+  $column-layout-expand-decoration-image: var(--fdFlexibleColumnLayoutToggleImage);
+
   $toggle-button-height: 2rem;
 
   @mixin both-pseudo-selectors() {
@@ -43,12 +45,12 @@ $block: #{$fd-namespace}-flexible-column-layout;
 
     &::before {
       bottom: calc(50% + #{$toggle-button-height} / 2);
-      background-image: linear-gradient(to top, var(--sapHighlightColor), transparent);
+      background-image: linear-gradient(to top, var(--fdFlexibleColumnLayoutToggleImageColor), transparent);
     }
 
     &::after {
       top: calc(50% + #{$toggle-button-height} / 2);
-      background-image: linear-gradient(to bottom, var(--sapHighlightColor), transparent);
+      background-image: linear-gradient(to bottom, var(--fdFlexibleColumnLayoutToggleImageColor), transparent);
     }
 
     @include fd-hover() {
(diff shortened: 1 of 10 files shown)


src/flexible-column-layout.scss
```diff
@@ -0,0 +1,98 @@
+@import "./new-settings";
+@import "./mixins";
+
+$block: #{$fd-namespace}-flexible-column-layout;
+
+.#{$block} {
+
+  $column-layout-expand-decoration-image: var(--fdFlexibleColumnLayoutToggleImage);
+
+  $toggle-button-height: 2rem;
```
InnaAtanasova, line 10: these 2 should have the `fd-` prefix in the naming

src/flexible-column-layout.scss
```diff
@@ -0,0 +1,98 @@
...
+
+  @mixin both-pseudo-selectors() {
+    &::after,
+    &::before {
+      @content;
+    }
+  }
```
InnaAtanasova, line 17: maybe move this to the `_mixins.scss` file

src/flexible-column-layout.scss
```diff
@@ -0,0 +1,98 @@
...
+      @include both-pseudo-selectors() {
+        height: 7rem;
+      }
+    }
+  }
+
+  .#{$block}__button {
+    height: $toggle-button-height;
+    width: 1.5rem;
+    min-width: 1.5rem;
```
InnaAtanasova, line 60: The value should be a variable

src/flexible-column-layout.scss
```diff
@@ -0,0 +1,98 @@
...
+      display: block;
+      width: 0.0625rem;
+      height: 4rem;
+      transition: all 0.1s ease-in;
+
+      @include fd-absolute-horizontal-center();
+    }
+
+    &::before {
+      bottom: calc(50% + #{$toggle-button-height} / 2);
```
InnaAtanasova, line 41: This can be a variable and used in bottom and top

src/mixins/_mixins.scss
```diff
@@ -324,6 +324,21 @@
...
+  @content;
+}
+
+@mixin fd-absolute-center {
+  position: absolute;
+  left: 50%;
+  top: 50%;
+  transform: translate(-50%, -50%);
+  @content;
+}
```
InnaAtanasova, line 340: I don't think @content is appropriate in these cases

InnaAtanasova (changes requested):

JKMarkowski:
@InnaAtanasova Thank You for checking, you comments has been addressed, could you take a look again? :)

stefanoScalzo:
Screen Shot 2020-11-17 at 12 11 42 PM
Screen Shot 2020-11-17 at 12 11 38 PM
The button is going behind

katekozlowska (approved):
Good job with this component! Only one thing - required value `min-width: 20rem,` makes presented example look broken (f.e. ratio 1:2 in two-columns layout looks incorrect), but it works fine, when I'm testing it separately on different examples.

src/mixins/_mixins.scss
```diff
@@ -351,6 +364,13 @@
   font-size: $fd-size;
 }
 
+@mixin both-pseudo-selectors() {
```
InnaAtanasova, line 367: sorry, could you add `fd-` for consistency?

InnaAtanasova (approved):

stefanoScalzo:
LGTM

stefanoScalzo (approved):

Issue #1400, opened 2020-08-18
List - Item Counter


JKMarkowski:
List has missing item counter element, it should be included as standalone element, which has some width restrictions and specified color

PR #1897, opened 2020-11-18
feat: add Item Counter to List component

InnaAtanasova:
## Related Issue
Closes SAP/fundamental-styles#1400

## Description
introduce an enhancement to List component - Item Counter

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation


droshev (approved):

salarenko (approved):

JKMarkowski (approved):
LGTM

PR #1898, opened 2020-11-19
feat: introduce Quick View component

EvilAlexei:
## Related Issue
No related issue

## Description
This PR introduces the Quick View component

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation


dimamarksman:
Looks like an extra element:

image

dimamarksman:
Two nested elements with the same selector:

image

dimamarksman:
It would be great to add an example in RTL mode

Lokanathan-k:
HI @EvilAlexei 
content are not truncating instead expanding the popover.

Expected: 

image

Actual implemented:

image

stories/quick-view/quick-view.stories.js
```diff
@@ -0,0 +1,257 @@
...
+                        </div>
+                        <div class="fd-quick-view__group-item-content">
+                            781 Main Street <br>
+                            Anytown, SD 57401, USA
+                        </div>
+                    </div>
+                </div>
+                
+                <div class="fd-quick-view__group">
+                    <div class="fd-quick-view__group__title">
```
katekozlowska, line 66: it should be `-title` instead of `__title`

katekozlowska (changes requested):

EvilAlexei:
> Looks like an extra element:
> 
> image

This element is a content holder inside flex layout, also there is a fix for flex width :)

dimamarksman:
> > Looks like an extra element:
> 
> This element is a content holder inside flex layout, also there is a fix for flex width :)

If it's just a container to follow a flex layout we should name it properly cause `fd-quick-view__text` does not describe the purpose it's used for.

InnaAtanasova:
Hi Oleksii,
from what I see in the specs Quick View is a composite component consisting of Header+Subheader+Content. Header and Subheader are extending the functionality of Bar component and the Content is extending the functionality of Form-Display Mode. This means that this component should contain classes that overwrite the classes of the mentioned components (if needed). For example if the header of the Quick view is different from the Bar header, then you create a modifier class `fd-bar fd-bar--header fd-quick-view__header` and `fd-quick-view__header` does not bring the reset. In the specs I didn't see any rule overwritten, which makes me think that this component should go into the "Patterns" category and be built out of already existing components.

stefanoScalzo:
Screen Shot 2020-12-14 at 11 47 03 AM
z index should be fixed

EvilAlexei:
> > > Looks like an extra element:
> > 
> > 
> > This element is a content holder inside flex layout, also there is a fix for flex width :)
> 
> If it's just a container to follow a flex layout we should name it properly cause `fd-quick-view__text` does not describe the purpose it's used for (`fd-quick-view__text` looks like a common thing for any text element in scope of this component).

It's a name based on semantic content, and it's a common name for such content. 
Also, we already have similar class naming in a similar structure, in the card component: https://sap.github.io/fundamental-ngx/#/core/card#standard

In my opinion semantic name better than styling naming. It's higher chances that styles will be changed, but the type of content has fewer chances to be changed.

stefanoScalzo:
Screen Shot 2020-12-14 at 11 47 55 AM
we should have some type of overflow for the title

EvilAlexei:
> Screen Shot 2020-12-14 at 11 47 55 AM
> we should have some type of overflow for the title

It's based on the content used for the title. For example, if we use title https://prnt.sc/w2kp19

EvilAlexei:
> Screen Shot 2020-12-14 at 11 47 03 AM
> z index should be fixed

This is a popover issue, not related to the quick view styles.

dimamarksman:
> > > > Looks like an extra element:
> > > 
> > > 
> > > This element is a content holder inside flex layout, also there is a fix for flex width :)
> > 
> > 
> > If it's just a container to follow a flex layout we should name it properly cause `fd-quick-view__text` does not describe the purpose it's used for (`fd-quick-view__text` looks like a common thing for any text element in scope of this component).
> 
> It's a name based on semantic content, and it's a common name for such content.
> Also, we already have similar class naming in a similar structure, in the card component: https://sap.github.io/fundamental-ngx/#/core/card#standard
> 
> In my opinion semantic name better than styling naming. It's higher chances that styles will be changed, but the type of content has fewer chances to be changed.

From provided `card` example the class name is `fd-card__header-text` that means `text` area in `header` scope. But here you use `fd-quick-view__text` that is more general thing. Let's use `fd-quick-view__header-text` or something similar to point the scope this `text` is belongs to.

dimamarksman (approved):

src/quick-view.scss
```diff
@@ -0,0 +1,100 @@
...
+
+      &-content {
+        @include fd-reset();
+
+        padding-top: 0.25rem;
+        padding-bottom: 0.25rem;
+      }
+    }
+  }
+}
```
JKMarkowski, line 100: Could you try to reuse current [form group](https://fundamental-styles.netlify.app/iframe.html?id=components-forms-form-group--primary&viewMode=docs#group-header) in examples, but in readonly mode (so just change inputs to spans)

src/quick-view.scss
```diff
@@ -0,0 +1,100 @@
...
+$block: #{$fd-namespace}-quick-view;
+
+$fd-quick-view-avatar-indent: 0.75rem;
+$fd-quick-view-title-bottom-indent: 0.5rem;
+
+.#{$block} {
+  @include fd-reset();
+
+  min-width: 20rem;
+  max-width: 30rem;
```
JKMarkowski, line 13: Where are those limits from?

src/quick-view.scss
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  &__header {
+    @include fd-reset();
+
+    > *:first-child {
+      border-top-right-radius: $fd-popover-border-radius;
+      border-top-left-radius: $fd-popover-border-radius;
+    }
+  }
```
JKMarkowski, line 28: Could you use in examples bar with `--header` modifier, instead of having separated `__header` element?

src/quick-view.scss
```diff
@@ -0,0 +1,100 @@
...
+
+    > *:first-child {
+      margin-right: $fd-quick-view-avatar-indent;
+    }
+
+    &-text {
+      max-width: 100%;
+      min-width: 0; // fix for flex width https://github.com/philipwalton/flexbugs#flexbug-17
+    }
+  }
```
JKMarkowski, line 49: Could you reuse other bar with `--header-with-subheader` in examples, instead of having this element?

src/quick-view.scss
```diff
@@ -0,0 +1,100 @@
...
+
+  &__title {
+    @include fd-reset();
+
+    font-size: var(--sapFontHeader5Size);
+    color: var(--sapTile_TitleTextColor);
+    margin-bottom: $fd-quick-view-title-bottom-indent;
+  }
```
JKMarkowski, line 57: Let's reuse title component, instead of having it

JKMarkowski (changes requested):
Hello @EvilAlexei I added some comments, that will drastically change component structure. This should be more composition of components, not standalone component. I can't see any specification for quick view. PM me know if you willl have any question

src/quick-view.scss
EvilAlexei: Changed

src/quick-view.scss
EvilAlexei: I wasn't able to find any info about those limits in the design docs,
but this component has already pre-defined content and I think those limits are necessary for the view perception.

For `min-width` I used a width value from [examples], (https://sapui5.hana.ondemand.com/#/entity/sap.m.QuickView/sample/sap.m.sample.QuickView) They all have `320px` width.

I wasn't able to use max-width as well from those examples because it is dynamic there,
so `30rem` is my assumption

src/quick-view.scss
EvilAlexei: changed

src/quick-view.scss
EvilAlexei: changed

src/quick-view.scss
EvilAlexei: changed

src/quick-view.scss
JKMarkowski: Could you then put it as a hardcoded value into examples? You can use inline-styling or check how it's done with dialog

src/quick-view.scss
```diff
@@ -0,0 +1,93 @@
...
+    }
+  }
+
+  min-width: 20rem;
+  max-width: 30rem;
+
+  > *:first-child {
+    border-top-right-radius: $fd-popover-border-radius;
+    border-top-left-radius: $fd-popover-border-radius;
+  }
```
JKMarkowski, line 27: I would go for `overflow: hidden` instead of this

src/quick-view.scss
```diff
@@ -0,0 +1,93 @@
...
+  @include fd-reset();
+
+  @include fd-rtl() {
+    .#{$fd-namespace}-bar {
+      .#{$fd-namespace}-avatar {
+        margin-right: 0;
+        margin-left: $fd-quick-view-avatar-indent;
+      }
+    }
+  }
```
JKMarkowski, line 19: Could you move it to `42` line, so it will be 
```
      .#{$fd-namespace}-avatar {
            margin-right: $fd-quick-view-avatar-indent;
           @include fd-rtl() { ....  }
      }
```

src/quick-view.scss
```diff
@@ -0,0 +1,93 @@
...
+    padding: 0;
+    border: none;
+  }
+
+  .#{$fd-namespace}-dialog__body & {
+    &__content {
+      padding-top: 0;
+      padding-bottom: 0;
+    }
+  }
```
JKMarkowski, line 80: I don't think if it's used anywhere

stories/quick-view/quick-view.stories.js
```diff
@@ -0,0 +1,258 @@
...
+                        <span class="fd-input">Weekly</span>
+                    </div>
+                </div>
+            </div>
+        </div>
+    </div>
+</div>
+`;
+
+RTLMode.storyName = 'RTL mode';
```
JKMarkowski, line 251: Let's also remove RTL mode example, there is a way to control direction on docs header
image

JKMarkowski (changes requested):
Hello @EvilAlexei Great job with refactoring this component. You are almost there, I added some code related comments

src/quick-view.scss
EvilAlexei: It's used in the dialog to prevent double padding, because of the `fd-dialog__body` padding.

src/quick-view.scss
EvilAlexei: The implemented version, in my opinion, has more common selectors
Implemented:
```
[dir="rtl"] .fd-quick-view .fd-bar--header-with-subheader .fd-avatar,
.fd-quick-view[dir="rtl"] .fd-bar--header-with-subheader .fd-avatar {...}
```

Suggested:
```
[dir="rtl"] .fd-quick-view .fd-bar--header-with-subheader .fd-avatar,
.fd-quick-view .fd-bar--header-with-subheader .fd-avatar[dir="rtl"] {...}
```
Just I see no use in `.fd-avatar[dir="rtl"]`

But I think now that it could be better to be change bar selector to be more specific:
```
[dir="rtl"] .fd-quick-view .fd-bar--header-with-subheader .fd-avatar,
.fd-quick-view[dir="rtl"] .fd-bar--header-with-subheader .fd-avatar {...}
```

src/quick-view.scss
EvilAlexei: I thought about `overflow: hidden`, it's easier, but it could cause issues for inner content, for example for tooltips, or popups.
I think border-radius will be better in this case.

EvilAlexei:
Hello @JKMarkowski, thank you for your comments!

About `min-width` and `max-width`, just I think it's better to have these limitations for a better view,
because of pre-defined content and structure.

About min-width:
For example, it's ok for the popover to have no such limitations, It could be used with any sort of content,
but the dialog has pre-defined `min-width` because you have an assumption about content there.

About max-width:
I see two almost the same ways setting width here, to set a specific width, or to have min/max-width.
With the specific width, a user has to set it all the time.
But with such limitations, it could be more dynamic, and a user still able to set width, but it's not necessary.

stories/quick-view/quick-view.stories.js
EvilAlexei: changed

stefanoScalzo:
> > Screen Shot 2020-12-14 at 11 47 55 AM
> 
> > we should have some type of overflow for the title
> 
> It's based on the content used for the title. For example, if we use fd-title
> download

Then maybe we should use fd-title, there should be some type of ellipses standard for text overflow

EvilAlexei:
> > > Screen Shot 2020-12-14 at 11 47 55 AM
> > 
> > 
> > > we should have some type of overflow for the title
> > 
> > 
> > It's based on the content used for the title. For example, if we use fd-title
> > download
> 
> Then maybe we should use fd-title, there should be some type of ellipses standard for text overflow

Yeap, I already used it with `fd-title`, and `fd-link` with title needs additional adjustments, I added styles for that, now it should work as requested.

JKMarkowski:
> About `min-width` and `max-width`, just I think it's better to have these limitations for a better view,
> because of pre-defined content and structure.
> 
> About min-width:
> For example, it's ok for the popover to have no such limitations, It could be used with any sort of content,
> but the dialog has pre-defined `min-width` because you have an assumption about content there.
> 
> About max-width:
> I see two almost the same ways setting width here, to set a specific width, or to have min/max-width.
> With the specific width, a user has to set it all the time.
> But with such limitations, it could be more dynamic, and a user still able to set width, but it's not necessary.


Hello @EvilAlexei, thanks for complex response, I totally agree with you that those limitations will be a little handy for users, but it's same situation as with dialogs, or card.  We provide such a limits on docs level, because specification doesn't specify them. Let's keep it same with this component. 

Everything else looks great for me. Good job!

JKMarkowski (approved):
LGTM

katekozlowska (approved):
Looks good!

PR #1905, opened 2020-11-22
chore: update theme names

droshev:
## Description
Update the theme names to match the Fiori 3 naming

- [N/A] All values are in `rem`
- [N/A] RTL support
3. Testing
- [N/A] Updated tests
4. Documentation


InnaAtanasova (commented):
1. Maybe these names should be variables since they are used in 3 different places. 
2. Chromatic is failing
3. Please check the UI tests

droshev:
> 1. Maybe these names should be variables since they are used in 3 different places.
> 2. Chromatic is failing
> 3. Please check the UI tests

1. the names are just labels
2. fixed
3. checked

InnaAtanasova (approved):
@droshev I approve as we discussed my comments won't be addressed.

Issue #1896, opened 2020-11-18
Message box missing header icons 


salarenko:
**Describe the bug**
Missing semantic icons in the message box header.

**Expected behavior**
Message box used with one of the semantic types should have an icon next to the header title.

**Screenshots**
How it is:
![image](https://user-images.githubusercontent.com/17496353/99519803-9e232700-2992-11eb-956b-4ad7b010afb4.png)

How it should be:
![image](https://user-images.githubusercontent.com/17496353/99520065-f0fcde80-2992-11eb-8691-963cbea8fda8.png)

PR #1910, opened 2020-11-24
fix: icons for message-box

stefanoScalzo:
## Related Issue
Closes SAP/fundamental-styles#1896

## Description
Fixes the icons so that they are displayed with message box
### Before:
Screen Shot 2020-11-24 at 10 38 11 AM

### After:
Screen Shot 2020-11-24 at 10 37 51 AM

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/message-box.scss
+++ b/src/message-box.scss
@@ -5,6 +5,7 @@
 $block: #{$fd-namespace}-message-box;
 $bar: #{$fd-namespace}-bar;
 $button: #{$fd-namespace}-button;
+$message: #{$fd-namespace}-message;
 
 $fd-message-box-icon-text-spacing: 0.5rem;
 $fd-message-box-title-icon-text-size: 1rem;
@@ -99,8 +100,8 @@ $message-box-states: (
         box-shadow: map_get($state-set, "boxShadow");
       }
 
-      .#{$block}__title {
-        @include fd-icon-base () {
+      .#{$message}__title {
+        @include fd-icon-base ("before") {
           position: relative;
           top: 0.0625rem;
           font-size: $fd-message-box-title-icon-text-size;
@@ -111,18 +112,18 @@ $message-box-states: (
       }
 
       @include fd-rtl() {
-        .#{$block}__title::before {
+        .#{$message}-title::before {
           margin-right: 0;
           margin-left: $fd-message-box-icon-text-spacing;
         }
       }
+    }
+  }
 
-      &.#{$block}--no-icon {
-        .#{$block}__title::before {
-          content: '';
-          margin: 0;
-        }
-      }
+  &--no-icon {
+    .#{$message}__title::before {
+      content: none;
+      margin: 0;
     }
   }
 }
--- a/src/message-box.scss
+++ b/src/message-box.scss
@@ -99,7 +99,7 @@ $message-box-states: (
         box-shadow: map_get($state-set, "boxShadow");
       }
 
-      .#{$block}__title {
+      .#{$block}__icon {
         font-style: normal;
 
         @include fd-icon-base () {
@@ -113,7 +113,7 @@ $message-box-states: (
       }
 
       @include fd-rtl() {
-        .#{$block}__title::before {
+        .#{$block}__icon::before {
           margin-right: 0;
           margin-left: $fd-message-box-icon-text-spacing;
         }
@@ -122,7 +122,7 @@ $message-box-states: (
   }
 
   &--no-icon {
-    .#{$block}__title::before {
+    .#{$block}__icon::before {
       content: none;
       margin: 0;
     }
--- a/src/message-box.scss
+++ b/src/message-box.scss
@@ -100,7 +100,7 @@ $message-box-states: (
         box-shadow: map_get($state-set, "boxShadow");
       }
 
-      .#{$message}__title {
+      .#{$block}__title {
         @include fd-icon-base ("before") {
           position: relative;
           top: 0.0625rem;
--- a/src/message-box.scss
+++ b/src/message-box.scss
@@ -27,6 +27,9 @@ $message-box-states: (
 .#{$block} {
   @include fd-reset();
 
+  position: relative;
+  z-index: 1;
+
   @extend %dialog;
 
   display: none;
--- a/src/message-box.scss
+++ b/src/message-box.scss
@@ -29,14 +29,22 @@ $message-box-states: (
 
   position: relative;
   z-index: 1;
+  background-color: transparent;
 
   @extend %dialog;
 
   display: none;
 
+  &::before {
+    background-color: transparent;
+  }
+
   // ELEMENTS
   &__content {
     @include fd-reset();
+
+    position: relative;
+    z-index: 1;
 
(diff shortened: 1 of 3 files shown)


src/message-box.scss
```diff
@@ -99,8 +100,8 @@ $message-box-states: (
         box-shadow: map_get($state-set, "boxShadow");
       }
 
-      .#{$block}__title {
-        @include fd-icon-base () {
+      .#{$message}__title {
```
InnaAtanasova, line 103: Why are you bringing `fd-message` component here? It should be the way it was before.

InnaAtanasova (changes requested):
I am pretty sure you are referencing the wrong issue in your PR template. You are referencing an issue about a border on the sub-header that shouldn't be there.

src/message-box.scss
InnaAtanasova: Also, we are no longer referencing other components this way. We rely on specificity.

stefanoScalzo: fixed

src/message-box.scss
```diff
@@ -5,6 +5,7 @@
 $block: #{$fd-namespace}-message-box;
 $bar: #{$fd-namespace}-bar;
 $button: #{$fd-namespace}-button;
+$message: #{$fd-namespace}-message;
```
InnaAtanasova, line 8: Do you still need this?

src/message-box.scss
```diff
@@ -111,18 +112,18 @@ $message-box-states: (
       }
 
       @include fd-rtl() {
-        .#{$block}__title::before {
+        .#{$message}-title::before {
```
InnaAtanasova, line 115: I think this should be `#{$block}__title `.

src/message-box.scss
```diff
@@ -111,18 +112,18 @@ $message-box-states: (
...
+  }
 
-      &.#{$block}--no-icon {
-        .#{$block}__title::before {
-          content: '';
-          margin: 0;
-        }
-      }
+  &--no-icon {
+    .#{$message}__title::before {
```
InnaAtanasova, line 124: I think this should be `#{$block}__title `.

stories/message-box/__snapshots__/message-box.stories.storyshot
```diff
@@ -137,7 +137,7 @@ exports[`Storyshots Components/Message Box No icon 1`] = `
             
                     
             <h2
-              class="fd-title fd-title--h5"
+              class="fd-title fd-title--h5 fd-message__title"
```
InnaAtanasova, line 140: I think you haven't updated the storyshot as this should be ` fd-message-box__title`, not `fd-message__title`

InnaAtanasova (changes requested):
Please check again if you are referencing the correct issue in the PR template.

src/message-box.scss
```diff
@@ -100,7 +101,7 @@ $message-box-states: (
       }
 
       .#{$block}__title {
-        @include fd-icon-base () {
+        @include fd-icon-base ("before") {
```
JKMarkowski, line 104: It can be removed, `before` is set by default

JKMarkowski (commented):
Let's apply Inna's comments, I added something else. Can you also run `npm run test:update` to refresh storyshots? Rest LGTM

src/message-box.scss
salarenko, line 104: Hey, some time ago we agreed no to create icons this way. Icons should be moved to separate `<i>`

salarenko (changes requested):

InnaAtanasova:
Again, this PR is not fixing the issue linked in the PR template.

stories/message-box/message-box.stories.js
```diff
@@ -248,11 +254,12 @@ Information | \`fd-message-box--information\` | Information messages provide inf
 
 export const noIcon = () =>
     `
-<div class="fd-message-box-docs-static fd-message-box fd-message-box--no-icon fd-message-box--success fd-message-box--active">
+<div class="fd-message-box-docs-static fd-message-box fd-message-box--success fd-message-box--no-icon fd-message-box--active">
     <section class="fd-message-box__content">
         <header class="fd-bar fd-bar--header fd-message-box__header">
             <div class="fd-bar__left">
                 <div class="fd-bar__element">
+                    <i class="fd-message-box__title"></i>
```
JKMarkowski, line 262: I don't think that `__title` is proper name for icon there, can you change it to basic `__icon`?

stefanoScalzo: fixed

src/message-box.scss
stefanoScalzo: fixed

mikerodonnell89 (changes requested):
Why can't I highlight text or click the buttons in these examples?  And the button text isn't centered

Screen Shot 2020-11-30 at 6 42 10 PM

JKMarkowski:
@stefanoScalzo could you address @mikerodonnell89 Issue there a well?
examples wrappers have `position: static;`, which are supposed to be `relative`.

stefanoScalzo:
> examples wrappers have `position: static;`, which are supposed to be `relative`.

Can we tackle this in another PR? @JKMarkowski @mikerodonnell89

stefanoScalzo:
> > examples wrappers have `position: static;`, which are supposed to be `relative`.
Screen Shot 2020-12-01 at 2 56 10 PM
We use static or else when you scroll with storybook it produces this effect

@JKMarkowski

InnaAtanasova:
Please rebase with latest master and check if the issues mentioned above are still relevant. Thanks

salarenko (approved):

stefanoScalzo:
> Please rebase with latest master and check if the issues mentioned above are still relevant. Thanks

I've rebased and fix comments. Master has an issue with failing grid list test

mikerodonnell89 (approved):

InnaAtanasova:
Closing in favour of https://github.com/SAP/fundamental-styles/pull/1945

stories/message-box/message-box.stories.js
```diff
@@ -51,12 +52,13 @@ Note: Include two action buttons in the message box when the user's decision is
 
 const messageBoxHeight = 200;
 
-export const structure = () => `<div class="fd-message-box-docs-static fd-message-box fd-message-box--active">
```
JKMarkowski, line 54: Don't remove `fd-message-box-docs-static`, It's needed there. I would add `z-index: 1` to this class, not to the lib

src/message-box.scss
```diff
@@ -27,14 +27,25 @@ $message-box-states: (
 .#{$block} {
   @include fd-reset();
 
+  position: relative;
+  z-index: 1;
```
JKMarkowski, line 31: I don't think if it's good idea to put it there, can you add it to `fd-message-box-docs-static` this class?

JKMarkowski (changes requested):
Examples LGTM, but check my comments about the code

Issue #1893, opened 2020-11-17
Message and Status Icons


InnaAtanasova:
Fundamental-styles is missing Semantic Notification/ Message Icons which is used in Message Popover component

droshev:
ping @droshev or @InnaAtanasova or @mikerodonnell89 for the visual specs

droshev:
we will create new issues once we plan this component

PR #1912, opened 2020-11-24
feat: icon message component

stefanoScalzo:
## Related Issue
Closes SAP/fundamental-styles#1893

## Description
Add the icon message component
the user has the option to just use an icon or just use text

### After:

Screen Shot 2020-11-24 at 4 33 56 PM
Screen Shot 2020-11-24 at 4 33 50 PM
Screen Shot 2020-11-24 at 4 33 45 PM
Screen Shot 2020-11-24 at 4 33 40 PM
Screen Shot 2020-11-24 at 4 33 35 PM

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation


stories/icon-message/icon-message.stories.js
```diff
@@ -0,0 +1,80 @@
...
+* You want to show a title using a semantic color
+            `,
+        components: ['icon-message']
+    }
+};
+
+export const success = () => `
+<div>
+<div class="fd-icon-message-success"></div>
+<div class="fd-icon-message-success fd-icon-message-success--no-icon">Success</div>
```
JKMarkowski, line 20: Could you change it to
```suggestion
<i arial-label="success" class="fd-icon-message-success"></i>

<br/><br/>

<span class="fd-icon-message-success fd-icon-message-success--no-icon">Success</span>
```

JKMarkowski (changes requested):
In general - Great Job with `@each` usage, I'm just wondering if it's worth to have `fd-icon-message` component, when this component doesn't have `icon`. I checked the naming
image
Maybe consider changing is somehow to semantic-notification? Or message status?

stories/icon-message/icon-message.stories.js
stefanoScalzo: i fixed it with different naming convention but same items

stories/semantic-color/semantic-color.stories.js
```diff
@@ -0,0 +1,87 @@
...
+* You want to show a title using a semantic color
+            `,
+        components: ['semantic-color']
+    }
+};
+
+export const information = () => `
+<div>
+    <i arial-label="information" class="fd-semantic-color__icon"></i>
+    <br>
```
jbadan, line 21: Can we avoid using `<br />` tags for styling?

salarenko (changes requested):
I don't think this should be structured like this:
image

Doesn't it make more sense to go with:
```
<div class="fd-semantic-color">
    <i class="fd-semantic-color__icon sap-icon--*"></i>
    <span class="fd-semantic-color__text"></span>
</div>
```

Is this really an "Semantic Color" component? Because that's how its named in the Storybook, and here in the PR its "Icon Message".

stefanoScalzo:
> I don't think this should be structured like this:
> image
> 
> Doesn't it make more sense to go with:
> 
> ```
> <div class="fd-semantic-color">
>     <i class="fd-semantic-color__icon sap-icon--*"></i>
>     <span class="fd-semantic-color__text"></span>
> </div>
> ```
> 
> Is this really an "Semantic Color" component? Because that's how its named in the Storybook, and here in the PR its "Icon Message".

Fixed the naming and the ay the documentation is displayed for more clarity

stories/semantic-color/semantic-color.stories.js
stefanoScalzo: Fixed so i dont need line breaks anymore

InnaAtanasova (changes requested):
This should not be a stand-alone component. It's part of the icons. Even in the Wiki specs it's shown as part of the icons.

stefanoScalzo:
> This should not be a stand-alone component. It's part of the icons. Even in the Wiki specs it's shown as part of the icons.

I put it inside the icon component like the wiki

mikerodonnell89 (approved):

JKMarkowski (approved):
Looks good for me now

stories/icon/icon.stories.js
```diff
@@ -56,3 +56,36 @@ availableIcons.parameters = {
     },
     skipRTLSnapshot: true
 };
+
+
+export const iconSemantics = () => `
+<div>
+    <i arial-label="information" class="fd-icon-message__icon"></i>
```
JKMarkowski, line 63: I think it doesn't follow BEM rules

JKMarkowski (changes requested):
There one things that should be changed

InnaAtanasova:
This component is expecting design changes so I close it for now. Thanks for the work @stefanoScalzo

Issue #1800, opened 2020-10-21
Alignment of Title and Title content in Dynamic Page


kavya-b:
**Describe the bug**
The alignment for Title and Title content needs to be changed per designer's suggestion.
As per current spec, title has a top padding while title content does not, which makes their alignment look off.
![image](https://user-images.githubusercontent.com/53509521/96679224-d8e47000-1390-11eb-9a97-8825fac08702.png)
On communicating this iissue with designers, they suggested the following:
![image](https://user-images.githubusercontent.com/53509521/96679389-3082db80-1391-11eb-95f7-8c58726d20f3.png)

> Is this issue related to a specific component?

Dynamic Page
> What version of the Fundamental Library Styles are you using?

latest rc
> What front-end framework are you implementing in, e.g., Angular, React, etc.?

Angular
**To Reproduce**
1. Go to https://fundamental-styles.netlify.app/?path=/docs/layouts-dynamic-page--desktop
2. Observe the title and title content('some random things')
3. They are not aligned 

**Expected behavior**
As per designer's suggestion, for text title content, the texts should be bottom aligned.

kavya-b:
There are a few things I noted while trying to achieve bottom alignment between title and title content when using `display: table` on the parent and `display:table-cell` on the children. 
1. Aligning to bottom with `vertical-align: bottom` doesn't really do the trick because the title has letters like 'g' and 'p', something not present in the proposed solution of Object Page alignment by designers. 
![image](https://user-images.githubusercontent.com/53509521/96710211-5a4ff880-13b9-11eb-84b4-e6c4123b4733.png)
`vertical-align: baseline` works better:
![image](https://user-images.githubusercontent.com/53509521/96710294-7e133e80-13b9-11eb-8578-d0ecd8921347.png)

2. The proposed solution takes the example of Object Page but only shows the alignment between the two containers on a single-line. Aligning to bottom on a multi-line of the title content container will again make this look off because overflow rules are not specified for title content:
![image](https://user-images.githubusercontent.com/53509521/96712888-6ccc3100-13bd-11eb-9ca9-8f3db14ae004.png)

we will need to provide overflow rules for the title content as well, so that longer texts can be contained in a single line.

Also, `display:table` is behaving a little strangely with respect to wrapping of the content in different browsers. I am looking into this presently.

kavya-b:
[...]

PR #1924, opened 2020-12-01
fix: align title and title content in Dynamic Page

kavya-b:
## Related Issue
Closes SAP/fundamental-styles#1800

## Description
This PR does the following:
- aligns title and title content in Dynamic Page 
- modifies z-index usage to fix look in mobile popover case
- keeps title content ellipsible and restricts to single line to maintain alignment
- uses variable for common value

Breaking change:
- removed modifier class `fd-dynamic-page__collapsible-header-container--collapsed`

### Before:
alignment: image
popover getting cut: image
title content multi-line: image

### After:
alignment: image
popover getting cut fixed: image
title-content single-line: image

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/dynamic-page.scss
+++ b/src/dynamic-page.scss
@@ -140,6 +140,7 @@ $block: #{$fd-namespace}-dynamic-page;
     }
 
     max-width: 100%;
+    width: 100%;
 
     &--sm {
       align-items: flex-start;
@@ -157,9 +158,8 @@ $block: #{$fd-namespace}-dynamic-page;
   &__breadcrumb-title-container {
     @include fd-reset();
 
-    display: table;
-    width: 100%;
-    table-layout: fixed;
+    flex: 1;
+    min-width: 0;
   }
 
   &__breadcrumb-container {
@@ -176,7 +176,6 @@ $block: #{$fd-namespace}-dynamic-page;
     color: var(--sapGroup_TitleTextColor);
     font-size: var(--sapFontHeader3Size);
     padding: 0.3125rem 0 0 0;
-    display: table-cell;
 
     @include fd-ellipsis();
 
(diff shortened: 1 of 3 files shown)


katekozlowska (changes requested):
There should be truncation on longer text, exampe from IE11:

image

kavya-b:
> There should be truncation on longer text, exampe from IE11:
> 
> image

unfortunately this is happening due to using `display: table-cell` . It expects a max-width value(but doesn't like % values) and I don't think max-width in rem should be provided when the content could be anything, and spec doesn't specify how much space title and title content should each take. 
something similar is happening with Wizard as well, @InnaAtanasova can you help a bit with this? Is this a known issue in IE where long text truncation in an element that is a `table-cell` has problems?

I am currently revisiting the earlier `display: flex` logic again; I remember it didn't align  well in some cases which is why I switched to `table-cell` in the first place 😓

kavya-b:
@katekozlowska fixed in IE with `display:flex` usage. Please check.

salarenko (approved):

src/dynamic-page.scss
```diff
@@ -272,6 +282,7 @@ $block: #{$fd-namespace}-dynamic-page;
 
   .#{$fd-namespace}-toolbar.#{$block}__title-content-toolbar {
```
InnaAtanasova, line 283: Can you remove the `.#{$fd-namespace}-toolbar` ? This is not making the component self-contained. If the rules in `.#{$block}__title-content-toolbar` are not taking precedence over the ones coming from ` .#{$fd-namespace}-toolbar` you should apply more specificity.

kavya-b: done, also removed mixed scope from breadcrumb and tabs.

katekozlowska (approved):

InnaAtanasova (approved):

PR #1929, opened 2020-12-01
fix: add fix for missing space between Menu Button text and expand arrow

InnaAtanasova:
## Description
this PR adds `!important` to the rule that sets spacing between the elements inside a Button. Without this rule, the behaviour depends on the import of the components. In Fundamental-ngx, the icons styling was overwriting the margin-left specified in the Button and was causing the issue where the expand icon was rendered right next to the text. 

### Before:
Screen Shot 2020-12-01 at 5 08 04 PM

### After:
Screen Shot 2020-12-01 at 5 08 26 PM

- [x] RTL support
- [NA] only one top level `fd-*` class is used in the file
- [NA ] Checked if current components can be reused, instead of having new code.
3. Testing
-  x] tested Storybook examples with "CSS Resources" `normalize` option 
- [NA] Updated tests
4. Documentation

### DIFF
--- a/src/button.scss
+++ b/src/button.scss
@@ -232,11 +232,11 @@ $block: #{$fd-namespace}-button;
 }
 
 @mixin iconTextMargin() {
-  margin-left: $fd-button-icon-margin;
+  margin-left: $fd-button-icon-margin !important;
 
   @include fd-rtl() {
-    margin-left: 0;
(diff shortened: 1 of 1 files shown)


JKMarkowski (approved):
I don't like importants, but without mixing namespaces there is nothing we can do with it

PR #1932, opened 2020-12-02
fix: remove token max-width

mikerodonnell89:
Fixes several issues that bubbled to the Angular implementation of the token/tokenizer. The issue with growing tokens is resolved by replacing `flex: 1 0 auto` with `flex-shrink: 0`, setting the token max-width does not fix the problem stated in the previous defect hunting (screenshot below).  If we want to limit the width of the token, it should be done so by modifying the fd-token__text class to avoid causing problems with the Angular implementations.

Screen Shot 2020-12-02 at 11 04 11 AM

### DIFF
--- a/src/tokenizer.scss
+++ b/src/tokenizer.scss
@@ -29,7 +29,6 @@ $block: #{$fd-namespace}-tokenizer;
 
   .fd-token {
     margin-right: $fd-tokenizer-spacing;
-    max-width: calc(100% - #{$fd-tokenizer-input-width} - #{$fd-tokenizer-spacing});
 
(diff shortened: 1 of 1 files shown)


InnaAtanasova (approved):

salarenko (commented):
Removing this changes the proper behavior of the token in the tokenizer. Token should take up to all available width in the tokenizer before wrapping the text with `...`. To be honest I don't see another way of doing this, but if this solution is problematic we can reverse it.

InnaAtanasova:
Closing this for now as we discussed during the scrum. Will try to find a better solution before we apply this one.

Issue #1907, opened 2020-11-23
Defect Hunting and Release - Sprint 51


InnaAtanasova:
List all bugs found during the Defect Hunting as comments below using the following format:
- [x] short description, browser 
attached image or gif

salarenko:
- [ ] Calendar `Date | Time` button should be a segmented button
![image](https://user-images.githubusercontent.com/17496353/100743917-426a8c00-33dd-11eb-9a86-31d599bf6c73.png)

- [ ] Calendar months and years buttons text is not centered
![image](https://user-images.githubusercontent.com/17496353/100744047-7a71cf00-33dd-11eb-8cb8-4cd40ad1de42.png)

- [ ] Misaligned icons in shellbar buttons
![image](https://user-images.githubusercontent.com/17496353/100747179-0128ab00-33e2-11eb-8251-273de80ca920.png)

- [ ] Vertically and horizontally misaligned select and date-picker controls (IE11)
![image](https://user-images.githubusercontent.com/17496353/100747797-cd01ba00-33e2-11eb-934c-f5e88012c543.png)

- [ ] Missing padding in two column combo-box items list
![image](https://user-images.githubusercontent.com/17496353/100748313-78ab0a00-33e3-11eb-9e75-8a29ed19b964.png)

JKMarkowski:
- [ ] calendar - button - all browsers
![image](https://user-images.githubusercontent.com/26483208/100749256-b78d8f80-33e4-11eb-89ff-0a0a77a4b7f2.png)

JKMarkowski:
- [ ] IE11 - tiles- one word is cut - I see initial width value is used. IE11 doesn't like initial.

![image](https://user-images.githubusercontent.com/26483208/100749624-35ea3180-33e5-11eb-88bd-b948320155ca.png)

mikerodonnell89:
Shellbar button hover state misaligned

![Screen Shot 2020-12-01 at 11 04 16 AM](https://user-images.githubusercontent.com/2471874/100779001-f57abb80-33c4-11eb-8a14-1e91fc13ec48.png)

![Screen Shot 2020-12-01 at 11 04 23 AM](https://user-images.githubusercontent.com/2471874/100779014-f9a6d900-33c4-11eb-9fe6-e1e5d6f659ad.png)

katekozlowska:
- [ ] Multi ComboBox - different styling of checkboxes

[...]

PR #1942, opened 2020-12-04
fix: form controls in Table

salarenko:
## Related Issue
Part of: #1907

## Description
This PR adds `full-width` modifier to popover to easily extend popover width when the popover and its content needs to fill the container.

The vertical misalignment seems to be "caused" by `display: inline-block` on `fd-popover` level. I don't know if there is a way to fix it.

## Before
image

## After
image

### DIFF
--- a/src/popover.scss
+++ b/src/popover.scss
@@ -224,6 +224,10 @@ $block: #{$fd-namespace}-popover;
     }
   }
 
+  &--full-width {
+    width: 100%;
+  }
+
   &__popper {
     @include fd-reset();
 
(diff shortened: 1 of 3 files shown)


InnaAtanasova:
This PR is introducing Breaking change that needs to be documented in wiki and the PR template

mikerodonnell89 (approved):
Looks good, @InnaAtanasova this doesn't really "break" anything just adds an improvement, do we still need to add changes like this to the wiki?

JKMarkowski (approved):
LGTM

PR #1945, opened 2020-12-07
fix: add missing icons to semantic Message Box

InnaAtanasova:
## Related Issue
Closes SAP/fundamental-styles#1896

## Description
1. Add semantic icons to the Message Box title
2. Remove `.fd-message-box--no-icon` modifier class as by default the Message Box title does not have an icon.
3. In some Message Box examples it was not possible to click on the buttons or highlight the text. This issue was fixed by changing the position from static to relative (documentation class change)
4. Correct the documentation where instead of `.fd-message-box__xxx`, `.fd-message__xxx` was used.

BREAKING CHANGE: 
`.fd-message-box--no-icon` modifier class is removed
Semantic icons are added separately: `<i class="sap-icon--message-error"></i>` 

### Before:
Screen Shot 2020-12-07 at 3 48 33 PM

### After:
Screen Shot 2020-12-07 at 3 47 18 PM

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/.storybook/custom/custom.scss
+++ b/.storybook/custom/custom.scss
@@ -288,7 +288,7 @@ code.docs-code-grey {
   .fd-dialog,
   &.fd-message-box,
   .fd-message-box {
-    position: static;
+    position: relative;
     background-color: transparent;
 
     &::before {
@@ -300,7 +300,7 @@ code.docs-code-grey {
   .fd-dialog__content,
   &.fd-message-box__content,
   .fd-message-box__content {
-    position: static;
+    position: relative;
     transform: translate(0%, 0%);
     width: 30rem;
     margin: auto;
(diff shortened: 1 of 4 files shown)


mikerodonnell89:
@InnaAtanasova duplicate of https://github.com/SAP/fundamental-styles/pull/1910/files ?

mikerodonnell89 (approved):

JKMarkowski:
It looks good for me, but what about icons in cozy(mobile) mode?
image

InnaAtanasova:
> It looks good for me, but what about icons in message-box in cozy(mobile) mode?
> image

The icons for Message box are optional, could be with or without them. Before we had a modifier class to remove them, but since we are now using `<i class="sap-icon--message-error"></i>` to add an icon, by default the title has no icon. One example shows Message boxes with icons (Semantic types) and the other example shows Message boxes without icons (Mobile)

JKMarkowski (approved):

PR #1946, opened 2020-12-07
docs: updated table documentation

jessmetivier:
## Description
{{ Updated **Table** component docs. }}

NS = need support
The Responsive Table example needs to be reviewed to better understand its use case, and if it's really different from the default table examples.


jessmetivier:
The **Responsive Table** example needs to be reviewed to better understand its use case, and if it's really different from the default table examples.

droshev (approved):

Issue #1869, opened 2020-11-09
Revisit Switch and review class composition


salarenko:
**Describe the bug**
Switch component class structure is shaped in following structure:
```
.fd-switch__label
  .fd-switch__text
  .fd-switch
        .fd-switch__input
        .fd-switch__wrapper
            .fd-switch__track
                .fd-switch__handle
```
which does not really applies to BEM methodology and creates confusion when maintaining the code.

**Expected behavior**
Structure created according to BEM methodoligy.

PR #1949, opened 2020-12-08
fix: improve class structure for Switch component

salarenko:
## Related Issue
Closes: #1869

## Description
This PR:
- improves switch component class structure.
- removes unnecessary SCSS variables

BREAKING CHANGE:
- On `.fd-switch__label` level should be used `.fd-switch` class
- On `.fd-switch` level should be used `.fd-switch__control` class
- On `.fd-switch` level should be used `.fd-switch__control` class
- Instead of  class should be used `.fd-switch__slider`
- `.fd-switch__label` and `.fd-switch__wrapper` have been removed

### Before:
```
.fd-switch__label
    .fd-switch__text
    .fd-switch
        .fd-switch__input
        .fd-switch__wrapper
            .fd-switch__track
                .fd-switch__icon
                .fd-switch__handle
```

### After:
```
.fd-switch
    .fd-switch__text
    .fd-switch__control
        .fd-switch__input
        .fd-switch__slider
            .fd-switch__track
                .fd-switch__icon
                .fd-switch__handle
```

3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/switch.scss
+++ b/src/switch.scss
@@ -10,50 +10,27 @@ $block: #{$fd-namespace}-switch;
 
 .#{$block} {
 
+  $fd-switch-off-opacity: 0;
+  $fd-switch-transition: all 0.1s;
   $fd-switch-active-border-color: var(--sapButton_Selected_BorderColor);
-  $fd-switch-active-background-color: var(--sapButton_Selected_Background);
   $fd-switch-inactive-border-color: var(--sapContent_ForegroundBorderColor);
   $fd-switch-inactive-handle-background: var(--sapButton_Background);
 
   $fd-switch-hover-border-color: var(--sapButton_Hover_BorderColor);
   $fd-switch-hover-selected-border-color: var(--sapButton_Selected_Hover_BorderColor);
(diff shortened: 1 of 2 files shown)


stefanoScalzo:
Screen Shot 2020-12-08 at 4 18 53 PM
Not quite sure whats going on here but its weird

salarenko:
@stefanoScalzo I'm also not quite sure what is this comment about, its weird 😛

If this is about doubled "Off" examples, yes my mistake.

katekozlowska (approved):
Looks much more clear now, good job!
My minor suggestion is to keep BEM order when adding more classes, so instead of 
`class="fd-switch__icon--off fd-switch__icon sap-icon--decline"` we should use `class="fd-switch__icon fd-switch__icon--off  sap-icon--decline"`

salarenko:
@katekozlowska Good suggestion. I've reordered the classes.

JKMarkowski (approved):

InnaAtanasova (approved):
Good job on this PR

PR #1951, opened 2020-12-09
fix: change min width when a user image will disappear Feed Input

genereg1:
## Related Issue
Closes
https://github.com/SAP/fundamental-ngx/issues/4020
## Description
The issue raised in the ngx but actually related to styles

### Before:
101314835-64518c00-387f-11eb-83ab-c34363bb3cfa (1)

### After:

Kapture 2020-12-09 at 15 10 32

- [x] All values are in `rem`
- [na] RTL support
3. Testing
- [na] Updated tests
4. Documentation

### DIFF
--- a/src/feed-input.scss
+++ b/src/feed-input.scss
@@ -69,7 +69,7 @@ $block: #{$fd-namespace}-feed-input;
     }
   }
 
-  @media (min-width: 400px) {
+  @media (min-width: 25rem) {
     margin-top: 1rem;
     margin-bottom: 1rem;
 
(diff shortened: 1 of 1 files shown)


InnaAtanasova (commented):
The value should be in `rem`. 
Screen Shot 2020-12-09 at 8 23 58 AM

InnaAtanasova:
For the checkboxes in the PR template, could you either delete them or better write [na] instead of [x]?
`Meargable` will fail with "There are incomplete TODO task(s) unchecked" if you don't check all checkboxes.

InnaAtanasova (approved):

Issue #1343, opened 2020-08-11
Double focus outline for checkbox and radio button when used with empty labels


InnaAtanasova:
When no label is provided (empty) the checkbox and the radio button get double focus outline

<img width="130" alt="Screen Shot 2020-08-11 at 12 27 42 PM" src="https://user-images.githubusercontent.com/39598672/89923157-2cd31b80-dbce-11ea-8249-d20df9020c8c.png">
<img width="133" alt="Screen Shot 2020-08-11 at 12 27 35 PM" src="https://user-images.githubusercontent.com/39598672/89923160-2d6bb200-dbce-11ea-9385-a8712ee6fd8d.png">

PR #1954, opened 2020-12-09
fix: checkbox focus with margin

stefanoScalzo:
## Related Issue
Closes SAP/fundamental-styles#1343

## Description
Changes the margin to the text element of the checkbox. 
This will remove the unwanted spacing for focus when there is no label

### Before:
Screen Shot 2020-12-09 at 1 21 06 PM

### After:
Screen Shot 2020-12-09 at 1 24 50 PM

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/checkbox.scss
+++ b/src/checkbox.scss
@@ -57,18 +57,9 @@ $block: #{$fd-namespace}-checkbox;
       align-items: center;
       font-family: "SAP-icons";
       background-color: var(--sapField_Background);
-      margin-right: $fd-checkbox-margin;
-      margin-left: 0;
       color: var(--fdCheckbox_Selected_Color);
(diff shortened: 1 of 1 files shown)


InnaAtanasova:
The linked issue was about having double focus line, not having margin on empty level. I think the double focus line was fixed. Anyways, the fix you added was still relevant.

InnaAtanasova (approved):

stefanoScalzo:
> The linked issue was about having double focus line, not having margin on empty level. I think the double focus line was fixed. Anyways, the fix you added was still relevant.

I assumed that the fix that fixed the issue created the issue, that's why I referenced it :)

JKMarkowski (approved):
Good Job!

Issue #1892, opened 2020-11-17
outline color different on breadcrumb hover


stefanoScalzo:
**Describe the bug**
![Screen Shot 2020-11-17 at 11 39 09 AM](https://user-images.githubusercontent.com/50607147/99418945-8d5aae80-28c9-11eb-975b-7d7bb07d1841.png)
The color of the outline is different on the down icon and the 3 dots outline
> Is this issue related to a specific component?
Breadcrumb
> What version of the Fundamental Library Styles are you using?
Latest

PR #1969, opened 2020-12-14
fix: outline color of breakcrumb icon overflow

stefanoScalzo:
## Related Issue
Closes SAP/fundamental-styles#1892

## Description
Make all states have the same color for both items
### Before:
image

### After:
Screen Shot 2020-12-14 at 1 05 46 PM

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/breadcrumb.scss
+++ b/src/breadcrumb.scss
@@ -54,6 +54,30 @@ $fd-breadcrumb-border-top-radius: 0.125rem !default;
 
     display: inline-block;
     transition: all $fd-animation-speed ease-in;
+
+    &:focus {
+      .#{$block}__dropdown-icon {
+        color: var(--sapLink_Active_Color, #0a6ed1);
+      }
+    }
+
+    &:active {
+      .#{$block}__dropdown-icon {
+        color: var(--sapLink_Active_Color, #0a6ed1);
+      }
+
+      &:hover {
+        .#{$block}__dropdown-icon {
+          color: var(--sapLink_Active_Color, #0a6ed1);
+        }
+      }
+    }
+
+    &:hover {
+      .#{$block}__dropdown-icon {
+        color: var(--sapLink_Hover_Color, #0854a0);
+      }
+    }
   }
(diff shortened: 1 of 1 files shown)


JKMarkowski (approved):

src/breadcrumb.scss
```diff
@@ -54,6 +54,30 @@ $fd-breadcrumb-border-top-radius: 0.125rem !default;
...
+      &:hover {
+        .#{$block}__dropdown-icon {
+          color: var(--sapLink_Active_Color, #0a6ed1);
+        }
+      }
+    }
+
+    &:hover {
+      .#{$block}__dropdown-icon {
+        color: var(--sapLink_Hover_Color, #0854a0);
```
InnaAtanasova, line 78: Why are you adding the colors this way? 
Should be just color: `var(--sapLink_Hover_Color);` or `color: var(--sapLink_Active_Color)

InnaAtanasova: and since `var(--sapLink_Active_Color)` is repeated 3 times maybe put it in a variable.

InnaAtanasova (changes requested):

src/breadcrumb.scss
stefanoScalzo: I noticed we have fallbacks in some components but ill change it :)

InnaAtanasova (approved):

src/breadcrumb.scss
InnaAtanasova: That's not the proper way to add a fallback in general. You will need:
```
color:  #0854a0;
color: var(--sapLink_Hover_Color, #0854a0);
```

InnaAtanasova: Also, I searched the project and couldn't find such case :)
There's one usage in `custom.scss` file which is for the documentation. I think you are talking about fundamental-ngx, but this is another case.

Issue #1803, opened 2020-10-21
Grid nogap vetical and nogap horizontal


stefanoScalzo:
**Is your feature request related to a problem? Please describe.**
I should be able to have no gap on grids horizontally, vertically or both.

> Is this issue related to a specific component?
Layout Grids
> What browser are you working in?
Chrome
> What version of Fundamental Library Styles are you using?
Lastest
> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

> What front-end framework are you implementing in, e.g., Angular, React, etc.?

**Describe the solution you'd like**
Add a modifier to the container for horizontal and vertical

PR #1970, opened 2020-12-14
feat: more control over no gap for grid layout

stefanoScalzo:
BREAKING CHANGES: `.fd-layout-grid--no-gap` removed
`.fd-layout-grid--no-vertical-gap` added
`.fd-layout-grid--no-horizontal-gap` added
## Related Issue
Closes SAP/fundamental-styles#1803

## Description
This PR adds more control to having no gap for layout grids.
The user can now choose between having no vertical gap which removes space in between rows, no horizontal gap which removes spacing between rows or can have no gap at all.
### Before:
After consists of before and more

### After:
Screen Shot 2020-12-14 at 2 24 43 PM

Screen Shot 2020-12-14 at 2 24 27 PM
Screen Shot 2020-12-14 at 2 24 57 PM

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/fundamental-styles.scss
+++ b/src/fundamental-styles.scss
@@ -57,6 +57,7 @@
 @import "./text";
 @import "./textarea";
 @import "./tile";
+@import "./title";
 @import "./time";
 @import "./token";
 @import "./tokenizer";
--- a/src/fundamental-styles.scss
+++ b/src/fundamental-styles.scss
@@ -57,7 +57,6 @@
 @import "./text";
 @import "./textarea";
 @import "./tile";
-@import "./title";
 @import "./time";
 @import "./token";
 @import "./tokenizer";
--- a/src/layout-grid.scss
+++ b/src/layout-grid.scss
@@ -144,11 +144,18 @@ $cols: 12;
       margin-top: 0;
     }
 
-    &--no-gap {
+    &--no-horizontal-gap {
+      .fd-col {
+        padding-left: 0;
+        padding-right: 0;
+      }
+    }
+
+    &--no-vertical-gap {
       margin: 0;
 
       .fd-col {
-        padding: 0;
+        padding-top: 0;
       }
     }
   }
--- a/stories/layout-grid/layout-grid.stories.js
+++ b/stories/layout-grid/layout-grid.stories.js
@@ -290,7 +290,93 @@ offset.parameters = {
     }
 };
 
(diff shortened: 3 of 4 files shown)


JKMarkowski (approved):
LGTM

src/fundamental-styles.scss
```diff
@@ -57,6 +57,7 @@
 @import "./text";
 @import "./textarea";
 @import "./tile";
+@import "./title";
```
InnaAtanasova, line 60: Why is this added?

InnaAtanasova (changes requested):

src/layout-grid.scss
```diff
@@ -144,11 +144,18 @@ $cols: 12;
       margin-top: 0;
     }
 
-    &--no-gap {
+    &--no-horizontal-gap {
```
InnaAtanasova, line 147: maybe `.row--no-gap` is more suitable (similar to .row-no-gutters in Bootstrap)

src/layout-grid.scss
```diff
@@ -144,11 +144,18 @@ $cols: 12;
...
+      .fd-col {
+        padding-left: 0;
+        padding-right: 0;
+      }
+    }
+
+    &--no-vertical-gap {
```
InnaAtanasova, line 154: `.col--no-gap` ?

stories/layout-grid/layout-grid.stories.js
```diff
@@ -290,7 +290,93 @@ offset.parameters = {
     }
 };
 
-export const noGap = () => `<div class="fd-container fd-container--no-gap">
+export const noHorizontalGap = () => `<div class="fd-container fd-container--no-horizontal-gap">
```
InnaAtanasova, line 293: Can't we move it to the `fd-row` level?

src/fundamental-styles.scss
stefanoScalzo: removed

src/layout-grid.scss
stefanoScalzo: As discussed, we can't due to negative margines

stories/layout-grid/layout-grid.stories.js
stefanoScalzo: As discussed, we can't due to negative margines

InnaAtanasova (approved):

Issue #1891, opened 2020-11-17
fix: focus with select


stefanoScalzo:
**Describe the bug**
The focus on the select

> Is this issue related to a specific component?
Select
> What version of the Fundamental Library Styles are you using?
Latest

**Screenshots**
![Screen Shot 2020-11-17 at 11 34 32 AM](https://user-images.githubusercontent.com/50607147/99418724-4f5d8a80-28c9-11eb-8860-0eaa09ae6454.png)

PR #1973, opened 2020-12-15
fix: focus of select

stefanoScalzo:
BREAKING CHANGES: fd-select__text removed
## Related Issue
Closes SAP/fundamental-styles#1891

## Description
Adds a before pseudo element to show focus so it can show on the whole element
Changes the padding to the actual text or the before element won't span the whole width properly

### Before:
Screen Shot 2020-12-15 at 9 27 26 AM

### After:
Screen Shot 2020-12-15 at 9 09 27 AM

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/select.scss
+++ b/src/select.scss
@@ -45,7 +45,7 @@ $block: #{$fd-namespace}-select;
 
     @include fd-form-states();
 
-    &:focus {
+    @include fd-focus() {
       outline: 0;
 
       &::before {
@@ -105,15 +105,10 @@ $block: #{$fd-namespace}-select;
     }
   }
 
-  &__text {
-    @include fd-reset();
-
-    font-size: var(--sapFontSize);
-  }
-
   &__text-content {
     @include fd-reset();
 
+    font-size: var(--sapFontSize);
     padding-left: $fd-select-padding-x;
     min-width: 5rem;
 
@@ -127,7 +122,7 @@ $block: #{$fd-namespace}-select;
     .#{$block}__control {
       height: $fd-form-input-height--compact;
 
-      .#{$block}.#{$block}__text-content {
+      .#{$block}__text-content {
         padding-left: $fd-select-padding-compact-x;
       }
 
@@ -135,14 +130,14 @@ $block: #{$fd-namespace}-select;
         text-align: right;
         padding-left: 0;
 
-        .#{$block}.#{$block}__text-content {
+        .#{$block}__text-content {
           padding-left: 0;
           padding-right: $fd-select-padding-compact-x;
         }
       }
 
       &.is-readonly {
-        .#{$block}.#{$block}__text-content {
+        .#{$block}__text-content {
           padding-right: $fd-select-padding-compact-x;
           padding-left: $fd-select-padding-compact-x;
         }
--- a/src/select.scss
+++ b/src/select.scss
@@ -17,6 +17,7 @@ $block: #{$fd-namespace}-select;
   text-shadow: var(--fdSelect_Text_Shadow);
 
   &__control {
+    @include fd-reset();
     @include fd-form-base();
 
     height: $fd-form-input-height;
@@ -26,7 +27,7 @@ $block: #{$fd-namespace}-select;
     align-items: center;
     justify-content: space-between;
     padding-right: 0;
-    padding-left: $fd-select-padding-x;
+    box-sizing: content-box;
 
     .#{$block}__button {
       align-items: center;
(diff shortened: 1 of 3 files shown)


src/select.scss
```diff
@@ -81,34 +96,56 @@ $block: #{$fd-namespace}-select;
...
+        padding-right: $fd-select-padding-x;
+        padding-left: $fd-select-padding-x;
+      }
+
       cursor: text;
     }
   }
 
   &__text {
+    @include fd-reset();
```
InnaAtanasova, line 109: I don't see where we use this class

src/select.scss
```diff
@@ -81,34 +96,56 @@ $block: #{$fd-namespace}-select;
...
+      .#{$block}.#{$block}__text-content {
+        padding-left: $fd-select-padding-compact-x;
+      }
 
       @include fd-rtl() {
         text-align: right;
-        padding-right: $fd-select-padding-compact-x;
         padding-left: 0;
+
+        .#{$block}.#{$block}__text-content {
```
InnaAtanasova, line 138: I didn't find a usage of `fd-select fd-select__text-content`

src/select.scss
```diff
@@ -81,34 +96,56 @@ $block: #{$fd-namespace}-select;
...
+          padding-left: 0;
+          padding-right: $fd-select-padding-compact-x;
+        }
       }
 
       &.is-readonly {
-        padding-right: $fd-select-padding-compact-x;
-        padding-left: $fd-select-padding-compact-x;
+        .#{$block}.#{$block}__text-content {
```
InnaAtanasova, line 145: Are you sure there's such usage?

src/select.scss
```diff
@@ -42,12 +43,26 @@ $block: #{$fd-namespace}-select;
       }
     }
 
-    @include fd-fiori-focus();
     @include fd-form-states();
 
+    &:focus {
```
InnaAtanasova, line 48: maybe use fd-focus mixin

InnaAtanasova (dismissed):

src/select.scss
```diff
@@ -26,7 +27,7 @@ $block: #{$fd-namespace}-select;
     align-items: center;
     justify-content: space-between;
     padding-right: 0;
-    padding-left: $fd-select-padding-x;
+    box-sizing: content-box;
```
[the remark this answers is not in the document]
stefanoScalzo: Without it, the outline goes to the edge due to the nesting

src/select.scss
stefanoScalzo: true i fixed it to just .#{$block}__text-content

src/select.scss
stefanoScalzo: changed to mixin

src/select.scss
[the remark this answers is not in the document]
stefanoScalzo: Screen Shot 2020-12-16 at 9 23 34 AM
this is what happens without it due to the button that was placed inside with the down arrow

src/select.scss
stefanoScalzo: fixed

src/select.scss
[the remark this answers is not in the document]
stefanoScalzo: i changed this with right: 0.0234375 which makes it equal on both sides, its the offset / 8

src/select.scss
stefanoScalzo: it was there previously, so instead fo causing breaking change, I changed everything to __Text and not __text-content

InnaAtanasova: But renaming `__text` to `__text-content` is a Breaking change

InnaAtanasova (commented):
I remove my requested changes to unblock the PR while I'm on vacation

stories/select/select.stories.js
```diff
@@ -304,7 +304,7 @@ export const semanticStates = () => `<div style="height: 200px">
                             "
                             aria-expanded="false"
                             aria-haspopup="listbox">
-                            <span id="b4546C40Value" class="fd-select__text-content">Error</span>
```
JKMarkowski, line 307: IS there any reason to change `text` to `text-content`?

src/select.scss
```diff
@@ -42,13 +42,34 @@ $block: #{$fd-namespace}-select;
...
+        content: '';
+        outline-offset: -0.1875rem;
+        outline-width: var(--sapContent_FocusWidth);
+        outline-color: var(--sapContent_FocusColor);
+        outline-style: var(--sapContent_FocusStyle);
+        height: 100%;
+        width: 100%;
+        z-index: 1;
+        right: 0.0234375rem;
+        position: absolute;
```
JKMarkowski, line 60: Maybe it will be better to have left/top/right/bottom as 0.1875rem and border instead of outline?

JKMarkowski (changes requested):
Hello @stefanoScalzo in examples it looks good, but there are things ot be changed in code

src/select.scss
stefanoScalzo: changed to border

JKMarkowski (approved):

JKMarkowski:
image
Is it done on purpose?

JKMarkowski (approved):

Issue #1721, opened 2020-09-28
Button - Text Overflow


JKMarkowski:
![image](https://user-images.githubusercontent.com/26483208/94436212-75ee2780-019c-11eb-939d-3bc8532cbb80.png)
When split button has some limited width, the text overflow is incorrect. 
The fix can be applied by wrapping text with `fd-button__text` element, but this should have changed styling as well.

PR #1975, opened 2020-12-15
fix: split button overflow

stefanoScalzo:
## Related Issue
Closes  https://github.com/SAP/fundamental-styles/issues/1721

## Description
Add text overflow to the text of split button
For regular cosy button 9.75rem is used because the other button takes 2.25rem and the max-width is always 12rem
For compact 10rem is used to keep this same 12rem max-width

### Before:
image

### After:
Screen Shot 2020-12-15 at 1 16 52 PM

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/button-split.scss
+++ b/src/button-split.scss
@@ -77,7 +77,7 @@ $fd-button-split-margin: 0.125rem / 2;
     }
   }
 
-  .fd-button__text {
+  .#{$fd-namespace}-button__text {
     @include fd-ellipsis();
 
     max-width: 9.75rem;
--- a/src/button-split.scss
+++ b/src/button-split.scss
@@ -76,4 +76,16 @@ $fd-button-split-margin: 0.125rem / 2;
       }
     }
   }
+
+  &__text {
+    max-width: 9.75rem;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+    display: block;
+
+    &--compact {
+      max-width: 10rem;
+    }
+  }
 }
--- a/src/button-split.scss
+++ b/src/button-split.scss
@@ -78,10 +78,9 @@ $fd-button-split-margin: 0.125rem / 2;
   }
 
   &__text {
+    @include fd-ellipsis();
+
     max-width: 9.75rem;
-    overflow: hidden;
-    text-overflow: ellipsis;
-    white-space: nowrap;
     display: block;
 
     &--compact {
--- a/src/button-split.scss
+++ b/src/button-split.scss
@@ -77,7 +77,7 @@ $fd-button-split-margin: 0.125rem / 2;
     }
   }
 
-  &__text {
+  .fd-button__text {
     @include fd-ellipsis();
 
     max-width: 9.75rem;
--- a/src/button-split.scss
+++ b/src/button-split.scss
@@ -77,9 +77,67 @@ $fd-button-split-margin: 0.125rem / 2;
     }
   }
 
-  .#{$fd-namespace}-button__text {
+  & > :first-child {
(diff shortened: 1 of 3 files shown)


src/button-split.scss
```diff
@@ -76,4 +76,16 @@ $fd-button-split-margin: 0.125rem / 2;
       }
     }
   }
+
+  &__text {
```
InnaAtanasova, line 80: Can't you target the first child instead of adding an additional modifier class?

src/button-split.scss
```diff
@@ -76,4 +76,16 @@ $fd-button-split-margin: 0.125rem / 2;
...
+    max-width: 9.75rem;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
```
InnaAtanasova, line 84: There's a mixin for this :D

src/button-split.scss
salarenko: But text is not wrapped in any element, and you cannot really target a `TextNode`.

salarenko (commented):
Where does the max-width come from? I don't see any requirements in the specs.
Shouldn't standard button component cover such cases?

stefanoScalzo:
> Where does the max-width come from? I don't see any requirements in the specs.
> Shouldn't standard button component cover such cases?

Screen Shot 2020-12-18 at 9 29 20 AM

stories/button/button.stories.js
```diff
@@ -284,7 +284,7 @@ menuButton.parameters = {
 
 export const splitMenuButton = () => `
 <div class="fd-button-split fd-has-margin-right-small" role="group" aria-label="button-split">
-  <button class="fd-button" aria-label="button">Button with text</button>
+  <button class="fd-button fd-button-split__text" aria-label="button">Button with a lot of text</button>
```
JKMarkowski, line 287: Could you just wrap the text inside `<span class="fd-button__text">Button with a lot of text</span>`?
And inside split-button just add `fd-button__text { max-width: ... }`

stefanoScalzo: fixed

src/button-split.scss
```diff
@@ -76,4 +76,15 @@ $fd-button-split-margin: 0.125rem / 2;
       }
     }
   }
+
+  .fd-button__text {
```
JKMarkowski, line 80: Let's add `fd-namespace` instead of `fd-`

stefanoScalzo: fixed

JKMarkowski (approved):
LGTM

src/button-split.scss
```diff
@@ -76,4 +76,15 @@ $fd-button-split-margin: 0.125rem / 2;
       }
     }
   }
+
+  .#{$fd-namespace}-button__text {
+    @include fd-ellipsis();
```
InnaAtanasova, line 81: Shouldn't this have a reset?

InnaAtanasova (changes requested):
I think it's a Breaking change and should be documented

src/button-split.scss
InnaAtanasova: Ok, I understand what you did. You don't need the reset coz it's `.fd-button__text`, but if this is the case, you can't just use it here. Split button is a separate component from Button and this violates the self-contained rule. You will need a modifier class for these rules

stefanoScalzo: fixed

InnaAtanasova:
Tests are failing

InnaAtanasova (approved):

PR #1979, opened 2020-12-16
docs: update form grid documentation

jessmetivier:
## Description
{{ _Updated **Form Grid** docs._ }}

Comment: I think it's best to remove the last example, since we shouldn't be providing an example of layouts that are unrecommended. Instead, I would like to add a note that advises against them in the "Recommended layouts" example.


InnaAtanasova (approved):

JKMarkowski (approved):
LGTM

droshev (approved):

PR #1989, opened 2020-12-21
fix: a11y changes to tabs

stefanoScalzo:
BREAKING CHANGES: change anchor element to button element on tab

## Description
Change the anchor element to a button element
Adds border bottom to remove the white overlapping the container border
fixes the offset on the navigation line to compensate
adds role=link for a11y

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/tabs.scss
+++ b/src/tabs.scss
@@ -25,7 +25,7 @@ $block: #{$fd-namespace}-tabs;
       transition: all $fd-tabs-link-transition-params;
       content: "";
       position: absolute;
-      bottom: -$fd-tabs-link-padding-y;
+      bottom: -0.9375rem;
       left: -$fd-tabs-active-link-line-offset-x;
       display: inline-block;
       height: $fd-tabs-link-border-width;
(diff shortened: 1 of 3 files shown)


JKMarkowski:
image
Counters are centered here, could you check it ?

JKMarkowski:
LEt's wait for @jbadan approval

InnaAtanasova (changes requested):
The role should not be `link`. Please check again the link provided in the issue description

stefanoScalzo:
> The role should not be `link`. Please check again the link provided in the issue description

fixed all a11y

mikerodonnell89 (changes requested):
I believe the changes to border bottom are causing this problem with the high contrast version

Screen Shot 2021-01-08 at 9 18 54 AM

mikerodonnell89:
Also I'm still seeing `role="link"` everywhere

InnaAtanasova:
Closing the PR, needs more work

Issue #107, opened 2019-03-26
Pagination component: underlying HTML changes for a11y


jbadan:
Pagination component: https://sap.github.io/fundamental/components/pagination.html

 `<a>` tags should be `<button>`

mikerodonnell89:
@jbadan Anchor is the correct element: https://a11y-style-guide.com/style-guide/section-navigation.html#kssref-navigation-pagination

droshev:
We will close it for now and reopen if the issue emerges again

PR #1990, opened 2020-12-21
fix: pagination button to anchor element

stefanoScalzo:
BREAKING CHANGES: change anchor element to button element for pagination links

## Related Issue
Closes SAP/fundamental-styles#107

## Description
Change all anchor elements to button elements with role="link" for a11y

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/stories/pagination/__snapshots__/pagination.stories.storyshot
+++ b/stories/pagination/__snapshots__/pagination.stories.storyshot
@@ -23,14 +23,14 @@ exports[`Storyshots Components/Pagination First page 1`] = `
       aria-label="Previous"
       class="fd-pagination__link fd-pagination__link--previous"
       href="#"
-      role="link"
+      role="navigation"
     />
     
       
     <button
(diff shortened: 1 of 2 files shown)


JKMarkowski:
Code looks good to me, but is it valid way for a11y? 
https://a11y-style-guide.com/style-guide/section-navigation.html#kssref-navigation-pagination

stefanoScalzo:
> https://a11y-style-guide.com/style-guide/section-navigation.html#kssref-navigation-pagination

@jbadan Are we sure we should be changing this markup to button?

mikerodonnell89:
I think it is fine to use button but they should have `role="navigation"` not link

mikerodonnell89 (changes requested):
edit: removed

mikerodonnell89:
Closing this as the original implementation is correct

PR #1999, opened 2020-12-29
fix: reworked styles for button and related components

mykolavorobiov:
## Related Issue
https://github.com/SAP/fundamental-ngx/issues/3458

## Description
Fix for input group component focus state. Required for https://github.com/SAP/fundamental-ngx/pull/4204
Small rework for button component and related components (combobox, carousel, input group, tile)

### Before:
image

### After:
image

- [x] RTL support
3. Testing
4. Documentation

### DIFF
--- a/src/carousel.scss
+++ b/src/carousel.scss
@@ -9,16 +9,13 @@ $button: #{$fd-namespace}-button;
   $fd-carousel-page-indicator-border: 0.0625rem solid var(--sapPageFooter_BorderColor) !default;
   $fd-carousel-button-size: 2.125rem !default;
   $fd-carousel-touchable-button-size: 2.625rem !default;
-  $fd-carousel-button-focus-outline-offset: -0.125rem !default;
+  $fd-carousel-button-focus-outline-offset: 0.0625rem !default;
   $fd-carousel-button-content-offset: 0.5rem !default;
   $fd-carousel-dot-size: var(--fdCarousel_Dot_Size) !default;
   $fd-carousel-dot-active-size: 0.5rem !default;
 
   @mixin after-position-overwrite() {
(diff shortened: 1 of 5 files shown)


mykolavorobiov:
@stefanoScalzo 
I saw fix related to combobox only in your PR https://github.com/SAP/fundamental-styles/pull/1973. But this issue related to input group and button in general. Please check.

stefanoScalzo:
> @stefanoScalzo
> I saw fix related to combobox only in your PR #1973. But this issue related to input group and button in general. Please check.

Well it should be for all elements that use select

mykolavorobiov:
@stefanoScalzo 
> Well it should be for all elements that use select
oh, `select` and `combobox` 're different components.
But `select` looks like `combobox`.
Anyway, I fixed button component and input group.

JKMarkowski:
IMO It should be applied to input group, because almost every pattern uses it.
The problem is that IE11 doesn't support `outline-offset`:
image
So I'm not sure if it can be approved.
@InnaAtanasova @mikerodonnell89

stefanoScalzo:
> IMO It should be applied to input group, because almost every pattern uses it (I would add `&::focus-within` selector, but it's not supported by IE11).
> The problem is that IE11 doesn't support `outline-offset`:
> image
> So I'm not sure if it can be approved.
> @InnaAtanasova @mikerodonnell89

I think in general we need a PR that refactors from outline to border for focus

InnaAtanasova:
Could be a Netlify preview issue but I don't see the focus in any of the examples.

mykolavorobiov:
@stefanoScalzo I think that `box-shadow` is better than border (`box-shadow` doesn't affect the size of the element and siblings)

droshev (approved):
🚢 LGTM

Issue #2009, opened 2021-01-04
fix: add focus for Object status as a link


Betrozov:
**Describe the bug**
Needs to add focus for Object status as a link.

**To Reproduce**
1. Go to https://sap.github.io/fundamental-styles/?path=/docs/components-object-status--primary
2. Go to Clickable Object Status example
3. UI5 example shows focus for clickable object status, but it is not present here

**Screenshots**
UI5: 
![image](https://user-images.githubusercontent.com/53509521/94402097-3407b600-0189-11eb-803c-be2b945a6de7.png)

PR #2010, opened 2021-01-04
fix: object status link focus

Betrozov:
## Related Issue
Closes https://github.com/SAP/fundamental-styles/issues/2009
Part of https://github.com/SAP/fundamental-ngx/issues/3414

## Description
Added focus for Object status as a link.

### Before:
Screenshot at 14-38-09

### After:
Screenshot at 14-38-48

- [X] All values are in `rem`
- [X] RTL support
3. Testing
- [X] Updated tests
4. Documentation

### DIFF
--- a/src/object-status.scss
+++ b/src/object-status.scss
@@ -291,6 +291,8 @@ $inverted-color-accents: (
 
   // CLICKABLE OBJECT STATUS
   &--link {
+    @include fd-fiori-focus-within(0);
+
     text-decoration: none;
     cursor: pointer;
 
--- a/src/object-status.scss
+++ b/src/object-status.scss
@@ -291,7 +291,7 @@ $inverted-color-accents: (
 
   // CLICKABLE OBJECT STATUS
   &--link {
-    @include fd-fiori-focus-within(0);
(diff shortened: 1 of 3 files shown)


olshansky (approved):

src/object-status.scss
```diff
@@ -291,6 +291,8 @@ $inverted-color-accents: (
 
   // CLICKABLE OBJECT STATUS
   &--link {
+    @include fd-fiori-focus-within(0);
```
JKMarkowski, line 294: Is there any reason of using focus within, instead of basic focus?

JKMarkowski: Could you also add `0.0625rem`, instead of 0? Just to follow focus visual concept from fiori 3

Betrozov: forgot to change,
@JKMarkowski  thanks
changed to fd-fiori-focus

Betrozov: @JKMarkowski fixed

InnaAtanasova (approved):

stefanoScalzo (approved):
LGTM

Issue #1357, opened 2020-08-13
Delta Theming Support


droshev:
Where there are different properties for some of the themes. Those properties are not handled with theming parameters and fundamental styling has to handle.
- [x] POC for delivering theme specific css file(s)
- [x] list all components that need delta
- [x] custom theming handling

The current components:
- [x] Inverted ObjectStatus
- [x] Panel
- [x] Avatar
- [x] Checkbox
- [x] Contextual menu (arrow color) |Inna: there's no theme parameter for the border color, the popover body uses box-shadow and we need border color. Will be discussed again with the designers|
- [x] radio button (borders)
- [x] tab 
- [x] Calendar Theme specific updates |Done by Mike D 3 months ago|
- [x] Button
- [x] Cards
- [x] Badge on Card
- [x] Carousel
- [x] Forms - header, input, message, textarea
- [x] Dialog - overlay
- [x] Feed Input - Inna
- [x] Tiles - Inna (needs info ofrom the designers)
- [x] Link - Inna
- [x] List - nav.indication, growing list, buttons - Inna
- [x] Message Box - Inna
- [x] Message Strip - Jedrzej
- [x] Multi Input - Jedrzej
- [x] Product Switch - Jedrzej
- [x] Select - Jedrzej
- [x] ShellBar - Jedrzej
- [x] Switch - Mike
- [x] Table- Mike
- [x] Time- Mike (needs info from designers) - https://github.com/SAP/fundamental-styles/issues/2081
- [x] Toolbar- Mike (needs info from designers - should be a quick fix if any changes required) - https://github.com/SAP/fundamental-styles/issues/2082
- [x] Page- Mike
- [x] Rating Indicator - see [#1394](https://github.com/SAP/fundamental-styles/issues/1394) | Done by Deno 4 months ago |

notes:
`element.sapThemeMetaData-Base-baseLib` -> 
example of a custom theme:
```
.sapThemeMetaData-Base-baseLib{background-image: url('data:text/plain;utf-8,{"Path": "Base.baseLib.denotheme.css_variables", "PathPattern": "/%frameworkId%/%libId%/%themeId%/%fileId%.css", "Extends": ["sap_fiori_3_dark","sap_fiori_3","sap_base_fiori","baseTheme"], "Tags": ["Fiori_3","DarkColorScheme"], "Version": { "Build":"11.1.20.20200713203344", "Source": "11.1.20", "Engine": "1.52.0"}}');}
```
Time: 1w

[...]

PR #2015, opened 2021-01-05
fix: add delta theming for Combobox and Multi Input

JKMarkowski:
## Related Issue
Part of #1357

## Description
There is added border on side for button inside multi input/combobox, on hover/active state, for HCB/HCW

- [x] RTL support
3. Testing


droshev (approved):

InnaAtanasova:
shouldn't this PR be a `chore`?

InnaAtanasova (approved):

Issue #2045, opened 2021-01-12
Defect Hunting sprint 53


mikerodonnell89:
- [x] Buttons docs - buttons should have some vertical margin (not new to this sprint) - Mike https://github.com/SAP/fundamental-styles/pull/2054

![Screen Shot 2021-01-12 at 10 04 55 AM](https://user-images.githubusercontent.com/2471874/104347248-a04ec100-54bd-11eb-9e4c-812262870228.png)

- [x] Field Set - Need space between checkbox and label

![Screen Shot 2021-01-12 at 10 09 02 AM](https://user-images.githubusercontent.com/2471874/104347745-34208d00-54be-11eb-9348-2c039c4eb8d5.png)

- [x]  Product switch small screen - UI5 centers the title vertically if there is no subtitle, our lib does not.  Unclear in the specs what should really happen, should probably talk to designers **(Inna: I agree, it's not very clear from the specs what should happen. I will add this to the topics to be discussed and will move this to a separate issue)** 
Opened an issue https://github.com/SAP/fundamental-styles/issues/2052

UI5
![Screen Shot 2021-01-12 at 10 38 28 AM](https://user-images.githubusercontent.com/2471874/104351188-51eff100-54c2-11eb-8a65-e3ea420fcf35.png)

fd-styles
![Screen Shot 2021-01-12 at 10 39 52 AM](https://user-images.githubusercontent.com/2471874/104351341-82378f80-54c2-11eb-8cb4-364a6fb9ce6d.png)

- [x] Select popover body width should match width of control unless min/max explicitly set - **It shouldn't be same by default - Jedrzej**

![Screen Shot 2021-01-12 at 10 46 35 AM](https://user-images.githubusercontent.com/2471874/104352077-74363e80-54c3-11eb-803e-e45b3cc58261.png)

- [x] I do not think the shellbar user menu button should have this border - Mike
![Screen Shot 2021-01-12 at 10 52 14 AM](https://user-images.githubusercontent.com/2471874/104352705-3dacf380-54c4-11eb-870a-e9f510989064.png)

- [x] Table docs - Contextual menu popovers not working - **Inna** https://github.com/SAP/fundamental-styles/pull/2056
![Screen Shot 2021-01-12 at 10 53 53 AM](https://user-images.githubusercontent.com/2471874/104352894-79e05400-54c4-11eb-9f49-5d068964f5e1.png)

InnaAtanasova:
[...]

PR #2054, opened 2021-01-13
docs: button docs improvements

mikerodonnell89:
part of #2045 

Improves the look of button docs


InnaAtanasova (changes requested):
It's not applied on all of the buttons:
Screen Shot 2021-01-13 at 4 33 40 PM
Screen Shot 2021-01-13 at 4 33 25 PM
Screen Shot 2021-01-13 at 4 33 18 PM
Screen Shot 2021-01-13 at 4 33 06 PM

.storybook/custom/custom.scss
```diff
@@ -286,6 +286,15 @@ code.docs-code-grey {
   background-color: #f8f8f8;
 }
 
+.fddocs-button-container {
```
InnaAtanasova, line 289: Won't this work if you apply it on the `fddocs-container` without adding additional class? And prob should be `fd-docs-container`?

mikerodonnell89: `fddocs-container` was already in place so I just kept it

JKMarkowski:
For me it works fine, but not for split button, which I guess has stronger selector:
image

InnaAtanasova (approved):

PR #2057, opened 2021-01-13
docs: add fix for missing box-shadow for Dialog in IE11

InnaAtanasova:
## Related Issue
Part of SAP/fundamental-styles#2045

## Description
In IE11 the top box-shadow was missing for Dialog. I changed the custom docs css rules responsible for rendering static dialogs

NA
NA
3. Testing
NA
4. Documentation
NA


JKMarkowski (approved):
Idk how you discovered that this is the problem, but the issue doesn't exist anymore :D

PR #2061, opened 2021-01-15
chore: update the rules for running chromatic in PRs

droshev:
## Description
The PR brings changes how the visual regression testing is done as part of the PRs.
By default the visual regression won't be triggered. In order to trigger it a commit has to have the following string `[ci visual]`

If we agree to go this way then we need to update the documentation and how a PR should be opened and maintained.

The [wiki](https://github.com/SAP/fundamental-styles/wiki/Visual-testing-with-Chromatic) has been updated.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -41,6 +41,7 @@ Closes SAP/fundamental-styles#
 - [ ] tested Storybook examples with "CSS Resources" `unnormalize` option 
 - [ ] Verified all styles in IE11
 - [ ] Updated tests
+- [ ] last commit message should have `[ci visual]` so it can trigger chromatic visual regression
 4. Documentation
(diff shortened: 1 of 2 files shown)


InnaAtanasova (changes requested):
Please add an example commit string + a checkbox in the PR template (with example) so new contributors are not confused

JKMarkowski (approved):

droshev:
> Please add an example commit string + a checkbox in the PR template (with example) so new contributors are not confused

done

InnaAtanasova (approved):

InnaAtanasova:
> > Please add an example commit string + a checkbox in the PR template (with example) so new contributors are not confused
> 
> done

I would add an example commit message in the template. I am sure many people will have questions from the sort of "can I add more text to `[ci visual]`?", "Should the message start with `[ci visual]`?", etc.

Issue #2030, opened 2021-01-07
Documentation - all compact bar examples should have all compact buttons


mikerodonnell89:
There are a few bar examples throughout the documentation where some buttons will have the cozy size.  Need to change these to compact (which is the default for bar) 

![Screen Shot 2021-01-06 at 5 37 16 PM](https://user-images.githubusercontent.com/2471874/103836270-d3203180-5045-11eb-918f-53c493d3621d.png)

PR #2064, opened 2021-01-16
docs: update buttons in compact bars

droshev:
## Related Issue
fixes: #2030

## Description
use compact buttons in compact bars

### Before:

Screen Shot 2021-01-17 at 6 56 16 PM

### After:

Screen Shot 2021-01-17 at 6 57 13 PM

2. Testing
- [x] Updated tests
3. Documentation

### DIFF
--- a/stories/bar/__snapshots__/bar.stories.storyshot
+++ b/stories/bar/__snapshots__/bar.stories.storyshot
@@ -199,172 +199,360 @@ exports[`Storyshots Components/Bar Cozy 1`] = `
 `;
 
 exports[`Storyshots Components/Bar Default 1`] = `
-<div
-  class="fd-bar"
->
+<section>
   
-    
+
+  <p>
+    <b>
+      Compact bar with compact elements
+    </b>
+  </p>
+  
+
   <div
-    class="fd-bar__left"
+    class="fd-bar"
   >
     
-        
+    
     <div
-      class="fd-bar__element"
+      class="fd-bar__left"
     >
       
-            
-      <button
-        aria-label="button"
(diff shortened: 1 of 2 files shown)


InnaAtanasova (changes requested):
This is NOT a Breaking change, it's just an example of what can go inside a Bar component. The implementation libraries do not depend on this additional class. Also, the buttons were left cosy intentionally as the user is not prevented from mixing cozy/compact mode inside Bar component. For me personally this PR (the issue) does not bring value but if we decide to merge it, it should not be a Breaking change PR.

droshev:
> This is NOT a Breaking change, it's just an example of what can go inside a Bar component. The implementation libraries do not depend on this additional class. Also, the buttons were left cosy intentionally as the user is not prevented from mixing cozy/compact mode inside Bar component. For me personally this PR (the issue) does not bring value but if we decide to merge it, it should not be a Breaking change PR.

I agree with you. On the other hand we should show the best practices and the right approach.

JKMarkowski (approved):

InnaAtanasova (approved):

droshev:
> This is NOT a Breaking change, it's just an example of what can go inside a Bar component. The implementation libraries do not depend on this additional class. Also, the buttons were left cosy intentionally as the user is not prevented from mixing cozy/compact mode inside Bar component. For me personally this PR (the issue) does not bring value but if we decide to merge it, it should not be a Breaking change PR.

@InnaAtanasova I have updated one of the examples explaining that the we can mix compact bar with cozy elements. Thank you for pointing out.

Issue #2023, opened 2021-01-06
Avatar Group - styles implemenation


manjunathanagaraj:
Define styles as per Fiori3 specifications:

https://wiki.wdf.sap.corp/wiki/pages/viewpage.action?pageId=2107190068

PR #2066, opened 2021-01-19
feat: introduce Avatar Group component

olshansky:
## Related Issue
Closes: https://github.com/sap/fundamental-styles/issues/2023

## Description
This PR introduces the Avatar Group component

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation


src/avatar-group.scss
```diff
@@ -0,0 +1,103 @@
...
+    "group": (
+      "xs": ("margin-right": -0.75rem),
+      "s": ("margin-right": -1.25rem),
+      "m": ("margin-right": -1.625rem),
+      "l": ("margin-right": -2rem),
+      "xl": ("margin-right": -2.75rem)
+    )
+  );
+
+  $fd-const-ratio: 1rem;
```
EvilAlexei, line 27: nit: Maybe it's better not to separate variables with an empty line?
For me, it's kind of a logical separation and it's a bit confusing.

stories/avatar-group/avatar-group.stories.js
```diff
@@ -0,0 +1,437 @@
...
+
+`,
+        docs: { iframeHeight: 250 },
+        tags: ['a11y', 'f3', 'theme'],
+        components: ['avatar-group', 'avatar', 'icon', 'button', 'popover', 'bar', 'layout-grid', 'title', 'quick-view', 'title', 'link', 'form-layout-grid', 'form-group', 'form-item', 'form-label', 'input']
+    }
+};
+
+export const individualType = () => `
+<div class="fd-avatar-group fd-avatar-group--individual fd-avatar-group--s">
```
EvilAlexei, line 48: please set them in one lime, to fix empty storyshot
```
export const individualType = () => `<div class="fd-avatar-group fd-avatar-group--individual fd-avatar-group--s">
```

EvilAlexei (changes requested):

EvilAlexei (approved):

src/avatar-group.scss
```diff
@@ -0,0 +1,102 @@
...
+    @include fd-reset();
+
+    display: inline-block;
+  }
+
+  &__more-button {
+    @include fd-reset();
+
+    .fd-button__text {
+      font-size: inherit;
```
InnaAtanasova, line 63: Cannot import button component inside Avatar Group component. If you need to change the properties of a button you need a modifier class like: `.fd-avatar-group__button` and then on your button element you will have `<button class="fd-button fd-avatar-group__button">...</button>`

src/avatar-group.scss
```diff
@@ -0,0 +1,102 @@
...
+
+      bdi {
```
InnaAtanasova, line 65: what is this?

src/avatar-group.scss
```diff
@@ -0,0 +1,102 @@
...
+  @include fd-reset();
+
+  &--individual {
+    .#{$block}__item {
+      @include fd-fiori-focus();
+      @include action-cursor();
+    }
+  }
+
+  &--group {
```
InnaAtanasova, line 44: Please select another name. `fd-avatar-group--group` is a bit confusing. Maybe use the type from the specs - Single Interaction Area | Individual Interaction Areas  or at least use the word type: `fd-avatar-group--group-type`  and `fd-avatar-group--individual-type`

InnaAtanasova (changes requested):
Screen Shot 2021-01-20 at 11 57 25 AM
Screen Shot 2021-01-20 at 11 57 11 AM
Please fix the focus line

InnaAtanasova:
Screen Shot 2021-01-20 at 12 02 02 PM

In the examples, where you show the Popover body for the "middle"  Avatars (aka all Avatars except the first and the last), could you show the Popover arrow in the middle?

src/avatar-group.scss
olshansky: @InnaAtanasova that's exactly what I've done: 

```<button class="fd-button fd-avatar-group__more-button fd-avatar-group__more-button--s"></button>```

https://github.com/SAP/fundamental-styles/pull/2066/files/9a4a0051f56d36b4816bc2bc8d40c815c7f6db11#diff-c517d72c23b1a1678fc295227f77f8c2ddf4657e02f3bdbb976615056dccfe96R233

And `.fd-button-text` have this styles changes just if it placed inside `.fd-avatar-group__more-button`

File: src/avatar-group.scss (hunk 4 above)
Comment (olshansky) [reply src/avatar-group.scss:65]: @InnaAtanasova 

> The `<bdi>` element is used to wrap a span of text and instructs the bidirectional algorithm to treat this text in isolation from its surroundings. This works in two ways:
> 
> - The directionality of text embedded in `<bdi>` does not influence the directionality of the surrounding text.
> - The directionality of text embedded in `<bdi>` is not influenced by the directionality of the surrounding text.

 In our case to prevent changes from `+9` to `9+`

Comment (JKMarkowski) [reply src/avatar-group.scss:65]: Could you add some class with `fd-reset` and inherit `font-size`/`color`/`weight`? Or even use `bdi` with `fd-button__text` , so you can get rid of his selector

**Review by JKMarkowski — COMMENTED** [review]
There is only 1 thing from my side. Rest LGTM, Great Job!

File: src/avatar-group.scss (hunk 3 above)
Comment (InnaAtanasova) [reply src/avatar-group.scss:63]: I disagree with you :) I see that you use specificity and in other applications this would work perfectly, but that's not how it works in styles. `.fd-button__text` means that an element `__text` belongs to a block `.fd-button` aka button. This means that ANY styling related to this element should only be present in the fd-button scss file. These are the rules for being self-contained. IF in another component (Avatar, Bar, Card, etc.) you use button, but you need to overwrite a little bit the text, you need an additional modifier class and you cannot rely on the specificity for this. So, in your case you will have `fd-button__text .fd-avatar-group__button-text` and inside src/avatar-group.scss file you will only have:
```
.fd-avatar-group__button-text {
     font-size: inherit;
}
```

In general, you know you are not violation the self-contained rules if inside your scss file you have only the block element `.fd-avatar-group` and inside its curly braces ALL block codes start with `__some-element` or `--some-modifier`. Hope this is clear.

Comment (olshansky) [reply src/avatar-group.scss:63]: @InnaAtanasova now I got it :smiley:  Thank you!

**Review by mikerodonnell89 — CHANGES_REQUESTED** [review]
Looking at the Fiori designs, there should be some margin/padding around each avatar in this popover

Screen Shot 2021-01-22 at 8 16 19 AM

File: stories/avatar-group/avatar-group.stories.js [hunk 6]
```diff
@@ -0,0 +1,435 @@
...
+        tags: ['a11y', 'f3', 'theme'],
+        components: ['avatar-group', 'avatar', 'icon', 'button', 'popover', 'bar', 'layout-grid', 'title', 'quick-view', 'title', 'link', 'form-layout-grid', 'form-group', 'form-item', 'form-label', 'input']
+    }
+};
+
+export const individualType = () => `<div class="fd-avatar-group fd-avatar-group--individual-type fd-avatar-group--s">
+    <div class="fd-popover">
+        <div class="fd-popover__control">
+            <div class="fd-avatar-group__item" tabindex="0" aria-haspopup="true" aria-expanded="false" onclick="onPopoverClick('popover_avatar_5z28edb');">
+                <span class="fd-avatar fd-avatar--s fd-avatar--circle fd-avatar--border" aria-label="Wendy Wallace">WW</span>
```
mikerodonnell89, line 51: This appears as a grey border in Quartz mode and I believe it should not be there at all.  HCW/HCB look good though

mikerodonnell89 (changes requested):

InnaAtanasova:
> Looking at the Fiori designs, there should be some margin/padding around each avatar in this popover
> Screen Shot 2021-01-22 at 8 17 19 AM

I didn't find anything in the design about the spacing of the avatars inside a Popover. I will write it in the agenda for the next meeting with the designers. For now I guess it's ok with the current implementation.

stories/avatar-group/avatar-group.stories.js
InnaAtanasova: This element has `fd-avatar--border` which adds a `border: 0.0625rem solid --sapGroup_ContentBorderColor`
Screen Shot 2021-01-22 at 4 16 25 PM

InnaAtanasova (approved):

mikerodonnell89 (approved):

EvilAlexei:
@olshansky for Chromatci build, please update this branch from the `main`, it's already updated with PostCSS

Issue #1742, opened 2020-10-01
Object Identifier - weak link selector, wrong `__table` class


JKMarkowski:
- Table - Current implementation forces user to put `fd-object-identifier__table` class as a wrapper for `fd-object-identifier` component. It should be changed to modifier, because it's only about the font size.

Table doesn't match example from Fiori3
![image](https://user-images.githubusercontent.com/26483208/94813205-ffe5fc80-03f7-11eb-91c7-7d8294c52edb.png)
![image](https://user-images.githubusercontent.com/26483208/94813226-06747400-03f8-11eb-8b0a-f6cf280f29f8.png)

It seems like the height should be auto + some padding, when object identifier is used. This should be consulted with designers.

- Link - For `@ngx/core` library the `fd-object-identifier__link` element, which in example is used with `fd-link` element has too weak selector to set own values. If it's not meant to be used only with `fd-link`, then it should have `fd-reset()` mixin. 

Also current [object identifier link](https://fundamental-styles.netlify.app/iframe.html?id=components-object-identifier--title-only&viewMode=docs#link) example has 2 ways to build ObjectIdentifier as link. We should keep the one built with `fd-link` in my opinion.

InnaAtanasova:
0.5day

PR #2067, opened 2021-01-20
fix: Simplify object identifier

JKMarkowski:
## Related Issue
Closes SAP/fundamental-styles#1742

## Description
There is simplification of object identifier.
Now link has to be used with `fd-link`. Previously there was 2 ways to make `link` object identifier. which didn't make any sense.

Also `medium` size is not restricted to table anymore

## Breaking Changes
Removed Classes:
 - `fd-object-identifier__title--link`
 - `fd-object-identifier__table`

Added:
- `fd-object-identifier--medum`

No visual changes.

- [x] All values are in `rem`
- [NA] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/object-identifier.scss
+++ b/src/object-identifier.scss
@@ -7,28 +7,20 @@ $block: #{$fd-namespace}-object-identifier;
   $fd-object-identifier-title-font-size: var(--sapFontLargeSize);
   $fd-object-identifier-title-font-weight: bold;
 
+  .#{$fd-namespace}-link {
+    font-size: inherit;
+    font-weight: inherit;
+  }
+
   &__title {
     @include fd-reset();
 
     color: var(--sapList_TextColor);
     font-size: $fd-object-identifier-title-font-size;
 
-    &--link {
-      color: var(--sapLinkColor);
-    }
-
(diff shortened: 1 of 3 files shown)


src/object-identifier.scss
```diff
@@ -7,46 +7,31 @@ $block: #{$fd-namespace}-object-identifier;
   $fd-object-identifier-title-font-size: var(--sapFontLargeSize);
   $fd-object-identifier-title-font-weight: bold;
 
+  .#{$fd-namespace}-link {
+    font-size: inherit;
+    font-weight: inherit;
+  }
+
```
InnaAtanasova, line 14: We should not bring link component inside Object Identifier. You need to create a modifier class to overwrite the link properties.

InnaAtanasova (changes requested):

src/object-identifier.scss
JKMarkowski: Done

mikerodonnell89 (approved):
LGTM

InnaAtanasova (approved):

Issue #2063, opened 2021-01-16
List Component improvements


droshev:
There are several issues related to list component. It is better if they are combined in 1 issue so we can tackle everything at the same time. Here is the list of those improvements needed:

### Done:
- [x] For "No Data" styles - Necessary styles are there - `--no-data` modifier
- [x] semantic indication in the beginning of the list - **I couldn't find it in any spec**
- [x] list Byline is changing its height when bottom border is added #1596 - I Can't reproduce it anymore
- [x] On touch of multi-select list item non fiori color is highlighted #1712 - On styles there is no way to make it work without JS. For NGX it works fine

#### Dnd Styles: - will be implemented in NGX (NA)
- [x] Drag and Drop styles - Can be taken from NGX 

#### Interractive states: - Jedrzej ( 1day )
- [x] Interactive States -Active/down state for non actionable item
- [x] Selection list with simple list item #1588 - `.is-selected` should work fine even on basic `fd-list__item`

#### Growing list: (1/2day) - Inna - https://github.com/SAP/fundamental-styles/pull/2099
- [x] growing list - TODO Intergration with busy-indicator

#### Byline: Mike ( 2days ) - https://github.com/SAP/fundamental-styles/pull/2074
- [x] wrapping behaviour - Missing in byline 
- [x] object status in standard list should be aligned on the right #1906 - Byline
- [x] list item counter - Done for standard list on styles, missing in byline list - `fd-list__item-counter`

#### Actions: (2days) - Jedrzej
- [x] [Buttons](https://wiki.wdf.sap.corp/wiki/display/visualcore/List+%28Fiori+3%29+-+Buttons?src=contextnavpagetreemode) 
- [x] [Deletion](https://wiki.wdf.sap.corp/wiki/display/visualcore/List+%28Fiori+3%29+-+Deletion?src=contextnavpagetreemode) 
- [x] Delete icon position is not middle in standard list with byline #1590 - There should be added example with centered button on byline list

#### Styles Inconsistency: (1day) - Inna
- [x] [Atachment List](https://wiki.wdf.sap.corp/wiki/display/visualcore/List+%28Fiori+3%29+-+Attachment+List) - [PR](https://github.com/SAP/fundamental-styles/pull/2103)
- [x] Group header got some issues with truncation - [PR](https://github.com/SAP/fundamental-styles/pull/2101)
[...]

Issue #1334, opened 2020-08-06
List - Interactive States -Active/down state is not present for  non actionable item


sKudum:
**Describe the bug**
Active/down state styles are not getting displayed on List styles on nonactionable items.

Nonactionable List

![Nonactionablelist (1)](https://user-images.githubusercontent.com/53487468/89535801-8a551a00-d814-11ea-90a4-3bdf3e6cbace.gif)

Actionable List

![Actionable List (1)](https://user-images.githubusercontent.com/53487468/89535693-6560a700-d814-11ea-8878-d05da2999025.gif)

> Is this issue related to a specific component?
List

> What version of the Fundamental Library Styles are you using?
0.11

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

> What front-end framework are you implementing in, e.g., Angular, React, etc.?
Angular

**Expected behavior**
Interactive states should match fiori3
<img width="262" alt="Screen Shot 2020-08-06 at 3 15 40 PM" src="https://user-images.githubusercontent.com/53487468/89517705-ba41f480-d7f7-11ea-86a6-6964b3a725b6.png">

droshev:
@sKudum is this still valid issue?

sKudum:
Hi @droshev , Yes please

![2020-12-11_12-06-02 (1)](https://user-images.githubusercontent.com/53487468/101871315-5cf4f000-3ba9-11eb-89b3-69917bece0a3.gif)

droshev:
will be tackled as part of [#2063](https://github.com/SAP/fundamental-styles/issues/2063)

PR #2068, opened 2021-01-20
feat: list interractive states

JKMarkowski:
## Related Issue
part of https://github.com/SAP/fundamental-styles/issues/2063

## Description
There is added: 
- `fd-list__item--interractive` modifier for interractive (hover/active) states for list item
- `is-selected` is now supported by basic `fd-list__item`

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/_listBase.scss
+++ b/src/_listBase.scss
@@ -38,6 +38,10 @@
       }
     }
 
+    &--interractive {
+      @include list-item-states();
+    }
+
     &--action {
       padding: 0;
       text-align: center;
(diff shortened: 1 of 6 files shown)


InnaAtanasova (approved):

mikerodonnell89:
In this comment: https://github.com/SAP/fundamental-styles/issues/1334#issuecomment-743005239

@sKudum looks like he wants to see it for byline too, however I'm not seeing this mentioned in the specs.  Thoughts?

mikerodonnell89 (commented):
In this comment: https://github.com/SAP/fundamental-styles/issues/1334#issuecomment-743005239

@sKudum looks like he wants to see it for byline too, however I'm not seeing this mentioned in the specs.  Thoughts?

sKudum:
> In this comment: [#1334 (comment)](https://github.com/SAP/fundamental-styles/issues/1334#issuecomment-743005239)
> 
> @sKudum looks like he wants to see it for byline too, however I'm not seeing this mentioned in the specs. Thoughts?

Hi @mikerodonnell89 , I agree with you it is not mentioned in case of byline as well.

Concern was raised by QA team that For interaction it should be same like list-item, because no other variant has mentioned this styles specifically.

The same way how focus, disable styles are mentioned. Similarly interactive states needs to be followed until it is mentioned specially other behaviour .

JKMarkowski:
@sKudum, @mikerodonnell89 This can be achieved in byline same as in standard list - by adding `fd-list__item--interractive` class to item. I will create example with it

PR #2069, opened 2021-01-20
fix: Quick View title gaps

olshansky:
## Related Issue
Related to https://github.com/SAP/fundamental-ngx/issues/4400

## Description
If you use `fd-bar__left` or `fd-bar__right` instead of `fd-bar__middle` inside `fd-bar--header` your popover header (quick view title as well) will have no gaps.

### Before:
image

### After:
image

- [x] All values are in `rem`
- [n/a] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/quick-view.scss
+++ b/src/quick-view.scss
@@ -29,24 +29,24 @@ $fd-quick-view-group-item-indent: 0.75rem;
     @include fd-reset();
 
     padding: 1rem;
-  }
 
-  .#{$fd-namespace}-bar {
-    padding: 0;
+    .#{$fd-namespace}-bar {
+      padding: 0;
 
-    &--header-with-subheader {
-      height: auto;
-      padding-bottom: $fd-quick-view-group-item-indent;
+      &--header-with-subheader {
+        height: auto;
+        padding-bottom: $fd-quick-view-group-item-indent;
 
-      .#{$fd-namespace}-avatar {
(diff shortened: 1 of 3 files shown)


mykolavorobiov (approved):

EvilAlexei (approved):

sKudum (approved):

src/quick-view.scss
```diff
@@ -29,24 +29,24 @@ $fd-quick-view-group-item-indent: 0.75rem;
...
-      }
+        .#{$fd-namespace}-avatar {
+          margin-right: $fd-quick-view-avatar-indent;
+        }
 
-      .#{$fd-namespace}-bar {
-        &__left,
-        &__middle,
-        &__right {
-          align-items: flex-start;
```
InnaAtanasova, line 49: I don't see any change here except formatting

InnaAtanasova (changes requested):

src/quick-view.scss
olshansky: @InnaAtanasova all `.#{$fd-namespace}-bar { ... }` block moved into `&__content {... <here now>}`

InnaAtanasova: ah, ok! Thanks for explaining. I totally missed this sneaky } :D

InnaAtanasova (approved):

PR #2071, opened 2021-01-21
chore: update browserlist package version

droshev:
update browserlist package version


JKMarkowski:
There is only `package-lock.json` changed. Could you add `package.json` there as well?

InnaAtanasova:
> There is only `package-lock.json` changed. Could you add `package.json` there as well?

and resolve the conflicts in `package-lock.json`

droshev:
> There is only `package-lock.json` changed. Could you add `package.json` there as well?

actually this is done by a script and this is what was changed

JKMarkowski (approved):

Issue #2021, opened 2021-01-06
Status Indicator - styles implementation


manjunathanagaraj:

manjunathanagaraj:
Define styles as per Fiori3 specifications:
https://wiki.wdf.sap.corp/wiki/display/visualcore/Status+Indicator

manjunathanagaraj:
Target sprint for PR merge is sprint55

PR #2072, opened 2021-01-21
feat: introduce status indicator component

Lokanathan-k:
introduce status indicator component

## Related Issue
https://github.com/SAP/fundamental-styles/issues/2021

## Description
 Introduce status indicator component styles 

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation


stories/status-indicator/status-indicator.stories.js
```diff
@@ -0,0 +1,637 @@
...
+					c-0.9-0.7-1.5-1.7-2-3H6.4l0.5-1.9h1V12H6.4l0.5-1.9h1.3C8.6,8.9,9.3,8,10.3,7.3c1-0.8,2.3-1.1,3.8-1.1c1.1,0,2,0.2,2.8,0.7L17.3,7
+					l-0.6,2.3l-0.5-0.4c-0.4-0.2-0.9-0.4-1.4-0.5c-0.6-0.1-1.1-0.1-1.7,0c-0.6,0.1-1,0.3-1.5,0.7c-0.3,0.3-0.6,0.6-0.7,1H16.4z" stroke-width="0.25" stroke="#6a6d70" fill="transparent">
+				</path>
+			</svg>
+		</svg>
+	</svg>
+</div>
+</div>
+<div class="example-container">
+<span style="min-width: 150px;">Critical Filling :</span>
```
kavya-b, line 206: inline styling is not encouraged. Can you place documentation related custom styling in custom.scss?

src/status-indicator.scss
```diff
@@ -0,0 +1,65 @@
...
+  color: $fd-status-indicator-default-color;
+  stroke: $fd-status-indicator-default-color-stroke;
+  stroke-width: $fd-status-indicator-default-color-stroke-width;
+  fill: $fd-status-indicator-default-color-fill;
+  flex-direction: column;
+  align-items: center;
+
+  @include fd-reset();
+
+  &-htext {
```
kavya-b, line 38: this should probably be a modifier

kavya-b:
all examples not showing correct dotted focus: 
image
image

same titles for examples? 
image

quartz dark doesn't work right in IE, could you check?
image

sKudum:
Hi lok,

**Query**.
 Are we targeting to provide svgs to users? as per my understanding application's UX team will create it and user has to tag them in html code and apply our modifier classes.

**Few observations.**
1).Having examples for Cirulcar fill, filling sequence, grouping, thresholds will help users.
2).How about radial and angular mechanisms, could not find any.
3).Sharing few details about usage will help, for example how can achieve required stroke, which class will help me to achieve radial mechanism, how I need to use it.
4). How I have to use these classes when I have a svg image shared from ux team?
           Try to use sketch for having one svg stored. in local, point out how a svg file and apply the classes for desired mechanisum.

_As mentioned before, we can take help of fiori-moments. how those guys are doing it._

sKudum:
Few Quires:

1). On click , on hover, active etc these svgs shouldn't have a11y features?
2).Are we targeting to give classes which help users to apply desired color and size in svg.  Everthing else its has to take help from svg?
3).Is it possible to have documentation with more details.
            we can state which options are from lib and from svg to obtain desired behaviour.?

Lokanathan-k:
> On click , on hover, active etc these svgs shouldn't have a11y features?

@puru-hk i have added the a11y feature as part of the pr, pls review this if it needs more details on the a11y for the above component.

Lokanathan-k:
> Few Quires:
> 
> 1). On click , on hover, active etc these svgs shouldn't have a11y features?
these on hover or active etc are not defined in the spec.
> 2).Are we targeting to give classes which help users to apply desired color and size in svg. Everthing else its has to take help from svg?
yes, we can only achieve the size and color other filling is done by lineargradient and radialgradient.
> 3).Is it possible to have documentation with more details.
added
> we can state which options are from lib and from svg to obtain desired behaviour.?

DeepakSap14 (approved):

sKudum (approved):
Approving it to avoid blocking, expecting queries be addressed with respected person consultation.

puru-hk:
Tested: From accessibility point of view able access view keyboard, tooltip is implemented for icon. 
But regarding the screen reader it is not reading the image icon.

olshansky (approved):

JKMarkowski (changes requested):
Hello @Lokanathan-k 
Could you add commit with `[ci visual]` to trigger chromatic? After it I can approve - Good Job!

puru-hk:
Looks fine for me

puru-hk (approved):

JKMarkowski (approved):

PR #2074, opened 2021-01-21
fix: byline list improvements

mikerodonnell89:
part of #2063 

Need some help with the "MORE" button for text that is too long, can't figure out why hover/click isn't affecting the style.  Using the developer console to set those selectors does adjust the style correctly


InnaAtanasova (commented):
Screen Shot 2021-01-22 at 3 20 28 PM
Screen Shot 2021-01-22 at 3 20 23 PM

In the specs for wrapping behaviour is specified that the min-width should be 2rem but I see it's 2.5rem in fundamental.
Screen Shot 2021-01-22 at 3 23 05 PM

I think it's a good idea to show one example with `LESS` link

JKMarkowski:
> Need some help with the "MORE" button for text that is too long, can't figure out why hover/click isn't affecting the style. Using the developer console to set those selectors does adjust the style correctly

@mikerodonnell89 This is because of pseudo element created, when list item is focused. You can add 
`pointer-events: none;` to this pseudo element.
Or change it, by just adding outline instead of whole pseudo element - but this is not well interpreted by IE11

InnaAtanasova (approved):

JKMarkowski (changes requested):
There are some issues with selecting text and clicking on `more` link, could you add  `pointer-events: none;` to focus indicator pseudo element ?

mikerodonnell89:
> There are some issues with selecting text and clicking on `more` link, could you add `pointer-events: none;` to focus indicator pseudo element ?

The cursor behaves as expected on the "more" link and I'm confused by what you mean for selecting text, these styles already have `pointer-events: none`, do you mean they should have `pointer-events: all`?

JKMarkowski (approved):
LGTM

Issue #2020, opened 2021-01-06
Introduce Facets(Form, Key-Value, etc) as component


kavya-b:
### Component name
> Propose a new or variant name

Facet

### Detailed description
> Describe the request

Facets are used with Object Page and Dynamic Page usually, but the facet types can also be used independently. Therefore, we can introduce Facets as their own component.

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

ngx

> What front-end framework are you implementing in, e.g., Angular, React, etc.?

Angular

### Use case
> Describe when to use (and when not to use)

- most frequently used with Object Page and Dynamic Page as part of the header content.
- the facet types Form Facets, Key Value Facets, Image Facets etc. can also be used independently

### Visual style
> Describe specific needs related to color, typography, iconography, borders, space, size, and other visual properties.

as per Fiori3 visual spec.

> Describe how to arrange multiple instances and other alignment specifics.

as per Fiori3 visual spec.

> How should this respond to common contexts, i.e., touch, RTL, small screens.

as per Fiori3 visual spec.

> Describe interactive states, e.g., disabled, expanded, pressed.

as per Fiori3 visual spec.

### Expected behavior
> How should this handle events, outcomes, and transitions

as per Fiori3 visual spec.

> What [states should be available](https://medium.com/swlh/the-nine-states-of-design-5bfe9b3d6d85), i.e., nothing, loading, none, one, some, too many, incorrect, correct, done.

as per Fiori3 visual spec.

## Additional information

* Screenshots or code
![image](https://user-images.githubusercontent.com/53509521/103755919-a0ba0880-5034-11eb-9f73-cbbd09f6fba0.png)

* Notes
Each facet has composing components but also has differences or additional styles. 
Should we use the underlying components as a pattern but add the additional styles with their own element/modifier classes? Or should we just add these deltas as modifiers to each underlying component and use these components directly as a pattern?

kavya-b:
08/01- in progress

kavya-b:
11/01- in progress, few queries/doubts on implementing styles, mailed to designers.

kavya-b:
12/01- in progress, received clarifications to queries from designers. To be noted, images in the visual spec are not always correct, best to cross-refer against the following url for facets:
https://sapui5.hana.ondemand.com/#/entity/sap.uxap.HeaderFacetPattern/sample/sap.uxap.sample.Sample2

kavya-b:
13/01- in progress

[...]

PR #2077, opened 2021-01-22
feat: introduce Facets- Form, Key-value, Image, and Rating Indicator

kavya-b:
## Related Issue
Closes SAP/fundamental-styles#2020

## Description
This PR introduces Facets. Facets are used with Object Page and Dynamic Page usually, but the facet types can also be used independently. Therefore, we are introducing Facets as a component by itself, which consists of different facet types- form facet, key-value facet, image facet, and rating indicator facet.

It also introduce a couple of new classes in Dynamic Page to work with the image facet when used in the header.

## Screenshots
image

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/facet.scss
+++ b/src/facet.scss
@@ -4,144 +4,43 @@
 $block: #{$fd-namespace}-facet;
 
 .#{$block} {
-  $fd-facet-horizontal-spacing: 2rem;
-  $fd-facet-vertical-spacing: 1rem;
-  $fd-facet-title-vertical-spacing: 1rem;
-  $fd-form-kv-facet-horizontal-spacing: 0.5rem;
-  $fd-image-form-facet-bottom-spacing: 0.5rem;
-  $fd-image-facet-mobile-spacing: 1rem;
-  $fd-image-facet-header-horizontal-spacing: 1rem;
-  $fd-object-status-facet-font-size: 1.5rem;
-
-  @mixin facet-spacing() {
-    margin-right: $fd-facet-horizontal-spacing;
-    margin-bottom: $fd-facet-vertical-spacing;
-
-    @include fd-rtl() {
-      margin-right: 0;
-      margin-left: $fd-facet-horizontal-spacing;
-    }
-  }
-
-  @mixin image-facet-mobile-spacing() {
-    margin-right: $fd-image-facet-mobile-spacing;
-    margin-bottom: $fd-image-form-facet-bottom-spacing;
-
-    @include fd-rtl() {
-      margin-right: 0;
-      margin-left: $fd-image-facet-mobile-spacing;
-    }
-  }
-
   @include fd-reset();
 
-  @include fd-flex() {
-    flex-wrap: wrap;
-    align-items: flex-start;
-  }
-
-  &__container {
+  &-group {
     @include fd-reset();
-    @include facet-spacing();
 
-    &--image {
-      @include fd-flex();
-
-      // by itself, only right spacing is applied
-      margin-right: $fd-facet-horizontal-spacing;
-      margin-bottom: 0;
-
-      // when used in object/dynamic page header title, different spacings are applied
-      &-header-title {
-        margin-right: $fd-image-facet-header-horizontal-spacing;
-        margin-bottom: 0;
-        align-items: center;
-
-        @include fd-rtl() {
-          margin-right: 0;
-          margin-left: $fd-image-facet-header-horizontal-spacing;
-        }
-      }
-
-      // when used in object/dynamic page header content, different spacings are applied
-      &-header-content {
-        @media (min-width: 0) {
-          @include image-facet-mobile-spacing();
-        }
-
-        @media (min-width: 37rem) {
-          @include facet-spacing();
-        }
-      }
-
-      // when used in object/dynamic page header content in mobile
-      &-header-content-sm {
-        @include image-facet-mobile-spacing();
-      }
-    }
-
-    &--form {
-      @include fd-flex(column);
-    }
-
-    &--key-value {
-      .#{$block}__object-status {
-        font-family: var(--sapFontFamily);
-        font-size: $fd-object-status-facet-font-size;
-        // font-weight: var(--sapFontHeaderWeight);  // @sapFontHeaderWeight is not part of the current theming base content and is only available as part of UI5. Currently under evaluation.
-        font-weight: normal;
-
-        & > [class*=sap-icon] {
-          font-size: $fd-object-status-facet-font-size; // of icon
-          margin-right: $fd-form-kv-facet-horizontal-spacing;
-
-          @include fd-rtl() {
-            margin-right: 0;
-            margin-left: $fd-form-kv-facet-horizontal-spacing;
-          }
-        }
-      }
-
-      .#{$block}__object-number-text, .#{$block}__object-number-unit {
-        line-height: 1;
-        padding-top: 0;
-      }
-    }
-
-    &--rating-indicator {
-      display: inline-flex;
-      justify-content: flex-start;
+    @include fd-flex() {
+      flex-wrap: wrap;
       align-items: flex-start;
-      flex-direction: column;
     }
   }
 
-  &__title-container {
-    @include fd-reset();
-    @include fd-flex(column);
-
-    margin-bottom: $fd-facet-title-vertical-spacing;
-  }
-
-  &__title {
-    @include fd-reset();
-
-    font-size: var(--sapFontLargeSize);
-  }
-
-  &__content-container {
-    @include fd-reset();
+  &--image {
     @include fd-flex();
 
-    &--form {
-      margin-bottom: $fd-image-form-facet-bottom-spacing;
-
-      &:last-child {
-        margin-bottom: 0;
-      }
+    // when used in object/dynamic page header title, we need different alignment
+    &-header-title {
+      align-self: flex-end;
     }
+  }
 
-    &--rating-indicator {
+  &--form {
+    @include fd-flex(column);
+  }
+
+  &--key-value {
+    .#{$block}__object-status {
+      font-family: var(--sapFontFamily);
+    }
+  }
+
+  &--rating-indicator {
+    display: inline-flex;
+    justify-content: flex-start;
+    align-items: flex-start;
+    flex-direction: column;
+
+    .#{$block}__container {
       flex-direction: column;
 
       .#{$block}__rating-container {
@@ -150,9 +49,13 @@ $block: #{$fd-namespace}-facet;
 
       .#{$block}__rating-dynamic-text {
         align-self: flex-end;
-        margin-top: 0.5rem;
       }
     }
+  }
+
+  &__container {
+    @include fd-reset();
+    @include fd-flex();
   }
 
   &__form-value {
@@ -162,14 +65,5 @@ $block: #{$fd-namespace}-facet;
     font-size: var(--sapFontSize);
     font-weight: normal;
     color: var(--sapGroup_TitleTextColor);
-
-    .#{$block}__link {
-      margin-left: $fd-form-kv-facet-horizontal-spacing;
-
-      @include fd-rtl() {
-        margin-left: 0;
-        margin-right: $fd-form-kv-facet-horizontal-spacing;
-      }
-    }
   }
 }
--- a/src/facet.scss
+++ b/src/facet.scss
@@ -1,0 +1,171 @@
+@import "./new-settings";
+@import "./mixins";
+
+$block: #{$fd-namespace}-facet;
+
+.#{$block} {
+  $fd-facet-horizontal-spacing: 2rem;
+  $fd-facet-vertical-spacing: 1rem;
+  $fd-facet-title-vertical-spacing: 1rem;
+  $fd-form-kv-facet-horizontal-spacing: 0.5rem;
+  $fd-image-form-facet-bottom-spacing: 0.5rem;
+  $fd-image-facet-mobile-spacing: 1rem;
+  $fd-image-facet-header-horizontal-spacing: 1rem;
+  $fd-object-status-facet-font-size: 1.5rem;
+
+  @mixin facet-spacing() {
+    margin-right: $fd-facet-horizontal-spacing;
+    margin-bottom: $fd-facet-vertical-spacing;
+
+    @include fd-rtl() {
+      margin-right: 0;
+      margin-left: $fd-facet-horizontal-spacing;
+    }
+  }
+
+  @mixin image-facet-mobile-spacing() {
+    margin-right: $fd-image-facet-mobile-spacing;
+    margin-bottom: $fd-image-form-facet-bottom-spacing;
+
+    @include fd-rtl() {
+      margin-right: 0;
+      margin-left: $fd-image-facet-mobile-spacing;
+    }
+  }
+
+  @include fd-reset();
+
+  @include fd-flex() {
+    flex-wrap: wrap;
+  }
+
+  &__title-container {
+    @include fd-reset();
+    @include fd-flex(column);
+
+    margin-bottom: $fd-facet-title-vertical-spacing;
+  }
+
+  &__title {
+    @include fd-reset();
+
+    font-size: var(--sapFontLargeSize);
+  }
+
+  &__image {
+    @include fd-reset();
+    @include fd-flex();
+
+    @media (min-width: 0) {
+      @include image-facet-mobile-spacing();
+    }
+
+    @media (min-width: 37rem) {
+      @include facet-spacing();
+    }
+
+    // when used in object/dynamic page header
+    &--header {
+      margin-right: $fd-image-facet-header-horizontal-spacing;
+      margin-bottom: 0;
+      align-items: center;
+
+      @include fd-rtl() {
+        margin-right: 0;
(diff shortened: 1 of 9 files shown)


InnaAtanasova:
Hey @kavya-b I reviewed your PR but I have some points I want to discuss with my team. 
My biggest concern is that this component does not exist in Fiori 3 specs and I'm not sure we can introduce it in Fundamentals. It should prob go to thee `Layouts` section as it's not really a Pattern. 
The other thing I would improve is the structure of the classes. 
Your block element is `.fd-facet`. Then you can have a `.fd-facet__container` class where you bring the reset and the margins bottom and right. You can have a modifier class to overwrite the defaults depending on the element. For example, for rating indicator you will have:
```
.fd-facet
    .fd-facet__container.fd-facet__container--rating-indicator
```

For form it will be:
```
.fd-facet
    .fd-facet__container.fd-facet__container--form
```

src/facet.scss
```diff
@@ -0,0 +1,171 @@
...
+    @include facet-spacing();
+
+    display: inline-flex;
+    justify-content: flex-start;
+    align-items: flex-start;
+    flex-direction: column;
+  }
+
+  &__rating-indicator {
+    @include fd-reset();
```
InnaAtanasova, line 158: This is used like this `fd-rating-indicator fd-facet__rating-indicator` and if you reset it here, you will reset the styling coming from `fd-rating-indicator`

InnaAtanasova (changes requested):

stories/facets/facets.stories.js
```diff
@@ -0,0 +1,819 @@
...
+            </div>
+            <span class="fd-object-number fd-object-number--large fd-object-number--informative fd-facet__object-number">
+                <span class="fd-object-number__text fd-facet__object-number-text">579</span><span class="fd-object-number__unit fd-facet__object-number-unit">EUR</span>
+            </span>
+        </div>
+
+        <div class="fd-facet__rating-indicator-container" role="group" aria-labelledby="ratingFacetTitle1">
+            <div class="fd-facet__title-container">
+                <h1 class="fd-facet__title" id="ratingFacetTitle1">Average User Rating</h1>
+                <h2 class="fd-facet__subtitle fd-form-label">6 reviews</h2>
```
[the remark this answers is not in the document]
InnaAtanasova: +1 to this comment. There's no `fd-facet__subtitle` in the facets scss file.

kavya-b:
> Hey @kavya-b I wrote to the designer about this "component". Our next sync up meeting with them is on Feb.2nd. We discussed it today during our team meeting and we think it can remain a stand-alone component but it should be hidden as Numeric Content and Object Attribute and you can provide examples of it in Object Page and/or Dynamic Page. More information I will provide after the meeting with the designers.

I see. From what I understand now, we will continue having Facets as a component(incorporating above comments on the structure as well), but will not show it in the documentation pages. Instead, we would include it directly in the Dynamic Page example similar to [this](https://deploy-preview-2077--fundamental-styles.netlify.app/?path=/docs/components-facets--combined#facets-in-object-page). Is my understanding correct? Can I start working on making these changes?

cc: @manjunathanagaraj

stories/facets/facets.stories.js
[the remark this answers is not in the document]
kavya-b: thanks for noticing, this was not needed. 😅

kavya-b:
> The other thing I would improve is the structure of the classes.
> Your block element is `.fd-facet`. Then you can have a `.fd-facet__container` class where you bring the reset and the margins bottom and right. You can have a modifier class to overwrite the defaults depending on the element. For example, for rating indicator you will have:
> 
> ```
> .fd-facet
>     .fd-facet__container.fd-facet__container--rating-indicator
> ```
> 
> For form it will be:
> 
> ```
> .fd-facet
>     .fd-facet__container.fd-facet__container--form
> ```
> 
> This will make the angular implementation easier as the type of the component can be specified with an input parameter.

I was looking into this and I'm not sure I understand how much easier implementation would get making this change. Sure we can take `type` from user, say `'rating-indicator'`, but within that, we have the title, the subtitle/label, and then the actual rating indicator. And this syntax of title->label->content format is not the same across all facets. For example, for key-value facet, we only have a title and a value; and a form facet has title, label and corresponding value. So even after we take `type` from user, we would still need more specific information to map to the correct facet, right?

InnaAtanasova:
> > Hey @kavya-b I wrote to the designer about this "component". Our next sync up meeting with them is on Feb.2nd. We discussed it today during our team meeting and we think it can remain a stand-alone component but it should be hidden as Numeric Content and Object Attribute and you can provide examples of it in Object Page and/or Dynamic Page. More information I will provide after the meeting with the designers.
> 
> I see. From what I understand now, we will continue having Facets as a component(incorporating above comments on the structure as well), but will not show it in the documentation pages. Instead, we would include it directly in the Dynamic Page example similar to [this](https://deploy-preview-2077--fundamental-styles.netlify.app/?path=/docs/components-facets--combined#facets-in-object-page). Is my understanding correct? Can I start working on making these changes?
> 
> cc: @manjunathanagaraj

@kavya-b We have exactly the same case with Card component and Numeric content component. You can check [here](https://deploy-preview-2077--fundamental-styles.netlify.app/?path=/docs/components-card--card-anatomy#analytical-card). Numeric content is used in multiple other components but is not a stand-alone component and that's why is hidden from the documentation and examples of its usage are shown in Cards, Tiles, etc.

InnaAtanasova:
> > The other thing I would improve is the structure of the classes.
> > Your block element is `.fd-facet`. Then you can have a `.fd-facet__container` class where you bring the reset and the margins bottom and right. You can have a modifier class to overwrite the defaults depending on the element. For example, for rating indicator you will have:
> > ```
> > .fd-facet
> >     .fd-facet__container.fd-facet__container--rating-indicator
> > ```
> > 
> > 
> > For form it will be:
> > ```
> > .fd-facet
> >     .fd-facet__container.fd-facet__container--form
> > ```
> > 
> > 
> > This will make the angular implementation easier as the type of the component can be specified with an input parameter.
> 
> I was looking into this and I'm not sure I understand how much easier implementation would get making this change. Sure we can take `type` from user, say `'rating-indicator'`, but within that, we have the title, the subtitle/label, and then the actual rating indicator. And this syntax of title->label->content format is not the same across all facets. For example, for key-value facet, we only have a title and a value; and a form facet has title, label and corresponding value. So even after we take `type` from user, we would still need more specific information to map to the correct facet, right?

I don't know the details of the component but here's how I would approach it (this is just my point of view, don't insist on you agreeing with me :) )
I would make "universal classes" for the facet component:

```
.fd-facet
    .fd-facet__container
        .fd-facet__title
        .fd-facet__subtitle
        .fd-facet__header
        .fd-facet__label
        .fd-facet__content
        .fd-facet__something-else ......
```

There are would put the repeating (shared) css rules that are valid in all cases. Then I would add the modifier classes:
```
.fd-facet.fd-facet--rating-indicator
.fd-facet.fd-facet--form
.fd-facet.fd-facet--key-value
.fd-facet.fd-facet--something-else ......
```

and from there, based on the modifier class I would change the "children" of the facet:
```
.fd-facet.fd-facet--rating-indicator {
     .fd-facet__container {
         // code specific for the container of the rating indicator
     }

     .fd-facet__subtitle {
         // code specific for the subtitle of the rating indicator
     }
}
```
and
```
.fd-facet.fd-facet--form {
     .fd-facet__container {
         // code specific for the container of the form
     }

     .fd-facet__subtitle {
         // code specific for the subtitle of the form
     }
}
```

And it's ok if you have for example `.fd-facet__header` that is only part of rating indicator, the child doesn't have to be present in all components, but at least you can have a component structure that can live independently of other components and use the modifier classes for the specific needs.

That's how I would approach this component. Again, just a suggestion.

kavya-b:
> > > The other thing I would improve is the structure of the classes.
> > > Your block element is `.fd-facet`. Then you can have a `.fd-facet__container` class where you bring the reset and the margins bottom and right. You can have a modifier class to overwrite the defaults depending on the element. For example, for rating indicator you will have:
> > > ```
> > > .fd-facet
> > >     .fd-facet__container.fd-facet__container--rating-indicator
> > > ```
> > > 
> > > 
> > > For form it will be:
> > > ```
> > > .fd-facet
> > >     .fd-facet__container.fd-facet__container--form
> > > ```
> > > 
> > > 
> > > This will make the angular implementation easier as the type of the component can be specified with an input parameter.
> > 
> > 
> > I was looking into this and I'm not sure I understand how much easier implementation would get making this change. Sure we can take `type` from user, say `'rating-indicator'`, but within that, we have the title, the subtitle/label, and then the actual rating indicator. And this syntax of title->label->content format is not the same across all facets. For example, for key-value facet, we only have a title and a value; and a form facet has title, label and corresponding value. So even after we take `type` from user, we would still need more specific information to map to the correct facet, right?
> 
> I don't know the details of the component but here's how I would approach it (this is just my point of view, don't insist on you agreeing with me :) )
> I would make "universal classes" for the facet component:
> 
> ```
> .fd-facet
>     .fd-facet__container
>         .fd-facet__title
>         .fd-facet__subtitle
>         .fd-facet__header
>         .fd-facet__label
>         .fd-facet__content
>         .fd-facet__something-else ......
> ```
> 
> There are would put the repeating (shared) css rules that are valid in all cases. Then I would add the modifier classes:
> 
> ```
> .fd-facet.fd-facet--rating-indicator
> .fd-facet.fd-facet--form
> .fd-facet.fd-facet--key-value
> .fd-facet.fd-facet--something-else ......
> ```
> 
> and from there, based on the modifier class I would change the "children" of the facet:
> 
> ```
> .fd-facet.fd-facet--rating-indicator {
>      .fd-facet__container {
>          // code specific for the container of the rating indicator
>      }
> 
>      .fd-facet__subtitle {
>          // code specific for the subtitle of the rating indicator
>      }
> }
> ```
> 
> and
> 
> ```
> .fd-facet.fd-facet--form {
>      .fd-facet__container {
>          // code specific for the container of the form
>      }
> 
>      .fd-facet__subtitle {
>          // code specific for the subtitle of the form
>      }
> }
> ```
> 
> And it's ok if you have for example `.fd-facet__header` that is only part of rating indicator, the child doesn't have to be present in all components, but at least you can have a component structure that can live independently of other components and use the modifier classes for the specific needs.
> 
> That's how I would approach this component. Again, just a suggestion.

Thanks for your detailed feedback, @InnaAtanasova ! I came up with the following modified structure after taking some of your suggestions: 
image
so, we have 
```
.fd-facet //styles for when many facets are used together
     .fd-facet-container //common margins and fd-reset() for each facet container
        modifiers for .fd-facet-container:
        .fd-facet-container--form  //form-facet-specific container level styles
        .fd-facet-container--image //image-facet-specific container level styles
        .fd-facet-container--key-value  //key-value-facet-specific container level styles
        .fd-facet-container--rating-indicator  //rating-indicator-facet-specific container level styles

          .fd-facet__title-container //title and subtitle common classes
        .fd-facet__content-container with modifiers for fd-facet__content-container--form and fd-facet__content-container--rating-indicator since their contents also have specific styles.
       
```
We also have other classes like `.fd-facet__form-value, .fd-facet__object-status, .fd-facet__object-number, fd-facet__rating-container, fd-facet__rating-dynamic-text, .fd-facet-container--image-header ` that deal with very specific style changes(like style the form facet's value, or modify object status/number/rating indicator/image in object page header) and only take affect when used in the respective facet's context.
Let me know what you think; will push in the changes tomorrow.

kavya-b:
@InnaAtanasova @JKMarkowski 
- have addressed your comments in addition to changing the structure of the classes. 
- have also made Facets 'Dev' only so the stories won't be visible anymore on the documentation page. 
- have added example usage of facets in Dynamic Page documentation page.
Please check.

InnaAtanasova (commented):
1. The facet containers are all the same height. 
Fundamentals:
Screen Shot 2021-02-03 at 9 12 40 AM
Screen Shot 2021-02-03 at 9 12 29 AM

Fiori:
Screen Shot 2021-02-03 at 9 13 24 AM

2. Image/Avatar container has margin bottom, in Fiori only margin right is specified.
3. `fd-facet__form-value` is not specific to facets, it says in the specs it should follow Medium Text
4. I don't understand why there's a `fd-facet__title-container` and `fd-facet__title`. This can be a fundamentals title component with font-size of 1rem: `h5 class="fd-title fd-title--h5` and then add a modifier class to set the spacing

I see facets as containers that hold existing components (none of the components are specific to Facets) and just add spacings to these components. 
I will ask Manju to discuss this component during one of the components review meetings this week or next one.

src/facet.scss
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    &--form {
+      @include fd-flex(column);
+    }
+
+    &--key-value {
+      .#{$block}__object-status {
+        font-family: var(--sapFontFamily);
+        font-size: $fd-object-status-facet-font-size;
```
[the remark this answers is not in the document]
kavya-b: Object Status Large design does have 1.5rem font-size but the font-family and how the status is displayed with the label(horizontal or vertical) is also different:
image

image

the reason why extra styles were added to object number is because when used as is, the alignment was looking off:
image

kavya-b:
> 1. The facet containers are all the same height.
>    Fundamentals:
> 
> Screen Shot 2021-02-03 at 9 12 40 AM
> 
> Screen Shot 2021-02-03 at 9 12 29 AM
> 
> Fiori:
> Screen Shot 2021-02-03 at 9 13 24 AM
> 
Will check this.

> 1. Image/Avatar container has margin bottom, in Fiori only margin right is specified.

By itself, it should not have a margin bottom, but when used in Object/Dynamic page, there is a margin bottom specified:
image

> 2. `fd-facet__form-value` is not specific to facets, it says in the specs it should follow Medium Text

It does inherit from Medium text, but has overriding styles for the color:
image
spec:
image

> 3. I don't understand why there's a `fd-facet__title-container` and `fd-facet__title`. This can be a fundamentals title component with font-size of 1rem: `h5 class="fd-title fd-title--h5` and then add a modifier class to set the spacing
> 
 `fd-facet__title-container` was added mainly to support the subtitle for rating indicator facet. Sure I can remove the `__title` element and use the title component with a modifier(modifier to title component like `fd-title--facet-spacing`?) but I think I would still need the `__title-container` element for spacing the rating indicator facet and setting the flex-direction on it:
image

> I see facets as containers that hold existing components (none of the components are specific to Facets) and just add spacings to these components.
> I will ask Manju to discuss this component during one of the components review meetings this week or next one.

kavya-b:
1.  After incorporating margins/paddings and aligning to the new dynamic page structure, the facet structure has now changed/minimized to this:

- `fd-facet-group`
  -   `fd-facet fd-facet(--image or --form or --key-value or --rating-indicator)`
      -   `fd-facet__container` to give some styles to the content of a facet, like adding the correct `flex-direction`
          -    `fd-facet__rating-container` and `fd-facet__rating-dynamic-text` - related to alignment in rating indicator facet


2. Removed `fd-facet__title` and replaced with `fd-title` instead since the font-size values `--sapFontLargeSize` and `--sapFontHeader5Size` evaluate to the same value.
3. Removed `fd-facet__form-value` and used `fd-text` instead since the colors `--sapTextColor` and `--sapGroup_TitleTextColor` resolve to the same value. Please let me know if I should match the variable name too; in that case I would need a separate class for this.
4. Image facet by itself should not have a margin bottom, but when used in Object/Dynamic page, there is a margin bottom specified:
image 
That's why you will see a bottom margin in the documentation.

@InnaAtanasova @JKMarkowski please take a look again.

src/facet.scss
```diff
@@ -0,0 +1,71 @@
...
+  &--form {
+    @include fd-flex(column);
+  }
+
+  &--key-value {
+    .#{$block}__object-status {
+      .#{$block}__object-status-text {
+        font-family: var(--sapFontFamily);
+      }
+    }
```
JKMarkowski, line 36: Can this be simplified to
```suggestion
  &--key-value {
      .#{$block}__object-status-text {
        font-family: var(--sapFontFamily);
      }
```
?

src/facet.scss
```diff
@@ -0,0 +1,71 @@
...
+
+      .#{$block}__rating-container {
+        align-self: flex-start;
+      }
+
+      .#{$block}__rating-dynamic-text {
+        align-self: flex-end;
+      }
+    }
+  }
```
JKMarkowski, line 56: Can this structure be flatted? 
```suggestion
    .#{$block}__container {
      flex-direction: column;
      }

      .#{$block}__rating-container {
        align-self: flex-start;
      }

      .#{$block}__rating-dynamic-text {
        align-self: flex-end;
      }
  }
```

JKMarkowski (commented):
Hi @kavya-b I added 2 comments, rest LGTM, let me know what you think about it, or whether it's addressed, so I will approve PR

src/facet.scss
kavya-b: Unfortunately, no. I had tried this, but object status component takes higher specificity and its `font-family` gets applied instead.

src/facet.scss
kavya-b: fixed as suggested, please take a look @JKMarkowski , thanks!

InnaAtanasova (approved):

JKMarkowski (approved):
LGTM

PR #2078, opened 2021-01-22
docs: replace master with main

### LINKED ISSUE #2050: Inclusive Language
Тhe terms master/slave and whitelist/blacklist (and only those) must be replaced in external content, including UIs. When UIs are changed, the accompanying documentation needs to be updated as well (the UADs are aware of the topic). In a second step, further related content like tutorials, slide decks etc. should be updated as well. 

Estimation: 0.5day

mikerodonnell89:
fixes #2050


src/icons/_settings.scss
```diff
@@ -606,8 +606,8 @@ $fd-icons: (
   circle-task-2: "\e255",
   project-definition-triangle: "\e256",
   project-definition-triangle-2: "\e257",
-  master-task-triangle: "\e258",
-  master-task-triangle-2: "\e259",
+  main-task-triangle: "\e258",
```
droshev, line 609: this is not own by us. We should raise it to SAP UX as they own the icons.
We can remove the support

droshev:
@mikerodonnell89 all those changes come from design. It has to be addressed there first.

src/icons/_settings.scss
mikerodonnell89: Didn't mean to stage these

InnaAtanasova (approved):
Changed the PR title to `docs` as it only touches the documentation part.

stories/list/standard/standard-list.stories.js
```diff
@@ -2,7 +2,7 @@ export default {
     title: 'Components/List/Standard',
     parameters: {
         description: `
-In SAP Fiori, we distinguish between tables and lists. Both usually contain homogeneous data, but lists generally have rather basic data, while the data in tables tend to be more complex. Lists are mostly used in the master list for a master-detail scenario using the flexible column layout, as well as in **Popovers** or **Dialogs**.
+In SAP Fiori, we distinguish between tables and lists. Both usually contain homogeneous data, but lists generally have rather basic data, while the data in tables tend to be more complex. Lists are mostly used in the main list for a main-detail scenario using the flexible column layout, as well as in **Popovers** or **Dialogs**.
```
droshev, line 5: i think the design doc should evolve before we do so

InnaAtanasova:
Closing the PR, keeping the branch

PR #2084, opened 2021-01-25
feat: introduce expand example for the Text component

EvilAlexei:
## Description
After Core: Text [spec](https://app.gitbook.com/@fundamental-ngx/s/docs/core-text) review
it was decided to add expand option, "MORE" and "LESS" links to show more and less of the text.
The new example was based on similar functionality in [feed-list-items](https://sap.github.io/fundamental-ngx/#/core/feed-list-item#feed-list-item-simple)

#### Feed list items example
feed-list-items-example

#### Text expand example
text-expand-example

- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/text.scss
+++ b/src/text.scss
@@ -24,4 +24,11 @@ $block: #{$fd-namespace}-text;
   &--hyphenation {
     hyphens: auto;
   }
+
+  &__link {
+    &--more {
+      text-transform: uppercase;
+      cursor: pointer;
+    }
+  }
 }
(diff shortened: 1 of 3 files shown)


olshansky (approved):

mykolavorobiov (approved):

InnaAtanasova (approved):

InnaAtanasova:
You need `[ci visual]` in your commit to trigger Chromatic

PR #2087, opened 2021-01-26
fix: added color accents to Overflow Button in Avatar Group

olshansky:
## Related Issue
Related to https://github.com/SAP/fundamental-styles/issues/2023

- [x] All values are in `rem`
- [x] RTL support
3. Testing 
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/avatar-group.scss
+++ b/src/avatar-group.scss
@@ -32,6 +32,18 @@ $block: #{$fd-namespace}-avatar-group;
     "xl": ("ratio": ($fd-const-ratio * 7), "font-size": 3rem)
   );
   $fd-focus-outline-offset: 0.0625rem;
+  $fd-avatar-group-more-button-accent-colors: (
+    "1": ("background-color": var(--sapAccentColor1)),
+    "2": ("background-color": var(--sapAccentColor2)),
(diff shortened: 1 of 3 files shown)


mykolavorobiov (approved):

EvilAlexei (approved):

InnaAtanasova:
I can't request changes for this as I'm also not sure if it's a bug or not :D 
Here's what I noticed:
When you apply accent color, the border of the overflow button remains the same (blue)
Screen Shot 2021-01-26 at 9 42 09 AM
Screen Shot 2021-01-26 at 9 42 18 AM
Screen Shot 2021-01-26 at 9 42 01 AM

Here's the design:
Screen Shot 2021-01-26 at 9 45 38 AM

But I couldn't find info about it. Need to ask the designers about it. I will open an issue and add it here, but the PR as it is IMO can be merged.

InnaAtanasova:
https://github.com/SAP/fundamental-styles/issues/2091

InnaAtanasova (approved):

InnaAtanasova:
You need a commit containing `[ci visual]` to trigger chromatic

Issue #2089, opened 2021-01-26
Defect Hunting for Sprint 54 - Los Angeles


InnaAtanasova:
Please record the issues found during the defect hunting as a thread to this issue.

mikerodonnell89:
Would be great to get this fixed as the ngx fix is coming next sprint: https://github.com/SAP/fundamental-styles/issues/2083

mikerodonnell89:
- [ ] Split button focus weirdness  ![Screen Shot 2021-01-26 at 5 12 24 PM](https://user-images.githubusercontent.com/2471874/105923485-aa160f80-5ff9-11eb-82fe-dc6956dbdf7f.png)

JKMarkowski:
Initial Focus Trap doesn't work on popover 
![Peek 2021-01-27 12-28](https://user-images.githubusercontent.com/26483208/105985009-3298bc80-609b-11eb-930a-b61e1ba523f9.gif)

mikerodonnell89:
- [ ] Floating footer bar should have a z-index - in this example the input is placed above the footer - Mike
![Screen Shot 2021-01-27 at 3 14 49 PM](https://user-images.githubusercontent.com/2471874/106061259-66cba780-60b2-11eb-979f-fd9c5fe15ad7.png)

PR #2094, opened 2021-01-27
fix: positioning of overflowed shellbar notification counter

mikerodonnell89:
part of #2089 

The shellbar notifications, when moved into the overflow menu on small screens, were not positioned correctly.  This was not known in the styles library because we had only seen it in ngx thus far.  This PR fixes the positioning and adds an example

Before:
Screen Shot 2021-01-27 at 11 44 40 AM

After:
Screen Shot 2021-01-27 at 11 37 22 AM

### DIFF
--- a/src/shellbar.scss
+++ b/src/shellbar.scss
@@ -162,6 +162,12 @@ $block: #{$fd-namespace}-shellbar;
       vertical-align: middle;
     }
 
+    .#{$block}__counter--overflowed {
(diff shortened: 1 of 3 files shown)


JKMarkowski:
Hi @mikerodonnell89, code looks good for me, but could you add commit with `[ci visual]`?

InnaAtanasova (approved):

PR #2101, opened 2021-01-29
fix: add truncation for List group header text

InnaAtanasova:
## Related Issue
Part of SAP/fundamental-styles#2063

## Description
This PR adds truncation rules for the List component Group Header <li> element

BREAKING CHANGE:
change in the markup, now the text is wrapped in a span with `fd-list__title` class.
Affected components: combobox-input, standard list, multi combobox, multi input, select

Before:
```
<li role="listitem" class="fd-list__group-header">
    Group header 1
</li>
```
Now:
```
<li role="listitem" class="fd-list__group-header">
    <span class="fd-list__title">Group header 1</span>
 </li>
```

### Before:
Screen Shot 2021-01-29 at 11 39 19 AM

### After:
Screen Shot 2021-01-29 at 11 39 37 AM

- [x] All values are in `rem`
- [na] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/_listBase.scss
+++ b/src/_listBase.scss
@@ -168,10 +168,13 @@
     border-bottom: var(--sapList_BorderWidth) solid var(--sapList_GroupHeaderBorderColor);
     color: var(--sapList_GroupHeaderTextColor);
     align-items: flex-end;
-    font-size: $fd-list-normal-font-size;
-    font-weight: bold;
     height: 2.75rem;
-    line-height: 2rem;
+
+    .#{$block}__title {
+      font-size: $fd-list-normal-font-size;
+      font-weight: bold;
+      line-height: 2rem;
+    }
   }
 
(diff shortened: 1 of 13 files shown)


JKMarkowski (approved):

JKMarkowski (changes requested):
Hi @InnaAtanasova  Could you apply these changes in every occurence of `fd-list__group-header`? I see it's missing in select/combobox/multi input (This is the moment when chromatic is useful)

InnaAtanasova:
> Hi @InnaAtanasova Could you apply these changes in every occurence of `fd-list__group-header`? I see it's missing in select/combobox/multi input (This is the moment when chromatic is useful)

Updated

JKMarkowski (approved):
LGTM

Issue #2097, opened 2021-01-28
Card - Unwanted hover style on card header


mengbingguo:
**Describe the bug**
There is hover style change on the card header, even if no interaction is set on the card header.

> Is this issue related to a specific component?

Card

> What version of the Fundamental Library Styles are you using?

0.15.0

> What front-end framework are you implementing in, e.g., Angular, React, etc.?
Angular, React

**To Reproduce**
https://sap.github.io/fundamental-styles/?path=/docs/components-card--card-anatomy

**Expected behavior**
Should have the possibility to turn on/off card header interactibility.

Non-interactive card header example in SAPUI5:
https://sapui5.hana.ondemand.com/#/entity/sap.f.Card/sample/sap.f.sample.Card

- Background color shouldn't change on hover
- Cursor should be `default`, unless hovering over text in header, then it should be the `text` cursor

**Screenshots**

https://user-images.githubusercontent.com/75482561/106125558-99ca7580-615c-11eb-81a6-b5aaa6a90225.mov

droshev:
Action item:
- [ ] create examples for cards with header that navigates to something and cards with a header that doesn't do so. Make sure to use correct card type examples

PR #2113, opened 2021-02-02
feat: Add non-interactive state to card header

JKMarkowski:
## Related Issue
Closes https://github.com/SAP/fundamental-styles/issues/2097

## Description
There is added new `fd-card__header--non-interactive` class modifier. And description of its usage.

- [x] `fd-reset()` mixin is applied to all element
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/card.scss
+++ b/src/card.scss
@@ -41,27 +41,11 @@ $fd-card-content-avatar-text-spacing: 0.5rem !default;
     @include fd-reset();
     @include fd-flex();
 
-    @include fd-hover() {
-      background: var(--sapTile_Hover_Background);
-    }
-
-    @include fd-active() {
-      background: var(--sapTile_Active_Background);
-    }
-
-    @include fd-focus() {
-      outline-style: dotted;
-      outline-width: var(--sapContent_FocusWidth);
-      outline-color: var(--sapContent_FocusColor);
-      outline-offset: -0.125rem;
-    }
-
     padding: 1rem;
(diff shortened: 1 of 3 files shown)


src/card.scss
```diff
@@ -41,6 +42,13 @@ $fd-card-content-avatar-text-spacing: 0.5rem !default;
     @include fd-reset();
     @include fd-flex();
 
+    padding: 1rem;
+    background: $fd-card-default-body-background-color;
+    border-bottom: $fd-card-header-border;
+    border-radius: $fd-card-header-border-radius $fd-card-header-border-radius 0 0;
+    text-decoration: none;
+    cursor: pointer;
+
```
InnaAtanasova, line 51: why did you move these rules here?
not that it's a big deal but usually all @include statements are at the beginning (except rtl) and after that come the normal css rules

JKMarkowski: It's done because it is totally `fd-hover/active/selected` are totally different mixins than `flex/reset`. I treat them same as `&:hover/active/selected`. That's why those properties have been moved up

mikerodonnell89 (approved):

InnaAtanasova (approved):

src/card.scss
InnaAtanasova: of course they are different. But the order of the css should not be important :) it's just a common practice in scss world that the include statements are at the beginning.

PR #2115, opened 2021-02-02
fix: grid list styles

Betrozov:
## Related Issue
Closes https://github.com/SAP/fundamental-ngx/issues/4560

## Description
Fixed label focus style and added overflow for item

- [X] All values are in `rem`
- [X] RTL support
3. Testing
- [X] Updated tests
4. Documentation

### DIFF
--- a/src/list-grid.scss
+++ b/src/list-grid.scss
@@ -53,8 +53,6 @@ $block: #{$fd-namespace}-grid-list;
       }
 
       + .#{$block}__radio-label {
-        @include fd-fiori-focus(0);
-
         @include fd-empty() {
           padding: 0;
         }
--- a/src/list-grid.scss
+++ b/src/list-grid.scss
@@ -39,6 +39,7 @@ $block: #{$fd-namespace}-grid-list;
     box-shadow: $fd-grid-list-item-box-shadow;
     height: 100%;
     position: relative;
+    overflow: hidden;
 
     .#{$block}__radio-input {
       height: 1rem;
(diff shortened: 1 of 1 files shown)


IlyaSurmay (approved):

InnaAtanasova (approved):

InnaAtanasova:
You need a commit message containing `[ci visual]` to trigger Chromatic

src/list-grid.scss
```diff
@@ -47,10 +48,13 @@ $block: #{$fd-namespace}-grid-list;
       @include fd-focus() {
         + .#{$block}__radio-label {
           outline-offset: 0;
+          outline: var(--sapContent_FocusWidth) var(--sapContent_FocusStyle) var(--sapContent_FocusColor);
         }
       }
 
       + .#{$block}__radio-label {
+        @include fd-fiori-focus(0);
+
```
JKMarkowski, line 57:
```suggestion
      + .#{$block}__radio-label {
        @include fd-fiori-focus(0);
```

JKMarkowski (commented):
Hello @Betrozov I think there is a lot of unnecessary code, check my suggestion - `input[type=radio]` element is never visible

JKMarkowski (changes requested):
Check my comment above

src/list-grid.scss
```diff
@@ -47,10 +48,13 @@ $block: #{$fd-namespace}-grid-list;
       @include fd-focus() {
         + .#{$block}__radio-label {
           outline-offset: 0;
+          outline: var(--sapContent_FocusWidth) var(--sapContent_FocusStyle) var(--sapContent_FocusColor);
         }
       }
 
       + .#{$block}__radio-label {
+        @include fd-fiori-focus(0);
```
JKMarkowski, line 56: So this can be removed

JKMarkowski (approved):

Seamoo13 (approved):

PR #2118, opened 2021-02-03
fix: wizard - all stacked steps last:children should have padding-right

### LINKED ISSUE #2117: Wizard - topmost stacked step indicator cannot have completed styling
Working on the summary page for the wizard on NGX, I realized that the topmost step indicator on the furthest right cannot be set to status "completed", or it runs off the side of the page like so:

Should be able to set the fd-wizard__step--completed class on this step indicator. Applying the "upcoming" class instead places the indicator in the correct spot but of course is colored incorrectly.

1d

mikerodonnell89:
fixes #2117

### DIFF
--- a/src/wizard.scss
+++ b/src/wizard.scss
@@ -154,20 +154,21 @@ $fd-wizard-stacked-upcoming-step-offset: 2.375rem !default;
       width: 0.25rem;
       min-width: auto;
 
+      &:last-child {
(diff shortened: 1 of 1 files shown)


JKMarkowski (approved):
LGTM

InnaAtanasova (approved):

Issue #2091, opened 2021-01-26
Avatar Group Overflow Button border color design question


InnaAtanasova:
What is the border color of the Overflow button when semantics colors are applied?

olshansky:
> Regarding the spacing around Avatars when placed inside a Popover body (attached image) here’s what they answered:
> The Popover body should have an inner padding of 1 rem and the spacing between the Avatars should be 0.5rem.
> 
> For the overflow button which has a blue border in its default state, when accent colors are applied, the border should have the same color as the accent color.

PR #2121, opened 2021-02-05
feat: Avatar Group overflow spacing and overflow button border color

olshansky:
## Related Issue
Closes https://github.com/SAP/fundamental-styles/issues/2091
Part of https://github.com/SAP/fundamental-styles/issues/2153

## Description
Added overflow spacing and overflow button accent border. 

- [x] All values are in `rem`
- [x] RTL support
3. Testing 
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/avatar-group.scss
+++ b/src/avatar-group.scss
@@ -81,16 +81,14 @@ $block: #{$fd-namespace}-avatar-group;
     }
   }
 
-  &__overflow {
+  &__overflow--body {
     @include fd-reset();
 
-    &__body {
-      padding: 1rem;
-      white-space: initial;
+    padding: 1rem;
+    white-space: initial;
 
-      .#{$block}__item {
-        margin: 0.25rem;
-      }
+    .#{$block}__item {
+      margin: 0.25rem;
     }
   }
 
--- a/src/avatar-group.scss
+++ b/src/avatar-group.scss
@@ -47,6 +47,8 @@ $block: #{$fd-namespace}-avatar-group;
 
   @include fd-reset();
 
+  display: flex;
+
   &--individual-type {
     .#{$block}__item {
       @include fd-fiori-focus($fd-focus-outline-offset);
@@ -81,7 +83,7 @@ $block: #{$fd-namespace}-avatar-group;
     }
   }
 
-  &__overflow--body {
+  &__overflow-body {
     @include fd-reset();
 
     padding: 1rem;
--- a/src/avatar-group.scss
+++ b/src/avatar-group.scss
@@ -2,7 +2,7 @@
 @import "./avatar";
 
 /*!
-.fd-avatar-group+((--individual|--group), (--xxs|--xs|--s|--m|--l|--xl|--xxl)
+.fd-avatar-group+((--individual|--group), (--xs|--s|--m|--l|--xl)
 */
 $block: #{$fd-namespace}-avatar-group;
 
@@ -81,6 +81,19 @@ $block: #{$fd-namespace}-avatar-group;
     }
   }
 
+  &__overflow {
+    @include fd-reset();
+
+    &__body {
+      padding: 1rem;
+      white-space: initial;
+
+      .#{$block}__item {
+        margin: 0.25rem;
+      }
+    }
+  }
+
   @each $set-name, $size-set in $fd-avatar-group-more-button-sizes {
     &--#{$set-name} {
       .#{$block}__more-button {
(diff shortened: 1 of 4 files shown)


src/avatar-group.scss
```diff
@@ -81,6 +81,19 @@ $block: #{$fd-namespace}-avatar-group;
     }
   }
 
+  &__overflow {
+    @include fd-reset();
+
+    &__body {
```
mikerodonnell89, line 87: results in `fd-avatar-group__overflow__body` which is not valid BEM, `fd-avatar-group__overflow-body` is fine

mikerodonnell89 (changes requested):

src/avatar-group.scss
```diff
@@ -81,6 +81,19 @@ $block: #{$fd-namespace}-avatar-group;
     }
   }
 
+  &__overflow {
```
InnaAtanasova, line 84: why do you need this? It's used like this `<div class="fd-popover fd-avatar-group__overflow">` and it already has a reset coming from `fd-popover`. If you inspect the element, all of the rules are overwritten:
Screen Shot 2021-02-05 at 2 23 54 PM

You can only use `fd-avatar-group__overflow-body` where you apply the 1rem padding and use the specificity to add margin to the item.

InnaAtanasova (changes requested):

src/avatar-group.scss
olshansky: @mikerodonnell89 fixed

src/avatar-group.scss
olshansky: @InnaAtanasova fixed

src/avatar-group.scss
```diff
@@ -81,6 +81,17 @@ $block: #{$fd-namespace}-avatar-group;
     }
   }
 
+  &__overflow--body {
```
InnaAtanasova, line 84: should be `&__overflow-body`

InnaAtanasova: `__overflow--body` means that there's a class `__overflow` and `--body` is modifying this class

mikerodonnell89 (changes requested):
High contrast colors for the "+8" in group mode are incorrect

Screen Shot 2021-02-10 at 11 20 57 AM
Screen Shot 2021-02-10 at 11 21 56 AM

Also, the margin-right in individual mode is correctly set to 0.125rem, but there is some additional space before the next avatar so I think they may still be spread too far apart
Screen Shot 2021-02-10 at 11 24 49 AM

olshansky:
> High contrast colors for the "+8" in group mode are incorrect
> 
> Screen Shot 2021-02-10 at 11 20 57 AM
> Screen Shot 2021-02-10 at 11 21 56 AM

@mikerodonnell89 why it's incorrect? The button has a `color-accent-6` and as I see avatar has the same colors in HC mode 

> Also, the margin-right in individual mode is correctly set to 0.125rem, but there is some additional space before the next avatar so I think they may still be spread too far apart

fixed

src/avatar-group.scss
olshansky: @InnaAtanasova fixed

InnaAtanasova (approved):

mikerodonnell89:
> @mikerodonnell89 why it's incorrect? The button has a `color-accent-6` and as I see avatar has the same colors in HC mode

@olshansky screenshot from the specs: 
Screen Shot 2021-02-16 at 6 36 16 PM

So I think we will need a new style for this overflow avatar with the + text

olshansky:
> @olshansky screenshot from the specs:
> Screen Shot 2021-02-16 at 6 36 16 PM
> 
> So I think we will need a new style for this overflow avatar with the + text

@mikerodonnell89 screenshot from the examples (I just removed accent color class modifier). Looks correct, isn't it? 
FYI this button has a color different color accent in examples just to show that we could use color accents modifier classes for this button.
image

mikerodonnell89 (approved):

InnaAtanasova:
Screen Shot 2021-02-18 at 3 22 37 PM


@olshansky are these changes intended?

olshansky:
> @olshansky are these changes intended?

@InnaAtanasova old service for placeholder images don't work _(that's why I changed it at the [last commit](https://github.com/SAP/fundamental-styles/pull/2121/commits/77ec69df46c09c331b60317380f9296020ebea7c))_

droshev:
@olshansky can you address https://github.com/SAP/fundamental-styles/issues/2153 as part of this ticket?

olshansky:
@droshev should it be part of this PR? Because it is an issue in the popover component 
image

droshev:
> @droshev should it be part of this PR? Because it is an issue in the popover component
> image

what is the problem with the popover?

olshansky:
> what is the problem with the popover?

Popover header border color is different than the body border in Dark theme

droshev:
> > what is the problem with the popover?
> 
> @droshev Popover header border color is different than the body border in Dark theme

my comment was related to the arrow in RTL - it wasn't centered

olshansky:
> my comment was related to the arrow in RTL - it wasn't centered
> 
> https://user-images.githubusercontent.com/4380815/108005485-d86a8780-6fc6-11eb-9f0a-d0d91f6d454f.png

Oh yes, fixed

Issue #2126, opened 2021-02-08
Quick-view: Header title should be centre aligned


DeepakSap14:
**Describe the bug**

Currently header title is not centre aligned. but it should be centre aligned as per fiori spec.

current implementation:

<img width="561" alt="Screenshot 2021-02-08 at 11 11 53 AM" src="https://user-images.githubusercontent.com/57661487/107180358-8cd04080-69fe-11eb-95b2-73c87c24207d.png">

Fiori spec:
<img width="1161" alt="Screenshot 2021-02-08 at 11 02 20 AM" src="https://user-images.githubusercontent.com/57661487/107180379-9659a880-69fe-11eb-9afe-d0414273cd92.png">

> Is this issue related to a specific component?
Quick view

> What version of the Fundamental Library Styles are you using?
v0.16.0-rc.2

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

> What front-end framework are you implementing in, e.g., Angular, React, etc.?

**To Reproduce**
1. Go to https://fundamental-styles.netlify.app/?path=/docs/components-quick-view--popover

**Expected behavior**
Header title should be centre aligned

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

DeepakSap14:
This is not a bug, but "fundamental-ngx" documentation needs small correction.
this issue can be closed.

PR #2127, opened 2021-02-08
docs: centered Quick View title

olshansky:
## Related Issue
Closes https://github.com/SAP/fundamental-styles/issues/2126

- [n/a] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/stories/quick-view/__snapshots__/quick-view.stories.storyshot
+++ b/stories/quick-view/__snapshots__/quick-view.stories.storyshot
@@ -281,7 +281,7 @@ exports[`Storyshots Components/Quick View Popover 1`] = `
         
                 
         <div
-          class="fd-bar__left"
+          class="fd-bar__middle"
         >
           
                     
--- a/stories/quick-view/quick-view.stories.js
+++ b/stories/quick-view/quick-view.stories.js
@@ -13,7 +13,7 @@ export const Popover = () => `<div class="fd-popover">
     <div class="fd-popover__body" aria-hidden="false" id="popoverA1">
         <div class="fd-quick-view">
             <div class="fd-bar fd-bar--header">
-                <div class="fd-bar__left">
+                <div class="fd-bar__middle">
                     <div class="fd-bar__element">Company</div>
                 </div>
             </div>


DeepakSap14:
@olshansky 
if this is only docs change here in styles, then probably component in "fundamental-ngx" needs fix.
currently quick-view has header which is not centred in middle.

screenshot from "fundamental-ngx".
Screenshot 2021-02-08 at 6 00 29 PM

olshansky:
> @olshansky
> if this is only docs change here in styles, then probably component in "fundamental-ngx" needs fix.
> currently quick-view has header which is not centred in middle.
> 
> screenshot from "fundamental-ngx".

@DeepakSap14 It's not really a bug, it's just using incorrect classes to align the title.
Also, in fundamental-ngx it's just an example _(example from your screenshot has been created according to [this example](https://sapui5.hana.ondemand.com/#/entity/sap.m.QuickView/sample/sap.m.sample.QuickView))_. `<fd-quick-view-title align="left">{{ data.title }}</fd-quick-view-title>` as you can see, the title in the fundamental-ngx has `align` property which can help align a title to a proper place.
Also, fundamental-ngx needs bumped version of fundamental-styles for correct gaps.

DeepakSap14:
> > @olshansky
> > if this is only docs change here in styles, then probably component in "fundamental-ngx" needs fix.
> > currently quick-view has header which is not centred in middle.
> > screenshot from "fundamental-ngx".
> 
> @DeepakSap14 It's not really a bug, it's just using incorrect classes to align the title.
> 
> Also, in fundamental-ngx it's just an example _(example from your screenshot has been created according to [this example](https://sapui5.hana.ondemand.com/#/entity/sap.m.QuickView/sample/sap.m.sample.QuickView))_. `<fd-quick-view-title align="left">{{ data.title }}</fd-quick-view-title>` as you can see, the title in the fundamental-ngx has `align` property which can help align a title to a proper place.
> 
> Also, fundamental-ngx needs bumped version of fundamental-styles for correct gaps.

Okay..i missed to see the html code.
yes it's not a bug then.
May be, you can correct the doc in "fundamental-ngx". 
I am happy to close this bug in "fundamental-styles".

Betrozov (approved):

JKMarkowski (approved):

Issue #2119, opened 2021-02-04
Introduce Margins and Paddings


InnaAtanasova:
Introduce helper modifier classes that can be added to ANY element that needs some space. This functionality can be exposed in the documentation under the Layout section. 
More info to be provided soon

PR #2131, opened 2021-02-10
feat: introduce Margins and Paddings

kavya-b:
## Related Issue
Closes SAP/fundamental-styles#2119

## Description
This PR introduces helper modifier classes that can be added to any element that needs some spacings through paddings or margins.

### Before:
NA
### After:
Margins:
image
Paddings:
image

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/.storybook/custom/custom.scss
+++ b/.storybook/custom/custom.scss
@@ -379,6 +379,14 @@ code.docs-code-grey {
     height: 100%;
   }
 
+  &--padded {
+    padding: 1.5rem;
+  }
+
+  &--margin {
+    margin: 1.5rem;
+  }
+
   &--color-1 {
     background-color: #30c5d2;
     color: white;
@@ -608,5 +616,14 @@ body.sb-show-main {
   .card-six-dimensions {
     width: 20rem;
     height: 20rem;
+  }
+}
+
+.docs-column-flex {
+  display: flex;
+  flex-direction: column;
+
+  &--inline {
+    display: inline-flex;
   }
 }
--- a/.storybook/custom/custom.scss
+++ b/.storybook/custom/custom.scss
@@ -379,12 +379,8 @@ code.docs-code-grey {
     height: 100%;
   }
 
-  &--padded {
-    padding: 1.5rem;
-  }
-
-  &--margin {
-    margin: 1.5rem;
+  &--border {
+    border: 2px solid #000000;
   }
 
   &--color-1 {
@@ -623,8 +619,60 @@ body.sb-show-main {
   display: flex;
   flex-direction: column;
 
-  &--inline {
-    display: inline-flex;
+  &--align-start {
+    align-items: flex-start;
+  }
+}
+
+.docs-layout-margins-paddings {
+  background-color: #3d6299;
+  color: #ffffff;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  min-height: 4rem;
+  text-align: center;
+
+  &--color-13 {
+    background-color: #3d6299;
+    color: white;
+    border: 2px solid red;
+  }
+
+  &--color-14 {
+    background-color: #30c5d2;
+    color: white;
+    border: 2px solid black;
+  }
+
+  &--secondary-bg {
+    background-color: #dbdbdb;
+    display: inline-block;
+  }
+
+  &--padded {
+    padding: 1.5rem;
+  }
+
+  &--margin {
+    margin: 1.5rem;
+  }
+
+  // padding+width = actual width, so specifying different widths to show the backgrounds correctly
+  &--width {
+    width: 19rem; // for tiny 0.5rem each(horizontal padding)+19rem = 20rem total
+
+    &-sm {
+      width: 18rem; // for tiny 1rem each(horizontal padding)+18rem = 20rem total
+    }
+
+    &-md {
+      width: 16rem; // for tiny 2rem each(horizontal padding)+16rem = 20rem total
+    }
+
+    &-lg {
+      width: 14rem; // for tiny 3rem each(horizontal padding)+14rem = 20rem total
+    }
   }
 }
 
--- a/.storybook/custom/custom.scss
+++ b/.storybook/custom/custom.scss
@@ -627,3 +627,8 @@ body.sb-show-main {
     display: inline-flex;
   }
 }
+
+.fddocs-margin-end-none,
+[dir="rtl"] .fddocs-margin-end-none {
+  margin: 1.5rem;
+}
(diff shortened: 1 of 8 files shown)


kavya-b:
One thing I am not sure of is RTL handling. For the case of `No Margins`, if we have a margin of say 1.5rem initially, `fd-margin-end--none` will give `margin-right: 0 !important;` which is expected, but switching to RTL makes it `margin-left: 0 !important;`, but there is no way to actually retrieve the initial value of 1.5rem that the user set.
image
in RTL becomes 
image
because the `margin-right` value was already set to 0, and the original value is lost. How do we handle this?

InnaAtanasova:
> One thing I am not sure of is RTL handling. For the case of `No Margins`, if we have a margin of say 1.5rem initially, `fd-margin-end--none` will give `margin-right: 0 !important;` which is expected, but switching to RTL makes it `margin-left: 0 !important;`, but there is no way to actually retrieve the initial value of 1.5rem that the user set.
> image
> in RTL becomes
> image
> because the `margin-right` value was already set to 0, and the original value is lost. How do we handle this?

would `margin-right: initial;` work?

kavya-b:
> > One thing I am not sure of is RTL handling. For the case of `No Margins`, if we have a margin of say 1.5rem initially, `fd-margin-end--none` will give `margin-right: 0 !important;` which is expected, but switching to RTL makes it `margin-left: 0 !important;`, but there is no way to actually retrieve the initial value of 1.5rem that the user set.
> > image
> > in RTL becomes
> > image
> > because the `margin-right` value was already set to 0, and the original value is lost. How do we handle this?
> 
> would `margin-right: initial;` work?

I tried `initial`, `unset`, `inherit`, `revert`, `auto` with `!important` but none of them worked. Also, most of these [don't work with IE anyway.](https://caniuse.com/?search=initial)

JKMarkowski:
Hello @kavya-b Could you try to add some mixin, which will be named `fd-not-rtl`, or `fd-ltr`? Maybe you can combine selector `::not()` with RTL. 
So for example `margin-right` will be modified only when there is not RTL, which means you won't need to reset `margin-right` on RTL

kavya-b:
> Hello @kavya-b Could you try to add some mixin, which will be named `fd-not-rtl`, or `fd-ltr`? Maybe you can combine selector `::not()` with RTL.
> So for example `margin-right` will be modified only when there is not RTL, which means you won't need to reset `margin-right` on RTL

Hi @JKMarkowski that is a good idea and I have been trying this, but unfortunately it isn't working as expected. 
I applied `not()` in the following way:
```
&--none {
      :not([dir="rtl"]), :not([dir="rtl"]) &{
        margin-right: 0 !important;
      }

      @include fd-rtl() {
        margin-left: 0 !important;
      }
    }
```
`:not([dir="rtl"]) .fd-margin-end--none` somehow does not work as intended in RTL.
Screenshot 2021-02-12 at 2 10 10 PM

`:not([dir="rtl"] .fd-margin-end--none)` works, but is incorrect as it affects everything that is not `.fd-margin-end--none`.
Screenshot 2021-02-12 at 2 10 37 PM

Apparently, the `:not()` [does not work very well with complex selectors and has limited browser support as well.](https://developer.mozilla.org/en-US/docs/Web/CSS/:not)

kavya-b:
@InnaAtanasova @JKMarkowski since I was not able to get around the RTL issue for `fd-margin-begin--none` and `fd-margin-end--none` through the lib styles itself, I am now applying a workaround from the docs side, where I set the `[dir="rtl"]` for the user-specified class in `customs.scss`, while also removing `!important` from the corresponding lib modifier. [Please take a look](https://github.com/SAP/fundamental-styles/pull/2131/commits/8101a73092bd959c01d4c333d9a25aa617fd1729) if this is okay for now.

JKMarkowski:
@kavya-b I also wasn't able to combine `at-root / rtl` with `not`. The other thing is to force users to put `[dir="ltr"]` by default, but this is something that everyone should agree on 
@InnaAtanasova What do you think about it?

kavya-b:
> @kavya-b I also wasn't able to combine `at-root / rtl` with `not`. The other thing is to force users to put `[dir="ltr"]` by default, but this is something that everyone should agree on
> @InnaAtanasova What do you think about it?

Hi @InnaAtanasova what are your thoughts about this?

InnaAtanasova:
> > @kavya-b I also wasn't able to combine `at-root / rtl` with `not`. The other thing is to force users to put `[dir="ltr"]` by default, but this is something that everyone should agree on
> > @InnaAtanasova What do you think about it?
> 
> Hi @InnaAtanasova what are your thoughts about this?

@kavya-b I need to discuss this with my team but from my research I found out that other libraries do not consider additional paddings and margins applied on the element. Most of the other libraries do not apply RTL at all. I will write back to you when we have some decision about it.

InnaAtanasova:
@kavya-b we just discussed the issue and we agreed that any additional padding/margins applied on the element should be ignored.

InnaAtanasova:
Code looks good, I am just worried the documentation is a bit confusing. 
This image is illustrating the margins:
Screen Shot 2021-02-18 at 3 41 20 PM

But usually this is the way to illustrate the paddings. The containers should all have the same size and the margins are "outside" the containers. Something like this:
Screen Shot 2021-02-18 at 3 49 50 PM

Same for paddings. This is how it's now in the documentation:
Screen Shot 2021-02-18 at 3 41 25 PM

But it's the way of illustrating margins. This is how it should be:
Screen Shot 2021-02-18 at 3 48 51 PM

I believe the best way would be to make all containers the same size and play with the backgrounds and maybe add borders to show the paddings and the margins.

kavya-b:
> Code looks good, I am just worried the documentation is a bit confusing.
> This image is illustrating the margins:
> Screen Shot 2021-02-18 at 3 41 20 PM
> 
> But usually this is the way to illustrate the paddings. The containers should all have the same size and the margins are "outside" the containers. Something like this:
> Screen Shot 2021-02-18 at 3 49 50 PM
> 
> Same for paddings. This is how it's now in the documentation:
> Screen Shot 2021-02-18 at 3 41 25 PM
> 
> But it's the way of illustrating margins. This is how it should be:
> Screen Shot 2021-02-18 at 3 48 51 PM
> 
> I believe the best way would be to make all containers the same size and play with the backgrounds and maybe add borders to show the paddings and the margins.

@InnaAtanasova 
margins now look like this:

and paddings look like this:


hope this is okay.

InnaAtanasova:
@kavya-b looks great! Thanks for updating the docs <3

InnaAtanasova (approved):

JKMarkowski (approved):
LGTM. Great Job!

Issue #2090, opened 2021-01-26
Defect Hunting for Sprint 55 - Monaco


InnaAtanasova:
Issues found during the defect hunting.

JKMarkowski:
- [ ] All Browsers - Docs - Form Grid - select in examples doesn't have arrow
![image](https://user-images.githubusercontent.com/26483208/107378657-7c968f00-6aec-11eb-9615-1eb680fbbc3d.png)

InnaAtanasova:
- [ ] Link Tile - all browsers 
<img width="1194" alt="Screen Shot 2021-02-09 at 1 45 15 PM" src="https://user-images.githubusercontent.com/39598672/107411927-1c4c2100-6add-11eb-9e64-581b49d9af2b.png">

mikerodonnell89:
- [ ] Border radius on the last list item for mobile multi-input / combobox

![Screen Shot 2021-02-10 at 12 03 05 PM](https://user-images.githubusercontent.com/2471874/107558394-f041a600-6b97-11eb-9ee5-48329c2523ac.png)

mikerodonnell89:
- [ ] Tristate checkboxes not demonstrating 3rd option
![Screen Shot 2021-02-10 at 12 52 51 PM](https://user-images.githubusercontent.com/2471874/107563919-e40d1700-6b9e-11eb-9757-bf9d4a74886c.png)

mikerodonnell89:
- [ ] Need to import link in message page example
![Screen Shot 2021-02-10 at 1 39 01 PM](https://user-images.githubusercontent.com/2471874/107569259-57b22280-6ba5-11eb-9507-125d96ab6f8b.png)

- [ ] Notifications - this more button needs to be compact
![Screen Shot 2021-02-10 at 1 40 08 PM](https://user-images.githubusercontent.com/2471874/107569393-7e705900-6ba5-11eb-8ce6-aa1cf81cf6b4.png)

- [ ] Object list - all prices are critical, most should probably not be semantic but we can have some semantic examples
![Screen Shot 2021-02-10 at 1 41 51 PM](https://user-images.githubusercontent.com/2471874/107569599-bd061380-6ba5-11eb-801b-bef15c739737.png)

- [ ] Mobile select (copy Inna's solution for multi-input)
[...]

PR #2140, opened 2021-02-10
docs: remove radio readonly examples

mikerodonnell89:
part of #2090 

HTML radio buttons do not support "readonly" functionality like a text input would, so I'm removing these examples from the docs.


InnaAtanasova (approved):

JKMarkowski (approved):
LGTM

PR #2145, opened 2021-02-12
fix: Side Nav taking the whole height of the container

InnaAtanasova:
closes https://github.com/SAP/fundamental-styles/issues/2146

## Description
The Side Nav should take the whole width of the container. Height 100% is added.


JKMarkowski (approved):
LGTM - this is what we were missing

Issue #2128, opened 2021-02-08
Dynamic Page Toolbar revisit


JKMarkowski:
Dynamic Page toolbars.

Current implementation of toolbars (action/global) is too complex for such an easy case.
- `order` can be removed and added flex `reverse` property.
- `align-items/justify-content` should be rechecked

This would make implementation on NGX easier

PR #2154, opened 2021-02-16
fix: simplify dynamic page

JKMarkowski:
## Related Issue
Closes https://github.com/SAP/fundamental-styles/issues/2128

## Description
There is refactor  of dynamic page - A lot of classes have been removed / updated. 
Examples have been changed to keep cozy buttons / cozy footers.
Placement for toolbars with buttons has been changed. For
- Desktop - it's aligned with title in single line
- Tablet / Mobile - first row is aligned with breadcrumb, second is aligned with a the tite

## BREAKING CHANGE:
- Removed all of size modifier classes with `--sm/--md/--lg/--xl`
- Removed `fd-dynamic-page__actions-container` - Additional container is not needed anymore
- `fd-dynamic-page__toolbar-actions` changed to `fd-dynamic-page__toolbar--actions`
- `fd-dynamic-page__content-toolbar` changed to `fd-dynamic-page__toolbar--content`
- `fd-tab` size modifier are not needed anymore 
- Placement of toolbars in DOM with buttons changed positions in tablet and desktop mode: 
  Now should be wrapped inside `fd-dynamic-page__breadcrumb-container` or `fd-dynamic-page__title-container`

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/dynamic-page.scss
+++ b/src/dynamic-page.scss
@@ -228,6 +228,8 @@ $block: #{$fd-namespace}-dynamic-page;
   }
 
   .#{$block}__toolbar {
+    @include fd-reset-spacing();
+
     padding-left: $fd-dynamic-page-toolbar-space;
 
     @include fd-rtl() {
@@ -296,6 +298,8 @@ $block: #{$fd-namespace}-dynamic-page;
   }
 
   .#{$block}__tabs {
+    @include fd-reset-spacing();
+
     &--overflow {
       overflow: hidden;
     }
--- a/src/dynamic-page.scss
+++ b/src/dynamic-page.scss
@@ -18,6 +18,7 @@ $block: #{$fd-namespace}-dynamic-page;
   $fd-button-height: 1.5rem;
   $fd-button-clickable-height: 2rem;
   $fd-button-touchable-area: ($fd-button-clickable-height - $fd-button-height) / 2;
+  $fd-dynamic-page-toolbar-space: 1rem;
 
   @include fd-reset();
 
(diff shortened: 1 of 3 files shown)


src/dynamic-page.scss
```diff
@@ -458,38 +366,94 @@ $block: #{$fd-namespace}-dynamic-page;
...
 
-    &--md,
-    &--lg {
-      @include responsive-page($fd-dynamic-page-header-padding-top-bottom, 2rem);
+    .#{$block}__toolbar {
+      padding-left: 1rem;
+      margin-left: auto;
+
+      @include fd-rtl() {
+        margin-left: initial;
```
InnaAtanasova, line 401: `initial` does not work in IE11

src/dynamic-page.scss
```diff
@@ -211,89 +214,37 @@ $block: #{$fd-namespace}-dynamic-page;
...
-    &--md {
-      flex-flow: row;
+    margin-left: auto;
 
-      .#{$block}__toolbar-container {
-        flex-flow: column wrap;
-      }
+    @include fd-rtl() {
+      margin-right: auto;
+      margin-left: initial;
```
InnaAtanasova, line 225: `initial` does not work in IE11

src/dynamic-page.scss
```diff
@@ -458,38 +366,94 @@ $block: #{$fd-namespace}-dynamic-page;
...
-      @include responsive-page($fd-dynamic-page-header-padding-top-bottom, 2rem);
+    .#{$block}__toolbar {
+      padding-left: 1rem;
+      margin-left: auto;
+
+      @include fd-rtl() {
+        margin-left: initial;
+        margin-right: auto;
+        padding-left: 0;
+        padding-right: 1rem;
```
InnaAtanasova, line 404: this should be in a variable

src/dynamic-page.scss
```diff
@@ -458,38 +366,94 @@ $block: #{$fd-namespace}-dynamic-page;
...
+      }
+
+      &--content {
+        margin-left: initial;
```
InnaAtanasova, line 408: same here for IE11

src/dynamic-page.scss
```diff
@@ -458,38 +366,94 @@ $block: #{$fd-namespace}-dynamic-page;
...
+
+        @include fd-rtl() {
+          margin-right: initial;
```
InnaAtanasova, line 411: and here

InnaAtanasova (changes requested):

JKMarkowski:
@InnaAtanasova Thank you for review, you comments have been addressed

InnaAtanasova (approved):

droshev (approved):

PR #2188, opened 2021-03-08
fix: add fix for Dynamic Page title alignment issues

InnaAtanasova:
## Related Issue
Portal related issue

## Description
The elements in the Title container of Dynamic Page were not aligned vertically due to padding top added to the title and items being aligned `baseline`.

### Before:
Screen Shot 2021-03-08 at 3 17 06 PM
Screen Shot 2021-03-08 at 3 16 44 PM
Screen Shot 2021-03-08 at 3 15 41 PM
Screen Shot 2021-03-08 at 3 16 24 PM

### After:

Screen Shot 2021-03-08 at 3 14 40 PM
Screen Shot 2021-03-08 at 3 14 15 PM
Screen Shot 2021-03-08 at 3 13 41 PM
Screen Shot 2021-03-08 at 3 18 51 PM

### DIFF
--- a/src/dynamic-page.scss
+++ b/src/dynamic-page.scss
@@ -149,7 +149,7 @@ $block: #{$fd-namespace}-dynamic-page;
     @include fd-reset();
 
     @include fd-flex() {
-      align-items: baseline;
+      align-items: center;
     }
   }
 
@@ -179,7 +179,6 @@ $block: #{$fd-namespace}-dynamic-page;
 
     color: var(--sapGroup_TitleTextColor);
     font-size: var(--sapFontHeader3Size);
-    padding: 0.3125rem 0 0 0;
 
     @include fd-ellipsis();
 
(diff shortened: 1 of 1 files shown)


src/dynamic-page.scss
```diff
@@ -398,12 +397,8 @@ $block: #{$fd-namespace}-dynamic-page;
       padding: 0;
     }
 
-    .#{$block}__title {
-      padding: 0.6875rem 0 0 0;
```
kavya-b, line 402: Fiori3 spec has these values specified for `padding-top`. Is it okay to remove these values?

InnaAtanasova: Hi @kavya-b , these paddings are added to the title element to center it vertically in the container. We can achieve the centering with Flex. When only title is used, the padding top is causing problems with the vertical alignment.
See the example from another project:
Screen Shot 2021-03-09 at 10 51 41 AM

JKMarkowski (approved):
LGTM - now it will look better, even on various cozy/compact modes mixed with small/large

PR #2191, opened 2021-03-09
fix:  Color issue with left and right arrows and dots between the pages

itmilos:
## Related Issue
Closes SAP/fundamental-styles/issues/1860

## Description
left and right arrows for the disabled state will be displayed in  **rgba(51,51,51,0.5)** 
dots between pages will be displayed in **rgb(6, 56, 6);**

For arrows I have applied fd-disabled after fd-link as fd-link is widely used in the framework 

### Before:
ARROW
image

DOTS
image

### After:

ARROW Left
image

ARROW Right 
image

DOTS
image

- [x] All values are in `rem`
- [x] RTL support
- [na ] only one top level `fd-*` class is used in the file
3. Testing
- [na] Updated tests
4. Documentation

### DIFF
--- a/src/pagination.scss
+++ b/src/pagination.scss
@@ -44,7 +44,7 @@ $block: #{$fd-namespace}-pagination;
 
     &::before {
       content: "...";
-      color: var(--sapContent_LabelColor);
+      color: var(--sapLinkColor);
       font-weight: bold;
     }
   }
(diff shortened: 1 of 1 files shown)


droshev:
@itmilos even if you checked the item for the visual test commit you didn't title properly the commit message

droshev (approved):

droshev:
@itmilos you are part of the team so you don't need to use forks. Would you open a branch PR and we can merge it?

itmilos:
> @itmilos you are part of the team so you don't need to use forks. Would you open a branch PR and we can merge it?

Indeed,  please refer to https://github.com/SAP/fundamental-styles/pull/2195

PR #2192, opened 2021-03-10
feat: Add no-data, remove borders from responsive mode

JKMarkowski:
## Related Issue
part of https://github.com/SAP/fundamental-ngx/issues/4622
part of https://github.tools.sap/dxp/jukebox/issues/420

## Description
There are
- added no-data mode
- responsvie mode borders removed
- word break css included into cells

- [x] All values are in `rem`
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/table.scss
+++ b/src/table.scss
@@ -125,8 +125,6 @@ $block: #{$fd-namespace}-table;
   }
 
   &__text {
-    @include fd-reset();
-
     word-break: break-all;
     white-space: nowrap;
     overflow: hidden;
--- a/src/table.scss
+++ b/src/table.scss
@@ -130,6 +130,7 @@ $block: #{$fd-namespace}-table;
     height: $fd-table-cozy-cell-height;
     padding: 0 $fd-table-cell-padding;
     color: inherit;
+    word-break: break-all;
     border-right: $fd-table-border;
 
     &--valid {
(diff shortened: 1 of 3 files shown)


itmilos (approved):

InnaAtanasova:
> Screen Shot 2021-03-10 at 10 34 30 AM
> 
> Screen Shot 2021-03-10 at 10 34 13 AM
> 
> word-break: break-all will break the overflow rules. There might be cases where the user will want to add `...` and not break the words

this might be an option when max-width of the cell is provided

JKMarkowski:
@InnaAtanasova changes that we agreed on are applied

src/table.scss
```diff
@@ -124,6 +124,16 @@ $block: #{$fd-namespace}-table;
     }
   }
 
+  &__text {
+    word-break: break-word;
+    overflow: hidden;
+    text-overflow: ellipsis;
```
InnaAtanasova, line 130: you could use the mixin here

InnaAtanasova (approved):

src/table.scss
InnaAtanasova: @itmilos why did you resolve this comment?

InnaAtanasova: One of the Definitions of Done is `Mixins are used for repeatable code (fd-rtl, fd-ellipsis, fd-flex, fd-selected, fd-focus, ect.)` and there's a mixin `fd-ellipsis` that has these 3 rules

PR #2193, opened 2021-03-10
fix: Tabs component: underlying HTML

itmilos:
## Related Issue
Closes SAP/fundamental-styles#105

## Description
Changed anchor tags to button tags

- [na] All values are in `rem`
- [na ] focus state of the element follow design spec
- [na] RTL support
3. Testing
- [ ] Updated tests
4. Documentation
- [ ] Storybook documentation has been created/updated
- [ ] Breaking Changes [wiki](https://github.com/SAP/fundamental-styles/wiki/Breaking-Changes) has been updated in case of breaking changes.


droshev (changes requested):
You need to update the tests.
Can you mark and document this as a breaking change?
Also, can you check in other places if the tabs are used and update the markup?

itmilos:
I added **role="presentation"** to **li** element and **role="tab"** to **button** ref. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_presentation_role

itmilos:
> I added **role="presentation"** to **li** element and **role="tab"** to **button** ref. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_presentation_role

This approach does not meet the accessibility tests standard

itmilos:
## 0.17.0

#### Tabs - change element  from `a` to `button`

Before:
```
 <li role="tab" class="fd-tabs__item">
                            <a class="fd-tabs__link">
                                <span class="fd-tabs__tag">
                                    Link
                                </span>
                            </a>
                        </li>
...
```

After:
```
 <li role="tab" class="fd-tabs__item">
                            <button class="fd-tabs__link">
                                <span class="fd-tabs__tag">
                                    Link
                                </span>                           
 </button>
                        </li>
...
```

stories/notification/notification.stories.js
```diff
@@ -202,34 +202,31 @@ Notifications can display error alerts by adding the <code>fd-message-strip fd-m
 export const notificationGroup = () => `<div class="fd-notification fd-notification--group">
     <ul class="fd-tabs fd-tabs--l fd-notification--tabs" role="tablist">
         <li role="tab" class="fd-tabs__item" aria-selected="true">
-            <a 
+            <button
```
InnaAtanasova, line 205: Buttons are missing tabindex, also missing is the aria-selected

InnaAtanasova (changes requested):
The examples are no longer changing the tab panels. 
To reproduce:
- click on a different Tab
- the tab panel content is not updated

droshev:
@itmilos you can move to branch PR

Issue #1155, opened 2020-06-15
toolbar button not having any margin


stefanoScalzo:
**Describe the bug**
The buttons are stuck together in overflow mode and need padding
> Is this issue related to a specific component?

toolbar

> What version of the Fundamental Library Styles are you using?

latest

**Screenshots**

![Screen Shot 2020-06-15 at 2 13 00 PM](https://user-images.githubusercontent.com/50607147/84691595-6d861f80-af12-11ea-9267-2826cf72f4e7.png)

stefanoScalzo:
Toolbar component is under review by ui5, waiting for new specs to come out

PR #2196, opened 2021-03-12
fix: toolbar button not having any margin

itmilos:
## Related Issue
Closes SAP/fundamental-styles#1155

## Description
Added margins for buttons in toolbar overflow 

### Before:
image

### After:
image

- [x] All values are in `rem`
- [na] RTL support
3. Testing
- [na] Updated tests
4. Documentation

### DIFF
--- a/src/toolbar.scss
+++ b/src/toolbar.scss
@@ -165,6 +165,13 @@ $title-toolbar-height: 2.75rem;
           background-color: var(--sapToolbar_SeparatorColor);
         }
 
+        .#{$fd-namespace}-button {
+          margin-top: 0.1875rem;
+          &:last-of-type {
+            margin-bottom: 0.1875rem;
+          }
+        }
+
         .#{$block}__overflow__label {
           &.#{$fd-namespace}-label {
             margin: 0.375rem 0;
@@ -198,6 +205,13 @@ $title-toolbar-height: 2.75rem;
           padding: 0.25rem 0.5rem;
         }
       }
+      .#{$fd-namespace}-button {
+        margin-top: 0.1875rem;
+        &:last-of-type {
+          margin-bottom: 0.1875rem;
+        }
+      }
     }
+
   }
 }
--- a/src/toolbar.scss
+++ b/src/toolbar.scss
@@ -153,36 +153,20 @@ $title-toolbar-height: 2.75rem;
           display: block;
           text-align: left;
           width: auto;
+          margin: 0.1875rem 0;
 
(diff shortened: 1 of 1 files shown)


src/toolbar.scss
```diff
@@ -163,6 +163,16 @@ $title-toolbar-height: 2.75rem;
           margin: 0.375rem 0.1875rem;
           width: auto;
           background-color: var(--sapToolbar_SeparatorColor);
+          &+ .#{$fd-namespace}-button {
```
InnaAtanasova, line 166: including fd-button inside fd-toolbar is violating the self-contained rules. If inside your component (in this case fd-toolbar) you are overwriting the properties of another component (fd-button in this case) you can't target directly the element like you did, you need a modifier class. 
Basically you will need something like: `fd-toolbar__button` where you will modify the specific property of `fd-button`. But the problem here is that the visual specs do not specify spacing for the buttons in the overflow menu. The button is supposed to have "touchable area" that is bigger than the dimensions of the button and this touchable area is playing the role of "margin" around the button

InnaAtanasova (changes requested):

src/toolbar.scss
InnaAtanasova: From the specs:
Screen Shot 2021-03-12 at 3 26 30 PM

mikerodonnell89: @InnaAtanasova I think it may be safe to assume the same spacing as buttons in the action sheet menu?

itmilos: > including fd-button inside fd-toolbar is violating the self-contained rules. If inside your component (in this case fd-toolbar) you are overwriting the properties of another component (fd-button in this case) you can't target directly the element like you did, you need a modifier class.

we can apply the same spacing for all elements in the menu, that way we should remove some of classes like

.#{$block}__overflow__label {
          &.#{$fd-namespace}-label {
                 margin: 0.375rem 0;
          

> Basically you will need something like: `fd-toolbar__button` where you will modify the specific property of `fd-button`. But the problem here is that the visual specs do not specify spacing for the buttons in the overflow menu. The button is supposed to have "touchable area" that is bigger than the dimensions of the button and this touchable area is playing the role of "margin" around the button

I see now but why margin around the button has to do with "touchable area" by default padding is doing that?

droshev (approved):

InnaAtanasova (approved):

InnaAtanasova:
@itmilos you prob need to update the storyshot, the builds are failing

PR #2204, opened 2021-03-18
fix: change the naming of status indicator classes to follow BEM

Lokanathan-k:
## Related Issue

## Description
changing the BEM to the status indicator and adding link css

### Before:
fd-status-indicator--lg__text and fd-status-indicator--critical__text. 

### After:
fd-status-indicator__text--lg  fd-status-indicator__text--critical

- [NA] RTL support
- [NA] only one top level `fd-*` class is used in the file
3. Testing
- [NA] Updated tests
- [NA] last commit message should have `[ci visual]` so it can trigger chromatic visual regression (e.g. `test: run chromatic visual regression [ci visual]`)
4. Documentation

### DIFF
--- a/src/status-indicator.scss
+++ b/src/status-indicator.scss
@@ -46,12 +46,16 @@ $color-states: (
     flex-direction: unset;
   }
 
+  &--link {
+    cursor: pointer;
+  }
+
   @each $set-name, $color-set in $color-states {
     &--#{$set-name} {
       fill: map_get($color-set, "color");
     }
 
-    &--#{$set-name}__text {
+    &__text--#{$set-name} {
       color: map_get($color-set, "color");
     }
   }
@@ -64,7 +68,7 @@ $color-states: (
       width: map-get($size-set, 'width');
       height: map-get($size-set, 'height');
     }
-    &--#{$set-name}__text {
+    &__text--#{$set-name} {
       font-size: map-get($size-set, 'font-size');
       margin: map-get($size-set, 'margin');
     }
--- a/src/status-indicator.scss
+++ b/src/status-indicator.scss
@@ -33,14 +33,14 @@ $color-states: (
   flex-direction: column;
   align-items: center;
 
-  &:focus {
-    outline: 1px dotted #000;
-    outline-offset: 1px;
+  @include fd-focus() {
+    outline: var(--sapContent_FocusWidth) var(--sapContent_FocusStyle) var(--sapContent_FocusColor);
+    outline-offset: var(--sapContent_FocusWidth);
   }
 
   @include fd-reset();
 
-  &--htext {
+  &--horizontal-label {
     @include fd-reset();
 
     flex-direction: unset;
@@ -53,9 +53,6 @@ $color-states: (
   @each $set-name, $color-set in $color-states {
     &--#{$set-name} {
       fill: map_get($color-set, "color");
-    }
-
-    &__text--#{$set-name} {
       color: map_get($color-set, "color");
     }
   }
@@ -68,7 +65,7 @@ $color-states: (
       width: map-get($size-set, 'width');
       height: map-get($size-set, 'height');
     }
-    &__text--#{$set-name} {
+    &__label--#{$set-name} {
       font-size: map-get($size-set, 'font-size');
       margin: map-get($size-set, 'margin');
     }
--- a/src/status-indicator.scss
+++ b/src/status-indicator.scss
@@ -22,6 +22,8 @@ $color-states: (
   $fd-status-indicator-default-color-stroke-width: 0.25 !default;
   $fd-status-indicator-default-color-fill: transparent !default;
   $fd-status-indicator-default-color: var(--sapContent_LabelColor) !default;
+  $fd-status-indicator-default-font-size: var(--sapFontSmallSize, 0.75rem);
+  $fd-status-indicator-default-margin: 0.375rem !default;
 
   font-family: var(--sapFontFamily);
   border-color: var(--sapContent_ForegroundBorderColor);
@@ -50,6 +52,19 @@ $color-states: (
     cursor: pointer;
   }
 
+  &__label {
+    font-size: $fd-status-indicator-default-font-size;
+    margin: $fd-status-indicator-default-margin;
+
+    // sizes
+    @each $set-name, $size-set in $fd-status-indicator-sizes {
+      &--#{$set-name} {
+        font-size: map-get($size-set, 'font-size');
+        margin: map-get($size-set, 'margin');
+      }
+    }
+  }
+
   @each $set-name, $color-set in $color-states {
     &--#{$set-name} {
       fill: map_get($color-set, "color");
@@ -60,14 +75,10 @@ $color-states: (
   // sizes
   @each $set-name, $size-set in $fd-status-indicator-sizes {
     &--#{$set-name} {
-      @include fd-reset-spacing();
-
-      width: map-get($size-set, 'width');
-      height: map-get($size-set, 'height');
-    }
-    &__label--#{$set-name} {
-      font-size: map-get($size-set, 'font-size');
-      margin: map-get($size-set, 'margin');
+      .#{$block}__svg {
+        width: map-get($size-set, 'width');
+        height: map-get($size-set, 'height');
+      }
     }
   }
 }
--- a/stories/status-indicator/status-indicator.stories.js
+++ b/stories/status-indicator/status-indicator.stories.js
@@ -270,7 +270,7 @@ export const StatusIndicatorLabels = () => `
 <div class="example-container">
 <span style="min-width: 150px;">Top Label :</span>
 <div class="fd-status-indicator fd-status-indicator--critical fd-status-indicator--lg"  aria-roledescription="Status Indicator" role="progressbar" aria-valuetext="100%" tabindex=0 aria-label="Euro Status Indicator With Labelled On Top" focusable="true" title="100% with label on top">
-	<span class="fd-status-indicator__label--lg">100%</span>
+	<span class="fd-status-indicator__label fd-status-indicator__label--lg">100%</span>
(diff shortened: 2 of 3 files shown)


JKMarkowski (changes requested):
Hello @Lokanathan-k Could you add `[ci visual]` text to your commits message? Just to trigger chromatic. Also you need to indicate that it's BREAKING CHANGE.

InnaAtanasova (changes requested):
@Lokanathan-k , you checked in the template that you updated the breaking changes in the wiki but you didn't. The checks in the PR templates are not there just to be checked :)

InnaAtanasova:
Also there's something wrong with the BEM structure in the case of Status indicator with Label. 
Currently you have the following structure:
```
<div class="fd-status-indicator fd-status-indicator--critical fd-status-indicator--htext">
    <span class="fd-status-indicator__text--lg fd-status-indicator__text--critical">100%</span>
    <svg class="fd-status-indicator--lg" ...>....
....
</div>
```

1. What is `fd-status-indicator--htext`?
2. Why `--lg` modifiers are put on the span level and on the svg level? In the other examples the size modifiers are applied on the parent level: `fd-status-indicator fd-status-indicator--lg`. 
3. You are repeating the `critical` modifier on the parent level and on the child level. You can't have `fd-status-indicator--critical` and `fd-status-indicator__text--critical`. It's redundant and not a good practice.
4. Your svg has `<svg class="fd-status-indicator--lg" ...>` this is not possible. You can't have a modifier class without a base class.
5. `fd-status-indicator__text` should be `fd-status-indicator__label` as it makes more sense

The structure should be something like:

```
<div class="fd-status-indicator fd-status-indicator--critical fd-status-indicator--lg">
    <span class="fd-status-indicator__label">100%</span>
    <svg class="fd-status-indicator__svg" ...>....
....
</div>
```

then you rely on specificity to overwrite the children properties:
```
.fd-status-indicator--critical {
    .fd-status-indicator__label {...}
    .fd-status-indicator__svg {...}
}
```
and
```
.fd-status-indicator--lg {
    .fd-status-indicator__label {...}
    .fd-status-indicator__svg {...}
}
```

src/status-indicator.scss
```diff
@@ -46,12 +46,16 @@ $color-states: (
     flex-direction: unset;
```
InnaAtanasova, line 46: should be `row` if it's default. Inherit, initial, unset have problems with IE11

InnaAtanasova: &:focus {
    outline: 1px dotted #000;
    outline-offset: 1px;
  }

you can't have 1px and #000  hardcoded

InnaAtanasova: there's a mixin for the focus that you can use

Lokanathan-k: Fixed

Lokanathan-k:
> Also there's something wrong with the BEM structure in the case of Status indicator with Label.
> Currently you have the following structure:
> 
> ```
> <div class="fd-status-indicator fd-status-indicator--critical fd-status-indicator--htext">
>     <span class="fd-status-indicator__text--lg fd-status-indicator__text--critical">100%</span>
>     <svg class="fd-status-indicator--lg" ...>....
> ....
> </div>
> ```
> 
> 1. What is `fd-status-indicator--htext`? (maybe name it horizontal-label): **Changed**

> 2. Why `--lg` modifiers are put on the span level and on the svg level? In the other examples the size modifiers are applied on the parent level: `fd-status-indicator fd-status-indicator--lg`.

  **on using with the parent level the svg size gets reduced on add the label content. or the label has to be kept as a separate component. since the label size vary from sm to xl and the margin size is different  for different size.**

> 3. You are repeating the `critical` modifier on the parent level and on the child level. You can't have `fd-status-indicator--critical` and `fd-status-indicator__text--critical`. It's redundant and not a good practice: **changed**


> 4. Your svg has `<svg class="fd-status-indicator--lg" ...>` this is not possible. You can't have a modifier class without a base class.
  **on add this on the parent tag found that the images size reduces and becomes distorted. on adding it to the respective SVG element we can see the image as defined in the spec**

> 5. `fd-status-indicator__text` should be `fd-status-indicator__label` as it makes more sense: **Changed**
> 
> The structure should be something like:
> 
> ```
> <div class="fd-status-indicator fd-status-indicator--critical fd-status-indicator--lg">
>     <span class="fd-status-indicator__label">100%</span>
>     <svg class="fd-status-indicator__svg" ...>....
> ....
> </div>
> ```
> 
> then you rely on specificity to overwrite the children properties:
> 
> ```
> .fd-status-indicator--critical {
>     .fd-status-indicator__label {...}
>     .fd-status-indicator__svg {...}
> }
> ```
> 
> and
> 
> ```
> .fd-status-indicator--lg {
>     .fd-status-indicator__label {...}
>     .fd-status-indicator__svg {...}
> }
> ```

InnaAtanasova:
@Lokanathan-k , 
You can't have a structure like `<span class="fd-status-indicator__label--lg">100%</span>` or `<svg class="fd-status-indicator--lg"...>`. That's just not possible. Classes containing `--something` are MODIFIER classes, they modify something and are NEVER used alone. They are supposed to modify a property or properties that belong to the element class applied on the same level. 
So, if you have a class `.fd-status-indicator__label` that for example has color red and font-size 14px you will have:

```
.fd-status-indicator__label {
    color: red;
    font-size: 14px
}
```

Then in case you want to have a label that has bigger font-size you create a MODIFIER class and overwrite ONLY the font-size:

```
.fd-status-indicator__label--lg {
    font-size: 16px
}
```

But you need to always apply both, because if you apply only `.fd-status-indicator__label--lg` the element will not get the red color defined in the element class:

<h4 fd-status-indicator__label fd-status-indicator__label--lg>something</h4>

stories/status-indicator/status-indicator.stories.js
```diff
@@ -269,9 +269,9 @@ modifier class together with the \`fd-status-indicator\` class.
 export const StatusIndicatorLabels = () => `
 <div class="example-container">
 <span style="min-width: 150px;">Top Label :</span>
-<div class="fd-status-indicator fd-status-indicator--critical"  aria-roledescription="Status Indicator" role="progressbar" aria-valuetext="100%" tabindex=0 aria-label="Euro Status Indicator With Labelled On Top" focusable="true" title="100% with label on top">
-	<span class="fd-status-indicator--lg__text fd-status-indicator--critical__text">100%</span>
-	<svg id="__shape0__box9-24" class="fd-status-indicator--lg" data-sap-ui="__shape0-__box21-24" version="1.1" xlmns="http://www.w3.org/2000/svg" viewBox="0 0 26 25" preserveAspectRatio="xMidYMid meet" x="0" y="0" width="100%" height="100%">
+<div class="fd-status-indicator fd-status-indicator--critical fd-status-indicator--lg"  aria-roledescription="Status Indicator" role="progressbar" aria-valuetext="100%" tabindex=0 aria-label="Euro Status Indicator With Labelled On Top" focusable="true" title="100% with label on top">
+	<span class="fd-status-indicator__label--lg">100%</span>
```
InnaAtanasova, line 273: Should be:
```
<span class="fd-status-indicator__label fd-status-indicator__label--lg">100%</span>
```

stories/status-indicator/status-indicator.stories.js
```diff
@@ -330,14 +330,14 @@ export const StatusIndicatorLabels = () => `
 			</path>
 		</svg>	
 	</svg>
-	<span class="fd-status-indicator--lg__text fd-status-indicator--positive__text">100%</span>
+	<span class="fd-status-indicator__label--lg">100%</span>
```
InnaAtanasova, line 333: Should be:
```
<span class="fd-status-indicator__label fd-status-indicator__label--lg">100%</span>
```

stories/status-indicator/status-indicator.stories.js
```diff
@@ -330,14 +330,14 @@ export const StatusIndicatorLabels = () => `
...
 </div>
 </div>
 <div class="example-container">
 <span style="min-width: 150px;">Left Label :</span>
-<div class="fd-status-indicator fd-status-indicator--critical fd-status-indicator--htext " aria-roledescription="Status Indicator" role="progressbar" aria-valuetext="100%" tabindex=0 aria-label="Euro Status Indicator With Labelled On Left" focusable="true" title="100% with label on left">
-	<span class="fd-status-indicator--lg__text fd-status-indicator--critical__text">100%</span>
-	<svg id="__shape0__box12-24" class="fd-status-indicator--lg" data-sap-ui="__shape0-__box21-24" version="1.1" xlmns="http://www.w3.org/2000/svg" viewBox="0 0 36 25" preserveAspectRatio="xMidYMid meet" x="0" y="0" width="100%" height="100%">
+<div class="fd-status-indicator fd-status-indicator--critical fd-status-indicator--horizontal-label fd-status-indicator--lg" aria-roledescription="Status Indicator" role="progressbar" aria-valuetext="100%" tabindex=0 aria-label="Euro Status Indicator With Labelled On Left" focusable="true" title="100% with label on left">
+	<span class="fd-status-indicator__label--lg">100%</span>
```
InnaAtanasova, line 339: Should be:
```
<span class="fd-status-indicator__label fd-status-indicator__label--lg">100%</span>
```

stories/status-indicator/status-indicator.stories.js
```diff
@@ -396,7 +396,7 @@ export const StatusIndicatorLabels = () => `
 			</path>
 		</svg>	
 	</svg>
-	<span class="fd-status-indicator--lg__text fd-status-indicator--positive__text">100%</span>
+	<span class="fd-status-indicator__label--lg">100%</span>
```
InnaAtanasova, line 399: Should be:
```
<span class="fd-status-indicator__label fd-status-indicator__label--lg">100%</span>
```

we already discussed that a modifier class like `fd-status-indicator__label--lg` should not be used alone but with a base element class like `fd-status-indicator__label`. In this case `fd-status-indicator__label--lg`  is modifying the props set in `fd-status-indicator__label`.

InnaAtanasova (changes requested):

InnaAtanasova (approved):

PR #2214, opened 2021-03-24
fix: remove margin from Resizable-Card-Layout item

### LINKED ISSUE #2213: fix: remove margin from resizable card. handled in component with script
**Describe the bug**
margin between cards are handled in resizable-card-component using script. hence it to be removed from styles.
changes includes:
1. remove margin between cards.
2. add z-index for resize handlers
3. add xl layout size with correct values of layout.

Resizable Card Layout component

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

Angular

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

DeepakSap14:
## Related Issue
Closes #2213 

## Description
1. Margin between cards is handled inside component using script. hence to be removed from styles.
2. added z-index for resize handlers
3. added missing "xl" screen size layout padding

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/.storybook/custom/custom.scss
+++ b/.storybook/custom/custom.scss
@@ -603,19 +603,19 @@ body.sb-show-main {
   .card-one-dimensions {
     width: 40rem;
     height: 16rem;
-    margin: 0 0.5rem 0.5rem 0;
+    margin: 0.5rem;
   }
 
   .card-two-dimensions {
     width: 20rem;
     height: fit-content;
-    margin: 0 0.5rem 0.5rem 0.5rem;
+    margin: 0.5rem;
   }
 
   .card-three-dimensions {
     width: 20rem;
     height: 20rem;
-    margin: 0.5rem 0.5rem 0.5rem 0;
+    margin: 0.5rem;
   }
 
   .card-four-dimensions {
--- a/.storybook/custom/custom.scss
+++ b/.storybook/custom/custom.scss
@@ -600,40 +600,34 @@ body.sb-show-main {
   display: flex;
   flex-wrap: wrap;
 
-  .card-one-dimensions {
-    width: 40rem;
-    height: 16rem;
+  @mixin set-card-dimensions($width: 20rem, $height: 20rem) {
+    width: $width;
+    height: $height;
     margin: 0.5rem;
+  }
+
+  .card-one-dimensions {
+    @include set-card-dimensions(40rem, 16rem);
   }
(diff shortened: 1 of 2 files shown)


DeepakSap14:
@InnaAtanasova @JKMarkowski 
while developing Resizable card layout component in fundamental-ngx, some minor change in styles has happened.
Now component is taking care of margin between cards using script. so i have removed it from styles.
there are other small changes as well, please review.

src/resizable-card-layout.scss
```diff
@@ -112,6 +118,7 @@ $block: #{$fd-namespace}-resizable-card-layout;
 
     &--horizontal {
       // subtract resize-corner icon font-size/2
+      z-index: 2;
```
InnaAtanasova, line 121: z-index is used 3 times, it should be moved to a variable

.storybook/custom/custom.scss
```diff
@@ -603,31 +603,37 @@ body.sb-show-main {
   .card-one-dimensions {
     width: 40rem;
     height: 16rem;
+    margin: 0.5rem;
```
InnaAtanasova, line 606: this code is repeated in all 6 card dimensions. Please create a mixin that will set the width, the height and the margin:
```
@mixin set-card-dimensions($width: 20rem, $height: 20rem) {
     width: $width;
     height: $height:;
     margin: 0.5rem;
}
```

Then you will only have to pass values in the mixin when the width is not 20rem or the height is not 20rem.

```
.card-one-dimensions {
    @include set-card-dimensions(40rem, 16rem);
}

.card-three-dimensions {
    @include set-card-dimensions();
}
```

InnaAtanasova (changes requested):

src/resizable-card-layout.scss
DeepakSap14: used variable now

.storybook/custom/custom.scss
DeepakSap14: created mixin used it.

src/resizable-card-layout.scss
DeepakSap14: used variable now

InnaAtanasova (approved):

Issue #2153, opened 2021-02-16
Visual Issues - Avatar Group


droshev:
- [ ] Avatar Group - dark theme; menu header border color is different than the body border in Dark theme
<img width="433" alt="Screen Shot 2021-02-15 at 7 46 27 PM" src="https://user-images.githubusercontent.com/4380815/108005393-9a6d6380-6fc6-11eb-9698-7001cec407e9.png">

- [ ] Avatar Group - menu arrow LTR is not centered (https://github.com/SAP/fundamental-styles/pull/2121)
<img width="474" alt="Screen Shot 2021-02-15 at 7 48 24 PM" src="https://user-images.githubusercontent.com/4380815/108005484-d86a8780-6fc6-11eb-958f-e495a53b1497.png">
<img width="420" alt="Screen Shot 2021-02-15 at 7 48 30 PM" src="https://user-images.githubusercontent.com/4380815/108005485-d86a8780-6fc6-11eb-9f0a-d0d91f6d454f.png">

N1XUS:
@droshev as discussed:
* First item has been fixed
* Second item is not possible to achieve with static nature of the fundamental styles. Yet, it can be achieved in fundamental ngx. Further discussion on implementation for fundamental ngx is required.

PR #2217, opened 2021-03-24
fix: add fix for the Popover alignment in Avatar Group component

itmilos:
## Related Issue
Part of SAP/fundamental-styles#2153

## Description
Fix position for popover body 

### Before:
https://user-images.githubusercontent.com/4380815/108005485-d86a8780-6fc6-11eb-9f0a-d0d91f6d454f.png
image

### After:
image
image
image

- [x] All values are in `rem`
- [ ] RTL support

### DIFF
--- a/src/mixins/_popover.scss
+++ b/src/mixins/_popover.scss
@@ -1,6 +1,6 @@
 @import "mixins";
 
-$fd-popover-body-shadow-offset: var(--sapGroup_BorderWidth) !default;
+$fd-popover-body-shadow-offset: calc(#{var(--sapGroup_BorderWidth)} + 0.4275rem) !default;
 $fd-popover-background-color: var(--sapGroup_ContentBackground) !default;
 $fd-popover-border-radius: var(--sapElement_BorderCornerRadius) !default;
 $fd-popover-box-shadow-no-arrow: var(--sapContent_Shadow1) !default;


InnaAtanasova:
@itmilos please maintain the PR template (the checkboxes)
The PR title should follow the guidelines. Should start with `fix: some text`, `feat: some text`, `docs: some text`, `chore: some text`, etc.
Did you check if this will work in fundamental-ngx? You can temporary bring the styling there and see if it's working.

JKMarkowski:
Hi @itmilos Could you also add `[ci visual]`  to commit message - it will trigger chromatic

mikerodonnell89:
@itmilos so the problem with this issue is not that the popover body isn't centered, in fact it is in the correct place here (right aligned).  The problem is specifically the little triangle arrow I've circled here, and how it is not centered in the middle under the circle:

108005485-d86a8780-6fc6-11eb-9f0a-d0d91f6d454f

So you should be targeting the ::before pseudoelement that is the triangle.  This is a problem with popovers throughout the whole library so I'm not sure why the linked issue only mentions it for avatar group.

Maybe a topic we can discuss in scrum, but we need to be conscious of left-to-right and right-to-left displays, so where you add some offset in LTR mode you will need to use the RTL mixin and actually subtract that offset.  We can have a call about that as it gets confusing

Also, worth mentioning that your changes here affected popovers throughout the entire library so it's always good to manually check around when manipulating commonly used components

mikerodonnell89 (changes requested):
see comment above

PR #2233, opened 2021-04-06
fix: table fixed cells hover state

platon-rov:
## Related Issue
Relates to [Fundamental-NGX #4163](https://github.com/SAP/fundamental-ngx/issues/4163) - fix needed at the level of styles library;

## Description
* Fixed cell now inherits its parent row background-color
* Fixed cell now have the right color on hover when it's inside of selected rov

### Before:

image

Hover:

image

### After:

image

Hover: 

image

- [X] All values are in `rem`
- [X] RTL support
3. Testing
- [X] Updated tests
4. Documentation


InnaAtanasova (approved):

dimamarksman (approved):

puru-hk:
Looks fine

puru-hk (approved):

PR #2236, opened 2021-04-07
fix: move card styles from ngx to styles

### LINKED ISSUE #1729: negative numbers in RTL are incorrect
**Describe the bug**
Right now in RTL mode negative numbers are incorrect (minus goes after number)
It's needed to have a predefined wrapper or class to keep numbers (potentially phone numbers as well) in LTR even when parent is in RTL.

This problem can be seen in `object-number` and `card` examples.
Analytical Card:
image
Object Number:
image

No
0.12

**Expected behavior**
Negative numbers should remain in LTR

DeepakSap14:
## Related Issue
closes #1729 

## Description
Currently fundamental-ngx card component has some styles present in fundamental-ngx, which can be moved to fundamental-styles.

### Before:
Screenshot 2021-04-07 at 2 29 39 PM

Screenshot 2021-04-07 at 2 29 13 PM

### After:
Negative sign placed correctly
Screenshot 2021-04-07 at 2 32 39 PM

Table border removed
Screenshot 2021-04-07 at 2 35 36 PM

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/card.scss
+++ b/src/card.scss
@@ -231,6 +231,11 @@ $fd-card-default-body-background-color: var(--sapTile_Background) !default;
     @include fd-reset();
 
     color: var(--sapTile_TitleTextColor);
+
+    @include fd-rtl() {
+      direction: ltr;
+      text-align: right;
+    }
   }
 
   &__second-subtitle {
@@ -352,6 +357,8 @@ $fd-card-default-body-background-color: var(--sapTile_Background) !default;
   }
 
   &--table {
+    border: hidden;
+
     &.#{$block}--compact {
       .#{$block}__content {
         padding-bottom: 0.5rem;
@@ -363,5 +370,12 @@ $fd-card-default-body-background-color: var(--sapTile_Background) !default;
     .#{$block}__content {
       padding: 1rem;
     }
+  }
+
+  &__loader {
+    @include fd-reset();
+    @include fd-flex-center();
+
+    min-height: 5rem;
   }
 }
--- a/src/numeric-content.scss
+++ b/src/numeric-content.scss
@@ -56,6 +56,11 @@ $fd-numeric-content-elements-spacing-s: 0.5rem !default;
 
   .#{$block}__kpi {
     font-size: 1.5rem;
+
+    @include fd-rtl() {
+      direction: ltr;
+      text-align: right;
+    }
   }
 
   @content;


src/card.scss
```diff
@@ -231,6 +231,11 @@ $fd-card-default-body-background-color: var(--sapTile_Background) !default;
     @include fd-reset();
 
     color: var(--sapTile_TitleTextColor);
+
+    @include fd-rtl() {
+      direction: ltr;
```
InnaAtanasova, line 236: why are you adding this?

src/card.scss
```diff
@@ -364,4 +371,23 @@ $fd-card-default-body-background-color: var(--sapTile_Background) !default;
...
+    @include fd-reset();
+
+    flex: auto;
+  }
+
+  .#{$numeric-content}__kpi {
+    @include fd-reset();
+
+    font-size: 2.25rem;
+  }
```
InnaAtanasova, line 392: You can't have `numeric-content` classes inside Card component, this is violating the self-contained rules.

src/card.scss
```diff
@@ -364,4 +371,23 @@ $fd-card-default-body-background-color: var(--sapTile_Background) !default;
...
+    @include fd-reset();
+    @include fd-flex-center();
+
+    min-height: 5rem;
+  }
+
+  .#{$numeric-content}__scale-container {
+    @include fd-reset();
+
+    flex: auto;
```
InnaAtanasova, line 385: `flex: auto` is not working properly in IE11. Have you checked it?

InnaAtanasova (changes requested):

DeepakSap14:
@InnaAtanasova please check this PR along with PR https://github.com/SAP/fundamental-ngx/pull/5015.
on closer looking, some css changes present in #5015 was not needed and already proper class is available. so i have used that.
rest of css changes are migrated here.

src/card.scss
DeepakSap14: this class is not needed. now using correct css class in ngx.

src/card.scss
DeepakSap14: removed.

src/numeric-content.scss
```diff
@@ -56,6 +56,11 @@ $fd-numeric-content-elements-spacing-s: 0.5rem !default;
 
   .#{$block}__kpi {
     font-size: 1.5rem;
+
+    @include fd-rtl() {
+      direction: ltr;
+      text-align: right;
+    }
```
InnaAtanasova, line 63: would be good if this is moved to another mixin for future re-use

InnaAtanasova (approved):

InnaAtanasova:
Please trigger Chromatic checks

droshev (approved):

Issue #2232, opened 2021-04-06
Page footer: Implement style 


Lokanathan-k:
Implement the non fiori 3 page footer component.

PR #2237, opened 2021-04-07
feat: Introduce non fiori3 page footer component

Lokanathan-k:
## Related Issue
Closes SAP/fundamental-styles#2232

## Description
introduce non fiori3 page footer component, as part of the platform ngx component development

- [x] All values are in `rem`
- [X] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/page-footer.scss
+++ b/src/page-footer.scss
@@ -1,0 +1,100 @@
+@import "./new-settings";
+@import "./mixins";
+
+$block: #{$fd-namespace}-page-footer;
+
+$fd-page-footer-sizes: (
+  "xl": ("padding": 3rem),
+  "lg": ("padding": 2rem),
+  "md": ("padding": 1rem),
+  "sm": ("padding": 0.5rem)
+);
+
+.#{$block} {
+  @include fd-reset();
+
+  $fd-footer-margin-right: 1.5rem !default;
+  $fd-footer-label-padding-top: 0.75rem !default;
+
+  background: var(--sapBaseColor);
+  position: relative;
+  padding-top: 1.5rem;
+  padding-bottom: 2rem;
+
+  &__logo {
+    @include fd-reset();
+
+    width: 3rem;
+    height: 1.5rem;
+    overflow: hidden;
+  }
+
+  &__row {
+    @include fd-reset();
+
+    word-break: break-all;
+    width: 70%;
+
+    &-item {
+      margin-right: $fd-footer-margin-right;
+
+      @include fd-rtl() {
+        margin-right: 0;
+        margin-left: $fd-footer-margin-right;
+      }
+
+      &--image {
+        display: block;
+      }
+    }
+
+    &:not(:first-child) {
+      padding-top: 0.25rem;
+    }
+  }
+
+  &__text {
+    @include fd-reset();
+    @include fd-ellipsis();
+
+    align-self: flex-end;
+    width: 30%;
+    text-align: right;
+
+    @include fd-rtl() {
+      text-align: left;
+    }
+  }
+
+  // sizes
+  @each $set-name, $size-set in $fd-page-footer-sizes {
+    &--#{$set-name} {
+      padding-right: map-get($size-set, 'padding');
+      padding-left: map-get($size-set, 'padding');
+      .#{$block}__container {
+        @include fd-flex(column);
+
+        padding-top: $fd-footer-label-padding-top;
+      }
+    }
+  }
+
+  @media (max-width: 768px) {
+
+    .#{$block}__row {
+      margin-right: 0;
+      width: 100%;
+    }
+
+    .#{$block}__text {
+      width: 90%;
+      text-align: left;
+      padding-top: $fd-footer-label-padding-top;
+    }
+
+    .#{$block}__container {
+      display: inline-block;
+      flex-direction: row;
+    }
+  }
+}
--- a/src/page-footer.scss
+++ b/src/page-footer.scss
@@ -42,10 +42,10 @@ $fd-page-footer-sizes: (
         margin-right: 0;
         margin-left: $fd-footer-margin-right;
       }
+    }
 
-      &--image {
-        display: block;
-      }
+    &-image {
+      display: block;
     }
 
     &:not(:first-child) {
@@ -60,6 +60,7 @@ $fd-page-footer-sizes: (
     align-self: flex-end;
     width: 30%;
     text-align: right;
+    position: absolute;
 
     @include fd-rtl() {
       text-align: left;
@@ -84,6 +85,14 @@ $fd-page-footer-sizes: (
     .#{$block}__row {
       margin-right: 0;
       width: 100%;
+
+      &:not(:first-child) {
+        padding-top: 0;
+      }
+
+      &-item {
+        padding-top: 0.25rem;
+      }
     }
 
     .#{$block}__text {
@@ -95,6 +104,7 @@ $fd-page-footer-sizes: (
     .#{$block}__container {
       display: inline-block;
       flex-direction: row;
+      width: 100%;
     }
   }
 }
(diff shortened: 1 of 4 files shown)


InnaAtanasova (changes requested):
1. Please provide link to the specs
2. Shouldn't there be some vertical spacing for the links?
Screen Shot 2021-04-07 at 3 09 23 PM

3. Is this the expected behaviour on smaller screens?

Screen Shot 2021-04-07 at 3 09 40 PM

5. Overlapping of text in smaller screens
Screen Shot 2021-04-07 at 3 08 59 PM

6. Shouldn't the image and the text of the link be aligned?
Screen Shot 2021-04-07 at 3 07 56 PM
Screen Shot 2021-04-07 at 3 07 47 PM

InnaAtanasova:
Screen Shot 2021-04-07 at 3 13 20 PM

Added like this, the issue will not be closed automatically when the PR is merged.

src/page-footer.scss
```diff
@@ -0,0 +1,126 @@
...
+
+  $fd-footer-padding-top: 0.75rem !default;
+  $fd-footer-margin-right: 1.5rem !default;
+  $fd-footer-link-padding-top: 0.25rem !default;
+
+  @mixin fd-truncate-text {
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
```
InnaAtanasova, line 24: there's already a mixin for it. Please check the mixibs file

src/page-footer.scss
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    .#{$block}__text {
+      width: 90%;
+    }
+
+    .#{$block}__container {
+      display: inline-block;
+      flex-direction: row;
+    }
```
InnaAtanasova, line 124: all these rules are repeated in the other media query

stories/page-footer/page-footer.stories.js
```diff
@@ -0,0 +1,215 @@
...
+			<span class="sap-icon--download sap-icon--s"></span>
+			 icon and link
+			</a>
+		</div>
+		<div class="fd-page-footer__row">
+			<a href="#" id="fd-link-14" class="fd-page-footer__row__item fd-page-footer__row__item--image fd-link" tabindex="0">
+			<img src="https://picsum.photos/id/1021/50/50" alt="page footer clickable image link"/>
+			image link
+			</a>
+			<a href="#" id="fd-link-15" class="fd-page-footer__row__item fd-page-footer__row__item--image fd-link" tabindex="0">
```
InnaAtanasova, line 99: `fd-page-footer__row__item` is not a BEM naming

src/page-footer.scss
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  // sizes
+  @each $set-name, $size-set in $fd-page-footer-sizes {
+    &--#{$set-name} {
+      padding-right: map-get($size-set, 'padding');
+      padding-left: map-get($size-set, 'padding');
+      padding-top: 1.5rem;
+      padding-bottom: 2rem;
```
InnaAtanasova, line 81: these 2 should go on `fd-page-footer` level as they are the same for all cases, only the horizontal paddings are changing based on the modifier class.

src/page-footer.scss
```diff
@@ -0,0 +1,126 @@
...
+      .#{$block}__container {
+        display: flex;
+
+        @include fd-flex(column);
```
InnaAtanasova, line 85: Have you checked what this mixin is doing? It brings `display: flex;` so you don't need code above this (line 83 where you have `display: flex;`

src/page-footer.scss
```diff
@@ -0,0 +1,126 @@
...
+  "xl": ("padding": 3rem),
+  "lg": ("padding": 2rem),
+  "md": ("padding": 1rem),
+  "sm": ("padding": 0.5rem)
+);
+
+.#{$block} {
+  @include fd-reset();
+
+  $fd-footer-padding-top: 0.75rem !default;
```
InnaAtanasova, line 16: Used only once, no need to be a variable

src/page-footer.scss
```diff
@@ -0,0 +1,126 @@
...
+  $fd-footer-margin-right: 1.5rem !default;
+  $fd-footer-link-padding-top: 0.25rem !default;
```
InnaAtanasova, line 18: Used only once, no need to be a variable

InnaAtanasova (changes requested):

InnaAtanasova:
@Lokanathan-k please don't resolve the comments, it's done by the reviewers. Please fix the PR template where you address the issue.

src/page-footer.scss
Lokanathan-k: multiline-truncate() i see this mixin other than this there is no mixin available and this mixin add's unwanted margin on right. and there is no '...' at the end of the overflow text.
Screenshot 2021-04-14 at 11 47 41 AM

stories/page-footer/page-footer.stories.js
```diff
@@ -0,0 +1,215 @@
...
+	</div>
+	<div class="fd-page-footer__container">
+		<div class="fd-page-footer__row">
+			<a href="#" id="fd-link-13"  class="fd-page-footer__row--item fd-link" tabindex="0">
+			<span class="sap-icon--download sap-icon--s"></span>
+			 icon and link
+			</a>
+		</div>
+		<div class="fd-page-footer__row">
+			<a href="#" id="fd-link-14" class="fd-page-footer__row--item fd-page-footer__row--item--image fd-link" tabindex="0">
```
InnaAtanasova, line 95: Please check the BEM naming!
`fd-page-footer__row--item--image` is not BEM. You can't have 2 modifiers.
`fd-page-footer` is your BLOCK
`fd-page-footer__row` is your ELEMENT, that's why it has `__`
`fd-page-footer__row--item` is your modifier, that's why you have `--`

But you can't add a second modifier or a second element. 
You need to change `fd-page-footer__row--item` to `fd-page-footer__row-item`, in this case `__row-item` is your element and then have `fd-page-footer__row-item--image` as a modifier.

src/page-footer.scss
```diff
@@ -0,0 +1,101 @@
...
+    &--item {
+      margin-right: $fd-footer-margin-right;
+
+      @include fd-rtl() {
+        margin-right: 0;
+        margin-left: $fd-footer-margin-right;
+      }
+
+      &--image {
+        display: inline-grid;
```
InnaAtanasova, line 47: Please reconsider this. Grid is not working in IE11 and we need to support it.

InnaAtanasova (changes requested):

stories/page-footer/page-footer.stories.js
```diff
@@ -0,0 +1,215 @@
...
+	<div class="fd-page-footer__container">
+		<div class="fd-page-footer__row">
+			<a href="#" id="fd-link-13"  class="fd-page-footer__row-item fd-link" tabindex="0">
+			<span class="sap-icon--download sap-icon--s"></span>
+			 icon and link
+			</a>
+		</div>
+		<div class="fd-page-footer__row">
+			<a href="#" id="fd-link-14" class="fd-page-footer__row-item fd-link" tabindex="0">
+			<img class="fd-page-footer__row-item--image" src="https://picsum.photos/id/1021/50/50" alt="page footer clickable image link"/>
```
InnaAtanasova, line 96: What are you modifying with a modifier class if you don't have a base class?

You need the following format:
```
class="fd-block__element fd-block__element--modifier"
```

You can't have just:
```
class="fd-block__element--modifier"
```

A modifier class is used to change a property/properties of a base class

InnaAtanasova (changes requested):

InnaAtanasova (approved):

sKudum (approved):

Issue #2122, opened 2021-02-05
postCSS v.8 migration


droshev:
PostCSS released version 8 and it requires some migration:

```
add fallback plugin: postcss.plugin was deprecated. Migration guide:
https://evilmartians.com/chronicles/postcss-8-plugin-migration
remove font variables: postcss.plugin was deprecated. Migration guide:
https://evilmartians.com/chronicles/postcss-8-plugin-migration
```

PR #2248, opened 2021-04-12
chore: migrate "add fallback plugin" and "remove font variables" to postCSS v.8

InnaAtanasova:
## Related Issue
Closes SAP/fundamental-styles#2122

## Description
- Moved PostCSS 8 to peerDependencies in package.json;
- Replaced `module.exports = postcss.plugin(name, creator)` with `module.exports = creator`;
- Returned an object with postcssPlugin property containing a plugin name and the Once method;
- Moved plugin code to the Once method;
- Added `module.exports.postcss = true` to the end of the file;
- Removed postcss imports

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,6 @@
     "node-sass": "^5.0.0",
     "normalize.css": "^8.0.1",
     "npm-run-all": "^4.1.5",
-    "postcss": "^8.2.6",
     "postcss-banner": "^3.0.2",
     "postcss-clean": "^1.2.2",
     "postcss-cli": "^8.3.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -111,6 +110,9 @@
     "tocbot": "^4.12.2",
     "webpack-merge": "^5.7.3"
   },
+  "peerDependencies": {
```
droshev, line 113: does it need to be a peer dep?

InnaAtanasova: https://evilmartians.com/chronicles/postcss-8-plugin-migration

InnaAtanasova: Screen Shot 2021-04-13 at 8 52 40 AM

InnaAtanasova: This is what the guideline recommends.

InnaAtanasova: I put it back in the devDeps

droshev (approved):

PR #2267, opened 2021-04-19
fix: remove readonly attribute from File Uploader input and add examples of  File Uploader input states

manu-kr:
## Related Issue

## Description
- removed `readonly` attribute from File Uploader input and added some css rules associated with this move (text shadow and focus)
- introduced examples in the documentation of File Uploader with error, success, information and warning states.

### After:

image

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/file-uploader.scss
+++ b/src/file-uploader.scss
@@ -36,6 +36,7 @@ $fd-file-uploader-padding: 0.25rem;
       background-color: var(--sapField_WarningBackground);
       border-color: var(--sapField_WarningColor);
     }
+
     &.is-information {
       background-color: var(--sapField_InformationBackground);
       border-color: var(--sapField_InformationColor);
--- a/src/file-uploader.scss
+++ b/src/file-uploader.scss
@@ -9,7 +9,7 @@ $fd-file-uploader-padding: 0.25rem;
 .#{$block} {
   @include fd-reset();
   @include fd-flex-vertical-center();
-
+  
   &__input.#{$input} {
     width: auto;
     margin-right: $fd-file-uploader-padding;
@@ -26,13 +26,17 @@ $fd-file-uploader-padding: 0.25rem;
       margin-right: 0;
     }
 
+    &.is-success {
+      background-color: var(--sapField_SuccessBackground);
+      border-color: var(--sapField_SuccessColor);
+    }
+
     &.is-error {
       background-color: var(--sapField_InvalidBackground);
       border-color: var(--sapField_InvalidColor);
     }
 
-    &.is-warning,
-    &.is-alert {
+    &.is-warning {
       background-color: var(--sapField_WarningBackground);
       border-color: var(--sapField_WarningColor);
     }
--- a/src/file-uploader.scss
+++ b/src/file-uploader.scss
@@ -36,6 +36,10 @@ $fd-file-uploader-padding: 0.25rem;
       background-color: var(--sapField_WarningBackground);
       border-color: var(--sapField_WarningColor);
     }
+    &.is-information {
+      background-color: var(--sapField_InformationBackground);
+      border-color: var(--sapField_InformationColor);
+    }
   }
 
   &__hidden {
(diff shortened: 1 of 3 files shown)


src/file-uploader.scss
```diff
@@ -25,6 +25,22 @@ $fd-file-uploader-padding: 0.25rem;
...
       margin-right: 0;
     }
+
+    &.is-error {
+      background-color: var(--sapField_InvalidBackground);
+      border-color: var(--sapField_InvalidColor);
+    }
+
+    &.is-warning,
+    &.is-alert {
```
InnaAtanasova, line 35: I am not aware of an `is-alert` state.

src/file-uploader.scss
```diff
@@ -25,6 +25,22 @@ $fd-file-uploader-padding: 0.25rem;
...
+      background-color: var(--sapField_WarningBackground);
+      border-color: var(--sapField_WarningColor);
+    }
+
+    &.is-information {
+      background-color: var(--sapField_InformationBackground);
+      border-color: var(--sapField_InformationColor);
+    }
```
InnaAtanasova, line 43: It's missing the is-success state

InnaAtanasova: `.fd-file-uploader__input` is always used with `fd-input` and `fd-input` already has `is-warning`, `is-success`, etc... Why do you bring them again?

InnaAtanasova (changes requested):

stories/file-uploader/file-uploader.stories.js
```diff
@@ -131,3 +131,102 @@ compact.parameters = {
...
+    type="file"
+    onchange="selectFile(this,'browse_input6')">
+</div>
+`;
+
+
+status.parameters = {
+    docs: {
+        iframeHeight: 250,
+        storyDescription: 'For Status File Uploader apply the `--compact` modifier to the button and input elements.'
```
kavya-b, line 230: is this the correct description? `--compact` for Status example?

platon-rov:
@manu-kr that PR cannot close #5025 as you provide changes only to the part of the components set we want to refactor, at least my part isn't merged yet :)

src/file-uploader.scss
manu-kr: i have tried to use ``is-error`, `is-success' with fd-input but styles are not applied. so it required to get them here. 

>

src/file-uploader.scss
manu-kr: removed.

stories/file-uploader/file-uploader.stories.js
manu-kr: modified.

src/file-uploader.scss
InnaAtanasova: Screen Shot 2021-04-20 at 10 06 36 AM

Hi @manu-kr the modifiers are not applied because the input is `readonly`. You don't have to introduce states in the File Uploader but check the states of Input and fix the problem there.

kavya-b:
Probably can change the title to `fix: move hardcoded styles from NGX for File Uploader component`? Also, in PR description, instead of `closes` it can be `part of` since other components cleanup are also part of that issue. Just a suggestion :)

droshev:
@manu-kr do you think a stakeholder would understand what was introduced as a feature with title `Style cleanup task #5025 move hard coded style from ngx to style for file uploader component` ?

src/file-uploader.scss
manu-kr: As i discussed with you i have removed readonly attribute for input field. and make some additional required changes in css. and it requires some logical changes in ngx library after this PR merge i can make those changes.

InnaAtanasova (approved):
Screen Shot 2021-04-23 at 1 36 23 PM

You are still missing the success state and it's important to have it as the border width is 1px, unlike the other states and it's good too have it visually tested.
The rest looks ok.

I changed the PR title and the description of the PR. Please try in future to write more meaningful PR titles that should be understood by everyone, not just devs working on the project. Thanks

kavya-b (approved):

PR #2272, opened 2021-04-23
fix: cleanup unused classes in Notifications and minor code refactoring

InnaAtanasova:
## Related Issue
NA

## Description
- removed unused and unnecessary classes
- deleted the rule that removes the border bottom from the last notification in the notification list/group
- added margin for every child inside actions, not only for buttons as we can have a popover 
- fixed the RTL rules for the actions

BREAKING CHANGE:
- removed classes: `fd-notification__actions--dismiss`, `fd-notification--tabs` and `fd-notification__actions--popover`

- [x] RTL support
NA
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/notification.scss
+++ b/src/notification.scss
@@ -269,8 +269,6 @@ $block: #{$fd-namespace}-notification;
     }
   }
 
-  // MOBILE view
-
   &--mobile {
     padding: $fd-notification-reset-offset;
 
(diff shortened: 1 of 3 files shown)


droshev:
We are missing show more/less (example).
Screen Shot 2021-05-02 at 11 23 05 AM

droshev (approved):
apart from missing show less/more example everything is good

InnaAtanasova:
> apart from missing show less/more example everything is good

It's not in the scope of this PR. I created a separate issue for this: https://github.com/SAP/fundamental-styles/issues/2325

PR #2273, opened 2021-04-26
fix: avatar zoom icon issues

### LINKED ISSUE #2111: Avatar: zoom icon issues
In Fiori 3 visual core, the "Zoom icon" has the alignment rule: 
image
In our current version, we do not follow this alignment:
image

N1XUS:
## Related Issue
Closes #2111

## Description
Fixed:
* Position of the zoom icon for circle avatar
* Vertical centering for the icons and text inside avatar

### Before:
image

### After:
image

- [x] All values are in `rem`
- [x]hover state of the element follow design spec
- [x] RTL support
- [x]`fd-reset()` mixin is applied to all elements
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/avatar.scss
+++ b/src/avatar.scss
@@ -65,11 +65,9 @@ $block: #{$fd-namespace}-avatar;
 
   &__icon {
     @include fd-icon-element-base() {
-      // Added in order to trully center the icon inside it's container
-      display: flex;
-      align-content: center;
-      justify-content: center;
-      align-items: center;
+      @include fd-flex-center() {
+        align-content: center;
+      }
     }
   }
 
--- a/src/avatar.scss
+++ b/src/avatar.scss
@@ -64,7 +64,13 @@ $block: #{$fd-namespace}-avatar;
   }
 
   &__icon {
-    @include fd-icon-element-base();
+    @include fd-icon-element-base() {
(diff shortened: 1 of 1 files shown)


src/avatar.scss
```diff
@@ -64,7 +64,13 @@ $block: #{$fd-namespace}-avatar;
...
 
   &__icon {
-    @include fd-icon-element-base();
+    @include fd-icon-element-base() {
+      // Added in order to trully center the icon inside it's container
+      display: flex;
+      align-content: center;
+      justify-content: center;
+      align-items: center;
+    }
```
InnaAtanasova, line 73: The `_mixins.scss` file contains reusable mixins such as `fd-flex-center`. 
Screen Shot 2021-04-28 at 11 41 51 AM

You can call here the mixin like this:
```
@include fd-flex-center() {
        align-content: center;
}
```

InnaAtanasova: You can remove the comment on line 68

N1XUS: Done. Please re-check

InnaAtanasova (approved):

InnaAtanasova:
@N1XUS you need a commit containing the text [ci visual] to trigger chromatic visual regression. Without it the checks won't pass.

PR #2275, opened 2021-04-27
fix: Align borders properly for fixed columns

### LINKED ISSUE #1948: Bug: Fixed column header borders are not aligned properly in the Table component
**Describe the bug**
In the table component a fixed column header borders are not aligned properly.

1. There is a missing border between the fixed column and the next column.
2. The border between the fixed column and the next column looks to be a bit too wide.
3. The row alignment between the fixed column and the rest of the row looks to be one pixel off.

Now:
image

Table
0.13.1

**To Reproduce**
1. Go to the Table Fix Column Header example 
https://fundamental-styles.netlify.app/?path=/docs/components-table--primary#fix-column-header
4. See borders issue

**Expected behavior**
Fixed column borders should be aligned with others 

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser chrome
 - Version 87.0.4280.66 (Official Build) (64-bit)

N1XUS:
## Related Issue
Closes #1948 

## Description
Change fixed column border width.
Shift fixed columns down so that bottom border would be visible

### Before:
image

### After:
image

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/table.scss
+++ b/src/table.scss
@@ -65,6 +65,7 @@ $block: #{$fd-namespace}-table;
   $fd-table-compact-cell-height: 2rem;
   $fd-table-navigation-icon-width: 2rem;
   $fd-table-condensed-cell-height: 1.5rem;
+  $fd-table-fixed-cell-shift: 0.05rem;
 
   // Thanks to table-cell specification, if the width is set to small value, the cell will have content width
   $fd-table-cell-fit-content-width: 0.5rem;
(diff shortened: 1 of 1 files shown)


src/table.scss
```diff
@@ -206,6 +206,7 @@ $block: #{$fd-namespace}-table;
       position: absolute;
       background-color: inherit;
       border-bottom: $fd-table-border;
+      margin-top: 0.05rem;
```
InnaAtanasova, line 209: I'm not sure what exactly this is fixing. 
the border around the cell is still missing:
Screen Shot 2021-04-28 at 12 03 33 PM

The first Header Column cell should have a border around all sides just like the second Header Column cell

N1XUS: @InnaAtanasova that line of code actually fixes the bottom border for the cells itself. Previously they weren't visible due to browser's antialiasing or something like that. I'll add borders for the heading. Also, maybe it's a good idea to change darker right border to the same border color we use across table?

N1XUS: @InnaAtanasova added missing borders. Please check.

InnaAtanasova (approved):

PR #2276, opened 2021-04-27
fix: vertically align close button for compact token

### LINKED ISSUE #1731: Token in compact mode is not vertically aligned
Screen Shot 2020-09-29 at 3 00 33 PM

The close icon is not vertically aligned in compact Tokens

N1XUS:
## Related Issue
Closes #1731

## Description
Change close button padding to vertically center the icon inside.

### Before:
image

### After:
image

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/token.scss
+++ b/src/token.scss
@@ -103,13 +103,13 @@ $block: #{$fd-namespace}-token;
   @include fd-token-focus();
 
   &--compact {
-    padding: 0 0 $fd-token-compact-padding-vertical $fd-token-compact-padding-horizontal;
+    padding: $fd-token-compact-padding-vertical 0 $fd-token-compact-padding-vertical $fd-token-compact-padding-horizontal;
     height: $fd-token-compact-height;
     min-width: 3rem;
     line-height: 1.125rem;
 
     .#{$block}__close {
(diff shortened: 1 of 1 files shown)


src/token.scss
```diff
@@ -109,7 +109,7 @@ $block: #{$fd-namespace}-token;
     line-height: 1.125rem;
 
     .#{$block}__close {
-      padding: $fd-token-compact-padding-vertical $fd-token-compact-padding-horizontal;
+      padding: ($fd-token-compact-padding-vertical * 2) $fd-token-compact-padding-horizontal $fd-token-compact-padding-vertical;
```
InnaAtanasova, line 112: I'm not sure it's a good idea to change the paddings. You can easily fix the issue by changing the flex `alig-items` rule to normal:
Screen Shot 2021-04-28 at 12 19 22 PM

InnaAtanasova: You will get this:
Screen Shot 2021-04-28 at 12 21 14 PM

N1XUS: @InnaAtanasova done in a bit different way. Please re-check.

InnaAtanasova (approved):
@N1XUS you need a commit containing the text [ci visual] to trigger chromatic visual regression. Without it the checks won't pass.

PR #2278, opened 2021-04-28
feat: implement vertical nav

### LINKED ISSUE #2211: Implement Vertical Navigation
The Vertical Navigation consists technically of the Side Navigation component.
It can be used as an alternative to the Horizontal Navigation.
This navigation type is composed by two distinct item levels: First Level items and Second Level items.

- [ ] **Start with Navigation List Item** - states, interactivity, first-level and second-level, navigation indication
- [ ] Put a `deprecate` label on the current Side Navigation
- [ ] Implement a new component Side Navigation that will be used together with Horizontal Navigation and instead of the old Side Nav
- [ ] Visual states - expanded, condensed, off-screen
- [ ] Visual states - use-cases
- [ ] "Drawer"
- [ ] Grouping
- [ ] Navigation Indication - first-level and second-level

mikerodonnell89:
closes #2211 

## Description
Implements the vertical navigation and deprecates side navigation

### After:
Screen Shot 2021-04-28 at 9 06 26 AM
Screen Shot 2021-04-28 at 9 06 13 AM
Screen Shot 2021-04-28 at 9 06 08 AM
Screen Shot 2021-04-28 at 9 06 03 AM

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/_listNavigationItem.scss
+++ b/src/_listNavigationItem.scss
@@ -1,14 +1,18 @@
-@import "./listDefinitions.scss";
+@import './listDefinitions.scss';
 
-$fd-list-navigation-item-height: 2.75rem;
+$fd-list-navigation-item-height: 2.75rem !default;
+$fd-list-navigation-item-indicator-spacing: 0.3125rem !default;
+$fd-list-navigation-item-text-spacing: 2.5rem !default;
 
 .#{$block} {
   list-style-type: none;
   background: var(--sapList_AlternatingBackground);
 
   .#{$block}__navigation-item-indicator {
+    @include fd-reset();
+
     &::after {
-      content: "";
+      content: '';
       width: 0.1875rem;
       display: inline-block;
       position: absolute;
@@ -18,7 +22,7 @@ $fd-list-navigation-item-height: 2.75rem;
       background-color: var(--sapSelectedColor);
       color: var(--sapSelectedColor);
       border-radius: var(--sapField_BorderCornerRadius);
-      margin-right: 0.3125rem;
+      margin-right: $fd-list-navigation-item-indicator-spacing;
     }
 
     @include fd-rtl() {
@@ -26,7 +30,7 @@ $fd-list-navigation-item-height: 2.75rem;
         right: unset;
         margin-right: 0;
         left: 0;
-        margin-left: 0.3125rem;
+        margin-left: $fd-list-navigation-item-indicator-spacing;
       }
     }
   }
@@ -34,39 +38,30 @@ $fd-list-navigation-item-height: 2.75rem;
   &__navigation-item {
     @include fd-reset();
     @include fd-fiori-focus();
+    @include set-paddings-x(0.5rem, 1rem);
 
     background: transparent;
     height: $fd-list-navigation-item-height;
-    width: auto;
-    padding-left: 0.5rem;
-    padding-right: 1rem;
     position: relative;
-
-    @include fd-rtl() {
-      padding-right: 0.5rem;
-      padding-left: 1rem;
-    }
 
     .fd-list {
       display: none;
     }
 
     &--expandable {
-      margin-top: 0.25rem;
-      margin-bottom: 0.25rem;
+      margin: 0.25rem 0;
     }
 
     &.is-expanded {
-      margin-left: 0.25rem;
-      margin-right: 0.25rem;
+      @include set-margins-x(0.25rem, 0.25rem);
+
       border-radius: var(--sapElement_BorderCornerRadius);
       box-shadow: var(--sapContent_Shadow0);
-      height: auto;
+      height: 100%;
       overflow: hidden;
 
       .#{$block}__navigation-item {
-        width: auto;
-
+        width: 100%;
         .#{$block}__navigation-item-indicator {
           &::after {
             margin-right: 0;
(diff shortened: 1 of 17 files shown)


InnaAtanasova:
Visual comments:
1. The height of the item should be 44px but results in 45.31px
Screen Shot 2021-04-28 at 4 01 10 PM

2. Text truncation missing:
Screen Shot 2021-04-28 at 4 02 00 PM
Screen Shot 2021-04-28 at 4 02 33 PM

3. The "arrow" in my opinion should be a button, o.w. the user won't be able to navigate to it with the keyboard (a11y). It should also overlap with the right padding as shown in the image:
Screen Shot 2021-04-28 at 4 07 38 PM

How it's now:
Screen Shot 2021-04-28 at 4 09 51 PM

4. RTL is not handled properly:
Screen Shot 2021-04-28 at 4 39 03 PM
Screen Shot 2021-04-28 at 4 38 54 PM

5. In my opinion the "indicators" for navigation and second level should be added with modifier classes like `.block__element--navigared` or `.block__element--hasChild` and add them with ::before and ::after. This way it will be easy to handle RTL. You can check the current implementation of Side Nav where we have second level indicator.
Screen Shot 2021-04-28 at 4 41 36 PM

6. Expanded state:
The "drawer" is not implemented in default state:
Screen Shot 2021-04-28 at 4 52 30 PM

Screen Shot 2021-04-28 at 4 49 37 PM

How it is:
Screen Shot 2021-04-28 at 4 50 01 PM

7. Missing the "inner" radius:
Screen Shot 2021-04-28 at 4 59 20 PM
Screen Shot 2021-04-28 at 4 59 14 PM

src/_listNavigationItem.scss
```diff
@@ -0,0 +1,274 @@
...
+      top: 0.5625rem; // (height of nav item - height of indicator) / 2
+      height: 1.625rem;
+      background-color: var(--sapSelectedColor);
+      color: var(--sapSelectedColor);
+      border-radius: var(--sapField_BorderCornerRadius);
+      margin-right: 0.3125rem;
+    }
+  }
+
+  &__navigation-item {
```
InnaAtanasova, line 22: missing reset

src/_listNavigationItem.scss
```diff
@@ -0,0 +1,274 @@
...
+.#{$block} {
+  list-style-type: none;
+
+  .#{$block}__navigation-item-indicator {
+    &::after {
+      content: "";
+      width: 0.1875rem;
+      display: inline-block;
+      position: absolute;
+      right: 0;
```
InnaAtanasova, line 12: this should prob change in RTL

src/_listNavigationItem.scss
```diff
@@ -0,0 +1,274 @@
...
+      background-color: var(--sapSelectedColor);
+      color: var(--sapSelectedColor);
+      border-radius: var(--sapField_BorderCornerRadius);
+      margin-right: 0.3125rem;
+    }
+  }
+
+  &__navigation-item {
+    background: var(--sapList_AlternatingBackground);
+    min-height: 2.75rem;
```
InnaAtanasova, line 24: the height value should be in a variable

src/_listNavigationItem.scss
```diff
@@ -0,0 +1,274 @@
...
+    width: 13.5rem;
+    padding-left: 0.5rem;
+    padding-right: 1rem;
```
InnaAtanasova, line 27: these should be in variables and reused in RTL

src/_listNavigationItem.scss
```diff
@@ -0,0 +1,274 @@
...
+        padding-left: 0.375rem;
+      }
+
+      .#{$block}__navigation-item-icon {
+        padding-right: 0.25rem;
+      }
+
+      .#{$block}__navigation-item-arrow {
+        padding-left: 2.125rem;
+      }
```
InnaAtanasova, line 50: missing RTL for these

src/_listNavigationItem.scss
```diff
@@ -0,0 +1,274 @@
...
+      display: none;
+    }
+
+    &.is-expanded {
+      border-radius: var(--sapElement_BorderCornerRadius);
+      box-shadow: inset 0 0 0 0.0625rem var(--sapTile_BorderColor);
+      width: 13.25rem;
+      margin: 0.25rem auto;
+
+      .#{$block}__navigation-item-text {
```
InnaAtanasova, line 40: missing reset

src/_listNavigationItem.scss
```diff
@@ -0,0 +1,274 @@
...
+        }
+      }
+
+      .#{$block}__navigation-item-text {
+        display: none;
+      }
+
+      .#{$block}__navigation-item--second {
+        display: none;
+        width: auto;
```
InnaAtanasova, line 124: this won't work in IE11

src/_listNavigationItem.scss
```diff
@@ -0,0 +1,274 @@
...
+              margin-left: 0.175rem;
+              margin-right: 0.175rem;
+            }
+          }
+
+          &--second {
+            border-top-right-radius: 0;
+            border-top-left-radius: 0;
+            top: 2.75rem;
+            left: 3.25rem;
```
InnaAtanasova, line 176: missing RTL

src/_listNavigationItem.scss
```diff
@@ -0,0 +1,274 @@
...
+                  margin-right: 0;
+                  display: inline-block;
+                }
+              }
+            }
+          }
+
+          &--first {
+            top: 0;
+            left: 0.325rem;
```
InnaAtanasova, line 161: missing RTL

InnaAtanasova (commented):
Your implementation has the following structure:
```
<ul class="fd-list">
      <li class="fd-list__navigation-item--condensed"></li>
      <li class="fd-list__navigation-item--condensed"></li>
</ul>
```

You can't have a modifier class without a base class. You need to have something like:
```
<ul class="fd-list">
      <li class="fd-list__navigation-item fd-list__navigation-item--condensed"></li>
      <li class="fd-list__navigation-item fd-list__navigation-item--condensed"></li>
</ul>
```
where `fd-list__navigation-item--condensed` will overwrite the rules of `fd-list__navigation-item` but this way you will need to apply this modifier class on every single `li` item. You can avoid this with specificity, aka have the modifier class applied on the parent level. For example:
```
.fd-list.fd-list--condensed {
    .fd-list__navigation-item {
         ....
    }
}
```

Same is valid for `fd-list__navigation-item--second`. Instead of applying this modifier class on every child you can apply it on the parent `fd-list fd-list--second-level`.

droshev:
a minor focus issue in IE11
Screen Shot 2021-05-02 at 11 09 25 AM
Screen Shot 2021-05-02 at 11 09 19 AM

InnaAtanasova:
Closing in favour of [this PR ](https://github.com/SAP/fundamental-styles/pull/2349)because of build issues and rebase main didn't help.

PR #2280, opened 2021-04-29
fix: fix the alignment issue between input box and description/unit value for the step input component.

manu-kr:
## Related Issue
Part of SAP/fundamental-ngx#5025

## Description
Changes contains to move move hard coded style from ngx to  style to fix the alignment issue between input box and description.

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/step-input.scss
+++ b/src/step-input.scss
@@ -86,7 +86,7 @@ $block: #{$fd-namespace}-step-input;
 
   &__container {
     display: flex;
-	  align-items: center;
+    align-items: center;
 
     .fd-form-label {
       align-self: auto;
--- a/src/step-input.scss
+++ b/src/step-input.scss
@@ -84,6 +84,15 @@ $block: #{$fd-namespace}-step-input;
     }
   }
 
+  &__container {
+    display: flex;
+	  align-items: center;
+
+    .fd-form-label {
+      align-self: auto;
+    }
+  }
+
   /** Width is different due to different border width  */
   &.is-information,
   &.is-error,
--- a/src/step-input.scss
+++ b/src/step-input.scss
@@ -85,12 +85,7 @@ $block: #{$fd-namespace}-step-input;
   }
 
   &__container {
-    display: flex;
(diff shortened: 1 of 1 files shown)


droshev:
@manu-kr can you update the title of the PR to be more meaningful with what the change is about?

manu-kr:
> @manu-kr can you update the title of the PR to be more meaningful with what the change is about?

>@droshev i have update the PR, please let me know if still it require. changes.

src/step-input.scss
```diff
@@ -84,6 +84,15 @@ $block: #{$fd-namespace}-step-input;
     }
   }
 
+  &__container {
+    display: flex;
+    align-items: center;
```
InnaAtanasova, line 89: Hi @manu-kr , there are checkboxes in the PR template that are put there with a purpose, not just to waste people's time to check them when they open a PR. One of these checks is about mixins:

Screen Shot 2021-04-29 at 11 10 33 AM

In `_mixins.scss` file there's a mixin called `fd-flex-vertical-center` which is basically this:
Screen Shot 2021-04-29 at 11 13 33 AM

You can use it in your code.

src/step-input.scss
```diff
@@ -84,6 +84,15 @@ $block: #{$fd-namespace}-step-input;
...
+
+    .fd-form-label {
```
InnaAtanasova, line 91: You can't use this directly here as it's violating the self-contained rules.

InnaAtanasova (changes requested):

src/step-input.scss
manu-kr: Hi @InnaAtanasova , Sorry i have missed this. I have reused this now.

src/step-input.scss
manu-kr: removed.

InnaAtanasova (approved):

droshev (approved):

PR #2296, opened 2021-04-29
fix: Avatar group - do not overlay outer shadow with inner component's shadow [ci visual]

N1XUS:
## Related Issue
Closes https://github.com/SAP/fundamental-styles/issues/2153

## Description
Add `overflow: hidden` for inner elements of `fd-popover__body` which will prevent inner component's shadow to overlay outer one.

### Before:
image

### After:
image

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/popover.scss
+++ b/src/popover.scss
@@ -102,6 +102,11 @@ $block: #{$fd-namespace}-popover;
       z-index: $fd-popover-z-index + 1;
     }
 
+    // Inner shadows can affect outer shadow. This will discard such behavior
+    > * {
+      overflow: hidden;
+    }
+
     > *:first-child {
       border-top-right-radius: $fd-popover-border-radius;
       border-top-left-radius: $fd-popover-border-radius;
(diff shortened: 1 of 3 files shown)


src/popover.scss
```diff
@@ -102,6 +102,11 @@ $block: #{$fd-namespace}-popover;
       z-index: $fd-popover-z-index + 1;
     }
 
+    // Inner shadows can affect outer shadow. This will discard such behavior
+    > * {
+      overflow: hidden;
+    }
```
InnaAtanasova, line 108: We can't use this solution unfortunately. We have cases where the size of the popover body is limited and the scroll should appear, but if we add this solution the scroll no longer works. Please check "Scrollable" example of Popover. https://deploy-preview-2296--fundamental-styles.netlify.app/?path=/docs/components-popover--alignment#scrollable

The scroll is not working, if you remove the `overflow: hidden;` it will appear.

InnaAtanasova (changes requested):

N1XUS:
@InnaAtanasova I've updated the selector also checked with all examples from popover and corrected one demo component in the documentation. Now should be fine. Please re-check.

InnaAtanasova (approved):

Issue #2200, opened 2021-03-16
Wizard - stacking bug with specific case of step completion


mikerodonnell89:
While working on this PR I noticed a bug: https://github.com/SAP/fundamental-ngx/pull/4901

On the default example, when I navigate to step 3, then back to step 1, I get the following stacking behavior for the step indicators on the right:
![Screen Shot 2021-03-15 at 6 17 34 PM](https://user-images.githubusercontent.com/2471874/111237859-8d3e9700-85bb-11eb-9c5c-ecfe1000a23b.png)
![Screen Shot 2021-03-15 at 6 17 21 PM](https://user-images.githubusercontent.com/2471874/111237868-90d21e00-85bb-11eb-8029-0d569341eae4.png)

This should be a valid use case - stacked steps where the first is --completed but the two following are --upcoming

PR #2324, opened 2021-05-03
fix: Wizard - change styling priorities in order to apply correct styles for stacked steps

N1XUS:
## Related Issue
fixes https://github.com/SAP/fundamental-styles/issues/2200

## Description
change styling priorities in order to apply correct styles for stacked steps

### Before:
image

### After:
image

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/wizard.scss
+++ b/src/wizard.scss
@@ -69,20 +69,6 @@ $fd-wizard-stacked-upcoming-step-offset: 2.375rem !default;
     vertical-align: top;
 
     // --- STEP MODIFIERS ---
-    &--completed {
-      .#{$block}__step-indicator {
-        background: var(--sapObjectHeader_Background);
-        color: var(--fdWizard_Completed_Step_Color);
-        border: var(--fdWizard_Completed_Step_Border);
-
-        .#{$block}__icon {
-          @include fd-icon-selector() {
-            color: var(--fdWizard_Completed_Step_Icon_Color);
-          }
-        }
-      }
-    }
-
     &--current {
       .#{$block}__step-indicator {
         position: relative;
@@ -127,6 +113,20 @@ $fd-wizard-stacked-upcoming-step-offset: 2.375rem !default;
       .#{$block}__step-container {
         cursor: default;
         pointer-events: none;
+      }
+    }
+
+    &--completed {
+      .#{$block}__step-indicator {
+        background: var(--sapObjectHeader_Background);
+        color: var(--fdWizard_Completed_Step_Color);
+        border: var(--fdWizard_Completed_Step_Border);
+
+        .#{$block}__icon {
+          @include fd-icon-selector() {
+            color: var(--fdWizard_Completed_Step_Icon_Color);
+          }
+        }
       }
     }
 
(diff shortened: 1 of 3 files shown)


droshev:
@N1XUS Can you add an example so we can have it covered?

src/wizard.scss
```diff
@@ -130,6 +116,20 @@ $fd-wizard-stacked-upcoming-step-offset: 2.375rem !default;
...
+        color: var(--fdWizard_Completed_Step_Color);
+        border: var(--fdWizard_Completed_Step_Border);
+
+        .#{$block}__icon {
+          @include fd-icon-selector() {
+            color: var(--fdWizard_Completed_Step_Icon_Color);
+          }
+        }
+      }
+    }
```
InnaAtanasova, line 131: @N1XUS what is the change here? You moved the `--completed` modifier to be after `--upcoming`?

N1XUS: Yes, my intention is to keep the `upcoming` class with `completed`. The logic here is that even if the step is completed, but we moved backward, it will be upcoming to refill/skip. By moving the code, I've changed the priority of the styling, which should do the trick with the ngx PR

N1XUS:
Please refer to this section in documentation to see the implemented solution: https://deploy-preview-2324--fundamental-styles.netlify.app/iframe.html?id=components-wizard--default-example&viewMode=docs#reverted-steps

src/wizard.scss
mikerodonnell89: I think this is an unsafe way to address the problem that is prone to future problems, we should either create an additional class or have more specificity for "completed but also upcoming"

mikerodonnell89 (changes requested):

src/wizard.scss
InnaAtanasova: Agree with Mike. We experienced multiple problems in the past due to the order of the classes in the file or the order of file import. This can be easily solved with specificity.

N1XUS: Done. Please re-check.

src/wizard.scss
```diff
@@ -9,6 +9,20 @@ $fd-wizard-step-no-label-max-size: 4rem !default;
 $fd-wizard-step-indicator-stacked-offset: -1 * ($fd-wizard-step-indicator-size - $fd-wizard-step-indicator-spacing + 0.25rem) !default;
 $fd-wizard-stacked-upcoming-step-offset: 2.375rem !default;
 
+@mixin fd-wizard-completed-step($block) {
```
InnaAtanasova, line 12: do you need to pass $block as argument?

InnaAtanasova: You can also set the default value and then you don't have to pass it when you include the mixin.

N1XUS: @InnaAtanasova yes, $block was not needed. Removed.

InnaAtanasova (approved):

mikerodonnell89:
@N1XUS in the future you need the word "fixes" or "closes" before the related issue number in the PR description so that the corresponding issue closes automatically.  I've added it here.

I built this and copied it over to fundamental-ngx and used your branch here: https://github.com/SAP/fundamental-ngx/pull/5272

And when navigating the 3rd step, then back to step 1, I see this: 
Screen Shot 2021-05-06 at 12 04 58 PM

I believe step 2 needs to be styled blue as it has technically been completed (but is also still upcoming)

mikerodonnell89 (approved):
nice work!

droshev (approved):

PR #2326, opened 2021-05-04
feat: introduce Fixed Card Layout component

DeepakSap14:
## Related Issue
part of https://github.com/SAP/fundamental-ngx/issues/5025

## Description
Fixed card layout component styles was not created in styles and was present in Ngx.
This can be moved to styles now. when this is merged and available, styles from Ngx can be removed and can be imported from styles.

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/fixed-card-layout.scss
+++ b/src/fixed-card-layout.scss
@@ -1,0 +1,79 @@
+@import './new-settings';
+@import './mixins';
+
+$block: #{$fd-namespace}-fixed-card-group;
+
+.#{$block} {
+  $fd-border-on-drag: 0.125rem dashed var(--sapContent_DragAndDropActiveColor);
+  $fd-card-margin: 1rem;
+  $fd-card-min-width: 20rem;
+
+  @include fd-reset();
+
+  &--card-layout {
+    @include fd-reset();
+    @include fd-flex();
+
+    justify-content: center;
+
+    * {
+      word-break: break-all;
+    }
+  }
+
+  &--card-layout-column {
+    @include fd-reset();
+
+    flex: 1;
+    margin-right: $fd-card-margin;
+    margin-left: 0;
+
+    @include fd-rtl() {
+      margin-left: $fd-card-margin;
+      margin-right: 0;
+    }
+  }
+
+  &--card-layout-column:last-child {
+    margin-right: 0;
+    margin-left: 0;
+
+    @include fd-rtl() {
+      margin-left: 0;
+      margin-right: 0;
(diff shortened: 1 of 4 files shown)


src/fixed-card-layout.scss
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  &--card-layout-column:last-child {
+    margin-right: 0;
+    margin-left: 0;
+
+    @include fd-rtl() {
+      margin-left: 0;
+      margin-right: 0;
+    }
```
InnaAtanasova, line 44: do you need RTL when both values are the same and equal to 0?

src/fixed-card-layout.scss
```diff
@@ -0,0 +1,79 @@
...
+      word-break: break-all;
+    }
+  }
+
+  &--card-layout-column {
+    @include fd-reset();
+
+    flex: 1;
+    margin-right: $fd-card-margin;
+    margin-left: 0;
```
InnaAtanasova, line 29: do you need this? Isn't it 0 by default?

src/fixed-card-layout.scss
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  &--card {
+    @include fd-reset();
+
+    margin-bottom: $fd-card-margin;
+    margin-right: 0;
+    min-width: $fd-card-min-width;
+
+    .fd-card__content {
```
InnaAtanasova, line 54: can't do this

InnaAtanasova (changes requested):
None of the class names is following BEM

src/fixed-card-layout.scss
DeepakSap14: only margin-left: 0 is needed in rtl, because __card-layout-column is setting margin-left in rtl, but for last column it is not needed.

src/fixed-card-layout.scss
DeepakSap14: removed

src/fixed-card-layout.scss
DeepakSap14: removed this now. Earlier card data used to flow, but it has been fixed in card. so not needed.

DeepakSap14:
> None of the class names is following BEM

changed class name formatting. These class name will be needed to change in Ngx as well, post this PR is merged and used in Ngx.

InnaAtanasova (approved):

Issue #1640, opened 2020-09-17
outline broken for radio button when using voiceover


DeepakSap14:
**Describe the bug**
Outline for voiceover is broken for radio button. see screenshot below.

> Is this issue related to a specific component?
Radio button

> What version of the Fundamental Library Styles are you using?
0.11
> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

> What front-end framework are you implementing in, e.g., Angular, React, etc.?
Angular

**To Reproduce**
1. Go to 'https://sap.github.io/fundamental-styles/?path=/docs/components-forms-radio--primary'
2. start voiceover
3. tab on radio

**Expected behavior**
outline should be proper as shown in below screenshot.

**Screenshots**
When VoiceOver is on, the outline is broken.
<img width="176" alt="Screenshot 2020-09-16 at 16 39 19" src="https://user-images.githubusercontent.com/8318592/93345199-3558ea80-f83b-11ea-981c-2c663148998f.png">
But should be like this
<img width="324" alt="Screenshot 2020-09-16 at 16 41 22" src="https://user-images.githubusercontent.com/8318592/93345448-7f41d080-f83b-11ea-9494-56be30bd2a36.png">

**Desktop (please complete the following information):**
 - Browser [e.g. chrome, safari] 'chrome'

**Smartphone (please complete the following information):**

**Additional context**

PR #2332, opened 2021-05-07
fix: voiceover outline for radio buttons

aberikashvili:
## Related Issue
Closes https://github.com/SAP/fundamental-styles/issues/1640

## Description
PR fixes broken outline for radio buttons, when voiceover is on

### Before:
image

### After:
image

- [x] All values are in `rem`
- [N/A] RTL support
3. Testing
- [N/A ] Verified all styles in IE11
- [x] Updated tests
4. Documentation


package.json
```diff
@@ -8,7 +8,7 @@
     "build:default": "npm run style:compile && npm run style:theme && npm run style:postCSS && npm run style:assets",
     "build:prod": "cross-env NODE_ENV=production npm run build",
     "deploy": "gh-pages -d storybook-static/",
-    "lint:style": "stylelint '{src,.storybook}/**/*.(sa|sc|c)ss'",
+    "lint:style": "stylelint \"{src,.storybook}/**/*.(sa|sc|c)ss\"",
```
droshev, line 11: is this part of the issue or it happened with autoformatting?

droshev (changes requested):
@aberikashvili is this a WIP or ready for review?
If it is WIP then the the PR title should start with `[WIP]`, if not then you need to maintain the PR a little bit more and cover the items from the template

package.json
aberikashvili: @droshev this isn't part of issue, this is windows specific fix, I ran into the problem when I was running `npm run lint:style`

droshev: @aberikashvili then can you open this changes in a separate PR?

src/radio.scss
```diff
@@ -47,6 +49,11 @@ $block: #{$fd-namespace}-radio;
 
   @extend %fd-radio-input-hidden;
 
+  &__wrapper {
```
N1XUS, line 52: image
Same issue repeating with a checkbox. @droshev should it be fixed as a separate issue?

aberikashvili: @N1XUS I think so, checkbox is different component right?

mikerodonnell89:
Trying to find a way to do this without breaking changes, @InnaAtanasova any ideas?  Can't wrap my head around why the screenreader's focus outline would be different from the focus outline we've defined

InnaAtanasova (changes requested):
I can no longer navigate the radio buttons with the Tab key.

stories/radio/radio.stories.js
```diff
@@ -25,45 +25,57 @@ export const primary = () => `<div style="display:flex;justify-content:space-bet
         <legend class="fd-fieldset__legend">Radio Buttons Cozy Mode</legend>
         <div class="fd-form__group">
             <div class="fd-form-item">
-                <input type="radio" class="fd-radio" id="pDidh761" name="radio1" checked>
-                <label class="fd-radio__label" for="pDidh761">
-                    Field label
-                </label>
+                <div class="fd-radio__wrapper">
+                    <input type="radio" class="fd-radio" id="pDidh761" name="radio1" checked>
```
InnaAtanasova, line 29: I am not sure we are following BEM with this approach. In this case you have `fd-radio__wrapper` as a parent element for `fd-radio` which is a BEM violation. A class with `__something` must be an element of a block class

InnaAtanasova:
Closing this PR. Feel free to open it when the comments are addressed.

Issue #2338, opened 2021-05-08
accessibility issue missing the example with label busy indicator


Lokanathan-k:
Description: There is no visual representation ‘please wait’ for the busy indicator as highlighted in red.

Expected: There is no visual representation "Please wait" for the busy indicator as highlighted in red 

Attached the screenshot
![image](https://user-images.githubusercontent.com/32538291/91549532-b57ae700-e944-11ea-8489-7e92ac8db56c.png)
 
expected output:
![image](https://user-images.githubusercontent.com/53534379/117546788-d81fdc80-b049-11eb-80b9-5f7e075543a2.png)

Lokanathan-k:
closes as part of the PR:https://github.com/SAP/fundamental-styles/pull/2339

PR #2339, opened 2021-05-08
feat: add Extended Busy Indicator

Lokanathan-k:
## Related Issue
Closes:https://github.com/SAP/fundamental-styles/issues/2338

## Description
adding the example for the label with busy indicator

Introduced example:
image
 

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation


droshev (changes requested):
I think it requires more styling for the extended busy indicator:
Screen Shot 2021-05-09 at 11 02 59 AM

kavya-b:
- `width` value is missing
- longer text should be wrapped as per spec, but is getting truncated: 
image
- please avoid using `!important`. You can also probably use the generic padding modifiers from `paddings.scss`
- why is `fd-form-label` being used? It is overriding the text color:
image

src/busy-indicator.scss
```diff
@@ -79,6 +94,11 @@ $fd-dots-size-l: 2rem;
   height: $fd-dots-size-l;
 }
 
+.#{$block}--label {
+  display: block;
+  margin: auto !important;
```
kavya-b, line 99: avoid using `!important`

stories/busy-indicator/busy-indicator.stories.js
```diff
@@ -68,6 +69,37 @@ export const contrastMode = () => `<div style="display:flex;justify-content:cent
...
+        <div class="fd-busy-indicator--circle-1"></div>
+        <div class="fd-busy-indicator--circle-2"></div>
+		<span class=" fd-busy-indicator--enhanced-indicator-text fd-busy-indicator--label fd-form-label">loading data...</span>
+    </div>`;
+
+ExtendedIndicator.parameters = {
+    docs: {
+        iframeHeight: 200,
+        storyDescription:
+            'The standard busy indicator animates a sequence of cascading dots expanding and shrinking in a loop. The component comes with a label/message to be displayed when the data is being loaded. <code>fd-busy-indicator--label</code> is used to display the message.'
```
kavya-b, line 103: can you modify this to explain more specifically for this example?

stories/busy-indicator/busy-indicator.stories.js
```diff
@@ -68,6 +69,37 @@ export const contrastMode = () => `<div style="display:flex;justify-content:cent
...
+
+LabelledIndicator.parameters = {
+    docs: {
+        iframeHeight: 200,
+        storyDescription:
+            'The standard busy indicator animates a sequence of cascading dots expanding and shrinking in a loop. The component comes with a label/message to be displayed when the data is being loaded. <code>fd-busy-indicator--label</code> is used to display the message.'
+    }
+};
+export const ExtendedIndicator = () => `<div style="text-align: center">
+    <div class="fd-busy-indicator--l fd-busy-indicator--enhanced-indicator" aria-hidden="false" aria-label="Loading">
```
kavya-b, line 92: shouldn't this be used with the `fd-busy-indicator` class since these are modifiers? Same on line 96

stories/busy-indicator/busy-indicator.stories.js
```diff
@@ -68,6 +69,37 @@ export const contrastMode = () => `<div style="display:flex;justify-content:cent
...
+        <div class="fd-busy-indicator--circle-0"></div>
+        <div class="fd-busy-indicator--circle-1"></div>
+        <div class="fd-busy-indicator--circle-2"></div>
+		<span class=" fd-busy-indicator--enhanced-indicator-text fd-busy-indicator--label fd-form-label">loading data...</span>
+    </div>`;
+
+ExtendedIndicator.parameters = {
```
kavya-b, line 99: The modifier is called `enhanced-indicator` but the example name is `Extended Indicator`.  Can this be made consistent?

kavya-b (changes requested):

src/busy-indicator.scss
```diff
@@ -79,6 +88,16 @@ $fd-dots-size-l: 2rem;
   height: $fd-dots-size-l;
 }
 
+.#{$block}--label {
+  display: block;
+  margin: auto;
```
kavya-b, line 93: do we need `margin:auto` here?

stories/busy-indicator/busy-indicator.stories.js
```diff
@@ -68,6 +69,37 @@ export const contrastMode = () => `<div style="display:flex;justify-content:cent
 contrastMode.parameters = {
     docs: {
         iframeHeight: 200,
-        storyDescription: 'The busy indicator also comes in contrast mode and displays white dots against a dark background. To apply contrast mode, add <code>contrast</code> into the element i.e. <code>fd-busy-indicator--m contrast</code>.'
+        storyDescription:
+            'The busy indicator also comes in contrast mode and displays white dots against a dark background. To apply contrast mode, add <code>contrast</code> into the element i.e. <code>fd-busy-indicator--m contrast</code>.'
+    }
+};
+export const LabelledIndicator = () => `<div style="text-align: center">
```
kavya-b, line 76: why do we have this inline style here and not in the scss file?

stories/busy-indicator/busy-indicator.stories.js
```diff
@@ -26,7 +26,7 @@ The ongoing operation only covers part of a screen that has multiple controls, a
 - You need to block the screen because the user is not supposed to start another operation. In this case, use the **Busy Dialog** component.
         `,
         tags: ['f3', 'a11y', 'theme'],
-        components: ['busy-indicator']
+        components: ['busy-indicator', 'form-label']
```
kavya-b, line 29: if not used, can be removed

stories/busy-indicator/busy-indicator.stories.js
```diff
@@ -68,6 +69,37 @@ export const contrastMode = () => `<div style="display:flex;justify-content:cent
...
+        <div class="fd-busy-indicator--circle-1"></div>
+        <div class="fd-busy-indicator--circle-2"></div>
+		<span class="fd-busy-indicator--label">loading data...</span>
+    </div>`;
+
+ExtendedIndicator.parameters = {
+    docs: {
+        iframeHeight: 200,
+        storyDescription:
+            'The standard busy indicator animates a sequence of cascading dots expanding and shrinking in a loop. The component comes with a dialog and a label to be displayed when the data is being loaded. <code>fd-busy-indicator--extended-indicator</code> is used to wrap the busyindiactor in a dialog.'
```
kavya-b, line 103:
```suggestion
            'The standard busy indicator animates a sequence of cascading dots expanding and shrinking in a loop. The component comes with a dialog and a label to be displayed when the data is being loaded. <code>fd-busy-indicator--extended-indicator</code> is used to wrap the busy indicator in a dialog.'
```
Also, just thinking about the modifier names. We have `--label` and `--extended-indicator`; instead how does `--labelled` and `--extended` sound? Just a suggestion.
Actually,  what is the difference between the label and extended indicator? Spec only shows extended indicator. The one within dialog is a `busy dialog`, which has a button additionally, so I don't think this component should have these lines- `...comes with a dialog...` 
image

kavya-b (changes requested):

src/busy-indicator.scss
```diff
@@ -79,6 +88,16 @@ $fd-dots-size-l: 2rem;
   height: $fd-dots-size-l;
 }
 
+.#{$block}--label {
```
InnaAtanasova, line 91: This is not a modifier class. You can't have `fd-busy-indicator--label`. You need `fd-busy-indicator__label`. You then need to include the reset function.

src/busy-indicator.scss
```diff
@@ -79,6 +88,16 @@ $fd-dots-size-l: 2rem;
...
+  display: block;
+  margin: auto;
+  font-size: var(--sapFontSize);
+  font-family: var(--sapFontFamily);
+  font-weight: normal;
```
InnaAtanasova, line 96: These things are coming from the reset function that you are missing

stories/busy-indicator/busy-indicator.stories.js
```diff
@@ -68,6 +69,37 @@ export const contrastMode = () => `<div style="display:flex;justify-content:cent
...
+
+LabelledIndicator.parameters = {
+    docs: {
+        iframeHeight: 200,
+        storyDescription:
+            'The standard busy indicator animates a sequence of cascading dots expanding and shrinking in a loop. The component comes with a label/message to be displayed when the data is being loaded. <code>fd-busy-indicator--label</code> is used to display the message.'
+    }
+};
+export const ExtendedIndicator = () => `<div style="text-align: center">
+    <div class="fd-busy-indicator--l fd-busy-indicator--extended-indicator fd-padding" aria-hidden="false" aria-label="Loading">
```
InnaAtanasova, line 92: You can't have this: `<div class="fd-busy-indicator--l fd-busy-indicator--extended-indicator fd-padding">`
You need a base class to modify it:
<div class="fd-busy-indicator fd-busy-indicator--l fd-busy-indicator--extended-indicator fd-padding" aria-hidden="false" aria-label="Loading">

InnaAtanasova (dismissed):

stories/busy-indicator/busy-indicator.stories.js
Lokanathan-k: @InnaAtanasova  on checking the previous example found that it does not require any base class, so a new modifier class has been added. 

image

stories/busy-indicator/busy-indicator.stories.js
Lokanathan-k: the difference is the action/ link present in the dialog box, above screen shot is already added in the dialogue.

the current example was added on the suggestion from @droshev 
https://github.com/SAP/fundamental-styles/pull/2339#pullrequestreview-655087143

stories/busy-indicator/busy-indicator.stories.js
Lokanathan-k: changed fd-busy-indicator--extended-indicator to fd-busy-indicator__extended

src/busy-indicator.scss
Lokanathan-k: fixed

stories/busy-indicator/busy-indicator.stories.js
Lokanathan-k: fixed

stories/busy-indicator/busy-indicator.stories.js
Lokanathan-k: fixed

stories/busy-indicator/busy-indicator.stories.js
Lokanathan-k: fixed

src/busy-indicator.scss
Lokanathan-k: fixed

stories/busy-indicator/busy-indicator.stories.js
Lokanathan-k: this is retained since the the busy indicator dots has to aligned to the center of the screen.

stories/busy-indicator/busy-indicator.stories.js
Lokanathan-k: fixed

src/busy-indicator.scss
Lokanathan-k: fixed

src/busy-indicator.scss
Lokanathan-k: fixed

stories/busy-indicator/busy-indicator.stories.js
kavya-b: for extended indicator, this will need to be part of the library, because if not, it won't align to center:

stories/busy-indicator/busy-indicator.stories.js
kavya-b: not yet fixed. Should be used with `fd-busy-indicator` like 
`<div class="fd-busy-indicator fd-busy-indicator--l fd-busy-indicator--enhanced-indicator"` 
where `fd-busy-indicator` is the element, and `fd-busy-indicator--l` is modifying the element.

stories/busy-indicator/busy-indicator.stories.js
kavya-b: I understand the difference between extended indicator and busy dialog. I was wondering what the difference is between `__label` example and `extended indicator` example. 
Also, `extended indicator` is not using a dialog, so the description should also not be mentioning anything about a dialog, right?

kavya-b (approved):
My comments have been addressed. LGTM!

stories/busy-indicator/busy-indicator.stories.js
InnaAtanasova: It should be `class="fd-busy-indicator d-busy-indicator--l fd-busy-indicator--extended"`

`fd-busy-indicator__extended` means a separate element, in your case it's a modifier of `fd-busy-indicator`

InnaAtanasova: What is the purpose of extended busy indicator?

Lokanathan-k: busyindicator has a box-shadow property add to it giving it a view as shown in the screenshot.

image

InnaAtanasova:
@Lokanathan-k @droshev @kavya-b pushed some changes in the branch, please review again. Thanks

src/busy-indicator.scss
```diff
@@ -3,6 +3,7 @@
 @import "./mixins";
 
 $block: #{$fd-namespace}-busy-indicator;
+$message-toast: #{$fd-namespace}-message-toast;
```
kavya-b, line 6: doesn't this go against the `only one top level fd-* class is used in the file` rule?

stories/busy-indicator/__snapshots__/busy-indicator.stories.storyshot
```diff
@@ -34,6 +34,16 @@ exports[`Storyshots Components/Busy Indicator Contrast Mode 1`] = `
 </div>
 `;
 
+exports[`Storyshots Components/Busy Indicator Extended Busy Indicator 1`] = `
```
kavya-b, line 37: empty storyshots, the opening backquote \` should be next to the first element, say `<div>` in the stories, otherwise the storyshots don't generate at all.

stories/busy-indicator/busy-indicator.stories.js
```diff
@@ -68,6 +69,43 @@ export const contrastMode = () => `<div style="display:flex;justify-content:cent
...
+        <div class="fd-busy-indicator--circle-2"></div>
+    </div>
+    <div class="fd-busy-indicator-extended__label">loading data...</div>
+</div>`;
+
+extendedBusyIndicatorInsideMessageToast.parameters = {
+    docs: {
+        iframeHeight: 200,
+        storyDescription:
+            'At the Page level the Busy Indicator should always be placed in a container. The simplest form of container will be centred on the page and inherit the color values from Message Toast.'
```
kavya-b, line 109: It says `The simplest form of container will be centred on the page` but thee container is not centred in the example. Is this expected?

InnaAtanasova: yes, it's just a visual representation (see Message Toast in fundamental-styles). When we bring it to fund-ngx we will rely on the Message Toast logic to open and position it in the center of the screen.

src/busy-indicator.scss
InnaAtanasova: Hmmm I think I might have a better idea. Let me test it and will let you know.

InnaAtanasova: ok, I just pushed, should be available soon in the preview. Thanks for pushing me to think more.
I added more specificity by introducing additional modifier class. Now if you want to use only busy indicator with a label, without a container you need `fd-busy-indicator-extended` wrapper. It will center the content. When you use it with a message toast you will need `fd-message-toast fd-busy-indicator-extended fd-busy-indicator-extended--message-toast`. This is more specific and overwrites the rules set by Message Toast. In future if we need the same scenario for Dialog we can introduce a new modifier class.

src/busy-indicator.scss
```diff
@@ -51,6 +52,29 @@ $fd-dots-size-l: 2rem;
   background-color: currentColor;
 }
 
+.#{$block}-extended {
+  @include fd-flex-center() {
+    flex-direction: column;
+  };
+
+  &.#{$block}-extended--message-toast {
```
kavya-b, line 60: Do we need the entire prefix `&.#{$block}-extended`?  How about just `&--message-toast`: 
```suggestion
  &--message-toast {
```

src/busy-indicator.scss
kavya-b: Thank you for the changes, looks good! Just one minor comment, will approve it meanwhile.

kavya-b (approved):

src/busy-indicator.scss
InnaAtanasova: `&--message-toast` won't enforce the specificity.
we want to explicitly say that `fd-busy-indicator-extended` is used at the same level with `fd-busy-indicator-extended--message-toast`  aka we have `class="fd-busy-indicator-extended fd-busy-indicator-extended--message-toast"`.
This way, because 2 classes used together on a same level is stronger that `fd-message-toast` and that's why they overwrite its rules.

kavya-b: ah I get it now. Cool. Thanks for explaining!

InnaAtanasova (approved):

droshev (approved):

PR #2343, opened 2021-05-10
fix: List component cursor pointer corrections, move css to styles from Ngx

DeepakSap14:
## Related Issue
Related to Issue https://github.com/SAP/fundamental-ngx/issues/5025

## Description
Moving list css present in Ngx to Styles.
It is moving core list css and platform list css.

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation


src/_listByline.scss
```diff
@@ -101,7 +102,7 @@
       color: $fd-list-byline-section-text-color;
       padding-right: $fd-list-byline-elements-spacing;
       line-height: $fd-list-byline-text-line-height;
-      align-self: flex-end;
+      align-self: flex-start;
```
InnaAtanasova, line 105: why was this changed?

src/_listBase.scss
```diff
@@ -123,6 +137,14 @@
...
       flex-basis: auto;
     }
+
+    &--growing {
+      @include fd-flex-center();
+    }
+
+    &--unread {
+      font-weight: bold;
+    }
```
InnaAtanasova, line 147: where are these in the html code? I looked for `__title--growing` and `__title--unread` but couldn't find it. If it's present in NGX but not in styles, you should add examples in fund-styles so we can test it.

src/_listBase.scss
```diff
@@ -100,6 +100,19 @@
...
+    &--inactive {
+      pointer-events: none;
+
+      .#{$block}__title,
+      .#{$block}__secondary {
+        pointer-events: none;
+      }
+    }
+
+    &--no-seprator {
```
InnaAtanasova, line 112: couldn't find this anywhere in the code as an example

InnaAtanasova (changes requested):
The hover state for selected list item is not working. The background should change:
Screen Shot 2021-05-12 at 2 20 39 PM

src/_listBase.scss
DeepakSap14: added examples for all css changes.

src/_listBase.scss
DeepakSap14: added examples.

src/_listByline.scss
DeepakSap14: i think some of the css present in Ngx is un-necessary and could have been avoided. i am not making any change for that because it is already working and approved code.
This change is present in https://github.com/SAP/fundamental-ngx/blob/main/libs/core/src/lib/list/list.component.scss, so i have moved it to styles.

DeepakSap14:
> The hover state for selected list item is not working. The background should change:
> Screen Shot 2021-05-12 at 2 20 39 PM

This one i am looking, will take little time to identify the fix.

DeepakSap14:
> > The hover state for selected list item is not working. The background should change:
> > Screen Shot 2021-05-12 at 2 20 39 PM
> 
> This one i am looking, will take little time to identify the fix.

@InnaAtanasova ,hover state on selected list item is as well fixed now. please review.

InnaAtanasova:
What is the difference between these?
Screen Shot 2021-05-17 at 11 22 09 AM

Screen Shot 2021-05-17 at 11 22 02 AM

InnaAtanasova:
We already have growing list with busy indicator. Why did you add this one?

Screen Shot 2021-05-17 at 11 22 39 AM
Screen Shot 2021-05-17 at 11 22 31 AM

InnaAtanasova:
We have counter. Why did you add a new one?

Screen Shot 2021-05-17 at 11 22 53 AM
Screen Shot 2021-05-17 at 11 22 46 AM

src/_listBase.scss
```diff
@@ -302,4 +324,21 @@
...
+    font-size: 0.875rem;
+    color: var(--sapContent_MarkerTextColor);
+    padding-left: $fd-list-item-padding-x;
+    text-shadow: none;
+    flex: none;
+
+    @include fd-active() {
+      color: var(--sapHighlightTextColor);
+    }
+  }
```
InnaAtanasova, line 338: I think we already have a counter.

src/_listBase.scss
```diff
@@ -302,4 +324,21 @@
...
+
+  &__infinite-scroll {
+    overflow: scroll;
+    outline: none;
+  }
```
InnaAtanasova, line 343: I couldn't find an example for this

DeepakSap14:
Need to refactor platform list component in Ngx, so already existing classes can be used.
once that is completed, remaining css classes will be moved here.

src/_listBase.scss
DeepakSap14: added example

src/_listByline.scss
```diff
@@ -37,7 +37,7 @@
       border-radius: $fd-list-byline-thumbnail-border-radius;
       margin-right: $fd-list-byline-thumbnail-spacing;
       font-size: 2.5rem;
-      align-self: flex-start;
+      align-self: center;
```
DeepakSap14, line 40: here icon/avatar should be centre align, https://wiki.wdf.sap.corp/wiki/display/visualcore/List+%28Fiori+3%29+-+Standard+List+with+Byline.
Borderless byline has issue with flex-start.

src/_listByline.scss
DeepakSap14: reverted back now. i do not see any issue using 'flex-end', examples are working fine. so no need to bring Ngx change for this.

src/_listBase.scss
DeepakSap14: removed these duplicate classes

src/_listBase.scss
DeepakSap14: i have removed all the duplicate classes which was coming from platform list. i have verified Ngx as well using correct styles classes.

src/_listByline.scss
InnaAtanasova, line 40: I don't see this rule centering the icon horizontally. You still have the same:
Screen Shot 2021-05-20 at 1 25 43 PM

You can use `text-align: center`

src/_listDefinitions.scss
```diff
@@ -93,6 +94,7 @@ $semantic-color: (
       left: 0.125rem;
       right: 0.125rem;
       bottom: 0.125rem;
+      pointer-events: none;
```
InnaAtanasova, line 97: Can you explain what you are achieving with enabling/disabling pointer events here?

InnaAtanasova (changes requested):

src/_listByline.scss
DeepakSap14: you can observe in this example. (styles docs)
Screenshot 2021-05-20 at 11 04 14 PM

DeepakSap14: here Avatar is not centre aligned vertically.

src/_listDefinitions.scss
DeepakSap14: this fix was done by markowski in core Ngx at https://github.com/SAP/fundamental-ngx/blob/main/libs/core/src/lib/list/list.component.scss#L4. i am not very sure of the reason. i have moved it here from Ngx.

DeepakSap14: when navigation list item is in focus, cursor is not changing to "grabbing cursor". by making pointer-events: none, it has "grabbing" cursor.
list-cursor-pointer

InnaAtanasova (changes requested):
Screen Shot 2021-05-25 at 1 31 27 PM

The icon inside the thumbnail is still not centered horizontally

src/_listByline.scss
InnaAtanasova: You can remove this rule, you don't need to change it to center.

DeepakSap14:
> Screen Shot 2021-05-25 at 1 31 27 PM
> 
> The icon inside the thumbnail is still not centered horizontally

This has been removed. No change here.

DeepakSap14:
> > Screen Shot 2021-05-25 at 1 31 27 PM
> > The icon inside the thumbnail is still not centered horizontally
> 
> This has been removed. keeping flex-start.

InnaAtanasova (changes requested):
When you hover over the text you get the `text` cursor. It should be the default.

InnaAtanasova:
Screen Shot 2021-06-04 at 11 22 53 AM

The thumbnail is still not centered.

DeepakSap14:
> When you hover over the text you get the `text` cursor. It should be the default.

If you implementation of Core list in fundamental-ngx, `text` cursor is used.
Do you want me to revert that and use default cursor?

DeepakSap14:
> Screen Shot 2021-06-04 at 11 22 53 AM
> 
> The thumbnail is still not centered.

you did asked to remove it here https://github.com/SAP/fundamental-styles/pull/2343#discussion_r639021611.

Now i am placed it in vertical and horizontal center, fiori also says it. Hope this is correct.

InnaAtanasova:
> #2343 (comment)

Screen Shot 2021-06-07 at 9 43 11 AM

I did not ask such thing! I said that this rule that you applied is NOT centering the element HORIZONTALLY and even suggested how to fix this using `text-align: center`.

InnaAtanasova (approved):

DeepakSap14:
> > #2343 (comment)
> 
> Screen Shot 2021-06-07 at 9 43 11 AM
> 
> I did not ask such thing! I said that this rule that you applied is NOT centering the element HORIZONTALLY and even suggested how to fix this using `text-align: center`.

I think, I misunderstood and removed vertical centre as well. i think, you wanted to say to remove `align-self: center;` property and use mixin, which will place horizontal and vertical centre both.

PR #2349, opened 2021-05-12
feat: introduce Vertical Navigation component

### LINKED ISSUE #2211: Implement Vertical Navigation

InnaAtanasova:
closes #2211 

## Description
This is the same PR as https://github.com/SAP/fundamental-styles/pull/2278 opened by @mikerodonnell89  but because Travis was failing and nothing really helped I moved the code to this PR. Please check #2278 for the review comments.
Implements the vertical navigation and deprecates side navigation

### After:
Screen Shot 2021-05-13 at 11 41 09 AM
Screen Shot 2021-05-13 at 11 41 21 AM
Screen Shot 2021-05-13 at 11 41 35 AM
Screen Shot 2021-05-13 at 11 41 51 AM

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/.storybook/static/preview-head.js
+++ b/.storybook/static/preview-head.js
@@ -375,3 +375,44 @@ function toggleNestedListSubmenu(event) {
         icon.classList = 'sap-icon--navigation-down-arrow';
     }
 }
+
+function collectionHas(a, b) { //helper function (see below)
+    for(var i = 0, len = a.length; i < len; i ++) {
+        if(a[i] == b) return true;
+    }
+    return false;
+}
+
+function toggleCondensedVerticalNavSubmenu(event) {
+    let button = event.target;
+
+    while (button.id !== 'parentCalendarButton') {
+        button = button.parentNode;
+    }
+
+    if(!button.classList.contains('is-expanded')) {
+        button.classList.add('is-expanded');
+    } else if(button.classList.contains('is-expanded')) {
+        button.classList.remove('is-expanded');
+    }
+}
+
+function toggleVerticalNavSubmenu(event) {
+    let button = event.target;
+    if (button.tagName.toLowerCase() !== 'li') {
+        button = event.target.parentNode;
+    }
+    let arrowIcon = button.querySelector('.fd-list__navigation-item-arrow');
+
+    if(arrowIcon && arrowIcon.classList.contains('is-expanded')) {
+        button.classList.remove('is-expanded');
+        arrowIcon.classList.remove('is-expanded');
+        arrowIcon.classList.add('sap-icon--navigation-right-arrow');
+        arrowIcon.classList.remove('sap-icon--navigation-down-arrow');
+    } else if (arrowIcon) {
+        button.classList.add('is-expanded');
+        arrowIcon.classList.add('is-expanded');
+        arrowIcon.classList.remove('sap-icon--navigation-right-arrow');
+        arrowIcon.classList.add('sap-icon--navigation-down-arrow');
+    }
+}
--- a/.storybook/static/preview-head.js
+++ b/.storybook/static/preview-head.js
@@ -377,9 +377,10 @@ function toggleNestedListSubmenu(event) {
 }
 
 function collectionHas(a, b) { //helper function (see below)
-    for(var i = 0, len = a.length; i < len; i ++) {
-        if(a[i] == b) return true;
+    for (let i = 0, len = a.length; i < len; i ++) {
+        return a[i] === b;
     }
+
     return false;
 }
 
@@ -390,10 +391,10 @@ function toggleCondensedVerticalNavSubmenu(event) {
         button = button.parentNode;
     }
 
-    if(!button.classList.contains('is-expanded')) {
-        button.classList.add('is-expanded');
-    } else if(button.classList.contains('is-expanded')) {
+    if (button.classList.contains('is-expanded')) {
         button.classList.remove('is-expanded');
+    } else {
+        button.classList.add('is-expanded');
     }
 }
 
@@ -404,7 +405,7 @@ function toggleVerticalNavSubmenu(event) {
     }
     let arrowIcon = button.querySelector('.fd-list__navigation-item-arrow');
 
-    if(arrowIcon && arrowIcon.classList.contains('is-expanded')) {
+    if (arrowIcon && arrowIcon.classList.contains('is-expanded')) {
         button.classList.remove('is-expanded');
         arrowIcon.classList.remove('is-expanded');
         arrowIcon.classList.add('sap-icon--navigation-right-arrow');
@@ -416,3 +417,4 @@ function toggleVerticalNavSubmenu(event) {
         arrowIcon.classList.add('sap-icon--navigation-down-arrow');
     }
 }
+
--- a/.storybook/static/preview-head.js
+++ b/.storybook/static/preview-head.js
@@ -376,14 +376,6 @@ function toggleNestedListSubmenu(event) {
     }
 }
 
-function collectionHas(a, b) { //helper function (see below)
-    for (let i = 0, len = a.length; i < len; i ++) {
-        return a[i] === b;
-    }
-
-    return false;
-}
-
 function toggleCondensedVerticalNavSubmenu(event) {
     let button = event.target;
 
@@ -417,4 +409,3 @@ function toggleVerticalNavSubmenu(event) {
         arrowIcon.classList.add('sap-icon--navigation-down-arrow');
     }
 }
-
--- a/src/_listDefinitions.scss
+++ b/src/_listDefinitions.scss
@@ -202,8 +202,7 @@ $semantic-color: (
 }
 
 @mixin fake-arrow-focus() {
-  &:focus,
-  &.is-focus {
+  @include fd-focus() {
     outline: none;
 
     &::after {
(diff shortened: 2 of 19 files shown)


InnaAtanasova:
This PR is the same as https://github.com/SAP/fundamental-styles/pull/2278
The build was failing and the rebase didn't work so I opened this PR.
The author of this PR is @mikerodonnell89 I just cherry-picked his commits

mikerodonnell89:
Nice! Especially from a code cleanup standpoint.  Couple issues though:

Still missing the reverse-radius ::after connector that goes between the first-level and second-level popovers (screenshot from the specs. There are no specific specs described on this but I'm guessing we can assume it's just the inverted typical border-radius for the outer corners of these popovers):
Screen Shot 2021-05-13 at 9 52 31 AM


IE11 - missing top border on the first-level popover and right border on the second-level popover
Screen Shot 2021-05-13 at 9 47 15 AM

IE11 - still seeing this weird focus issue on the expand/collapse button:
Screen Shot 2021-05-13 at 9 48 10 AM

.storybook/static/preview-head.js
```diff
@@ -375,3 +375,44 @@ function toggleNestedListSubmenu(event) {
         icon.classList = 'sap-icon--navigation-down-arrow';
     }
 }
+
+function collectionHas(a, b) { //helper function (see below)
+    for(var i = 0, len = a.length; i < len; i ++) {
+        if(a[i] == b) return true;
+    }
+    return false;
```
aberikashvili, line 383: I'd put empty line before `return` for better readability

.storybook/static/preview-head.js
```diff
@@ -375,3 +375,44 @@ function toggleNestedListSubmenu(event) {
         icon.classList = 'sap-icon--navigation-down-arrow';
     }
 }
+
+function collectionHas(a, b) { //helper function (see below)
+    for(var i = 0, len = a.length; i < len; i ++) {
```
aberikashvili, line 380: Why not to use `let` instead of `var`?

also, I'd suggest putting empty space after `for` for better formatting

.storybook/static/preview-head.js
```diff
@@ -375,3 +375,44 @@ function toggleNestedListSubmenu(event) {
...
+        if(a[i] == b) return true;
```
aberikashvili, line 381: What is `a[i]` here? 
Isn't it better to use `===` instead of `==`?
I think this line can be simplified like: `return a[i] === b;`

.storybook/static/preview-head.js
```diff
@@ -375,3 +375,44 @@ function toggleNestedListSubmenu(event) {
...
+function toggleCondensedVerticalNavSubmenu(event) {
+    let button = event.target;
+
+    while (button.id !== 'parentCalendarButton') {
+        button = button.parentNode;
+    }
+
+    if(!button.classList.contains('is-expanded')) {
+        button.classList.add('is-expanded');
+    } else if(button.classList.contains('is-expanded')) {
```
aberikashvili, line 395: I think `else if` isn't necessary here, same result could be achieved the following way:
```
if(button.classList.contains('is-expanded')) {
    button.classList.remove('is-expanded');
} else {
    button.classList.add('is-expanded');
}
```

.storybook/static/preview-head.js
```diff
@@ -375,3 +375,44 @@ function toggleNestedListSubmenu(event) {
...
+
+function toggleVerticalNavSubmenu(event) {
+    let button = event.target;
+    if (button.tagName.toLowerCase() !== 'li') {
+        button = event.target.parentNode;
+    }
+    let arrowIcon = button.querySelector('.fd-list__navigation-item-arrow');
+
+    if(arrowIcon && arrowIcon.classList.contains('is-expanded')) {
+        button.classList.remove('is-expanded');
```
aberikashvili, line 408: I think this line could be extracted into separate function

.storybook/static/preview-head.js
```diff
@@ -375,3 +375,44 @@ function toggleNestedListSubmenu(event) {
...
+        arrowIcon.classList.remove('is-expanded');
+        arrowIcon.classList.add('sap-icon--navigation-right-arrow');
+        arrowIcon.classList.remove('sap-icon--navigation-down-arrow');
+    } else if (arrowIcon) {
+        button.classList.add('is-expanded');
```
aberikashvili, line 413: Same here

.storybook/static/preview-head.js
InnaAtanasova: I don't think it's necessary for 1 line of code

.storybook/static/preview-head.js
mikerodonnell89: `a` is the collection and `b` is an item that may or may not be in that collection and `i` is the index of the collection, `a[i]` means the item at index `i` in collection `a`

.storybook/static/preview-head.js
mikerodonnell89: Just the `button.classList.remove('is-expanded');` part? Seems a bit overkill to me

.storybook/static/preview-head.js
mikerodonnell89: Also I just realized this function is unused so it can be deleted

src/_listDefinitions.scss
```diff
@@ -200,3 +200,27 @@ $semantic-color: (
     }
   }
 }
+
+@mixin fake-arrow-focus() {
+  &:focus,
+  &.is-focus {
```
N1XUS, line 206: Can we use fd-focus mixin here?

src/_listNavigationItem.scss
```diff
@@ -0,0 +1,352 @@
...
+  .#{$block}__navigation-item-indicator {
+    @include fd-reset();
+
+    &::after {
+      content: '';
+      width: 0.1875rem;
+      display: inline-block;
+      position: absolute;
+      right: 0;
+      top: 0.5625rem; // (height of nav item - height of indicator) / 2
```
N1XUS, line 20: Maybe use math function instead of hardcoded value, in case later we change variable values?

src/_listNavigationItem.scss
```diff
@@ -0,0 +1,352 @@
...
+              left: unset;
+              right: 0.325rem;
+            }
+          }
+
+          &--second-level {
+            border-top-right-radius: 0;
+            border-top-left-radius: 0;
+            top: $fd-list-navigation-item-height;
+            left: 3.25rem;
```
N1XUS, line 221: I would move this repeated value to separate variable

N1XUS (changes requested):
Great job. Found some minor issues in styling, maybe you can review them.

src/_listNavigationItem.scss
InnaAtanasova: done

src/_listNavigationItem.scss
InnaAtanasova: done

src/_listDefinitions.scss
InnaAtanasova: done

.storybook/static/preview-head.js
InnaAtanasova: deleted the function

mikerodonnell89 (approved):
I am fine to merge this and address any ie11 issues when storybook is up and running on that browser

N1XUS (approved):

stories/vertical-navigation/vertical-navigation.stories.js
```diff
@@ -0,0 +1,252 @@
+export default {
+    title: 'Components/Vertical Navigation',
+    parameters: {
+        description: `
+The Vertical Navigation can be used as an alternative to the Horizontal Navigation.<br>
+This navigation type is composed by two distinct item levels:<br>
+
+- First Level items - always visible, can be either navigable in and of themselves or be a container for Second Level items (it should not be both)
```
kari-leg, line 8: Rephrase syntax: 

First Level items →  Always visible, can be either navigable in and of themselves or be a container for Second Level items (it should not be both).

Second Level items → Are always contained by a First Level item, and can have their visibility toggled by this First Level item. They are always only navigable.

stories/vertical-navigation/vertical-navigation.stories.js
```diff
@@ -0,0 +1,252 @@
...
+- Second Level items → Are always contained by a First Level item, and can have their visibility toggled by this First Level item. They are always only navigable
+
+##Usage      
+**Use the vertical navigation if:**
+
+- You need to display global navigation structures of up to two levels.
+- Your scenarios are in the tooling or administration space.
+- If you want the entries to change as though they are dynamic content.
```
kari-leg, line 16: Rephrase:
- You want the entries to change as though they are dynamic content.

Issue #2360, opened 2021-05-18
add class for removing outer border for table component


DeepakSap14:
**Describe the bug**
A table can be displayed inside a card, and that time outer border of table should not visible, otherwise it will reserve some space. This PR adds `fd-tabel--no-border` class inside table component, which hides the outer border when applied

Requirement coming from below comment: 
https://github.com/SAP/fundamental-ngx/pull/5015#pullrequestreview-657228575

> Is this issue related to a specific component?
Table

> What version of the Fundamental Library Styles are you using?

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

> What front-end framework are you implementing in, e.g., Angular, React, etc.?
Angular

**To Reproduce**

**Expected behavior**

**Screenshots**
Before
<img width="928" alt="before" src="https://user-images.githubusercontent.com/57661487/118605511-fe751300-b7d3-11eb-9b9f-82922b0babbf.png">

After:
<img width="932" alt="after" src="https://user-images.githubusercontent.com/57661487/118605521-02a13080-b7d4-11eb-89c4-c4afa991de2b.png">

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**
When table displayed under card content, outer border should not be visible.

PR #2361, opened 2021-05-18
fix: class added for no outer border for table

DeepakSap14:
## Related Issue
Closes SAP/fundamental-styles#2360

## Description
A table can be displayed inside a card, and that time outer border of the table should not visible, otherwise it will reserve some space. This PR adds fd-tabel-no border class inside table-component which hides the outer border when applied.

Requirement coming from below comment:
https://github.com/SAP/fundamental-ngx/pull/5015#pullrequestreview-657228575

### Before:
before

### After:
after

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/table.scss
+++ b/src/table.scss
@@ -593,6 +593,10 @@ $block: #{$fd-namespace}-table;
     }
   }
 
+  &--no-border {
+    border: hidden;
+  }
+
   @include fd-rtl() {
     .#{$block}__cell {
       text-align: right;
--- a/src/table.scss
+++ b/src/table.scss
@@ -593,7 +593,7 @@ $block: #{$fd-namespace}-table;
     }
   }
 
-  &--no-border {
+  &--no-outer-border {
     border: hidden;
   }
 
--- a/stories/table/table.stories.js
+++ b/stories/table/table.stories.js
@@ -230,12 +230,12 @@ Table can be displayed without borders that separate the columns and rows only, 
     }
 };
 
-export const borderlesstable = () => `
+export const noOuterBorder = () => `
 <div class="fd-toolbar fd-toolbar--solid fd-toolbar--title fd-toolbar-active">
     <h4 style="margin: 0;">Table Without Outer Borders</h4>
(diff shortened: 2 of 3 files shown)


src/table.scss
```diff
@@ -593,6 +593,10 @@ $block: #{$fd-namespace}-table;
     }
   }
 
+  &--no-border {
```
InnaAtanasova, line 596: please rename it to `--no-outer-border` as user can be confused that this is removing all the borders from the table. Thank you.

InnaAtanasova (commented):
Just a suggestion to rename the modifier class, the rest looks good.

src/table.scss
DeepakSap14: class renamed to --no-outer-border.

kari-leg (commented):
rephrased the description to: A table can be displayed inside a card. When this happens, the outer border of the table should not be visible, since it will reserve some additional space. This PR adds 'fd-table--no-border' class inside of the table component, which hides the outer border when applied. 

' ' back ticks just be added to codes/class names to look more organized.

stories/table/table.stories.js
```diff
@@ -230,6 +230,52 @@ Table can be displayed without borders that separate the columns and rows only,
...
+        </tr>
+    </tbody>
+</table>
+`;
+
+noOuterBorder.storyName = 'No outer Border';
+noOuterBorder.parameters = {
+    docs: {
+        storyDescription: `
+Table can be displayed without outer borders, might be needed when used inside some other element. To display a table without outer border, add the \`fd-table--border-less\` modifier class to the main element.
```
dimamarksman, line 274:
```suggestion
Table can be displayed without outer borders, might be needed when used inside some other element. To display a table without outer border, add the \`fd-table--no-outer-border\` modifier class to the main element.
```

dimamarksman (approved):
One comment regarding class-name in the doc, otherwise LGTM

DeepakSap14:
> One comment regarding class-name in the do, otherwise LGTM

corrected class name in description.

PR #2367, opened 2021-05-20
fix: removal of extra margin on the labelled status indicator

Lokanathan-k:
## Related Issue
Closes: https://github.com/SAP/fundamental-styles/issues/2331

## Description
Removal of extra margin on the labelled status indicator

### Before:
116866104-76660980-ac28-11eb-8415-3606b75d2b1b

### After:

Screenshot 2021-05-20 at 3 13 49 PM

- [N/A]  All values are in `rem`
- [N/A]  Text elements follow the truncation rules
- [N/A] RTL support
3. Testing
- [N/A] Updated tests
4. Documentation

### DIFF
--- a/src/status-indicator.scss
+++ b/src/status-indicator.scss
@@ -44,7 +44,7 @@ $color-states: (
   &--horizontal-label {
     @include fd-reset();
 
-    flex-direction: unset;
+    flex-direction: row;
   }
 
   &--link {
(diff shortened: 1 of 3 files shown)


DeepakSap14 (approved):

InnaAtanasova (changes requested):
You need to handle RTL support for `--right` and `--left`. For the label and for the indicator.
Screen Shot 2021-05-20 at 1 15 56 PM
Screen Shot 2021-05-20 at 1 15 43 PM
Screen Shot 2021-05-20 at 1 15 23 PM
Screen Shot 2021-05-20 at 1 16 11 PM

InnaAtanasova (approved):

InnaAtanasova (changes requested):
I noticed the label does not have RTL, the indicator also is not changed in RTL
Screen Shot 2021-05-21 at 6 01 45 PM
Screen Shot 2021-05-21 at 6 01 57 PM


Screen Shot 2021-05-21 at 6 02 05 PM
Screen Shot 2021-05-21 at 6 01 33 PM

Lokanathan-k:
> I noticed the label does not have RTL, the indicator also is not changed in RTL
> Screen Shot 2021-05-21 at 6 01 45 PM
> Screen Shot 2021-05-21 at 6 01 57 PM
> 
> Screen Shot 2021-05-21 at 6 02 05 PM
> 
> Screen Shot 2021-05-21 at 6 01 33 PM

@InnaAtanasova this is just a span tag to define a type of indicator, not a label is used this label is not part of the component.

InnaAtanasova:
Ok, for the label. Maybe it was better to use fundamental label but it's ok to keep it like this. What about the other problem? In the specs the illustration is always centered in the "container". But in your case, for left and right, the illustration is not centered. 
Wiki specs:
Screen Shot 2021-05-25 at 1 50 40 PM

Your case:
Screen Shot 2021-05-25 at 1 51 11 PM
Screen Shot 2021-05-25 at 1 51 04 PM

Lokanathan-k:
> Ok, for the label. Maybe it was better to use fundamental label but it's ok to keep it like this. What about the other problem? In the specs the illustration is always centered in the "container". But in your case, for left and right, the illustration is not centered.
> Wiki specs:
> Screen Shot 2021-05-25 at 1 50 40 PM
> 
> Your case:
> Screen Shot 2021-05-25 at 1 51 11 PM
> Screen Shot 2021-05-25 at 1 51 04 PM

@InnaAtanasova Maybe it was better to use fundamental label but it's ok to keep it like this, _**will change this**_.

In the specs the illustration is always centered in the "container". But in your case, for left and right, the illustration is not centered.

for the exmple with right and left label if the image is aligned in center then the space between the svg and label will be more, Because the image position is defined viewbox property of the SVG ,and SAPui5 example was used to as a reference to achieve the same.

SapUI5 exmaple:
Screenshot 2021-05-26 at 10 16 48 AM

SVG without label are aligned to the centered to container.
image

InnaAtanasova (approved):

PR #2371, opened 2021-05-21
fix: form grid aligns items vertically to the top

### LINKED ISSUE #2262: Form Grid alignment
**Describe the bug**
Form Grid aligns items vertically to the center when according to the guideline it should align to the top.

*When you have more than one column and the number of the fields is not equal for each column fields are aligned to the middle instead of to the top*:

image

*According to the guideline it should be aligned to the top*:

image

sKudum:
## Related Issue
Closes  #2262 

## Description

BREAKING CHANGE:
Added a new modifier class to fd-row to achieve even alignment of its elements:`fd-row--top` in a form. 

Form Grid aligns items vertically to the center when according to the guideline it should align to the top.

*When you have more than one column and the number of the fields is not equal for each column fields are aligned to the middle instead of to the top*:

image

*According to the guideline it should be aligned to the top*:

image

### Before:

image

### After:

image

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/form-label.scss
+++ b/src/form-label.scss
@@ -105,4 +105,12 @@ $block: #{$fd-namespace}-form-label;
     justify-content: space-between;
     align-items: center;
   }
+
+  &--horizontal {
+    line-height: 2.8rem;
+  }
+
+  &--compact {
+    line-height: 2rem;
+  }
 }
--- a/src/form-label.scss
+++ b/src/form-label.scss
@@ -105,12 +105,4 @@ $block: #{$fd-namespace}-form-label;
     justify-content: space-between;
     align-items: center;
   }
-
-  &--horizontal {
-    line-height: 2.8rem;
-  }
-
-  &--compact {
-    line-height: 2rem;
-  }
 }
--- a/src/form-layout-grid.scss
+++ b/src/form-layout-grid.scss
@@ -31,10 +31,8 @@ $cols: 11;
     }
 
     .#{$blockRow} {
-      align-items: center;
       margin-top: 0;
       margin-bottom: 0.625rem;
-
       .#{$blockCol} {
         padding: 0 0.25rem;
         overflow: hidden;
@@ -192,9 +190,19 @@ $cols: 11;
 @media (min-width: 1440px) {
   .#{$blockContainer}.#{$block}-container {
     .#{$blockRow} {
+      align-items: start;
+
       .#{$blockCol} {
         .fd-form-label {
           padding-bottom: 0;
+
+          &--horizontal {
+            line-height: 2.8rem;
+          }
+
+          &--compact {
+            line-height: 2rem;
+          }
         }
 
         @for $n from 0 through $cols {
--- a/src/form-layout-grid.scss
+++ b/src/form-layout-grid.scss
@@ -195,14 +195,6 @@ $cols: 11;
       .#{$blockCol} {
         .fd-form-label {
           padding-bottom: 0;
-
-          &--horizontal {
-            line-height: 2.8rem;
-          }
-
-          &--compact {
-            line-height: 2rem;
-          }
         }
 
         @for $n from 0 through $cols {
--- a/src/form-layout-grid.scss
+++ b/src/form-layout-grid.scss
@@ -31,8 +31,14 @@ $cols: 11;
     }
 
     .#{$blockRow} {
+      align-items: center;
       margin-top: 0;
       margin-bottom: 0.625rem;
+
+      &--top {
+        align-items: start;
+      }
+
       .#{$blockCol} {
         padding: 0 0.25rem;
         overflow: hidden;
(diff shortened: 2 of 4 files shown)


Betrozov (approved):

InnaAtanasova (approved):

InnaAtanasova (changes requested):
@sKudum this change is introducing another bug. Now the label and the input are not aligned vertically.
Before:
Screen Shot 2021-05-25 at 1 20 32 PM

Now:
Screen Shot 2021-05-25 at 1 20 00 PM

InnaAtanasova (commented):
baseline is not centering vertically the label and the input field:
Screen Shot 2021-05-26 at 1 47 50 PM
Screen Shot 2021-05-26 at 1 47 45 PM

sKudum:
> baseline is not centering vertically the label and the input field:
> Screen Shot 2021-05-26 at 1 47 50 PM
> Screen Shot 2021-05-26 at 1 47 45 PM

Hi @InnaAtanasova , 
  When  explored came to know that for horizontal and compact line-hight are causing issues with alignment of label with elements, To overcome it created a modifiers  for both of them with basic heights. 
  Class and examples are updated accordingly with details. 

 Please let me know if this approach is acceptable or not, If not then please let me know the suggested approach.

Thank You

src/form-layout-grid.scss
```diff
@@ -192,9 +190,19 @@ $cols: 11;
...
         .fd-form-label {
           padding-bottom: 0;
+
+          &--horizontal {
+            line-height: 2.8rem;
+          }
+
+          &--compact {
+            line-height: 2rem;
+          }
```
InnaAtanasova, line 205: These 2 modifier classes belong to the `fd-form-label` and they should bee in the corresponding scss file. Also this PR template should be updated with the Breaking changes as you are changing the HTML structure. The wiki pagee should bee updated, too.

InnaAtanasova (changes requested):

src/form-layout-grid.scss
sKudum: Done

src/form-label.scss
```diff
@@ -105,4 +105,12 @@ $block: #{$fd-namespace}-form-label;
     justify-content: space-between;
     align-items: center;
   }
+
+  &--horizontal {
+    line-height: 2.8rem;
```
platon-rov, line 110: It's clear with the compact - you want compact input, use the **compact** label for it.

But when you use regular input it still needs to use the `horizontal`  label and it's not fully clear. Could you please rework it so "horizontal" will be the default? Thank you

sKudum: Hi @platon-rov , we have vertical as default. but when we want label to be aligned horizontally  with its element and top of the grid mainly in non matching no of elements we need to add this class.

hope I answered your query. If not please could you please share more details about query, that will help.

InnaAtanasova (approved):

Seamoo13 (approved):

InnaAtanasova:
@sKudum I think you need to update the images for the tests. Some of them show the label not aligned in the middle. I think it's from the prev commit. You can see the screenshots I denied.

sKudum:
> @sKudum I think you need to update the images for the tests. Some of them show the label not aligned in the middle. I think it's from the prev commit. You can see the screenshots I denied.

Hi @InnaAtanasova ,

  As per observation and doing reliazed we don't need to add form-label--horizontal or  form-label--compact classes instead added fd-row--top class to fix this issue by which we can avoid breaking any changes. We can remove breaking change from this PR.

Please let me know your view on it.Will update accordingly.

InnaAtanasova:
> > @sKudum I think you need to update the images for the tests. Some of them show the label not aligned in the middle. I think it's from the prev commit. You can see the screenshots I denied.
> 
> Hi @InnaAtanasova ,
> 
> As per observation and doing reliazed we don't need to add form-label--horizontal or form-label--compact classes instead added fd-row--top class to fix this issue by which we can avoid breaking any changes. We can remove breaking change from this PR.
> 
> Please let me know your view on it.Will update accordingly.

It's def a better solution. Thanks for digging more. You still need to add the Breaking changes, tho. It's a new modifier class that needs to be reflected in the implementation libraries.

sKudum:
> > > @sKudum I think you need to update the images for the tests. Some of them show the label not aligned in the middle. I think it's from the prev commit. You can see the screenshots I denied.
> > 
> > 
> > Hi @InnaAtanasova ,
> > As per observation and doing reliazed we don't need to add form-label--horizontal or form-label--compact classes instead added fd-row--top class to fix this issue by which we can avoid breaking any changes. We can remove breaking change from this PR.
> > Please let me know your view on it.Will update accordingly.
> 
> It's def a better solution. Thanks for digging more. You still need to add the Breaking changes, tho. It's a new modifier class that needs to be reflected in the implementation libraries.

Hi @InnaAtanasova , Updated breaking change information.

PR #2384, opened 2021-05-26
fix: table fixed (freeze) column border fix

platon-rov:
## Related Issue

Part of https://github.com/SAP/fundamental-ngx/issues/4415

## Description

We have applied a special right border to the last fixed column by `.class:last-of-type` selector. That's working since we have only one fixed column. Basically, that selector should not be applied to the class as it's working with the type (of the element). The additional class created for that purpose. 

## Screenshots

Nothing changed in the examples. It's more like improvements for the future (when we will provide the ability to use multiple fixed columns), for the NGX library (where we do provide multiple fixed columns but there is a special border not shown).

- [X] All values are in `rem`
- [X] RTL support
3. Testing
- [X] Updated tests
4. Documentation

### DIFF
--- a/src/table.scss
+++ b/src/table.scss
@@ -126,9 +126,12 @@ $block: #{$fd-namespace}-table;
       &--fixed {
         margin-top: -$fd-table-fixed-cell-shift;
         border-top: $fd-table-border;
-        border-right: $fd-table-fixed-cell-border;
         border-left: $fd-table-border;
         height: $fd-table-cozy-cell-height + $fd-table-fixed-cell-shift;
+
+        &-last {
+          border-right: $fd-table-fixed-cell-border;
+        }
 
         @include fd-rtl() {
           border-left: $fd-table-fixed-cell-border;
@@ -229,7 +232,7 @@ $block: #{$fd-namespace}-table;
         background-color: var(--sapList_Hover_Background);
       }
 
-      &:last-of-type {
+      &-last {
         border-right: $fd-table-fixed-cell-border;
       }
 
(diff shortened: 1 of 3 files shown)


src/table.scss
```diff
@@ -229,18 +232,18 @@ $block: #{$fd-namespace}-table;
...
       }
 
-      &:last-of-type {
+      &-last {
         border-right: $fd-table-fixed-cell-border;
+
+        @include fd-rtl() {
+          border-right: $fd-table-border;
+          border-left: $fd-table-fixed-cell-border;
+        }
```
InnaAtanasova, line 241: The exact same code is repeated twice. Can you move it to a placeholder? Thank you

olshansky (approved):

dimamarksman (approved):

InnaAtanasova (approved):

InnaAtanasova:
@platon-rov in the UI tests you can see that the Slider component is affected by your changes. Could you please rebase main and see if the changes will disappear? The Slider component should not be affected by these changes.

platon-rov:
> @platon-rov in the UI tests you can see that the Slider component is affected by your changes. Could you please rebase main and see if the changes will disappear? The Slider component should not be affected by these changes.

Hmm, already done. I guess it's the issue about chromatic itself, there are no changes in the slider component.

N1XUS (approved):

InnaAtanasova (approved):

maxym-valor:
Tested, everything is ok!

Issue #2327, opened 2021-05-04
BTP vs. Fiori


droshev:
- [x] Review BTP design
- [x] Identify the differences and evaluate the changes needed

Documentation Requests:
- [ ] buttons - regular buttons - update the documentation to map and reflect primary/secondary/tertiary vs. button types
- [ ] List item - show more/less - update the doc
- [ ] Vertical Nav - add examples with collapsed/expanded sections - update the doc

Visual Requests:
- [ ] buttons - toggle and segmented button - centred text, button width should be equal regardless of text length
- [ ] list - vertical spacing: should Fundamental match with BTP design to reduce vertical spacing for improved readability?
- [ ] card-based layout - vertical spacing:  should Fundamental match with BTP design to reduce vertical spacing for improved readability?

_To be implemented in Fundamental if needed_
- [ ] color picker - arrows design, few design diff. **design question**, missing in fundamentals
- [ ] Tabs - instance tab - to be implemented in fundamental
- [ ] wizard - embedding wizard in the dialog component
- [ ] list - in-place editing; show examples
- [ ] multi-input - new pattern introduced in BTP to fit five lines
- [ ] splitter; split layout

BTP Related:
- [ ] input examples in BTP: UI texts should reflect the semantic message
- [ ] switch - color differences/ width (awaiting response from BTP team)

Resolved:
- [x] timepicker - arrows up/down **design question**
- [x] select - arrow icons - two weights available
- [x] slider - ring size - two weights available
- [x] panel - slightly diff. design/padding - check with Fiori 3 design: two options: (a) having the line under the header (b) having the line under the section. both can be used
- [x] List item - avatar examples: avatar can be in either a circle or a rectangular shape, depends on use cases; in some cases it is flexible for the user to choose
- [x] wizard - text colors, spacing and line width, empty spacing
- [x] user menu - 2 additional icons in Fundamental: only exist in spec but never has the use case for it in implementation; badge design
- [x] calendar - the state which the date is on light blue block with dark blue border: not documented in Fundamental

I546502:
For details on these requests, please refer to [this Figma file](https://www.figma.com/file/ToIXWKKNShFMOfO0MLGYlB/BTP-Analysis?node-id=0%3A1).

- [ ] button - regular button
[...]

PR #2387, opened 2021-05-26
docs: update documentation for button

hanjunt:
Update documentation for buttons

## Related Issue
Part of SAP/fundamental-styles#2327 - Documentation Requests:Button

## Description
"Fundamental buttons should be renamed to Primary/Secondary/Tertiary"
You can find the request at https://github.com/SAP/fundamental-styles/issues/2327 under ```Documentation Requests: Button```

## Screenshots
Screen Shot 2021-05-26 at 11 21 46 AM
Screen Shot 2021-05-26 at 11 22 29 AM

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/stories/button/__snapshots__/button.stories.storyshot
+++ b/stories/button/__snapshots__/button.stories.storyshot
@@ -22,23 +22,23 @@ exports[`Storyshots Components/Button Button styles 1`] = `
    
         
   <button
-    class="fd-button"
+    class="fd-button fd-button--emphasized"
   >
-    Default Button
+    Primary Button
   </button>
   
         
   <button
-    class="fd-button fd-button--emphasized"
+    class="fd-button"
   >
-    Emphasized Button
+    Secondary Button
   </button>
   
         
   <button
     class="fd-button fd-button--ghost"
   >
-    Ghost Button
+    Secondary Button
   </button>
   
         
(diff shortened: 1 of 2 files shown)


stories/button/__snapshots__/button.stories.storyshot
```diff
@@ -22,23 +22,23 @@ exports[`Storyshots Components/Button Button styles 1`] = `
    
         
   <button
-    class="fd-button"
+    class="fd-button fd-button--emphasized"
   >
-    Default Button
```
kari-leg, line 27: What about these styles?

Default button: is used for neutral or informative actions (secondary action on the screen).
Emphasized button: is used to indicate the primary action on the screen. 
Ghost button: is used to trigger secondary actions However, if a page also requires an important section highlighted in a content toolbar, then use the ghost style.
Attention button: is used to trigger a semantic action that needs attention and should always be accompanied by text. 
Transparent button: is used to trigger a negative path action within a header or footer bar, and secondary actions in toolbars.

kari-leg (changes requested):

mikerodonnell89:
Closing to go with https://github.com/SAP/fundamental-styles/pull/2389

PR #2389, opened 2021-05-26
docs: update regular button

mikerodonnell89:
part of #2327 

This is a new PR duplicate of https://github.com/SAP/fundamental-styles/pull/2387 originally by @hanjunt - reopening this PR on a separate branch due to rebase commit message reword not cooperating with the new lint commit hooks.

All text below from @hanjunt :
Update documentation for buttons

## Related Issue
Part of SAP/fundamental-styles#2327 - Documentation Requests:Button

## Description
"Fundamental buttons should be renamed to Primary/Secondary/Tertiary"
You can find the request at https://github.com/SAP/fundamental-styles/issues/2327 under ```Documentation Requests: Button```

## Screenshots
Screen Shot 2021-05-26 at 11 21 46 AM
Screen Shot 2021-05-26 at 11 22 29 AM

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/stories/button/button.stories.js
+++ b/stories/button/button.stories.js
@@ -41,40 +41,38 @@ export const primary = () => `
  */
 
 export const types = () => `<div class="fddocs-container fddocs-button-container"> 
-        <button class="fd-button">Default Button</button>
-        <button class="fd-button fd-button--emphasized">Emphasized Button</button>
-        <button class="fd-button fd-button--ghost">Ghost Button</button>
+        <button class="fd-button fd-button--emphasized">Primary Button</button>
+        <button class="fd-button">Secondary Button</button>
+        <button class="fd-button fd-button--ghost">Secondary Button</button>
         <button class="fd-button fd-button--positive">Positive Button</button>
         <button class="fd-button fd-button--negative">Negative Button</button>
         <button class="fd-button fd-button--attention">Attention Button</button>
-        <button class="fd-button fd-button--transparent">Transparent Button</button>
+        <button class="fd-button fd-button--transparent">Tertiary Button</button>
 </div>`;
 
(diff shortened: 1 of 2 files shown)


stories/button/button.stories.js
```diff
@@ -41,40 +41,38 @@ export const primary = () => `
  */
 
 export const types = () => `<div class="fddocs-container fddocs-button-container"> 
-        <button class="fd-button">Default Button</button>
-        <button class="fd-button fd-button--emphasized">Emphasized Button</button>
-        <button class="fd-button fd-button--ghost">Ghost Button</button>
+        <button class="fd-button fd-button--emphasized">Primary Button</button>
```
droshev, line 44: I think we need to explain the mapping between the 2 types - one is the use of the buttons and the other is its design. not just to replace the current.
Maybe in another example document which of the designs are primary, secondary, tertiary buttons

stories/button/button.stories.js
```diff
@@ -41,40 +41,38 @@ export const primary = () => `
...
         <button class="fd-button fd-button--negative">Negative Button</button>
         <button class="fd-button fd-button--attention">Attention Button</button>
-        <button class="fd-button fd-button--transparent">Transparent Button</button>
+        <button class="fd-button fd-button--transparent">Tertiary Button</button>
 </div>`;
 
 types.storyName = 'Button styles';
 types.parameters = {
     docs: {
-        storyDescription: `
```
droshev, line 56: I don't think it is a good idea to remove that. It would better to have both designs in parallel

droshev (changes requested):

PR #2390, opened 2021-05-27
fix: dialog z-index styles fixes

platon-rov:
**Please read last [comment](https://github.com/SAP/fundamental-styles/pull/2390#issuecomment-852035614)**

## Related Issue

Relates to [NGX #4518](https://github.com/SAP/fundamental-ngx/issues/4518).

## Description

`z-index` dialog styles moved from NGX library to styles library, fix applied to have the ability to show dropdowns in dialogs.

### Before:

image

### After:

image

- [X] All values are in `rem`
- [X] RTL support
3. Testing
- [X] Updated tests
4. Documentation


olshansky (commented):

Betrozov (approved):

src/dialog.scss
```diff
@@ -47,6 +47,7 @@ $button: #{$fd-namespace}-button;
   @extend %dialog;
 
   display: none;
+  z-index: 999;
```
dimamarksman, line 50: not sure if we are safe to hardcode z-index.
we already have [this mixing](https://github.com/SAP/fundamental-styles/blob/7c93b2c83dcedbd68b5c189f35fdefe787edb32b/src/_new-settings.scss#L16).
@InnaAtanasova please advise the approach to use.

dimamarksman: We already have something similar with [popover](https://github.com/SAP/fundamental-styles/blob/7c93b2c83dcedbd68b5c189f35fdefe787edb32b/src/popover.scss#L10)

InnaAtanasova: We avoid adding z-index unless it's really necessary and then we avoid using numbers like 999 and 1001. When our lib is used in apps with other UI components the z-indexes introduce "bugs".

platon-rov:
> As far as I understand the original issue is about a relation of `angular-cdk-overlay z-index` in conjuction with `fd-dialog z-index` + their order in the DOM

Yea, exactly, as both of them have `z-index: 1000;` so the top element - which is placed below in the DOM. To avoid that I've set z-index: 999 for dialog to be able have any dropdowns (overlays) on top of it.

src/dialog.scss
platon-rov: It's made to have ability to have overlays over the dialogs, [some more information](https://github.com/SAP/fundamental-styles/pull/2390#issuecomment-850813451).

N1XUS: I would introduce variable $dialog-initial-z-index: 999, and use it for incrementing the necessary z-index, depending on the element itself, at least this will give us more control in the future

dimamarksman:
> > As far as I understand the original issue is about a relation of `angular-cdk-overlay z-index` in conjuction with `fd-dialog z-index` + their order in the DOM
> 
> Yea, exactly, as both of them have `z-index: 1000;` so the top element - which is placed below in the DOM. To avoid that I've set z-index: 999 for dialog to be able have any dropdowns (overlays) on top of it.

Why do we need to move this `z-index` to the styles library?
Isn't better to keep it on the consumer side itself (how it's done now) cause the consumer only is aware of all dependencies it's relying on so can decide which particular z-index is suitable to avoid `z-index` issues.
As you can see `fundamental-styles` gives us only pure dialog component, it doesn't care how a consumer is going to position it and where it appears in the DOM. If so why the `dialog` component should be delivered with `z-index` at all?

platon-rov:
I would agree with @dimamarksman's reply about this PR. From the one side - it's good to have more styles in the styles library and have fewer in the NGX library. But from the other - these styles related only to the styles library in connecting with the NGX library and don't have much sense without it. 

@droshev @InnaAtanasova please take a look.

N1XUS (approved):

InnaAtanasova:
I agree with Dima. Please move it to fund-ngx. If it's related to Angular cdk it's ok to keep the styles there.

PR #2392, opened 2021-05-27
[WIP] fix: update UI text on list byline

hanjunt:
## Related Issue
part of #2327

## Description
Change list item button "MORE" and "LESS" to "Show more" and "Show less"

## Screenshots
Screen Shot 2021-05-27 at 11 33 51 AM

- [x] All values are in `rem`
- [ ] hover state of the element follow design spec
- [ ] focus state of the element follow design spec
- [ ] active state of the element follow design spec
- [ ] selected state of the element follow design spec
- [ ] selected hover state of the element follow design spec
- [ ] pressed state of the element follow design spec
- [ ] Responsiveness rules - the component has modifier classes for all breakpoints
- [ ] RTL support
- [ ] only one top level `fd-*` class is used in the file
- [ ] BEM naming convention is used
- [ ] Mixins are used for repeatable code (`fd-rtl`, `fd-ellipsis`, `fd-flex`, `fd-selected`, `fd-focus`, ect.)
- [ ] A11y support - keyboard support, screenreader support, proper ARIA attributes, etc.
- [ ] `fd-reset()` mixin is applied to all elements
- [ ] Variables are used, if some value is used more than twice.
- [ ] Checked if current components can be reused, instead of having new code.
3. Testing
- [ ] Updated tests
4. Documentation

### DIFF
--- a/src/_listByline.scss
+++ b/src/_listByline.scss
@@ -82,7 +82,6 @@
 
     .#{$block}__link {
       &--more {
-        text-transform: uppercase;
         pointer-events: all;
       }
     }
--- a/stories/list/byline/byline-list.stories.js
+++ b/stories/list/byline/byline-list.stories.js
@@ -604,7 +604,7 @@ export const longText = () => `
     <span class="fd-list__thumbnail"><i role="presentation" class="sap-icon--activate"></i></span>
     <div class="fd-list__content">
       <div class="fd-list__title">List Item Title</div>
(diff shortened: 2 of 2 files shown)


InnaAtanasova:
@hanjunt 
when you type `Closes #2327` in the PR template, once the PR is merged, the issue will be closed automatically. We don't want this for issues that have multiple tasks within. In your case you need to write `Part of #2327`. The issue will be closed once all checkboxes are checked and verified. 

The checkboxes in the PR template should be checked, o.w. mergeable check will fail. We also don't review PRs that have unchecked checkboxes as we assume the developer is still working on the PR.

If your PR is still work in progress, you can mark it as draft or add [WIP] in front on the PR title.

The last point is about the change you made. I see that in List specs, for wrapping behaviour, the MORE and LESS words are uppercase. Why are you changing it to lowercase?

hanjunt:
@InnaAtanasova I just came to understand what you meant. I'll close this PR so the buttons still follow Fiori specs.

PR #2431, opened 2021-06-04
fix: show navigation button on hover in carousel content area, carousel component css moved from ngx

DeepakSap14:
## Related Issue
Related to Issue SAP/fundamental-ngx#5025

## Description
Moving css present for carousel component in Ngx to styles.
navigation button in content area will be visible on hover on carousel.

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/.storybook/custom/custom.scss
+++ b/.storybook/custom/custom.scss
@@ -700,3 +700,8 @@ body.sb-show-main {
 .fddocs-list-height {
   height: 11rem;
 }
+
+.fddocs-horizontal-center {
+  display: flex;
+  justify-content: center;
+}
--- a/.storybook/custom/custom.scss
+++ b/.storybook/custom/custom.scss
@@ -705,3 +705,15 @@ body.sb-show-main {
   display: flex;
   justify-content: center;
 }
+
+.fddoc-carousel-example {
+  margin-bottom: 3rem;
+  width: 30rem;
+  height: 15.5rem;
+}
+
+.fddoc-horizontal-carousel-example {
+  margin-bottom: 3rem;
+  width: 60rem;
+  height: 15.5rem;
+}
--- a/.storybook/custom/custom.scss
+++ b/.storybook/custom/custom.scss
@@ -700,20 +700,3 @@ body.sb-show-main {
 .fddocs-list-height {
   height: 11rem;
 }
-
-.fddocs-horizontal-center {
-  display: flex;
-  justify-content: center;
-}
-
-.fddoc-carousel-example {
-  margin-bottom: 3rem;
-  width: 30rem;
-  height: 15.5rem;
-}
-
-.fddoc-horizontal-carousel-example {
-  margin-bottom: 3rem;
-  width: 60rem;
-  height: 15.5rem;
-}
--- a/src/carousel.scss
+++ b/src/carousel.scss
@@ -21,8 +21,12 @@ $button: #{$fd-namespace}-button;
   @include fd-reset();
   @include fd-flex(column);
 
-  @include fd-focus() {
-    outline: 0.0625rem dotted;
+  @include fd-hover() {
+    .#{$block}__content {
+      .#{$block}__button.#{$button} {
+        display: block;
+      }
+    }
   }
 
   width: 100%;
@@ -44,16 +48,13 @@ $button: #{$fd-namespace}-button;
       margin: 0;
       top: calc(50%);
       transform: translateY(-50%);
+      display: none;
 
       &.#{$block}__button--left {
         left: $fd-carousel-button-content-offset;
 
-        @include fd-focus() {
-          right: auto;
-        }
-
         @include fd-rtl() {
-          left: auto;
+          left: initial;
           right: $fd-carousel-button-content-offset;
         }
       }
@@ -61,12 +62,8 @@ $button: #{$fd-namespace}-button;
       &.#{$block}__button--right {
         right: $fd-carousel-button-content-offset;
 
-        @include fd-focus() {
-          right: auto;
-        }
-
         @include fd-rtl() {
-          right: auto;
+          right: initial;
           left: $fd-carousel-button-content-offset;
         }
       }
@@ -162,6 +159,12 @@ $button: #{$fd-namespace}-button;
         right: -0.25rem;
         height: $fd-carousel-touchable-button-size;
         width: $fd-carousel-touchable-button-size;
+      }
+
+      @include fd-rtl() {
+        & > [class*=sap-icon] {
+          transform: rotate(180deg);
+        }
       }
 
       @include fd-focus() {
--- a/src/carousel.scss
+++ b/src/carousel.scss
@@ -185,7 +185,7 @@ $button: #{$fd-namespace}-button;
 
   &--no-navigation {
     .#{$block}__button.#{$button} {
-      display: none;
+      display: none !important;
     }
   }
 
--- a/src/carousel.scss
+++ b/src/carousel.scss
@@ -72,6 +72,14 @@ $button: #{$fd-namespace}-button;
     &--display {
       overflow: hidden;
     }
+
+    &--horizontal {
+      @include fd-flex();
+    }
+
+    &--error-message {
+      @include fd-flex-center();
+    }
   }
 
   &__item {
@@ -180,12 +188,6 @@ $button: #{$fd-namespace}-button;
           @include after-position-overwrite();
         }
       }
-    }
-  }
-
-  &__horizontal {
-    @include fd-flex-vertical-center() {
-      width: 100%;
     }
   }
 
--- a/src/carousel.scss
+++ b/src/carousel.scss
@@ -185,7 +185,15 @@ $button: #{$fd-namespace}-button;
 
   &--no-navigation {
     .#{$block}__button.#{$button} {
-      display: none !important;
+      display: none;
+    }
+
+    @include fd-hover() {
+      .#{$block}__content {
+        .#{$block}__button.#{$button} {
+          display: none;
+        }
+      }
     }
   }
 
--- a/src/carousel.scss
+++ b/src/carousel.scss
@@ -22,7 +22,7 @@ $button: #{$fd-namespace}-button;
   @include fd-flex(column);
 
   @include fd-focus() {
-    outline: 1px dotted;
+    outline: 0.0625rem dotted;
   }
 
   width: 100%;
--- a/src/carousel.scss
+++ b/src/carousel.scss
@@ -16,10 +16,15 @@ $button: #{$fd-namespace}-button;
 
   @mixin after-position-overwrite() {
     outline-offset: $fd-carousel-button-focus-outline-offset;
+    top: 50%;
   }
 
   @include fd-reset();
   @include fd-flex(column);
+
+  @include fd-focus() {
+    outline: 1px dotted;
+  }
 
   width: 100%;
   min-width: 15.5rem;
@@ -44,8 +49,12 @@ $button: #{$fd-namespace}-button;
       &.#{$block}__button--left {
         left: $fd-carousel-button-content-offset;
 
+        @include fd-focus() {
+          right: auto;
+        }
+
         @include fd-rtl() {
-          left: initial;
+          left: auto;
           right: $fd-carousel-button-content-offset;
         }
       }
@@ -53,11 +62,19 @@ $button: #{$fd-namespace}-button;
       &.#{$block}__button--right {
         right: $fd-carousel-button-content-offset;
 
+        @include fd-focus() {
+          right: auto;
+        }
+
         @include fd-rtl() {
-          right: initial;
+          right: auto;
           left: $fd-carousel-button-content-offset;
         }
       }
+    }
+
+    &--display {
+      overflow: hidden;
     }
   }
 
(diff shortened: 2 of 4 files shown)


src/carousel.scss
```diff
@@ -172,7 +185,13 @@ $button: #{$fd-namespace}-button;
 
   &--no-navigation {
     .#{$block}__button.#{$button} {
-      display: none;
+      display: none !important;
```
aberikashvili, line 188: Is important really necessary here? can we use specificity instead of `!important` here?

You can see this article why using `!important` is bad practise https://uxengineer.com/css-specificity-avoid-important-css/

aberikashvili (commented):
Please see my comment regarding `!important`

DeepakSap14:
> Please see my comment regarding `!important`

using specificity instead of `!important` now.

src/carousel.scss
```diff
@@ -21,6 +21,14 @@ $button: #{$fd-namespace}-button;
   @include fd-reset();
   @include fd-flex(column);
 
+  @include fd-hover() {
+    .#{$block}__content {
+      .#{$block}__button.#{$button} {
```
InnaAtanasova, line 26: Do you need `.#{$button}`? You can avoid this by adding more specificity.

src/carousel.scss
```diff
@@ -174,5 +187,19 @@ $button: #{$fd-namespace}-button;
     .#{$block}__button.#{$button} {
       display: none;
     }
+
+    @include fd-hover() {
+      .#{$block}__content {
+        .#{$block}__button.#{$button} {
```
InnaAtanasova, line 193: same comment as above

stories/carousel/carousel.stories.js
```diff
@@ -477,3 +476,100 @@ carouselNoNavigation.parameters = {
...
     }
 };
+
+export const horizontalCarousel = () => `<div style="display: flex; flex-direction: column; align-items: center; background: #CCD1D1;">
+    <h4>Items in horizontal direction</h4>
+    <div
+        class="fd-carousel"
+        data-ride="carousel"
+        style="margin-bottom: 3rem; max-width: 60rem;">
+        <div class="fd-carousel__content fd-carousel--horizontal" style="text-align: center; padding: 1rem; min-height: 15.5rem;">
```
InnaAtanasova, line 486: wrong BEM

InnaAtanasova (changes requested):

src/carousel.scss
DeepakSap14: `.#{$button}` not needed. It works fine with only `.#{$block}__button`.
corrected.

src/carousel.scss
DeepakSap14: corrected, same as above.

stories/carousel/carousel.stories.js
DeepakSap14: using __horizontal now.

stories/carousel/carousel.stories.js
```diff
@@ -477,3 +476,100 @@ carouselNoNavigation.parameters = {
...
     }
 };
+
+export const horizontalCarousel = () => `<div style="display: flex; flex-direction: column; align-items: center; background: #CCD1D1;">
+    <h4>Items in horizontal direction</h4>
+    <div
+        class="fd-carousel"
+        data-ride="carousel"
+        style="margin-bottom: 3rem; max-width: 60rem;">
+        <div class="fd-carousel__content fd-carousel__horizontal" style="text-align: center; padding: 1rem; min-height: 15.5rem;">
```
InnaAtanasova, line 486: It's wrong again

InnaAtanasova (changes requested):

stories/carousel/carousel.stories.js
DeepakSap14: have moved it inside __content. so the class name now is __content--horizontal.
have created class for --error-message, as it was getting broken after above change.

aberikashvili (approved):
LGTM

stories/carousel/carousel.stories.js
```diff
@@ -477,3 +476,133 @@ carouselNoNavigation.parameters = {
...
+                    role="img"
+                    aria-label="Carousel Image 5">
+                </div>
+            </div>
+
+            <div class="fd-carousel__item fd-carousel__item--active">
+                <div
+                    style="
+                        width:100%;
+                        background-image: url(assets/images/backgrounds/Godafoss_waterfall_in_northern_Iceland.jpg);
```
InnaAtanasova, line 503: Is this image free from copyrights?

stories/carousel/carousel.stories.js
```diff
@@ -477,3 +476,133 @@ carouselNoNavigation.parameters = {
...
+        Displaying item 2 of 3
+    </div>
+</div>
+`;
+
+horizontalCarousel.storyName = 'Items in horizontal direction';
+horizontalCarousel.parameters = {
+    docs: {
+        iframeHeight: 900,
+        storyDescription: `Carousel will have items which will spread horizontally and will be visible on navigation. For translation to work, all items should be in DOM.
```
InnaAtanasova, line 560: `translation` ?

stories/carousel/carousel.stories.js
```diff
@@ -477,3 +476,133 @@ carouselNoNavigation.parameters = {
...
+    }
+};
+
+export const error = () => `<div style="display: flex; flex-direction: column; align-items: center; background: #CCD1D1;">
+    <h4>Error in loading items</h4>
+    <div
+        class="fd-carousel fd-carousel--no-navigation"
+        data-ride="carousel"
+        style="margin-bottom: 3rem; max-width: 60rem;">
+        <div class="fd-carousel__content fd-carousel__content--error-message" style="text-align: center; padding: 1rem; min-height: 15.5rem;">
```
InnaAtanasova, line 571: Why do you have a new class `fd-carousel__content--error-message` AND custom css applied on the same element?

stories/carousel/carousel.stories.js
```diff
@@ -477,3 +476,133 @@ carouselNoNavigation.parameters = {
...
+            <div class="docs-column-flex">
+                <div class="fddocs-horizontal-center">
+                    <i class="sap-icon--document"></i>
+                </div>
+                Items could not be loaded
+            </div>
```
InnaAtanasova, line 577: This part will not be consumed in ngx and people will have to add custom css there

stories/carousel/carousel.stories.js
```diff
@@ -477,3 +476,133 @@ carouselNoNavigation.parameters = {
 `
     }
 };
+
+export const horizontalCarousel = () => `<div style="display: flex; flex-direction: column; align-items: center; background: #CCD1D1;">
+    <h4>Items in horizontal direction</h4>
+    <div
+        class="fd-carousel"
+        data-ride="carousel"
+        style="margin-bottom: 3rem; max-width: 60rem;">
```
InnaAtanasova, line 485: Shouldn't have inline styling

stories/carousel/carousel.stories.js
```diff
@@ -477,3 +476,133 @@ carouselNoNavigation.parameters = {
...
+        <div class="fd-carousel__content fd-carousel__content--horizontal" style="text-align: center; padding: 1rem; min-height: 15.5rem;">
```
InnaAtanasova, line 486: Shouldn't have inline styling

stories/carousel/carousel.stories.js
DeepakSap14: This image is already getting used in previous example of carousel.
https://sap.github.io/fundamental-styles/?path=/docs/components-carousel--carousel-bottom.
Not sure on copy rights, but we have few more images in `assets/images/backgrounds/` folder.

InnaAtanasova (changes requested):
Reconsider the usage of inline styling on the elements that have fundamental classes

stories/carousel/carousel.stories.js
DeepakSap14: corrected the sentence.
For translate animation effect to work, all items should be in DOM.

stories/carousel/carousel.stories.js
DeepakSap14: copy paste issue from above example.
removed inline styling.

stories/carousel/carousel.stories.js
DeepakSap14: removed inline styling.

stories/carousel/carousel.stories.js
DeepakSap14: removed inline styling.

stories/carousel/carousel.stories.js
DeepakSap14: It was decided during tech spec review that Error message will be coming in Ngx as <ng-content>.
so only placing of error message is take care.

formatting of error message like having Icon and Error message vertically and icon horizontally centred will have to be taken care by application.

Do you want to have that styling present here?
if so, the during application development, those classes will be exposed.

let me know your thought and correct me.

InnaAtanasova: Hi @DeepakSap14 the specs clearly mention that the Empty State (when no pages are loaded) should use the Message Page component. It's present in both fund-styles and fund-ngx. You don't have to use custom css, just use the component.
Screen Shot 2021-06-16 at 8 52 44 AM

InnaAtanasova:
Screen Shot 2021-06-16 at 8 49 20 AM
Screen Shot 2021-06-16 at 8 49 02 AM

I think the problem with using inline styling and/or docs styling is still not clear. This styling is NOT consumed by the implementation libraries. You just make the component looks good on the documentation page, but when imported in fund-ngx, for example, you will need again custom css. 
Here's an example:
Screen Shot 2021-06-16 at 8 49 02 AM
Screen Shot 2021-06-16 at 8 49 20 AM

When you turn of the styling from the docs, the component doesn't look like as should.

DeepakSap14:
> Screen Shot 2021-06-16 at 8 49 20 AM
> 
> Screen Shot 2021-06-16 at 8 49 02 AM
> 
> I think the problem with using inline styling and/or docs styling is still not clear. This styling is NOT consumed by the implementation libraries. You just make the component looks good on the documentation page, but when imported in fund-ngx, for example, you will need again custom css.
> 
> When you turn off the styling from the docs, the component doesn't look like as should.

@InnaAtanasova 
1. when we create carousel in Ngx, we provide width for carousel; which is width taken by single carousel item.
2. In case of horizontal carousel example, i have shown 3 carousel items; to show horizontal arrangement of items.
    ideally, only 1 item will be visible and carousel width will be equal to width taken by this carousel item, taking 100% width, as expected.
3. I am attaching a video here, where i am creating carousel of 20rem width. i am setting carousel item minimum-width to be 20rem to fill the carousel. and when only 1 item is visible, carousel looks fine.
4. let me know, if this implementation is not correct.
    

https://user-images.githubusercontent.com/57661487/122268866-374ef780-cefa-11eb-8324-d09e7b4211fe.mp4

InnaAtanasova:
> > Screen Shot 2021-06-16 at 8 49 20 AM
> > Screen Shot 2021-06-16 at 8 49 02 AM
> > I think the problem with using inline styling and/or docs styling is still not clear. This styling is NOT consumed by the implementation libraries. You just make the component looks good on the documentation page, but when imported in fund-ngx, for example, you will need again custom css.
> > When you turn off the styling from the docs, the component doesn't look like as should.
> 
> @InnaAtanasova
> 
> 1. when we create carousel in Ngx, we provide width for carousel; which is width taken by single carousel item.
> 2. In case of horizontal carousel example, i have shown 3 carousel items; to show horizontal arrangement of items.
>    ideally, only 1 item will be visible and carousel width will be equal to width taken by this carousel item, taking 100% width, as expected.
> 3. I am attaching a video here, where i am creating carousel of 20rem width. i am setting carousel item minimum-width to be 20rem to fill the carousel. and when only 1 item is visible, carousel looks fine.
> 4. let me know, if this implementation is not correct.
> 
>  2021-06-16_23-21-59.mp4

@Hi @DeepakSap14 the building blocks of Carousel are made in a dynamic way to fit any content, aka the width of the elements is `100%`. Here's what I mean:
In the image bellow you will see the changes I made on your example:
1. I removed the custom css that you added
2. I added width to the images. You had before 100% but 100% of what? The images will shrink to fit the available content. When you provide width of the inner elements, the images in this case, the Carousel elements will grow to fit them. 
3. Instead of adding margin bottom you can just provide height for the example container.

What you will achieve with this:
- less code in the documentation - you don't need an extra div and custom css
- you will not confuse the user when they copy the example and paste it in their code about the need if the custom css
- show the user how to manage the width of the content inside the Carousel 

I hope this makes sense.

Screen Shot 2021-06-16 at 4 26 35 PM

InnaAtanasova:
None of these css classes are needed in the documentation:
Screen Shot 2021-06-16 at 4 54 39 PM

droshev (dismissed):
@DeepakSap14 are you sure you are fixing the whole #5052? I doubt that

.storybook/custom/custom.scss
```diff
@@ -700,3 +700,20 @@ body.sb-show-main {
 .fddocs-list-height {
   height: 11rem;
 }
+
```
droshev, line 703: why is this added?

DeepakSap14:
> @DeepakSap14 are you sure you are fixing the whole #5052? I doubt that

corrected it.

DeepakSap14:
> None of these css classes are needed in the documentation:
> Screen Shot 2021-06-16 at 4 54 39 PM

@InnaAtanasova 
custom classes are removed.
for showing Error message, using message page component.
examples created correctly as suggested by you.

please see if it looks fine.

InnaAtanasova (approved):

Issue #2443, opened 2021-06-10
Provide focus feature for Icon, Avatar components for a11y


sKudum:
**Is your feature request related to a problem? Please describe.**

In popover when image, icon is used user is not able to access it using keyboard actions.

address  https://github.com/SAP/fundamental-ngx/issues/3139  logged in ngx.

> Is this issue related to a specific component?
Icon, Avatar

> What browser are you working in?
all

> What version of Fundamental Library Styles are you using?
0.18

> What front-end framework are you implementing in, e.g., Angular, React, etc.?
angular

**Describe the solution you'd like**

Provide focus style on inclusion of tabindex to avatar,icon components.

sKudum:
Closing this issue as the fix is merged and tested.

PR #2454, opened 2021-06-14
fix: provide focus for avatar and icon components

sKudum:
## Related Issue
Closes SAP/fundamental-styles #2443 

## Description
Providing focus options for icon and avatar components when used with tabindex for accessibility purpose.

### Before:
No Focus 

### After:

Displaying focus when tabindex="0"

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/avatar.scss
+++ b/src/avatar.scss
@@ -49,6 +49,12 @@ $block: #{$fd-namespace}-avatar;
   background-position: 50%;
   border-radius: var(--sapElement_BorderCornerRadius);
 
+  @include fd-focus() {
+    outline-width: var(--sapContent_FocusWidth);
+    outline-color: var(--sapContent_FocusColor);
+    outline-style: var(--sapContent_FocusStyle);
+  }
+
   // BLOCK ELEMENTS ************
   // zoom icon
   &__zoom-icon {
--- a/src/avatar.scss
+++ b/src/avatar.scss
@@ -49,11 +49,7 @@ $block: #{$fd-namespace}-avatar;
   background-position: 50%;
   border-radius: var(--sapElement_BorderCornerRadius);
 
-  @include fd-focus() {
-    outline-width: var(--sapContent_FocusWidth);
-    outline-color: var(--sapContent_FocusColor);
-    outline-style: var(--sapContent_FocusStyle);
-  }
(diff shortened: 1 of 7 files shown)


src/avatar.scss
```diff
@@ -49,6 +49,12 @@ $block: #{$fd-namespace}-avatar;
   background-position: 50%;
   border-radius: var(--sapElement_BorderCornerRadius);
 
+  @include fd-focus() {
+    outline-width: var(--sapContent_FocusWidth);
+    outline-color: var(--sapContent_FocusColor);
+    outline-style: var(--sapContent_FocusStyle);
```
Betrozov, line 55:
```suggestion
  @include fd-fiori-focus(0);
```
please use fd-fiori-focus instead of fd-focus

src/icon.scss
```diff
@@ -13,6 +13,12 @@ $blockTNT: sap-icon-TNT;
 
   font-style: normal;
   display: inline-block;
+
+  @include fd-focus() {
```
Betrozov, line 17: same

Betrozov (changes requested):

src/avatar.scss
sKudum: Hi Yuri,

It uses _outline-offset_ which is not supported in IE11, to avoid this issue used fd-focus.

src/icon.scss
sKudum: Hi Yuri,

It uses _outline-offset_ which is not supported in IE11, to avoid this issue used fd-focus.

Betrozov (approved):

src/mixins/_states.scss
```diff
@@ -51,7 +51,10 @@
 @mixin fd-fiori-focus($offset: -0.1875rem) {
   &:focus,
   &.is-focus {
-    outline-offset: $offset;
+    @if $offset != 0 {
+      outline-offset: $offset;
+    }
```
Betrozov, line 56: @InnaAtanasova please verify these changes

InnaAtanasova (approved):

Seamoo13 (approved):

Issue #2073, opened 2021-01-21
Implement IconTabBar component with UniversalIconTabBar and Badge features


InnaAtanasova:
- [ ] Single Click Area
- [ ] Two Click Areas
- [ ] Semantic tab bar (with icons and colors)
- [ ] Shell Navigation color delta
- [ ] Dropdown menu with/without separators
- [ ] IconTabBar with badges
- [ ] Badge inside dropdown meneu

PR #2455, opened 2021-06-15
feat: introduce new Icon Tab Bar component

InnaAtanasova:
## Related Issue
Closes SAP/fundamental-styles#2073

## Description
Introduce new Icon Tab Bar which is similar to the existing Tabs component with the following extra features
- Single Click Area
- Two Click Areas
- Semantic tab bar (with icons and colors and theming)
- Shell Navigation color delta
- Dropdown menu with/without separators
- IconTabBar with badges
- Badge inside dropdown mene

- [x] All values are in `rem`
- [x] Text elements follow the truncation rules - NOTE: THE TABS DO NOT TRUNCATE PER FIORI 3 SPECS
- [x] focus state of the element follow design spec - NOTE: the focus for overflow button to be confirmed later by the designers. As of today it's still WIP in the design
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/stories/icon-tab-bar/icon-tab-bar.stories.js
+++ b/stories/icon-tab-bar/icon-tab-bar.stories.js
@@ -1,0 +1,1431 @@
+export default {
+    title: 'Components/Icon Tab Bar',
+    parameters: {
+        description: `
+The Icon Tab Bar can be used for navigation within an object, or as a filter. Each tab of the component links to a different content area or view.
+
+##Usage
+**Use the icon tab bar if:**
+
+- Your business objects need to show multiple facets at the same time.
+- You want to allow the user to browse through these facets.
+- You need a prominent or very visual filter on top of a list.
+- You have clear-cut process steps that need to be visualized.
+
+**Do not use the icon tab bar if:**
+
+- You plan to use only a single tab.
+
+##Semantic colors
+Only use semantic colors if it is important for users to know that they need to take action (for example, to indicate errors or critical situations requiring action). Otherwise, use the neutral default colors. To add semantic color to a tab use a modifier class from the table below on <code>fd-icon-tab-bar\\_\\_item</code> level.
+
+| **Color** | **Modifier class** |
+| :--------- | :---------- |
+| positive &nbsp;&nbsp; | <code>fd-icon-tab-bar\\_\\_item--positive</code> |
+| negative &nbsp;&nbsp; | <code>fd-icon-tab-bar\\_\\_item--negative</code> |
+| critical &nbsp;&nbsp; | <code>fd-icon-tab-bar\\_\\_item--critical</code> |
+| informative &nbsp;&nbsp; | <code>fd-icon-tab-bar\\_\\_item--informative</code> |
+
+##Horizontal paddings
+You can add horizontal paddings by applying a modifier class to the container. For responsive horizontal paddings (based on the screen size) add the <code>fd-icon-tab-bar--responsive-paddings</code> modifier class. In this case the left and right spacing will change as the screen size changes.
+
+| **Size** | **Modifier class** |
+| :--------- | :---------- |
+| sm (1rem) &nbsp;&nbsp; | <code>fd-icon-tab-bar--sm</code> |
+| md (2rem) &nbsp;&nbsp; | <code>fd-icon-tab-bar--md</code> |
+| lg (2rem) &nbsp;&nbsp; | <code>fd-icon-tab-bar--lg</code> |
+| xl (3rem) &nbsp;&nbsp; | <code>fd-icon-tab-bar--xl</code> |
+| xxl (3rem) &nbsp;&nbsp; | <code>fd-icon-tab-bar--xxl</code> |
+
+##Background
+By default, the background for the Icon Tab Bar header and content is set to "solid". This will apply <code>--sapObjectHeader_Background</code> background color and <code>--sapContent_HeaderShadow</code> box-shadow to the header and <code>--sapBackgroundColor</code> to the container (panel). It can be changed to "translucent" or "transparent" using modifier classes. In translucent mode the header gets <code>--sapObjectHeader_Background</code> background color and the panel <code>--sapGroup_ContentBackground</code> background color. 
+
+| **Background** | **Modifier class** |
+| :--------- | :---------- |
+| solid &nbsp;&nbsp; | default |
+| translucent &nbsp;&nbsp; | <code>fd-icon-tab-bar--translucent</code> |
+| transparent &nbsp;&nbsp; | <code>fd-icon-tab-bar--transparent</code> |
+
+<br><br><br>
+      `,
+        tags: ['f3', 'theme', 'development'],
+        components: ['icon', 'icon-tab-bar', 'popover', 'menu', 'button', 'list']
+    }
+};
+
+
+export const textOnly = () => `<div class="fd-icon-tab-bar">
+    <ul role="tablist" class="fd-icon-tab-bar__header">  
+        <li role="presentation" class="fd-icon-tab-bar__item">
+            <a role="tab" class="fd-icon-tab-bar__tab" href="#section1" id="tab1">
+                <span class="fd-icon-tab-bar__tag">Info</span>
+            </a>
+        </li>
+        <li role="presentation" class="fd-icon-tab-bar__item">
+            <a role="tab" class="fd-icon-tab-bar__tab" href="#section2" aria-selected="true" id="tab2">
+                <span class="fd-icon-tab-bar__tag">Attachments (16)</span>
+            </a>
+        </li>
+        <li role="presentation" class="fd-icon-tab-bar__item">
+            <a role="tab" class="fd-icon-tab-bar__tab" href="#section3" id="tab3">
+                <span class="fd-icon-tab-bar__tag">Approvals (42)</span>
+            </a>
+        </li>
+    </ul> 
+    <section role="tabpanel" class="fd-icon-tab-bar__panel" id="section1" aria-labelledby="tab1">  
+        Section 1
+    </section>  
+    <section role="tabpanel" class="fd-icon-tab-bar__panel" id="section2" aria-labelledby="tab2" hidden>  
+        Section 2
+    </section>  
+    <section role="tabpanel" class="fd-icon-tab-bar__panel" id="section3" aria-labelledby="tab3" hidden>  
+        Section 3
+    </section>
+</div>
+`;
+textOnly.storyName = 'Text Only (Inline mode)';
+textOnly.parameters = {
+    docs: {
+        storyDescription: 'The text-only variant is one of the most common types. It allows longer labels, and can also display counters next to the text to indicate the number of items on the tab page. The labels of the tabs do not get truncated. The full text is always shown. As a result, you need to ensure that your labels do not become too long. They should still be easy to read on smaller screens. The control has the same look and feel in Cozy and Compact mode.'
+    }
+};
+
+export const textOnlySemantic = () => `<div class="fd-icon-tab-bar">
+    <ul role="tablist" class="fd-icon-tab-bar__header">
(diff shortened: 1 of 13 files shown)


N1XUS (approved):

stories/icon-tab-bar/icon-tab-bar.stories.js
```diff
@@ -0,0 +1,1434 @@
...
+- You want to allow the user to browse through these facets.
+- You need a prominent or very visual filter on top of a list.
+- You have clear-cut process steps that need to be visualized.
+
+**Do not use the icon tab bar if:**
+
+- You plan to use only one single tab.
+
+##Semantic colors
+Only use semantic colors if it is important for users to know that they need to take action (for example, to indicate errors or critical situations requiring action). Otherwise, use the neutral default colors. To add semantic color to a tab use a modifier class from the table below on <code>fd-icon-tab-bar\\_\\_item</code> level.
```
kari-leg, line 20: Rewrite: Only use semantic colors if it is important for users to know when they need to take action..

stories/icon-tab-bar/icon-tab-bar.stories.js
```diff
@@ -0,0 +1,1434 @@
...
+| **Size** | **Modifier class** |
+| :--------- | :---------- |
+| sm (1rem) &nbsp;&nbsp; | <code>fd-icon-tab-bar--sm</code> |
+| md (2rem) &nbsp;&nbsp; | <code>fd-icon-tab-bar--md</code> |
+| lg (2rem) &nbsp;&nbsp; | <code>fd-icon-tab-bar--lg</code> |
+| xl (3rem) &nbsp;&nbsp; | <code>fd-icon-tab-bar--xl</code> |
+| xxl (3rem) &nbsp;&nbsp; | <code>fd-icon-tab-bar--xxl</code> |
+
+##Background
+By default the background for the Icon Tab Bar header and content is set to "solid". This will apply <code>--sapObjectHeader_Background</code> background color and <code>--sapContent_HeaderShadow</code> box-shadow to the header and <code>--sapBackgroundColor</code> to the container (panel). It can be changed to "translucent" or "transparent" using modifier classes. In translucent mode the header gets <code>--sapObjectHeader_Background</code> background color and the panel <code>--sapGroup_ContentBackground</code> background color.
```
kari-leg, line 41: Rewrite: By default, the background for the Icon Tab Bar header and content is set as "solid".

stories/icon-tab-bar/icon-tab-bar.stories.js
```diff
@@ -0,0 +1,1434 @@
...
+                <div class="fd-icon-tab-bar__label">Overweight</div>
+            </a>
+        </li>
+    </ul>
+</div>`;
+filter.storyName = 'Filter';
+filter.parameters = {
+    docs: {
+        storyDescription: `The tab bar as a filter has two parts: <br>
+        - An “all” tab on the left - shows the total numbeer of items, and describes the type of item (for example, 189 Products). <br>
```
kari-leg, line 829: Rewrite: An “all” tab on the left - shows the total number of items..

stories/icon-tab-bar/icon-tab-bar.stories.js
```diff
@@ -0,0 +1,1434 @@
...
+                <span class="fd-icon-tab-bar__overflow-text">More</span>
+                <i class="sap-icon--slim-arrow-down" role="presentation"></i>
+            </button>
+        </li> 
+    </ul>
+</div>`;
+overflow.storyName = 'Overflow';
+overflow.parameters = {
+    docs: {
+        storyDescription: `Tabs can have an overflow to either of the sides of the control. Depending on if the order is fixed (process steps) or can be rearranged (tabs).
```
kari-leg, line 899: Rewrite: `Tabs can have an overflow to either sides of the control.

stories/icon-tab-bar/icon-tab-bar.stories.js
```diff
@@ -0,0 +1,1434 @@
...
+                </a>
+            </li>
+        </ul>
+    </div>
+</div>
+`;
+singleClick.storyName = 'Single Click Icon Tab Bar';
+singleClick.parameters = {
+    docs: {
+        storyDescription: `When there is only one click area per tab (also including tabs with sub-items), regular tabs get selected immediately on after the click is released. Tabs with sub-items trigger the expand of a menu (Popover) showing its sub-items. The parent tab title remains displayed when its child is select.
```
kari-leg, line 1050: Rewrite: `When there is only one click area per tab (also including tabs with sub-items), regular tabs get selected immediately after the click is released. Tabs with sub-items trigger the expansion of a menu (Popover) showing its sub-items...

stories/icon-tab-bar/icon-tab-bar.stories.js
```diff
@@ -0,0 +1,1434 @@
...
+                </a>
+            </li>
+        </ul>
+    </div>
+</div>
+`;
+multiClick.storyName = 'Multi Click Icon Tab Bar';
+multiClick.parameters = {
+    docs: {
+        storyDescription: 'In case of two click areas for tabs with sub-tabs, the behaviour for regular tabs (without sub-items) remains unchanged, like described above the tabs get selected immediately. Tabs with two click areas behave differently: when the main part of the tab is clicked, it is highlighted and selected immediately. Clicking the second area opens a menu (Popover) containing the sub-items. Once a sub-item is clicked, the main tab is highlighted as selected. '
```
kari-leg, line 1148: rewrite: ,it is highlighted and then selected immediately..

stories/icon-tab-bar/icon-tab-bar.stories.js
```diff
@@ -0,0 +1,1434 @@
...
+      </a>
+    </li>
+  </ul>
+</div>
+`;
+badge.storyName = 'Icon Tab Bar with Badge';
+badge.parameters = {
+    docs: {
+        storyDescription: `The Tab attention badge is applied as a visual eye-catcher to indicate a change within a Tab. <br>
+        The badge is a span element with <code>fd-icon-tab-bar\\_\\_badge</code> class. It can be animated by adding an additional <code> fd-icon-tab-bar\\_\\_badge--animated</code> modifier class. In this case the badge appears for a few miliseconds and disappears.`
```
kari-leg, line 1292: Rewrite: In this case the badge appears for a few milliseconds and then disappears.`

kari-leg (changes requested):
Just made some syntax & typo corrections- The rest LGTM!

stories/icon-tab-bar/icon-tab-bar.stories.js
InnaAtanasova: I didn't change this one.  Replacing `that` with `when` might confuse the user. The user in all cases should take action when the item has semantic color. If we use `when` we might make them think that the color is associated with some time.

kari-leg: okay, noted!

kari-leg (approved):

kavya-b (approved):
Incredible work! LGTM, just few minor comments.

stories/icon-tab-bar/icon-tab-bar.stories.js
```diff
@@ -0,0 +1,1431 @@
...
+            <li role="presentation" class="fd-icon-tab-bar__item">
+                <a role="tab" class="fd-icon-tab-bar__tab" href="#" id="tab4">
+                    <span class="fd-icon-tab-bar__tag">Section 1</span>
+                    <span class="fd-icon-tab-bar__badge"></span>
+                </a>
+            </li>
+            <li role="presentation" class="fd-icon-tab-bar__item fd-icon-tab-bar__item--single-click">
+                <div class="fd-popover">
+                    <div class="fd-popover__control">
+                        <a role="tab" class="fd-icon-tab-bar__tab" aria-selected="true" id="tab5" tabindex="0" aria-controls="popoverA1" aria-expanded="true" aria-haspopup="true" onclick="onPopoverClick('popoverA2');">
```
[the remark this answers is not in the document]
InnaAtanasova: Nice catch, thank you!

stories/icon-tab-bar/icon-tab-bar.stories.js
```diff
@@ -0,0 +1,1431 @@
...
+`;
+respPaddings.storyName = 'Responsive Paddings';
+respPaddings.parameters = {
+    docs: {
+        storyDescription: 'For an Icon Tab Bar with responsive paddings that adjust based on the screen size add the <code>fd-icon-tab-bar--responsive-paddings</code> modifier class.'
+    }
+};
+
+export const paddings = () => `<h4>SM (1rem) paddings</h4>
+<div class="fd-icon-tab-bar fd-icon-tab-bar--sm">
```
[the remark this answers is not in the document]
InnaAtanasova: I thought of using the helper classes (because that's why we added them 😄 ) but the thing is that they are not really consistent and will confuse the user.
This is the specs for IconTabBar:
Screen Shot 2021-06-18 at 2 58 29 PM
`sm` and `md` will work because they match the values for paddings:
Screen Shot 2021-06-18 at 3 03 28 PM

In these 2 cases we can use `fd-padding-begin-end--sm` and `fd-padding-begin-end--md` but then is the problem with LG, XL and XXL. `fd-padding-begin-end--lg` adds paddings left and right of 3rem but should be 2rem for IconTabBar. Also we don't have paddings for XL and XXL so we will need the additional modifier classes for IconTabBar such as `fd-icon-tab-bar--xxl` and `fd-icon-tab-bar--xl`. For this reason I added `fd-icon-tab-bar--SIZE` for all sizes so that it will be easily handled in the NGX implementation where we will just change the `--SIZE` part of the same class.

stories/icon-tab-bar/icon-tab-bar.stories.js
```diff
@@ -0,0 +1,1431 @@
...
+
+| **Color** | **Modifier class** |
+| :--------- | :---------- |
+| positive &nbsp;&nbsp; | <code>fd-icon-tab-bar\\_\\_item--positive</code> |
+| negative &nbsp;&nbsp; | <code>fd-icon-tab-bar\\_\\_item--negative</code> |
+| critical &nbsp;&nbsp; | <code>fd-icon-tab-bar\\_\\_item--critical</code> |
+| informative &nbsp;&nbsp; | <code>fd-icon-tab-bar\\_\\_item--informative</code> |
+
+##Horizontal paddings
+You can add horizontal paddings by applying a modifier class to the container. For responsive horizontal paddings (based on the screen size) add the <code>fd-icon-tab-bar--responsive-paddings</code> modifier class. In this case the left and rigth spacing will change as the screen size changes.
```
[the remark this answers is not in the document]
InnaAtanasova: Thank you!

stories/icon-tab-bar/icon-tab-bar.stories.js
```diff
@@ -0,0 +1,1431 @@
...
+    <section role="tabpanel" class="fd-icon-tab-bar__panel" id="section4" aria-labelledby="tab4" hidden>Section 4</section>
+</div>
+`;
+textOnlySemantic.storyName = 'Text Only (Inline mode) Semantic Colors';
+
+
+export const counter = () => `<div class="fd-icon-tab-bar fd-icon-tab-bar--counters">
+    <ul role="tablist" class="fd-icon-tab-bar__header">  
+        <li role="presentation" class="fd-icon-tab-bar__item">
+            <a role="tab" class="fd-icon-tab-bar__tab" href="#section1" id="tab1">
```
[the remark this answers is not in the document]
InnaAtanasova: We can, but I don't see much value in it. We don't have logic (JavaScript) involved in Fundamental-styles and the ids are added just for the sake of the example. Also each component is in a "isolation" in Storybook and Storybook does not detect a11y violations:
Screen Shot 2021-06-18 at 3 21 04 PM



If we don't have unique values in the same example then Storybook detects the a11y violation:
Screen Shot 2021-06-18 at 3 27 53 PM
Screen Shot 2021-06-18 at 3 28 21 PM

PR #2458, opened 2021-06-17
chore: remove old unused code

droshev:
## Description
 remove old unused code

it should not affect anything as it was not used


src/mixins/_mixins.scss
```diff
@@ -487,3 +487,20 @@
     padding-left: $right;
   }
 }
+
```
droshev, line 490: moved it from the delete file `_layouts.scss`

InnaAtanasova (approved):

PR #2473, opened 2021-06-22
fix: change tabs styles for routerLink

OleksiyDrozdov:
## Related Issue
 Part of [#5128](https://github.com/SAP/fundamental-ngx/pull/5735)

- [x] All values are in `rem`
- [n/a] RTL support
3. Testing
- [n/a] Updated tests
4. Documentation

### DIFF
--- a/src/tabs.scss
+++ b/src/tabs.scss
@@ -114,15 +114,14 @@ $block: #{$fd-namespace}-tabs;
     line-height: 1rem;
     padding: $fd-tabs-link-padding-y $fd-tabs-item-padding-x;
 
-    &.is-selected {
-      .fd-tabs__tag::after {
+    @include fd-selected() {
+      .#{$block}__icon {
+        color: var(--fdTabs_Selected_Icon_Color, var(--sapContent_ContrastIconColor));
         background-color: var(--fdTabs_Selected_Color);
       }
 
-      .fd-tabs__icon {
-        color: var(--fdTabs_Selected_Icon_Color, var(--sapContent_ContrastIconColor));
-        background-color: var(--fdTabs_Selected_Color);
-
+      .#{$block}__icon,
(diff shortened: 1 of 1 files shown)


src/tabs.scss
```diff
@@ -114,6 +114,21 @@ $block: #{$fd-namespace}-tabs;
     line-height: 1rem;
     padding: $fd-tabs-link-padding-y $fd-tabs-item-padding-x;
 
+    &.is-selected {
```
InnaAtanasova, line 117: You need to use `fd-selected()` mixin

src/tabs.scss
```diff
@@ -114,6 +114,21 @@ $block: #{$fd-namespace}-tabs;
...
+      .fd-tabs__tag::after {
+        background-color: var(--fdTabs_Selected_Color);
+      }
+
+      .fd-tabs__icon {
```
InnaAtanasova, line 122: .#{$block}__icon

src/tabs.scss
```diff
@@ -114,6 +114,21 @@ $block: #{$fd-namespace}-tabs;
     line-height: 1rem;
     padding: $fd-tabs-link-padding-y $fd-tabs-item-padding-x;
 
+    &.is-selected {
+      .fd-tabs__tag::after {
```
InnaAtanasova, line 118: .#{$block}__tag

InnaAtanasova (changes requested):
You are missing the word `closes` or `fixes` in front of the PR number and description with before/after screenshots. You need to add reviewers, labels, project, etc.

InnaAtanasova (approved):

InnaAtanasova:
@OleksiyDrozdov UI tests are not triggered

dimamarksman (approved):

Issue #2477, opened 2021-06-23
place element off screen such that it is not visible but can be read by screen reader


DeepakSap14:
### Component name
> offScreen

### Detailed description
> Describe the request
For aria-live to read any change in element, we need to place element but it should not be visible to user.
same time, it can not be made `hidden` or `display: none`, as screen reader will not be able to pick this element.

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?
NA

> What front-end framework are you implementing in, e.g., Angular, React, etc.?
Angular

### Use case
> Describe when to use (and when not to use)
For aria-live to read any change in element, we need to place element but it should not be visible to user.
same time, it can not be made `hidden` or `display: none`, as screen reader will not be able to pick this element.

### Visual style
> Describe specific needs related to color, typography, iconography, borders, space, size, and other visual properties.
NA
> Describe how to arrange multiple instances and other alignment specifics.
NA
> How should this respond to common contexts, i.e., touch, RTL, small screens.
NA
> Describe interactive states, e.g., disabled, expanded, pressed.
NA

### Expected behavior
> How should this handle events, outcomes, and transitions
NA
> What [states should be available](https://medium.com/swlh/the-nine-states-of-design-5bfe9b3d6d85), i.e., nothing, loading, none, one, some, too many, incorrect, correct, done.
NA

### Editorial
> Advise on labels, tone, length, punctuation, truncation.
NA

## Additional information

* Screenshots or code
* Notes
NA

PR #2478, opened 2021-06-23
feat: create component to place an element off screen

DeepakSap14:
## Related Issue
Closes SAP/fundamental-styles#2477

## Description
Requirement coming from https://github.com/SAP/fundamental-ngx/pull/5673#discussion_r653824449.

For aria-live to read any change in element, we need to place element but it should not be visible to user.
same time, it can not be made hidden or display: none, as screen reader will not be able to pick this element.

- [x] All values are in `rem`
- [NA] RTL support
3. Testing
- [x] Updated tests
4. Documentation


InnaAtanasova:
Screen Shot 2021-06-23 at 10 47 50 AM

@DeepakSap14 please move this to the `Dev` folder. This is a helper component.

InnaAtanasova (changes requested):
The component is not exported, it can't be used in fund-ngx

src/off-screen.scss
```diff
@@ -0,0 +1,26 @@
...
+$block: #{$fd-namespace}-off-screen;
+
+.#{$block} {
+  @include fd-reset();
+
+  /*
+  * Element with aria-live will be announced on change in value, but
+  * this elemen should not be visible on screen, but also can't be made hidden or display none because of
+  * for screen reader requirement
+  * please see: https://webaim.org/techniques/css/invisiblecontent/
```
kavya-b, line 18:
```suggestion
  * Element with aria-live will be announced on change in value, but
  * this element should not be visible on screen, and also can't be made `visibility: hidden` or `display: none` 
  * because of screen reader requirement
  * please see: https://webaim.org/techniques/css/invisiblecontent/
```

stories/off-screen/off-screen.stories.js
```diff
@@ -0,0 +1,22 @@
+export default {
+    title: 'Components/Off Screen',
+    parameters: {
+        description: `For aria-live to read any change in element, we need to place element but it should not be visible to user.
+            same time, it can not be made hidden or display: none, as screen reader will not be able to pick this element.
+            This component will make sure that element is not visible on screen but it can be read by screen reader.`,
```
kavya-b, line 6:
```suggestion
        description: `For aria-live to read any change in element, we need to place element but it should not be visible to user.
            At the same time, it can not be made \`visibility: hidden\` or \`display: none\`, as screen reader will not be able to pick this element.
            This component will make sure that element is not visible on screen but it can be read by screen reader.`,
```

src/off-screen.scss
DeepakSap14: fixed description

stories/off-screen/off-screen.stories.js
DeepakSap14: fixed

kavya-b (approved):

InnaAtanasova (approved):

PR #2480, opened 2021-06-24
fix: provide padding if title is first element in bar

### LINKED ISSUE #2479: Bar - if title is used as first control, then left/right padding should be 1rem
**Describe the bug**
Visual Core - says if title is used as first control, then left padding should be 1rem vice versa in rtl mode.

Bar Design SubHeader
Visual Core

Bar

latest

Angular

sKudum:
## Related Issue
Closes #2479 

## Description
BREAKING CHANGE:

Added modifier class `fd-bar__element--title` class to bar component when title is placed as first element to add 0.5rem padding to it as per visual core.

Before:
```
  <span aria-label="text">TEXT</span>
```
After:
```
  <span class="fd-bar__element fd-bar__element--title" aria-label="text">TEXT</span>
```

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/bar.scss
+++ b/src/bar.scss
@@ -28,6 +28,16 @@ $block: #{$fd-namespace}-bar;
     right: $margin-x;
   }
 
+  @mixin padding-left-right($left, $right) {
+    padding-left: $left;
+    padding-right: $right;
+
+    @include fd-rtl() {
+      padding-left: $right;
+      padding-right: $left;
+    }
+  }
+
   @include fd-reset();
   @include bar-design();
 
@@ -83,6 +93,12 @@ $block: #{$fd-namespace}-bar;
 
     & > *:first-child {
       flex-grow: 1;
+    }
+  }
+
+  &__left > &__element:first-child {
+    span:nth-child(1) {
+      @include padding-left-right(1rem, 0);
     }
   }
 
--- a/src/bar.scss
+++ b/src/bar.scss
@@ -28,16 +28,6 @@ $block: #{$fd-namespace}-bar;
     right: $margin-x;
   }
 
-  @mixin padding-left-right($left, $right) {
-    padding-left: $left;
-    padding-right: $right;
-
-    @include fd-rtl() {
-      padding-left: $right;
-      padding-right: $left;
-    }
-  }
-
   @include fd-reset();
   @include bar-design();
 
@@ -86,6 +76,10 @@ $block: #{$fd-namespace}-bar;
       margin-right: 0;
       margin-left: $fd-bar-element-spacing;
     }
+
+    &--title {
+      @include fd-set-paddings-x(0.5rem,0);
+    }
   }
 
   &__element--full-width {
@@ -93,12 +87,6 @@ $block: #{$fd-namespace}-bar;
 
     & > *:first-child {
       flex-grow: 1;
-    }
-  }
-
-  &__left > &__element:first-child {
-    span:nth-child(1) {
-      @include padding-left-right(1rem, 0);
     }
   }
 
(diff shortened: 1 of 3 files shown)


Betrozov (approved):

platon-rov (approved):

N1XUS (approved):

InnaAtanasova (changes requested):
Screen Shot 2021-06-25 at 3 28 20 PM

1. The Bar component in all cases has padding-left and padding-right of 0.5rem. The specs mention that when Title control is used as first child then the spacing should be 1rem. The way you implemented it is for all SPAN elements placed inside `fd-bar__element` that is a first child. The Title control is an h1-h6 element:
```
<h1 class="fd-title fd-title--h1">Title Level 1</h1>
<h2 class="fd-title fd-title--h2">Title Level 2</h2>
<h3 class="fd-title fd-title--h3">Title Level 3</h3>
<h4 class="fd-title fd-title--h4">Title Level 4</h4>
<h5 class="fd-title fd-title--h5">Title Level 5</h5>
<h6 class="fd-title fd-title--h6">Title Level 6</h6>
```

You need to just add extra 0.5rem padding left to achieve a total of 1rem as shown in the specs. The easiest way would be to add a modifier class to the `fd-bar__element` element. You can have `fd-bar__element fd-bar__element--title` and then have:
```
.fd-bar__element--title {
    @include fd-set-padding-left(0.5rem)
}
```

In ngx this can be done with an input prop of the `fd-bar-element` component. If the `title` prop is set to true, the modifier class is added and the additional spacing is achieved.

In ngx documentation this part is not a bug:
Screen Shot 2021-06-25 at 3 18 59 PM

This is just a normal text indicating the 3 different sections of the Bar, it's not a Title.

src/bar.scss
```diff
@@ -86,6 +96,12 @@ $block: #{$fd-namespace}-bar;
     }
   }
 
+  &__left > &__element:first-child {
+    span:nth-child(1) {
```
InnaAtanasova, line 100: this will be applied for all SPAN elements and actually never for Title component which is h1...h6 element.

src/bar.scss
```diff
@@ -28,6 +28,16 @@ $block: #{$fd-namespace}-bar;
...
 
+  @mixin padding-left-right($left, $right) {
+    padding-left: $left;
+    padding-right: $right;
+
+    @include fd-rtl() {
+      padding-left: $right;
+      padding-right: $left;
+    }
+  }
```
InnaAtanasova, line 39: There's already a mixin for this. Please check the `src/mixins/_mixins.scss` file and `fd-set-margins-x` mixin.

src/bar.scss
sKudum: removed it

sKudum:
Hi @InnaAtanasova ,

     Sorry for the delay was in sick off
     I have updated the changes as suggested above.

   Do I need to mark it as breaking change? as we are introducing a modifier here? Please do suggest.

Thank You

InnaAtanasova:
> Hi @InnaAtanasova ,
> 
> ```
>  Sorry for the delay was in sick off
>  I have updated the changes as suggested above.
> ```
> 
> Do I need to mark it as breaking change? as we are introducing a modifier here? Please do suggest.
> 
> Thank You

Yes, please add this as a Breaking change. Thank you!

InnaAtanasova (approved):

sKudum:
Hi @InnaAtanasova ,

     Updated it with breaking changes, ready for merge now. Shall I go ahead or do you want me to valor team.

Issue #2305, opened 2021-04-29
Create Theming components pages


droshev:
- [x] create a new npm package (e.g. `@fundamental-styles/theming-preview` - to be discussed)
- [x] create sample pages for the components (
- [x] include `testsuite.json` - we don't need to have `controls` part

PR #2481, opened 2021-06-25
docs: create theming preview sample pages

hanjunt:
## Related Issue
Part of SAP/fundamental-styles#2305

## Description
Create sample pages for the components

1. The output matches the design specs - N/A

2. The code follows fundamental-styles code standards and style - N/A

3. Testing
- N/A tested Storybook examples with "CSS Resources" `normalize` option 
- N/A tested Storybook examples with "CSS Resources" `unnormalize` option 
- N/A Verified all styles in IE11
- N/A Updated tests

4. Documentation
- N/A Storybook documentation has been created/updated
- N/A Breaking Changes [wiki](https://github.com/SAP/fundamental-styles/wiki/Breaking-Changes) has been updated in case of breaking changes.


droshev (changes requested):
@hanjunt thanks for starting the content.
Few comments:
 - the content doesn't match the theming preview components. Can you stick to their content and only add the variants that they don't have and we have(see my commit with semantic colors, where i added the large design)
 - each example should be wrapped in a panel

droshev (changes requested):
That is a good progress. 
Can we add few more examples:
1. Button 
-> add Attention Button (in button, menu button, toggle button
-> Segmented Button with icon and text
2. Form - use text for read-only inputs
3. Panel - can you add panel examples with toolbars
4. Indication Colors - can you add large design examples of object status
5. Shell, Tiles, Cards 
-> Can you use initials for the user menu icon, and remove the menu
-> card -> can you add initials for the card avatar
6. Table - tables are missing data
7. Link - add inverted colors link
8. Toolbar 
-> remove the hardcoded background color, replace the 50% width with left/right padding
-> add all toolbar types examples, overflow(without opened menu, separator

droshev (approved):

PR #2483, opened 2021-06-26
fix: add bottom border to panel content

### LINKED ISSUE #2482: panel missing bottom border when expanded
The current implementation of the panel is missing
- [x] the bottom border when expanded:
```
border-bottom: solid 0.0625rem --sapGroup_TitleBorderColor; 
```

droshev:
## Related Issue
fixes #2482 

## Description
add panel content bottom border

### Before:
Screen Shot 2021-06-26 at 6 31 26 PM
Screen Shot 2021-06-26 at 6 31 20 PM

### After:
Screen Shot 2021-06-26 at 6 30 56 PM
Screen Shot 2021-06-26 at 6 30 48 PM

- [X] All values are in `rem`
- [na] RTL support
3. Testing
- [na] Updated tests
4. Documentation

### DIFF
--- a/src/panel.scss
+++ b/src/panel.scss
@@ -37,6 +37,7 @@ $block: #{$fd-namespace}-panel;
   &__content {
     padding: 0.625rem 1rem 1.375rem 1rem;
     overflow: auto;
+    border-bottom: solid 0.0625rem var(--sapGroup_TitleBorderColor);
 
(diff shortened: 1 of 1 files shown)


InnaAtanasova (approved):

InnaAtanasova:
Please complete the tasks in the PR template.

PR #2524, opened 2021-07-02
fix: popover style fix for Nested list component. Move nested-list styles from Ngx

DeepakSap14:
## Related Issue

## Description
There are some styles changes done for popover inside Nested list component and it is currently in fundamental-ngx.
This change is specific to nested list component. so i have moved those styles to side-nav--condensed class.

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/side-nav.scss
+++ b/src/side-nav.scss
@@ -111,6 +111,10 @@ $has-child-content-arrow-margin: 0.125rem;
           }
         }
       }
+
+      &__popover-header {
+        box-shadow: var(--sapContent_HeaderShadow);
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


DeepakSap14:
personally i feel that these changes are related to popover component and it should not be moved to styles.
please let me know your thought.

src/side-nav.scss
```diff
@@ -72,6 +72,10 @@ $has-child-content-arrow-margin: 0.125rem;
   &--condensed {
     width: 2.75rem;
 
+    .#{$fd-namespace}-popover-custom {
+      display: block;
+    }
```
InnaAtanasova, line 77: Can't modify popover inside Side nav. This is against the rules for self-contained elements.

InnaAtanasova (changes requested):

src/side-nav.scss
DeepakSap14: created modifier class for this

PR #2525, opened 2021-07-02
fix: fix action button in Rtl mode for grid list component. Move grid list styles from Ngx.

DeepakSap14:
## Related Issue
part of https://github.com/SAP/fundamental-ngx/issues/5025

Breaking Change:
Added a modifier class `fd-grid-list__btn-navigation` to the navigation button to support RTL.

wiki link [here](https://github.com/SAP/fundamental-styles/wiki/Breaking-Changes#grid-list-2525)

## Description
This fixes and moves the grid list styles present in fundamental-ngx.
1. fix action button in Rtl mode
2. fix color for decline button on active hover
3. toolbar paddings

### Before:
Screenshot 2021-07-02 at 5 30 59 PM

Screenshot 2021-07-02 at 5 22 02 PM

### After:
Screenshot 2021-07-02 at 5 33 41 PM

Screenshot 2021-07-02 at 5 36 32 PM

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/list-grid.scss
+++ b/src/list-grid.scss
@@ -175,17 +175,17 @@ $block: #{$fd-namespace}-grid-list;
           }
         }
 
-        .fd-grid-list__item-input,
+        .#{$item}-input,
         &--actions,
         &--extra-content {
-          display: flex;
-          align-items: center;
+          @include fd-flex-vertical-center();
+
           height: 100%;
         }
       }
     }
 
-    .drop-area__line--vertical {
+    .fd-drop-area__line--vertical {
       height: 100%;
 
       &.before {
@@ -320,7 +320,7 @@ $block: #{$fd-namespace}-grid-list;
 
   &__filter,
   &__filter-button {
-    color: var(--sapHighlightTextColor);
+    color: var(--sapHighlightTextColor) !important;
   }
 
   &__filter-button {
--- a/src/list-grid.scss
+++ b/src/list-grid.scss
@@ -305,7 +305,7 @@ $blockToolbar: #{$fd-namespace}-toolbar;
 
     @include fd-active() {
       @include fd-hover() {
-        color: var(--sapHighlightTextColor);
+        color: var(--sapHighlightTextColor) !important;
       }
     }
   }
--- a/src/list-grid.scss
+++ b/src/list-grid.scss
@@ -259,23 +259,24 @@ $block: #{$fd-namespace}-grid-list;
     }
   }
 
-  &__filter {
-    @include fd-fiori-focus();
+  .#{$block}__filter {
+    &[class*="active"] {
+      @include fd-fiori-focus();
+
+      color: var(--sapHighlightTextColor);
+    }
   }
 
-  &__filter,
-  &__filter-button {
-    color: var(--sapHighlightTextColor) !important;
-  }
+  .#{$block}__filter-button {
+    color: var(--sapHighlightTextColor);
 
-  &__filter-button {
     @include fd-hover() {
-      color: var(--sapButton_TextColor) !important;
+      color: var(--sapButton_TextColor);
     }
 
     @include fd-active() {
       @include fd-hover() {
-        color: var(--sapHighlightTextColor) !important;
+        color: var(--sapHighlightTextColor);
       }
     }
   }
--- a/src/list-grid.scss
+++ b/src/list-grid.scss
@@ -5,7 +5,6 @@
 .fd-grid-list
 */
 $block: #{$fd-namespace}-grid-list;
-$blockToolbar: #{$fd-namespace}-toolbar;
 
 .#{$block} {
   $item: #{$block}__item;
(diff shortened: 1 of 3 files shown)


src/list-grid.scss
```diff
@@ -138,6 +142,34 @@ $block: #{$fd-namespace}-grid-list;
...
+        }
+
+        .#{$item}-input,
+        &--actions,
+        &--extra-content {
+          @include fd-flex-vertical-center();
+
+          height: 100%;
+        }
+      }
```
InnaAtanasova, line 172: You can't do this, it's violating the rules for self-contained elements. You can't modify the properties of one component (Toolbar) inside another (List-grid). You need a modifier class.

src/list-grid.scss
```diff
@@ -332,6 +370,7 @@ $block: #{$fd-namespace}-grid-list;
       width: 0.375rem;
       display: block;
       border-radius: $fd-grid-list-border-radius 0 0 $fd-grid-list-border-radius;
+      z-index: 1;
```
InnaAtanasova, line 373: why do you need this? We rely on z-index as a last resort

InnaAtanasova (changes requested):

src/list-grid.scss
DeepakSap14: current implementation of grid-list is not using `fd-toolbar__spacer`, so custom classes were introduced. when tested using `fd-toolbar__spacer`, these extra classes are not needed. so removed it.

src/list-grid.scss
DeepakSap14: In Ngx toolbar component, we have styles present for `fd-toolbar`, which is planned to move to `fundamenatl-styles`.
This class in `toolbar.component.scss` is leaking and causing issue in grid-list component. so `z-index=1` was used to fix.
I have removed it. it will not be needed.

src/list-grid.scss
```diff
@@ -342,4 +350,10 @@ $block: #{$fd-namespace}-grid-list;
       }
     }
   }
+
+  &__btn-navigation {
+    @include fd-rtl() {
+      transform: rotate(180deg);
+    }
```
InnaAtanasova, line 357: this is not used anywhere. Why do you need it?

InnaAtanasova (changes requested):

src/list-grid.scss
InnaAtanasova: If it's a class needed in fund-ngx you need to add an example in styles so we have a visual test to avoid bugs in future. Also, it's a new class and needs to be added in the Breaking changes wiki.

DeepakSap14: added example and breaking change information.

stories/list-grid/list-grid.stories.js
```diff
@@ -201,7 +201,7 @@ export const noneMode = () => `<div style="min-height: 500px;">
 
                         <span class="fd-grid-list__item-counter" aria-label="Item has 10 children">10</span>
 
-                        <button class="fd-button fd-button--compact fd-button--transparent" aria-label="Navigation">
+                        <button class="fd-button fd-button--compact fd-grid-list__btn-navigation fd-button--transparent fd-grid-list__btn-navigation" aria-label="Navigation">
```
InnaAtanasova, line 204: `fd-grid-list__btn-navigation` is used twice

stories/list-grid/__snapshots__/list-grid.stories.storyshot
```diff
@@ -4829,7 +4829,7 @@ exports[`Storyshots Components/Grid List None mode 1`] = `
                         
               <button
                 aria-label="Navigation"
-                class="fd-button fd-button--compact fd-button--transparent"
+                class="fd-button fd-button--compact fd-grid-list__btn-navigation fd-button--transparent fd-grid-list__btn-navigation"
```
InnaAtanasova, line 4832: `fd-grid-list__btn-navigation` is used twice

src/list-grid.scss
```diff
@@ -263,13 +265,19 @@ $block: #{$fd-namespace}-grid-list;
...
   }
 
   &__filter-button {
     @include fd-hover() {
       color: var(--sapButton_TextColor) !important;
     }
+
+    @include fd-active() {
+      @include fd-hover() {
+        color: var(--sapHighlightTextColor) !important;
```
InnaAtanasova, line 278: have you tried without `!important`?

stories/list-grid/list-grid.stories.js
DeepakSap14: removed duplicate class name.

stories/list-grid/__snapshots__/list-grid.stories.storyshot
DeepakSap14: fixed

src/list-grid.scss
DeepakSap14: yes, i tried without !important, which does not work. because we have already `!important` applied on `__filter-button` color. and that is necessary because transparent `fd-button` has text colored.

InnaAtanasova: That's why specificity exists in CSS. You can avoid using !important just by nesting the css classes and start relying on specificity. :

```
  .#{$block}__filter {
    &[class*="active"] {
      @include fd-fiori-focus();

      color: var(--sapHighlightTextColor);
    }
  }

  .#{$block}__filter-button {
    color: var(--sapHighlightTextColor);

    @include fd-hover() {
      color: var(--sapButton_TextColor);
    }

    @include fd-active() {
      @include fd-hover() {
        color: var(--sapHighlightTextColor);
      }
    }
  }

```

InnaAtanasova (changes requested):

src/list-grid.scss
DeepakSap14: using specificity to remove `!important`. working fine.
thanks.

InnaAtanasova (approved):

PR #2540, opened 2021-07-05
fix: refactor bar documentation as suggested

### LINKED ISSUE #2528: Bar refactoring #2480 needs refactoring
**Describe the bug**
#2480 needs refactoring

bar

`0.19.0-rc.24`

The markup of the Title was created:

```
<div class="fd-bar">
    <div class="fd-bar__left">
        <div class="fd-bar__element">
            <span class="fd-bar__element fd-bar__element--title" aria-label="text">TEXT</span>
        </div>
    </div>
</div>
```
A parent and a child elements have the same class `fd-bar__element`. That is not current best practice in fundamental-styles. It needs to be changed

sKudum:
## Related Issue
Closes #2528 

## Description

Updated documention examples with title component as first element in bar.

Before:
```
<div class="fd-bar__element">  
 <span class="fd-bar__element--title" aria-label="text">TEXT</span>
</div>
```
After:
```
<div class="fd-bar__element fd-bar__element--title">
<h6 class="fd-title fd-title--h6" aria-label="text">TEXT</h6>
</div>
```

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/stories/bar/bar.stories.js
+++ b/stories/bar/bar.stories.js
@@ -44,7 +44,7 @@ export const Default = () => `
 <div class="fd-bar">
     <div class="fd-bar__left">
         <div class="fd-bar__element fd-bar__element--title">
-            <h4 class="fd-title" aria-label="text">TEXT</h4>
+            <h6 class="fd-title fd-title--h6" aria-label="text">TEXT</h6>
         </div>
     </div>
     <div class="fd-bar__middle">
--- a/stories/bar/bar.stories.js
+++ b/stories/bar/bar.stories.js
@@ -35,7 +35,7 @@ export default {
 - Buttons are sorted by usage i.e. from frequently-used to seldomly-used.
         `,
         tags: ['f3', 'a11y', 'theme'],
-        components: ['button', 'icon', 'input', 'segmented-button', 'avatar', 'bar']
+        components: ['button', 'icon', 'input', 'segmented-button', 'avatar', 'bar', 'title']
     }
 };
 
@@ -43,8 +43,8 @@ export const Default = () => `
 <p><b>Compact bar with compact elements</b></p>
 <div class="fd-bar">
     <div class="fd-bar__left">
-        <div class="fd-bar__element">
-            <span class="fd-bar__element--title" aria-label="text">TEXT</span>
+        <div class="fd-bar__element fd-bar__element--title">
+            <h4 class="fd-title" aria-label="text">TEXT</h4>
         </div>
     </div>
     <div class="fd-bar__middle">
--- a/stories/bar/bar.stories.js
+++ b/stories/bar/bar.stories.js
@@ -44,7 +44,7 @@ export const Default = () => `
 <div class="fd-bar">
     <div class="fd-bar__left">
         <div class="fd-bar__element">
-            <span class="fd-bar__element fd-bar__element--title" aria-label="text">TEXT</span>
+            <span class="fd-bar__element--title" aria-label="text">TEXT</span>
         </div>
     </div>
     <div class="fd-bar__middle">
--- a/stories/bar/__snapshots__/bar.stories.storyshot
+++ b/stories/bar/__snapshots__/bar.stories.storyshot
@@ -220,16 +220,16 @@ exports[`Storyshots Components/Bar Default 1`] = `
       
         
       <div
-        class="fd-bar__element"
+        class="fd-bar__element fd-bar__element--title"
       >
         
             
-        <span
+        <h4
           aria-label="text"
-          class="fd-bar__element--title"
+          class="fd-title"
         >
           TEXT
-        </span>
+        </h4>
         
         
       </div>
--- a/stories/bar/__snapshots__/bar.stories.storyshot
+++ b/stories/bar/__snapshots__/bar.stories.storyshot
@@ -226,7 +226,7 @@ exports[`Storyshots Components/Bar Default 1`] = `
             
         <span
           aria-label="text"
-          class="fd-bar__element fd-bar__element--title"
+          class="fd-bar__element--title"
         >
           TEXT
         </span>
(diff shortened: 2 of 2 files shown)


stories/bar/bar.stories.js
```diff
@@ -44,7 +44,7 @@ export const Default = () => `
 <div class="fd-bar">
     <div class="fd-bar__left">
         <div class="fd-bar__element">
-            <span class="fd-bar__element fd-bar__element--title" aria-label="text">TEXT</span>
+            <span class="fd-bar__element--title" aria-label="text">TEXT</span>
```
InnaAtanasova, line 47: You can't have modifier class `fd-bar__element--title` without base class `fd-bar__element` on the same level

InnaAtanasova: Instead of:
```
<div class="fd-bar__element">
            <span class="fd-bar__element--title" aria-label="text">TEXT</span>
</div>
```

You need:
```
<div class="fd-bar__element fd-bar__element--title">
          <h4 class="fd-title fd-title--h4">Title</h4>
</div>
```

InnaAtanasova (changes requested):

stories/bar/bar.stories.js
sKudum: Hi @InnaAtanasova ,

  Added `fd-bar--title` ( instead of` fd-title`,` fd-title--h4` and `fd-bar__element--title` 3 classes) as shown below.

```
<div class="fd-bar__element">
          <h4 class="fd-bar--title">Title</h4>
</div>
```

Please do let me know, if this approrach is ok..will update the whole PR details accordingly .


Thank You

InnaAtanasova (changes requested):
Unfortunately you can't do this:
```
<span class="fd-bar__title" aria-label="text">TEXT</span>
```

Title is a specific Fiori 3 component that has its own styling. You can check here: https://fundamental-styles.netlify.app/?path=/docs/components-title--levels
It can't be a span as you did it. You have 2 options: 
1. Create a modifier class for `fd-bar__element` element and have something like:
`fd-bar__element fd-bar__element--title`
2. Create a modifier class for the title element:
`fd-title fd-title--h1 fd-bar__title`

sKudum:
> Unfortunately you can't do this:
> 
> ```
> <span class="fd-bar__title" aria-label="text">TEXT</span>
> ```
> 
> Title is a specific Fiori 3 component that has its own styling. You can check here: https://fundamental-styles.netlify.app/?path=/docs/components-title--levels
> It can't be a span as you did it. You have 2 options:
> 
> 1. Create a modifier class for `fd-bar__element` element and have something like:
>    `fd-bar__element fd-bar__element--title`
> 2. Create a modifier class for the title element:
>    `fd-title fd-title--h1 fd-bar__title`

Sure @InnaAtanasova , updated with option1.

stories/bar/bar.stories.js
```diff
@@ -43,8 +43,8 @@ export const Default = () => `
 <p><b>Compact bar with compact elements</b></p>
 <div class="fd-bar">
     <div class="fd-bar__left">
-        <div class="fd-bar__element">
-            <span class="fd-bar__element fd-bar__element--title" aria-label="text">TEXT</span>
+        <div class="fd-bar__element fd-bar__element--title">
+            <span aria-label="text">TEXT</span>
```
InnaAtanasova, line 47: Hi @sKudum the padding for Bar component changes ONLY when Title control is the first control. It won't work with a span. I mean it will, but it's not what the specs recommend. Please check the code snippets I provided in the prev review. You need to use Title component in the documentation.
Screen Shot 2021-07-07 at 10 22 24 AM

InnaAtanasova (changes requested):

InnaAtanasova:
> > Unfortunately you can't do this:
> > ```
> > <span class="fd-bar__title" aria-label="text">TEXT</span>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Title is a specific Fiori 3 component that has its own styling. You can check here: https://fundamental-styles.netlify.app/?path=/docs/components-title--levels
> > It can't be a span as you did it. You have 2 options:
> > 
> > 1. Create a modifier class for `fd-bar__element` element and have something like:
> >    `fd-bar__element fd-bar__element--title`
> > 2. Create a modifier class for the title element:
> >    `fd-title fd-title--h1 fd-bar__title`
> 
> Sure @InnaAtanasova , updated with option1.
> 
> Hope this will be considered as breaking change too, in [wikki](https://github.com/SAP/fundamental-styles/wiki/Breaking-Changes) shall I remove #2480 details? do avoid confusions (because it also speaks about same issue with different approach).
> 
> Thank You.

Yes, please update the wiki to only mention that we introduced a new modifier class for the case where Title component is the first component.

stories/bar/bar.stories.js
sKudum: Hi @InnaAtanasova ,

 Is it mandatory to have fd-bar__element--title/fd-bar__title classes?  

We can check for title component as  first child as below and add the padding for it. Then won't be breaking change too.

```
  &__element {
      =========
     > .fd-title:first-child {
      @include fd-set-paddings-x(0.5rem,0);
    }
  }
```

With <h> tags
        ```
<div class="fd-bar__element">
            <h4 class="fd-title" aria-label="text">TEXT</h4>
        </div>
```

With <span> tags

        ```
<div class="fd-bar__element">
            <span class="fd-title" aria-label="text">TEXT</span>
        </div>
```

Observation : Creating to `fd-bar__title` padding value display is not shown as expected.

InnaAtanasova: You can't do this:
```
  &__element {
      =========
     > .fd-title:first-child {
      @include fd-set-paddings-x(0.5rem,0);
    }
  }
```
as it's violating the self contained rules. You are modifying one component inside another component.

sKudum: Sure @InnaAtanasova , 

   In this case am updating documentation example and breaking changes details as well in wikki for PR #2480 (where fd-bar__element--title class is introduced.).

Betrozov (approved):

InnaAtanasova (approved):

Issue #2509, opened 2021-06-30
Multi Combobox and Multi Input issue


droshev:
- [ ] Multi Combobox and Multi Input - on hover, the popover control icon "jumps" a little bit
<img width="585" alt="Screen Shot 2021-05-20 at 3 59 15 PM" src="https://user-images.githubusercontent.com/39598672/119041346-84d05700-b984-11eb-93dc-dbc0b528784a.png">

PR #2547, opened 2021-07-06
fix: update border width for combobox button

N1XUS:
## Related Issue
Closes SAP/fundamental-styles#2509

## Description
Updated border width so it won't jump on hover or focus.

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/theming/sap_fiori_3.scss
+++ b/src/theming/sap_fiori_3.scss
@@ -108,7 +108,7 @@
 
   /* Input Group */
   --fdInputGroup_Text_Shadow: none;
-  --fdInputGroup_ControlButton_SideBorder: solid 0.0625rem transparent;
+  --fdInputGroup_ControlButton_SideBorder: solid var(--sapButton_BorderWidth) transparent;
 
   /* Input */
   --fdInput_Text_Shadow: none;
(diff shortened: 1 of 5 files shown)


src/theming/sap_fiori_3.scss
```diff
@@ -108,7 +108,7 @@
 
   /* Input Group */
   --fdInputGroup_Text_Shadow: none;
-  --fdInputGroup_ControlButton_SideBorder: none;
+  --fdInputGroup_ControlButton_SideBorder: solid 0.0625rem transparent;
```
InnaAtanasova, line 111: For the width you can use a variable. It can be either `--sapElement_BorderWidth` or `--sapField_BorderWidth` for Input fields:
Screen Shot 2021-07-06 at 11 26 17 AM
Screen Shot 2021-07-06 at 11 26 03 AM

InnaAtanasova:
@N1XUS could you trigger Chromatic? The rest looks good

N1XUS:
> @N1XUS could you trigger Chromatic? The rest looks good

Done

InnaAtanasova (approved):

PR #2549, opened 2021-07-06
fix: table fixed column borders fix, table improvements

platon-rov:
## Related Issue

Closes [Fundamental NGX #5097](https://github.com/SAP/fundamental-ngx/issues/5097).

## Description

Table:
* Fixed columns border styles fix.
* Condensed mode selection checkbox styles fix.
* Documentation title fix.

### Before:
1.
Screenshot 2021-07-06 at 21 00 13
2.
Screenshot 2021-07-06 at 21 03 46
3.
Screenshot 2021-07-06 at 21 06 11

### After:
1.
Screenshot 2021-07-06 at 21 03 21
2.
Screenshot 2021-07-06 at 21 05 11
3.
Screenshot 2021-07-06 at 21 06 49

- [X] All values are in `rem`
- [X] RTL support
3. Testing
- [X] Updated tests
4. Documentation

### DIFF
--- a/src/table.scss
+++ b/src/table.scss
@@ -63,9 +63,9 @@ $block: #{$fd-namespace}-table;
   $fd-table-first-cell-padding: 1rem;
   $fd-table-cozy-cell-height: 2.75rem;
   $fd-table-compact-cell-height: 2rem;
-  $fd-table-navigation-icon-width: 2rem;
   $fd-table-condensed-cell-height: 1.5rem;
-  $fd-table-fixed-cell-shift: 0.05rem;
+  $fd-table-navigation-icon-width: 2rem;
+  $fd-table-fixed-cell-shift: 0.0625rem; // 1px
 
   // Thanks to table-cell specification, if the width is set to small value, the cell will have content width
   $fd-table-cell-fit-content-width: 0.5rem;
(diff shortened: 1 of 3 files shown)


N1XUS (approved):

src/table.scss
```diff
@@ -598,6 +585,19 @@ $block: #{$fd-namespace}-table;
...
+    .#{$block}__body {
+      .#{$block}__cell {
+        &--checkbox {
+          position: relative;
+        }
+
+        .fd-checkbox + .fd-checkbox__label:empty {
+          top: -0.25rem;
+          position: absolute;
+        }
```
InnaAtanasova, line 598: You can't modify properties of one component (checkbox in this case) inside another (table). This is violating the self-contained rules.

InnaAtanasova (changes requested):

src/table.scss
platon-rov: I agree, but in this case, we're adjusting it only for the checkboxes inside the table.

We also don't have any condensed mode for the checkbox, that's why I done it like that.

InnaAtanasova: it doesn't matter that it's only for the checkboxes inside table. We have lots of cases like this. You can't modify props of checkbox inside table scss. You can always add additional styling inside checkbox or modifier class inside table. But this way we can't accept it.

InnaAtanasova (approved):

InnaAtanasova:
@platon-rov please update the wiki page. You are adding new classes that the implementation libraries need to add. Thanks

Issue #2209, opened 2021-03-23
Implement Tool Header Component


InnaAtanasova:
The Tool Header is a control that is found in most SAP Cloud Platform products, most commonly as a part of the Tool Page Layout. The Tool Header inherits its functionality from the **OverflowToolbar** and its visual design from the **ShellBar**.

- [x] decide if we will re-use the ShellBar classes or create new ones for ToolHeader (naming, code) EDIT: decided to create a new component with new class names
- [x] over-styled controls
- [x] buttons and their states - The ToolHeader supports only text and icon buttons in their Default, Transparent, Back, and Up types. The Menu button is supported in its Emphasized, Default, and Transparent styles.
- [x] Texts, Titles, and Labels - The ToolHeader supports only single-line text and truncation. This applies to Texts, Titles, and Labels. Consider using small and normal size texts or title headings of H4, H5, H6. Avoid using text wrapping.
- [x] Icons - icons + background
- [na] Tabs - uses Horizontal Navigation - will be implemented as part of Horizontal Navigation

PR #2550, opened 2021-07-07
feat: introduce Tool Header component from BTP design

InnaAtanasova:
## Related Issue
Closes SAP/fundamental-styles#2209

## Description
This PR introduces Tool Header component  from BTP design

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/tool-header.scss
+++ b/src/tool-header.scss
@@ -3,7 +3,6 @@
 
 $block: #{$fd-namespace}-tool-header;
 $fd-tool-header-height: 2.75rem !default;
-$fd-tool-header-screen-size-sm: 599px !default;
 
 $fd-tool-header-responsive-paddings: (
   'sm':  ('padding': 0 1rem),
@@ -108,7 +107,6 @@ $fd-tool-header-object-status-semantic-values: (
 
     width: auto;
     height: 1.625rem;
-    object-fit: cover;
     background-size: contain;
     background-repeat: no-repeat;
   }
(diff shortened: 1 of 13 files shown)


src/tool-header.scss
```diff
@@ -0,0 +1,295 @@
...
+    }
+  }
+
+  &__logo {
+    @include fd-reset();
+
+    width: auto;
+    height: 1.625rem;
+    object-fit: cover;
+    background-size: contain;
```
N1XUS, line 112: Maybe add a modifier for 'cover' option?

src/tool-header.scss
```diff
@@ -0,0 +1,295 @@
...
+
+        &[class*="inverted"] {
+          text-shadow: none;
+          color: map_get($set-params, 'iconColorInverted');
+
+          [class*="icon"] {
+            color: map_get($set-params, 'iconColorInverted');
+          }
+
+          background: map_get($set-params, 'bgColorInverted');
```
N1XUS, line 221: Maybe move this block into mixin?

src/tool-header.scss
```diff
@@ -0,0 +1,295 @@
...
+      padding: map_get($set-padding, 'padding');
+    }
+  }
+
+  &--responsive-paddings {
+    @media (max-width: $fd-tool-header-screen-size-sm) {
+      @include fd-set-paddings-x-equal(1rem);
+    }
+
+    @media (min-width: 600px) and (max-width: 1023px) {
```
N1XUS, line 283: Can we have these responsive breakpoints moved into some global variables for all styles, so we keep breakpoints as consistent as possible?

N1XUS (commented):
LGTM

Smiranin (approved):

src/tool-header.scss
InnaAtanasova: it's actually not needed, removed it.

src/tool-header.scss
InnaAtanasova: Mixins can not be defined within control directives or other mixins. In this case we are inside an `@each` loop and rely on variables like `$set-params` which can't be accessed outside the loop.

InnaAtanasova:
@N1XUS thanks for the review. Your comments have been addressed. Please check again.

Issue #2503, opened 2021-06-29
Slider in form row gets cutted


Smiranin:
**Describe the bug**

See related [issue](https://github.com/SAP/fundamental-ngx/issues/5529) in ngx repo.

PR #2552, opened 2021-07-08
fix: added additional panding for slider [ci visual]

Smiranin:
## Related Issue
Closes SAP/fundamental-styles#2503

## Description
Added additional padding to prevent hiding  of outline of handle inside block with overflow: hidden.

image

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/slider.scss
+++ b/src/slider.scss
@@ -6,6 +6,10 @@ $handle-desktop-dimensions: 1.25rem;
 $handle-hit-area-desktop-dimensions: 2rem;
 $handle-mobile-dimensions: 1.625rem;
 $handle-hit-area-mobile-dimensions: 2.75rem;
+$handle-outline-offset: 0.0625rem;
+$handle-outline-width: 0.0625rem;
+$slider-x-paddings: $handle-desktop-dimensions / 2;
+$slider-x-paddings--lg: $handle-mobile-dimensions / 2;
 
 .#{$block} {
   @include fd-reset();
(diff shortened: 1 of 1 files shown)


dimamarksman (approved):

src/slider.scss
```diff
@@ -6,6 +6,10 @@ $handle-desktop-dimensions: 1.25rem;
 $handle-hit-area-desktop-dimensions: 2rem;
 $handle-mobile-dimensions: 1.625rem;
 $handle-hit-area-mobile-dimensions: 2.75rem;
+$handle-outline-offset: 0.0625rem;
+$handle-outline-width: 1px;
```
platon-rov, line 10: You can use `0.0625rem` here also. The `rem` using is required.

src/slider.scss
```diff
@@ -16,11 +20,12 @@ $handle-hit-area-mobile-dimensions: 2.75rem;
...
   min-width: 4rem;
   max-width: 100%;
-  padding: 1rem 0.625rem;
+  padding: 1rem calc(#{$slider-x-paddings} + #{$handle-outline-offset} + #{$handle-outline-width});
   position: relative;
 
   &--lg {
-    @include fd-set-paddings-x-equal(0.8125rem);
+    padding-left: calc(#{$slider-x-paddings--lg} + #{$handle-outline-offset} + #{$handle-outline-width});
+    padding-right: calc(#{$slider-x-paddings--lg} + #{$handle-outline-offset} + #{$handle-outline-width});
```
platon-rov, line 28: Could we use `padding: 1rem calc(#{$slider-x-paddings--lg} + #{$handle-outline-offset} + #{$handle-outline-width});`?

Smiranin: Why? 1 rem for Y pending is already installed. We need to change just X pending.
But I have better idea to use fd-set-paddings-x-equal mixin for it.

src/slider.scss
Smiranin: Fixed

platon-rov (approved):

InnaAtanasova (approved):

PR #2553, opened 2021-07-08
fix: popover style fix for Nested list component. Move nested-list styles from Ngx

DeepakSap14:
## Related Issue

## Description
There are some styles changes done for popover inside Nested list component and it is currently in fundamental-ngx.
This change is specific to nested list component. so i have moved those styles to side-nav--condensed class.

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/side-nav.scss
+++ b/src/side-nav.scss
@@ -111,6 +111,10 @@ $has-child-content-arrow-margin: 0.125rem;
           }
         }
       }
+
+      &__popover-header {
+        box-shadow: var(--sapContent_HeaderShadow);
+      }
     }
   }
 
--- a/src/side-nav.scss
+++ b/src/side-nav.scss
@@ -72,7 +72,7 @@ $has-child-content-arrow-margin: 0.125rem;
   &--condensed {
     width: 2.75rem;
 
-    .#{$fd-namespace}-popover-custom {
+    &--popover-custom {
       display: block;
     }
 
--- a/src/side-nav.scss
+++ b/src/side-nav.scss
@@ -111,10 +111,6 @@ $has-child-content-arrow-margin: 0.125rem;
           }
         }
       }
-
-      &__popover-header {
-        box-shadow: var(--sapContent_HeaderShadow);
-      }
     }
   }
 
--- a/src/side-nav.scss
+++ b/src/side-nav.scss
@@ -72,6 +72,10 @@ $has-child-content-arrow-margin: 0.125rem;
   &--condensed {
     width: 2.75rem;
 
+    .#{$fd-namespace}-popover-custom {
+      display: block;
+    }
+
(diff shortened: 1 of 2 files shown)


DeepakSap14:
this PR recreated from original PR https://github.com/SAP/fundamental-styles/pull/2524

src/side-nav.scss
```diff
@@ -107,6 +111,10 @@ $has-child-content-arrow-margin: 0.125rem;
           }
         }
       }
+
+      &__popover-header {
+        box-shadow: var(--sapContent_HeaderShadow);
+      }
```
InnaAtanasova, line 117: can you update with an example the Condensed state so we catch problems like this in the visual regression tests? 
Where did you take this value from?

DeepakSap14: @InnaAtanasova 
on closely checking in Ngx, i don't find this class(`fd-nested-list__popover-header`) applied and there is no effect of adding/removing this class on any html element.
Screenshot 2021-07-12 at 3 11 05 PM

In Ngx, value has been used for popover-header-shadow. In styles i used corresponding variable from 
`https://github.com/SAP/theming-base-content/blob/master/content/Base/baseLib/sap_fiori_3/css_variables.css`.

This class (`fd-nested-list__popover-header`) can be removed safely from Ngx and does not need to be moved to Styles.

src/side-nav.scss
```diff
@@ -72,6 +72,10 @@ $has-child-content-arrow-margin: 0.125rem;
   &--condensed {
     width: 2.75rem;
 
+    &--popover-custom {
```
DeepakSap14, line 75: can't add example for this as we are not showing interaction on side-nav click. and this is modifier for `fd-popover-custom` class.

InnaAtanasova: You can by hardcoding the popover body to be always open. There are multiple examples in the documentation. You can check Action Bar `Contextual` example.

InnaAtanasova: And you can always add some JavaScript for the documentation purposed. You can check Vertical Navigation for exactly the same case: https://fundamental-styles.netlify.app/?path=/docs/components-vertical-navigation--cozy

DeepakSap14: i do not see any change on using `display: block` instead of using `display: inline-block` here. I think change in Ngx can be safely removed without migrating to Styles.

with `display: block`
Screenshot 2021-07-13 at 5 34 41 PM

with `display: inline-block`
Screenshot 2021-07-13 at 5 34 59 PM

Hence this PR can be closed.

DeepakSap14:
@InnaAtanasova I am closing this PR because i do not see any change effect by using styles change present in Ngx.
Styles in Ngx can be safely removed without migrating it to styles repo.

Issue #2210, opened 2021-03-23
Implement Horizontal Navigation


InnaAtanasova:
The Horizontal Navigation is an optional layout to the Vertical Navigation (aka: Side Navigation control).
The Horizontal Navigation is suitable for a maximum of seven to nine primary items. 

Dependent on:
- Universal IconTabBar
- Tool Header

Horizontal Navigation is a component extending the functionality of the UniversalIconTabBar
- [x] implement Drop Down Menu deltas
- [x] Tab filters with or without icon (text only)

PR #2556, opened 2021-07-09
feat: introduce navigation for IconTabBar

InnaAtanasova:
## Related Issue
Part of SAP/fundamental-styles#2210

## Description
Shell Navigation Tabs are a type of Icon Tab Bar with some color modifications so it can be used with Tool Header and Shell Bar. Once this PR and https://github.com/SAP/fundamental-styles/pull/2550 are merged I will create documentation for the entire Horizontal Navigation topic. There won't be more css needed, just putting 2 components together (Tool Header+Shell Navigation Tabs) in a documentation page.

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/icon-tab-bar.scss
+++ b/src/icon-tab-bar.scss
@@ -974,6 +974,131 @@ $fd-icon-tab-bar-semantic-values: (
     }
   }
 
+  &--navigation {
+    .#{$block}__header {
+      box-shadow: var(--sapShell_Shadow);
+      background: var(--sapShell_Navigation_Background);
+    }
+
+    .#{$block}__item {
+      color: var(--sapShell_Navigation_TextColor);
+    }
+
+    .#{$block}__tab {
+      .#{$block}__tag {
+        color: var(--sapShell_Navigation_TextColor);
+      }
+
+      .#{$block}__arrow {
+        [class*='sap-icon'] {
+          color: var(--sapShell_Navigation_TextColor);
+        }
+      }
+
+      @include fd-selected() {
+        .#{$block}__tag {
+          color: var(--sapShell_Navigation_Selected_TextColor);
+        }
+
+        .#{$block}__arrow {
+          [class*='sap-icon'] {
+            color: var(--sapShell_Navigation_Selected_TextColor);
+          }
+        }
+
+        &::after {
+          background: var(--sapShell_Navigation_SelectedColor);
+        }
+      }
+
+      @include fd-hover() {
+        .#{$block}__tag {
+          color: var(--sapShell_Navigation_Selected_TextColor);
+        }
+        .#{$block}__arrow {
+          [class*='sap-icon'] {
+            color: var(--sapShell_Navigation_Selected_TextColor);
+          }
+        }
+      }
+
+      @include fd-focus() {
+        .#{$block}__tag,
+        .#{$block}__icon,
+        .#{$block}__container,
+        .#{$block}__tab-container {
+          outline-color: var(--sapShell_Navigation_TextColor);
+        }
+      }
+    }
+
+    .#{$block}__button {
+      color: var(--sapShell_Navigation_SelectedColor);
+
+      @include fd-hover() {
+        color: var(--sapShell_Navigation_SelectedColor);
+        background: var(--sapShell_Navigation_Hover_Background);
+        border-color: var(--sapShell_Navigation_Hover_Background);
+      }
+
+      @include fd-active() {
+        color: var(--sapShell_Navigation_Active_TextColor);
+        background: var(--sapShell_Navigation_Active_Background);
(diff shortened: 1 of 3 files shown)


mikerodonnell89:
This feels weird to me, 2 of the overflow dots and no margin with this button? Screen Shot 2021-07-12 at 11 58 31 AM

This screenshot from the designs tells me that the button shouldn't be styled that way
Screen Shot 2021-07-12 at 12 06 18 PM

Also confused about what's going on with these >>> icons in the menu
Screen Shot 2021-07-12 at 12 36 59 PM

The designs have these up in the tab bar itself, but feel strange to me pushed to the overflow menu:
Screen Shot 2021-07-12 at 12 37 41 PM

Smiranin:
I noticed that there is used deprecated layout for "Counters and Text".

As I see in design [docs](https://experience.sap.com/fiori-design-web/icontabbar/#text-only)
This layout isn't recommended.

src/icon-tab-bar.scss
```diff
@@ -974,6 +974,131 @@ $fd-icon-tab-bar-semantic-values: (
...
+      background: var(--sapShell_Navigation_Background);
+    }
+
+    .#{$block}__item {
+      color: var(--sapShell_Navigation_TextColor);
+    }
+
+    .#{$block}__tab {
+      .#{$block}__tag {
+        color: var(--sapShell_Navigation_TextColor);
```
Smiranin, line 989: Do we really need nested selectors?
This provokes a prioritization war. It seems to me that this is not according to BEM. Now I can't just set a class and override the color. I need to give the appropriate priority to this selector otherwise it won't apply.
Sometimes nested selectors necessary, but I'm not sure if this is necessary in this case (maybe I'm wrong)

Smiranin (approved):

InnaAtanasova:
> 
> This screenshot from the designs tells me that the button shouldn't be styled that way
> Screen Shot 2021-07-12 at 12 06 18 PM
> 
> Also confused about what's going on with these >>> icons in the menu
> Screen Shot 2021-07-12 at 12 36 59 PM
> 
> The designs have these up in the tab bar itself, but feel strange to me pushed to the overflow menu:
> Screen Shot 2021-07-12 at 12 37 41 PM

Hi @mikerodonnell89 the `>>>` are following the specs guidelines for `Icons in Drop Down Menu`: https://wiki.wdf.sap.corp/wiki/pages/viewpage.action?pageId=2208148593
Screen Shot 2021-07-12 at 1 22 06 PM

InnaAtanasova:
> This feels weird to me, 2 of the overflow dots and no margin with this button? Screen Shot 2021-07-12 at 11 58 31 AM
> 
> This screenshot from the designs tells me that the button shouldn't be styled that way
> Screen Shot 2021-07-12 at 12 06 18 PM
> 

Universal IconTab Bar has 2 types: 
- Two Click Areas: 
Screen Shot 2021-07-12 at 1 26 39 PM
- Single Click Area:
Screen Shot 2021-07-12 at 1 27 13 PM

In the case of Single Click Area the "arrow" is part of the tab and the tab serves the purpose of opening the dropdown menu. In the other case, the Tab text is a navigation link and the arrow is a separate button which opens the dropdown menu.

The badge position is implemented following the guidelines from here:
Screen Shot 2021-07-12 at 1 29 41 PM

InnaAtanasova:
> I noticed that there is used deprecated layout for "Counters and Text".
> 
> As I see in design [docs](https://experience.sap.com/fiori-design-web/icontabbar/#text-only)
> This layout isn't recommended.
> 

@Smiranin yes, I also noticed it but the specs were last updated in May 26, 2021 so they are the most current documentation we should follow. I can raise this to the next design meeting we have with the designers, but for now we keep this option, too.

src/icon-tab-bar.scss
InnaAtanasova: yes, it's needed. It's due to the fact that `.fd-icon-tab-bar__tag` depends on the state of it's parent `.fd-icon-tab-bar__tab`. When `.fd-icon-tab-bar__tab` is selected, the color of the tag changes. Same is valid for the icon. 
```
 &__tab {
    @include fd-reset();

    @include fd-flex(column) {
      align-items: flex-start;
    }

    @include fd-selected() {
      @include fd-icon-tab-bar-selection-bar();
      ....
      .#{$block}__tag {
        color: var(--fdIcon_Tab_Bar_Selection_Color);
      }

      .#{$block}__icon {
        @include fd-icon-selector() {
          color: var(--sapContent_ContrastIconColor);
        }

        background-color: var(--sapSelectedColor);
      }
    }

```

For this reason, when we have the case with navigation we need to rely on the exact same specificity to change the color of the `.fd-icon-tab-bar__tag`

Issue #2444, opened 2021-06-12
Form Grid design corrections needed


droshev:
**Describe the bug**

> Is this issue related to a specific component?
Form Grid

**To Reproduce**
1. Go to https://fundamental-styles.netlify.app/?path=/docs/components-forms-form-grid--s-size
2. Go to https://fundamental-styles.netlify.app/?path=/docs/components-forms-form-group--primary

**Expected behavior**
- [ ] the spacing between fields and forms needs to be updates
- [ ] add Display Mode examples (horizontal and vertical). Maybe there are missing stylings for the display mode (e.g. padding, background)
- [ ] add Display Mode and Horizontal Layout examples for `Form Item`, `Form Label`, `Form Group`, and `Form Header`
- [ ] add examples with form items in `Form Header` page

**Screenshots**
![Screen Shot 2021-06-12 at 6 49 58 AM](https://user-images.githubusercontent.com/4380815/121773702-93f19180-cb4b-11eb-8a71-a2191513deba.png)
![Screen Shot 2021-06-12 at 6 58 52 AM](https://user-images.githubusercontent.com/4380815/121773718-ac61ac00-cb4b-11eb-9cbc-a5603de4d668.png)
![Screen Shot 2021-06-12 at 6 52 27 AM](https://user-images.githubusercontent.com/4380815/121773725-b8e60480-cb4b-11eb-8cb4-c2ab52255c48.png)

aberikashvili:
@droshev I noticed, when there's 2 columns and you delete once component from any column, other components are centered in the middle:

Before deletion:
![image](https://user-images.githubusercontent.com/3863691/125079977-4bd77780-e0d5-11eb-9dc9-ebbeecaa4c9b.png)
After deletion:
![image](https://user-images.githubusercontent.com/3863691/125080030-5d208400-e0d5-11eb-8d89-4f7b3458ef75.png)

Is this also a bug which need to be fixed?

droshev:
> @droshev I noticed, when there's 2 columns and you delete once component from any column, other components are centered in the middle:
> 
> Before deletion:
> ![image](https://user-images.githubusercontent.com/3863691/125079977-4bd77780-e0d5-11eb-9dc9-ebbeecaa4c9b.png)
> After deletion:
> ![image](https://user-images.githubusercontent.com/3863691/125080030-5d208400-e0d5-11eb-8d89-4f7b3458ef75.png)
> 
> Is this also a bug which need to be fixed?

[...]

PR #2568, opened 2021-07-12
[WIP] fix: form grid design correctionsa [ci visual]

aberikashvili:
## Related Issue
Closes SAP/fundamental-styles#2444

## Description

- [x] All values are in `rem`
- [N/A] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/form-layout-grid.scss
+++ b/src/form-layout-grid.scss
@@ -18,17 +18,7 @@ $cols: 11;
     margin: 0 -0.25rem;
     padding: 1rem;
 
-    &--translucent {
-      background-color: transparent;
-    }
-
-    &--transparent {
-      background-color: transparent;
-    }
-
-    &--solid {
-      background-color: var(--sapGroup_ContentBackground);
-    }
+    @include fd-form-backgrounds();
 
     & > .fd-row {
       & > :first-child & :not(.fd-col--wrap) {
--- a/src/form-layout-grid.scss
+++ b/src/form-layout-grid.scss
@@ -69,8 +69,9 @@ $cols: 11;
 
         .fd-form-label {
           display: inline-block;
-          margin-right: 0.5rem;
           padding-bottom: 0.125rem;
+          padding-top: 0.875rem;
+          padding-right: 1rem;
 
           @include fd-rtl() {
             padding-left: 0.5rem;
@@ -131,9 +132,7 @@ $cols: 11;
 
       .#{$blockCol} {
         .fd-form-label {
-          padding-bottom: 0.125rem;
-          padding-top: 0.875rem;
-          padding-right: 1rem;
+          padding-bottom: 0;
         }
 
         @for $n from 0 through $cols {
--- a/src/form-layout-grid.scss
+++ b/src/form-layout-grid.scss
@@ -70,7 +70,7 @@ $cols: 11;
         .fd-form-label {
           display: inline-block;
           padding-bottom: 0.125rem;
-          padding-top: 0.875rem;
+          padding-top: 0.275rem;
           padding-right: 1rem;
 
           @include fd-rtl() {
--- a/src/form-layout-grid.scss
+++ b/src/form-layout-grid.scss
@@ -74,7 +74,7 @@ $cols: 11;
           padding-right: 1rem;
 
           @include fd-rtl() {
-            padding-left: 0.5rem;
+            padding-left: 1rem;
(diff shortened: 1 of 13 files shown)


src/form-layout-grid.scss
```diff
@@ -211,7 +226,7 @@ $cols: 11;
               text-align: left;
             }
 
-            margin-bottom: 0;
+            // margin-bottom: 0;
```
InnaAtanasova, line 229: Please remove if not used

src/form-layout-grid.scss
```diff
@@ -169,7 +183,7 @@ $cols: 11;
               @include placement('left');
             }
 
-            margin-bottom: 0;
+            // margin-bottom: 0;
```
InnaAtanasova, line 186: Please remove if not used

src/form-layout-grid.scss
```diff
@@ -198,6 +212,7 @@ $cols: 11;
     .#{$blockRow} {
       .#{$blockCol} {
         .fd-form-label {
+          padding-top: 0.5rem;
```
InnaAtanasova, line 215: Is it the same value for compact and cozy?

src/form-layout-grid.scss
```diff
@@ -119,7 +131,9 @@ $cols: 11;
 
       .#{$blockCol} {
         .fd-form-label {
-          padding-bottom: 0;
+          padding-bottom: 0.125rem;
+          padding-top: 0.875rem;
+          padding-right: 1rem;
```
InnaAtanasova, line 136: Shouldn't these values be different for cozy and compact?

InnaAtanasova (changes requested):

src/form-layout-grid.scss
aberikashvili: removed

src/form-layout-grid.scss
aberikashvili: removed

src/form-layout-grid.scss
aberikashvili: @InnaAtanasova this value is for compact `min-width: 1024px`
image

InnaAtanasova:
Screen Shot 2021-07-20 at 1 46 30 PM

@aberikashvili make sure you add all the examples mentioned in the `Expected behaviour`. This will allow us to catch future bugs when we change the form elements.

src/form-layout-grid.scss
```diff
@@ -69,8 +69,9 @@ $cols: 11;
 
         .fd-form-label {
           display: inline-block;
-          margin-right: 0.5rem;
           padding-bottom: 0.125rem;
+          padding-top: 0.875rem;
+          padding-right: 1rem;
```
InnaAtanasova, line 74: shouldn't this have RTL, too? I see in RTL it's 0.5rem, not 1rem. Is that correct?

droshev (changes requested):
Can you also add examples as per the issue?

src/form-layout-grid.scss
aberikashvili: updated

aberikashvili:
image
image

aberikashvili:
@droshev I changed `0.875rem` to `0.275rem` because labels were looking too ugly

droshev (changes requested):
The vertical spacing doesn't look equal
Screen Shot 2021-07-21 at 6 22 34 PM

droshev:
@aberikashvili here are few things to be done:
- [ ] the spacing between form header/group/form items is off - can you check and fix according to the specs (you can see it [here](https://deploy-preview-2568--fundamental-styles.netlify.app/?path=/docs/components-forms-form-display-mode--primary) and [here](https://deploy-preview-2568--fundamental-styles.netlify.app/?path=/docs/components-forms-form-header--primary) )
- [ ] vertical spacing is not equal
Screen Shot 2021-07-21 at 10 35 21 PM
- [ ] vertical spacing in Display mode is not per design
- [ ] add form in display mode where the labels and texts are next to each other. (add examples also when the text goes to multiple lines
- [ ] add form in display mode with mulitple form groups using the form grid
- [ ] [form group page](https://deploy-preview-2568--fundamental-styles.netlify.app/?path=/docs/components-forms-form-group--primary) add display mode example


Let me know if you have more questions

PR #2572, opened 2021-07-13
fix: adding placeholder color and adding rtl to the button

Lokanathan-k:
## Related Issue
part of: https://github.com/SAP/fundamental-ngx/issues/5025

## Description
adding placeholder color for text area and adding rtl to the button
 

### Before:
Screenshot 2021-07-13 at 6 10 11 PM

### After:
Screenshot 2021-07-13 at 6 10 34 PM

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/feed-input.scss
+++ b/src/feed-input.scss
@@ -57,12 +57,6 @@ $block: #{$fd-namespace}-feed-input;
     padding: 0.3125rem 0.5rem;
   }
 
-  .fd-button {
-    @include fd-rtl() {
-      transform: rotate(180deg);
-    }
-  }
-
   // It's needed to have stronger selector to override fd-button styles
   .#{$block}__submit-button {
     margin-top: auto;
@@ -70,17 +64,10 @@ $block: #{$fd-namespace}-feed-input;
     margin-left: $fd-feed-input-btn-margin-start;
     margin-right: $fd-feed-input-btn-margin-end;
 
-    &[class*=sap-icon]::before {
-      display: inline-block;
-    }
-
     @include fd-rtl() {
       margin-left: $fd-feed-input-btn-margin-end;
       margin-right: $fd-feed-input-btn-margin-start;
-
-      &[class*=sap-icon]::before {
-        transform: rotate(180deg);
-      }
+      transform: rotate(180deg);
     }
   }
 
--- a/src/feed-input.scss
+++ b/src/feed-input.scss
@@ -27,6 +27,8 @@ $block: #{$fd-namespace}-feed-input;
 
   @include fd-disabled() {
     cursor: not-allowed;
+    pointer-events: initial;
+    opacity: initial;
   }
 
   .#{$block}__thumb {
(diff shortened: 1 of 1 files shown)


src/feed-input.scss
```diff
@@ -49,6 +51,35 @@ $block: #{$fd-namespace}-feed-input;
...
+
+      &:-ms-input-placeholder {
+        color: #32363a;
+        color: var(--sapTextColor, #32363a);
+      }
+
+      &:-moz-placeholder {
+        color: #32363a;
+        color: var(--sapTextColor, #32363a);
+      }
```
InnaAtanasova, line 75: - you don't need the fallback

src/feed-input.scss
```diff
@@ -49,6 +51,35 @@ $block: #{$fd-namespace}-feed-input;
...
+    }
+  }
+
+  .fd-button {
+    [dir="rtl"] &,
+    &[dir="rtl"] {
```
InnaAtanasova, line 81: There's a mixin for this

src/feed-input.scss
```diff
@@ -27,6 +27,8 @@ $block: #{$fd-namespace}-feed-input;
 
   @include fd-disabled() {
     cursor: not-allowed;
+    pointer-events: initial;
+    opacity: initial;
```
InnaAtanasova, line 31: why?

InnaAtanasova (changes requested):

src/feed-input.scss
InnaAtanasova: Do you really need `&::-webkit-input-placeholder`, `&::-moz-placeholder`, etc.? Have you tested without them?

src/feed-input.scss
Lokanathan-k: Without that the screen looks like this in NGX which is not seen in the style
Screenshot 2021-07-14 at 5 16 39 PM

src/feed-input.scss
Lokanathan-k: fixed

src/feed-input.scss
Lokanathan-k: Fixed, removed since its working in the latest version of firefox and other browser

src/feed-input.scss
```diff
@@ -27,6 +27,15 @@ $block: #{$fd-namespace}-feed-input;
...
+
+    textarea {
```
InnaAtanasova, line 33: `textarea::placeholder { ... }`

src/feed-input.scss
```diff
@@ -46,9 +55,11 @@ $block: #{$fd-namespace}-feed-input;
     max-height: $fd-feed-input-max-height;
     margin: 0;
     padding: 0.3125rem 0.5rem;
+  }
 
-    &::placeholder {
-      font-style: var(--fdFeed_Input_Placeholder_Font_Style);
+  .fd-button {
```
InnaAtanasova, line 60: Can't do this

InnaAtanasova (changes requested):

InnaAtanasova (approved):

src/feed-input.scss
kavya-b: will `initial` work in IE, have you checked? https://caniuse.com/?search=initial says no support for IE

PR #2573, opened 2021-07-14
fix: add active state background color for object list as per spec

### LINKED ISSUE #2520: Active state of Object List
Active state of Object List has the wrong. background color: - Jedrzej
Screen Shot 2021-03-25 at 5 21 41 PM

sKudum:
## Related Issue
Closes #2520 

## Description
Active state of Object List should has value as per fiori specifications.

### Before:

Screen Shot 2021-03-25 at 5 21 41 PM

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/stories/object-list/object-list.stories.js
+++ b/stories/object-list/object-list.stories.js
@@ -599,34 +599,34 @@ export const selectionWithNavigation = () => `
 <h4 id="objectListItemRowSelectionAndNavigation">Object List Item With Row Selection And Navigation</h4>
 <div role="navigation" style="max-width: 450px">
 <ul class="fd-list fd-object-list fd-list--navigation-object fd-list--navigation fd-list--selection" role="listbox" aria-labelledby="objectListItemRowSelectionAndNavigation">
-<li role="option" id="check" style="outline:#fff;" aria-selected="true" tabindex="0" class="fd-list__item fd-object-list__item fd-list__item--link is-selected">
-<a tabindex="0" class="fd-list__link" style="background:#0854a0;outline-offset:-0.1875rem;outline:0.0625rem dotted #fff;"> 
-<div class="fd-object-list__container">
-      <div class="fd-object-list__intro" style="color:#fff;text-shadow: none;">
+<li role="option" aria-selected="true" tabindex="0" class="fd-list__item fd-object-list__item fd-list__item--link is-selected">
+  <a tabindex="0" id="anchor" class="fd-list__link is-active"> 
+  <div class="fd-object-list__container">
+      <div class="fd-object-list__intro">
        Optional inline text</span>
       </div>
       <div class="fd-object-list__header">
         <span class="fd-avatar fd-avatar--s"
-         style="background-image: url('/assets/images/products/fitbit.png');color:#fff;text-shadow: none;">
+         style="background-image: url('/assets/images/products/fitbit.png')">
         </span>
         <div class="fd-object-list__header-left">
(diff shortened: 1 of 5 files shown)


sKudum:
Hi @InnaAtanasova ,

   fyi..
    PR:[2124](https://github.com/SAP/fundamental-styles/pull/2124),  has lead to this issue. 
    PR: [2222](https://github.com/SAP/fundamental-styles/pull/2222) does not fix it.

     respective file changes are considered in it along with the issues which they address.

Thank you

InnaAtanasova (commented):
Could you please add an example with hard-coded active state so we have the color in the visual tests and avoid such problems in future? Thank you!
The code looks good.

stories/object-list/object-list.stories.js
```diff
@@ -599,6 +599,71 @@ export const selectionWithNavigation = () => `
...
+          <p class="fd-object-identifier__title" style="color:#fff;text-shadow: none;">
+           Fitbit Versa Smart Watch, Black Fitbit Versa Smart Watch
+          </p>
+         </div>
+        </div>
+        <div class="fd-object-list__header-right">
+          <span class="fd-object-number fd-object-list__object-number">
+            <span class="fd-object-number__text" style="color:#fff;">457.00</span>
+            <span class="fd-object-number__unit" style="color:#fff;">Euro</span>
+          </span>
```
InnaAtanasova, line 623: Hi @sKudum this is not what I meant. You are not testing anythings right now as you are hardcoding the colors for active state. We need to test how the colors for the background and text change when the item becomes "active" (is clicked/pressed). This is achieved with `is-active` class. 
In this case you show a selected state of the element by hardcoding `aria-selected="true"`:
Screen Shot 2021-07-16 at 1 00 19 PM


You can achieve the same for active stat by adding `is-active` class to `fd-list__link`. You will have:
```
<a tabindex="0" class="fd-list__link is-active"> ....</a> 
```

This will lead to this view:
Screen Shot 2021-07-16 at 1 04 06 PM


The background and the text color automatically change to dark blue and white because of the additional modifier class. This is how we test different states of the components. Like I said, ignore the focus line. If you insist on testing it you can always create a modifier class for focussed state or write JS code to focus the element all the time, but we don't do it usually. Inline styling is not an option as it won't work with theming and is basically useless.

InnaAtanasova (changes requested):

stories/object-list/object-list.stories.js
sKudum: Hi @InnaAtanasova ,

 Thank You for the elaboration. I have removed the inline style and updated it as suggested. Please let me know if this is ok?

Thank You

olshansky (approved):

Betrozov (approved):

N1XUS (approved):

InnaAtanasova (approved):

InnaAtanasova:
Screen Shot 2021-07-19 at 2 05 45 PM

Could you please rebase main branch? I see Chromatic detected changes in components that shouldn't be affected by your changes.

sKudum:
> Screen Shot 2021-07-19 at 2 05 45 PM
> 
> Could you please rebase main branch? I see Chromatic detected changes in components that shouldn't be affected by your changes.

Hi @InnaAtanasova ,

     Rebased with main again..and I can see this.




where as in local also no diffs are getting listed.

Seamoo13 (approved):

Issue #2460, opened 2021-06-17
Defect Hunting for Sprint 65 - Bern


InnaAtanasova:
Please report the issues you found during the defect hunting with the following format:

```
- [ ] Short title containing the component name
- Description of the issue
- Ways to reproduce it
- Screenshot or gif
- Browser 
```

Before starting to work on the issue please write your name after the title in bold.
Once the PR is open, please link a link to the PR after your name
Mark this issue as done by checking the checkbox

Example:
<img width="836" alt="Screen Shot 2021-06-03 at 3 29 26 PM" src="https://user-images.githubusercontent.com/39598672/120701160-828bf380-c480-11eb-8aba-d84e08204da8.png">

InnaAtanasova:
Issues from previous sprints:
- [ ] https://github.com/SAP/fundamental-styles/issues/2194
- [ ] https://github.com/SAP/fundamental-styles/issues/2364

mikerodonnell89:
- [ ] Calendar in phone portrait mode does not look correct, it looks almost identical to the landscape mode
<img width="959" alt="Screen Shot 2021-06-23 at 1 40 07 PM" src="https://user-images.githubusercontent.com/2471874/123143384-87f9af80-d428-11eb-9b9c-0855b138f29c.png">

But here it is the actual design
<img width="869" alt="Screen Shot 2021-06-23 at 1 38 32 PM" src="https://user-images.githubusercontent.com/2471874/123143205-4ff26c80-d428-11eb-8c92-6310dedc0beb.png">

mikerodonnell89:
- [ ] Tabs - leftmost tab missing padding on the left (same problem but on the right in RTL)

<img width="300" alt="Screen Shot 2021-06-23 at 2 13 45 PM" src="https://user-images.githubusercontent.com/2471874/123147593-3acc0c80-d42d-11eb-83fe-367c0e39380b.png">

mikerodonnell89:
- [ ] Grid list - header that contains edit button is not tall enough.  This is readily apparent in the high contrast color themes. I think the class `fd-grid-list__item-toolbar` should also set the height.

From fundamental-styles:
<img width="305" alt="Screen Shot 2021-06-23 at 3 07 53 PM" src="https://user-images.githubusercontent.com/2471874/123154409-cac18480-d434-11eb-9c1d-70e1499ae773.png">

From the specs:
[...]

PR #2591, opened 2021-07-19
fix: center tristate icon horizontally

InnaAtanasova:
## Related Issue
Part of SAP/fundamental-styles#2460

## Description
Center horizontally the "square" icon used for tristate chackbox

### Before:
Screen Shot 2021-07-19 at 5 00 34 PM
Screen Shot 2021-07-19 at 5 00 52 PM

### After:

Screen Shot 2021-07-19 at 5 01 41 PM
Screen Shot 2021-07-19 at 5 01 35 PM

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [na] Updated tests
4. Documentation

### DIFF
--- a/src/checkbox.scss
+++ b/src/checkbox.scss
@@ -11,6 +11,17 @@ $block: #{$fd-namespace}-checkbox;
   opacity: 0.0001;
 }
 
+@mixin applyIndeterminatePadding() {
+  $offset: 0.0625rem;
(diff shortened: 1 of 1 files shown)


N1XUS:
image
MacBook Pro 16" Retina

droshev (approved):

Issue #2593, opened 2021-07-20
bar component: add focus to the bar element


Lokanathan-k:
add focus to the bar element, focus is not as per the fiori spec

**Describe the bug**

<img width="345" alt="Screenshot 2021-07-20 at 6 45 21 PM" src="https://user-images.githubusercontent.com/53534379/126330592-b6f45c21-28ba-4276-b7e3-4f4d8fbfcceb.png">

<img width="327" alt="focus error " src="https://user-images.githubusercontent.com/53534379/126330168-0ce7f0ad-c3e2-46d7-a270-24e95ceaaecf.png">

droshev:
@Lokanathan-k first of all, how is this related to the bar? Bar is just container component.

Lokanathan-k:
> how is this related to the bar? Bar is just container component


@droshev 
in the popover header and few other components use the fd-bar as the header container these container doesnot have a focus the the element as per fiori spec.

Lokanathan-k:
not an issue

PR #2594, opened 2021-07-20
fix: adding focus to the bar element

Lokanathan-k:
fix: adding focus to the bar element

## Related Issue
Closes SAP/fundamental-styles#https://github.com/SAP/fundamental-styles/issues/2593

## Description

adding focus to the bar element

### Before:

Screenshot 2021-07-20 at 6 45 21 PM

### After:

Screenshot 2021-07-20 at 7 07 14 PM

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [NA] Updated tests
4. Documentation

### DIFF
--- a/src/bar.scss
+++ b/src/bar.scss
@@ -72,6 +72,11 @@ $block: #{$fd-namespace}-bar;
     max-width: 100%;
     max-height: 100%;
 
+    @include fd-focus() {
+      outline: var(--sapContent_FocusWidth) var(--sapContent_FocusStyle) var(--sapContent_FocusColor);
+      outline-offset: var(--sapContent_FocusWidth);
+    }
+
     @include fd-rtl() {
       margin-right: 0;
       margin-left: $fd-bar-element-spacing;


src/bar.scss
```diff
@@ -72,6 +72,11 @@ $block: #{$fd-namespace}-bar;
     max-width: 100%;
     max-height: 100%;
 
+    @include fd-focus() {
+      outline: var(--sapContent_FocusWidth) var(--sapContent_FocusStyle) var(--sapContent_FocusColor);
+      outline-offset: var(--sapContent_FocusWidth);
+    }
+
```
InnaAtanasova, line 79: `fd-bar__element` is just a wrapper whose sole purpose is to apply some spacing on the right hand side (margin-right: $fd-bar-element-spacing;). This element SHOULD NOT get a focus. What gets the focus is the element inside it.

InnaAtanasova (changes requested):

src/bar.scss
Lokanathan-k: these focus change is not required

Issue #2442, opened 2021-06-09
Missing styles/functionality from vertical navigation


mikerodonnell89:
- [x] If an item is navigated to and has an icon but no children, the item text and icon should be colored blue/active.  If an item is navigated and is also a child itself, then the text should be blue.
- [x] Missing vertical navigation grouping
- [x] Cursor: pointer over navigable items

PR #2596, opened 2021-07-20
fix: add missing features for Vertical Navigation 

InnaAtanasova:
## Related Issue
Closes SAP/fundamental-styles#2442

## Description
- changed the cursor to pointer for navigable items
- added modifier class `fd-vertical-nav__group-header` for the `fd-list__group-header` class 
- added example with Grouping
- applied `!important` to the rules changing the color of the navigated/indicated items. Couldn't rely on specificity as it was already used once. 
- In the example added another item with text and icon that is navigated to illustrate the changing of the color.

### DIFF
--- a/src/_listNavigationItem.scss
+++ b/src/_listNavigationItem.scss
@@ -7,6 +7,18 @@ $fd-list-navigation-item-text-spacing: 2.5rem !default;
 $fd-list-second-level-popover-position: 3.25rem !default;
 
 .#{$block} {
+  @mixin applySelectedColor {
+    &.#{$block}__navigation-item--indicated {
+      color: var(--sapSelectedColor);
+
+      .#{$block}__navigation-item-text,
+      .#{$block}__navigation-item-icon,
+      .#{$block}__navigation-item-arrow {
+        color: var(--sapSelectedColor);
+      }
+    }
+  }
+
   .#{$block}__navigation-item-indicator {
     @include fd-reset();
 
@@ -38,6 +50,7 @@ $fd-list-second-level-popover-position: 3.25rem !default;
     @include fd-reset();
     @include fd-fiori-focus();
     @include fd-set-paddings-x(0.5rem, 1rem);
+    @include applySelectedColor();
 
     background: transparent;
     height: $fd-list-navigation-item-height;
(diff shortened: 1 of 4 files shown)


src/_listNavigationItem.scss
```diff
@@ -123,12 +124,12 @@ $fd-list-second-level-popover-position: 3.25rem !default;
       .#{$block}__navigation-item-text,
       .#{$block}__navigation-item-icon,
       .#{$block}__navigation-item-arrow {
-        color: var(--sapSelectedColor);
+        color: var(--sapSelectedColor) !important;
```
InnaAtanasova, line 127: Note: there's already specificity so I couldn't rely on it, had to use !important

src/_listNavigationItem.scss
```diff
@@ -123,12 +124,12 @@ $fd-list-second-level-popover-position: 3.25rem !default;
...
       }
     }
 
     .#{$block}__navigation-item--indicated .#{$block}__navigation-item-text {
-      color: var(--sapSelectedColor);
+      color: var(--sapSelectedColor) !important;
```
InnaAtanasova, line 132: Same note as above

src/vertical-nav.scss
```diff
@@ -16,27 +16,15 @@ $block: #{$fd-namespace}-vertical-nav;
...
     @include fd-reset();
   }
 
   &--condensed {
     width: 3.25rem;
   }
+
+  .#{$block}__group-header {
+    background: var(--sapList_AlternatingBackground);
+  }
```
InnaAtanasova, line 29: removed the other code as I used for base `.fd-list__group-header` and just added this class to change the background.

src/_listNavigationItem.scss
N1XUS: I wouldn't rely on `important`. Maybe add more level to it to make it prioritized if possible?

mikerodonnell89: I too am wondering why not just add another class

InnaAtanasova: @N1XUS I wouldn't rely on 4-5 levels of nesting to enforce specificity either :) It becomes a hell if refactoring is needed or if a new functionality is added. 
@mikerodonnell89 `fd-list__navigation-item--indicated` is the additional class added to change the color of the items. We can't have 2 classes doing the same to enforce specificity. 

Anyways, I added additional nesting. It's not a big deal, just something I don't consider a good practice :)

src/_listNavigationItem.scss
InnaAtanasova: Refactored to rely on specificity.

N1XUS (approved):

PR #2600, opened 2021-07-22
feat: introduce Horizontal Navigation pattern

InnaAtanasova:
## Related Issue
Closes SAP/fundamental-styles#2210

BREAKING CHANGE: 
- Added an additional modifier class `fd-icon-tab-bar__item--overflow` to `fd-icon-tab-bar__item` for the `li` element wrapping the overflow button
- Added an additional modifier class `fd-icon-tab-bar__popover-body` to `fd-popover__body` for the overflow

## Description
 Horizontal Navigation is a composite component built on extended functionality of Universal IconTabBar and Tool Header.

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/icon-tab-bar.scss
+++ b/src/icon-tab-bar.scss
@@ -132,10 +132,22 @@ $fd-icon-tab-bar-semantic-values: (
   }
 }
 
-%fd-tabs-focus {
+@mixin fd-tabs-focus($outlineColor: var(--sapContent_FocusColor)) {
+  outline-offset: 0.125rem;
+  outline-color: $outlineColor;
   outline-style: var(--sapContent_FocusStyle);
   outline-width: var(--sapContent_FocusWidth);
-  outline-color: var(--sapContent_FocusColor);
+}
+
+@mixin setNavigationTabColors($color: var(--sapShell_Navigation_Selected_TextColor)) {
+  [class*='sap-icon'] {
+    color: $color;
+    @content;
+  }
+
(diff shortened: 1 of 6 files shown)


src/icon-tab-bar.scss
```diff
@@ -132,10 +132,25 @@ $fd-icon-tab-bar-semantic-values: (
...
   outline-style: var(--sapContent_FocusStyle);
   outline-width: var(--sapContent_FocusWidth);
-  outline-color: var(--sapContent_FocusColor);
+
+  @if $offset {
+    outline-offset: 0.0625rem;
+  }
+}
+
+@mixin setNavigationTabColors($color: var(--sapShell_Navigation_Selected_TextColor)) {
```
platon-rov, line 145: I see that every other mixin named in kebab-case and only this one in camelCase, should it be unified?

InnaAtanasova: you are absolutely right. I don't know where my mind was when I did this :D Probably on vacation. Thanks for noticing.Fixing it right away  :)

platon-rov (approved):

InnaAtanasova:
BREAKING CHANGE:
Added an additional modifier class fd-icon-tab-bar__item--overflow to fd-icon-tab-bar__item for the li element wrapping the overflow button
Added an additional modifier class fd-icon-tab-bar__popover-body to fd-popover__body for the overflow

PR #2616, opened 2021-07-29
Added missing styles for icon-tab-bar

Smiranin:
## Description
These changes were made during the implementation of the icon-tab-bar component of ngx repo. Related [PR](https://github.com/SAP/fundamental-ngx/pull/6193) 

- Added missing styles for sub-tabs with icon type
- Added styles for drag and drop manipulation
- Added some styles for dynamic overflow behavior.
- Added an example for icon sub-tabs

- [x] All values are in `rem`
These changes were made during the implementation of the icon-tab-bar compenent of fundamental/ngx. 
Related [PR](https://github.com/SAP/fundamental-ngx/pull/6193)

- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -3,7 +3,7 @@ on: push
 
 jobs:
   test:
-    if: "contains(github.event.head_commit.message, '[ci visual]') || contains(github.event.head_commit.message, 'chore(deps-dev)')"
+    if: "contains(github.event.head_commit.message, '[ci visual]') || contains(github.event.head_commit.message, 'chore(deps-dev)') || github.event.ref == 'refs/heads/main'"
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
--- a/src/icon-tab-bar.scss
+++ b/src/icon-tab-bar.scss
@@ -264,7 +264,7 @@ $fd-icon-tab-bar-semantic-values: (
     position: relative;
 
     &--left-offset {
-      padding-left: 3.5rem;
+      @include fd-set-padding-left(3.5rem);
     }
   }
 
@@ -321,10 +321,19 @@ $fd-icon-tab-bar-semantic-values: (
     }
 
     &.#{$block}__item--overflow {
-      @include fd-reset-margins-x();
+      position: absolute;
+      top: 50%;
+      transform: translateY(-50%);
+      margin: 0;
 
-      flex: 1;
-      justify-content: flex-end;
+      &--left {
+        left: 0;
+
+        @include fd-rtl() {
+          left: auto;
+          right: 0;
+        }
+      }
 
       .#{$block}__popover {
         position: relative;
@@ -337,17 +346,6 @@ $fd-icon-tab-bar-semantic-values: (
 
     &--hidden {
       display: none;
-    }
-
-    &--extra-btn {
-      position: absolute;
-      top: 50%;
-      transform: translateY(-50%);
-      margin: 0;
-
-      &--left {
-        left: 0;
-      }
     }
   }
 
@@ -755,6 +753,7 @@ $fd-icon-tab-bar-semantic-values: (
   &__list {
     max-height: 20rem;
     overflow: auto;
+    width: 100%;
 
     .#{$block}__list-item {
       @for $i from 1 through 10 {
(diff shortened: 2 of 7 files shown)


droshev (changes requested):
@Smiranin `Updated tests` has not been updated

src/icon-tab-bar.scss
```diff
@@ -252,6 +261,11 @@ $fd-icon-tab-bar-semantic-values: (
     list-style: none;
     box-shadow: var(--sapContent_HeaderShadow);
     background: var(--sapObjectHeader_Background);
+    position: relative;
+
+    &--left-offset {
+      padding-left: 3.5rem;
```
mikerodonnell89, line 267: I am not seeing this class used anywhere - and typically any sort of padding-left will need its corresponding padding-right for RTL mode. See the new `fd-set-padding-left` mixin

src/icon-tab-bar.scss
```diff
@@ -320,6 +334,21 @@ $fd-icon-tab-bar-semantic-values: (
         }
       }
     }
+
+    &--hidden {
+      display: none;
+    }
+
+    &--extra-btn {
```
mikerodonnell89, line 342: This is unused too

src/icon-tab-bar.scss
```diff
@@ -722,6 +753,9 @@ $fd-icon-tab-bar-semantic-values: (
   }
 
   &__list {
+    max-height: 20rem;
```
mikerodonnell89, line 756: Wondering if we want to leave this up to the application developer

src/icon-tab-bar.scss
```diff
@@ -737,6 +771,14 @@ $fd-icon-tab-bar-semantic-values: (
         }
       }
 
+      min-width: 7rem;
```
mikerodonnell89, line 774: Also wondering if we want to leave this up to the application developer

src/icon-tab-bar.scss
```diff
@@ -1127,4 +1203,49 @@ $fd-icon-tab-bar-semantic-values: (
       @include fd-apply-responsive-paddings();
     }
   }
+
+  .#{block}-dnd-preview {
+    background: var(--sapObjectHeader_Background, #fff);
```
mikerodonnell89, line 1208: Any example of this?

mikerodonnell89 (commented):
Can you link (or create) an issue for this PR? Wondering what is the purpose behind some of these changes, maybe I missed something

src/icon-tab-bar.scss
Smiranin: +

src/icon-tab-bar.scss
Smiranin: Remove this class and applied its styles for --overflow modifier

src/icon-tab-bar.scss
Smiranin: I added default height, because without height/overflow: auto. Popover is looked ugly, It is cut off and climbs out of the page if there are a lot of entities. You can check it to remove max-height

src/icon-tab-bar.scss
Smiranin: I think you are right it is unnecessary

src/icon-tab-bar.scss
Smiranin: I added some classes for drag and drop actions. I don't know how to show it in examples without js. Do you have any ideas?

Smiranin:
> @Smiranin `Updated tests` has not been updated. Also the PR is not named properly
@droshev Updated. 


> Can you link (or create) an issue for this PR? Wondering what is the purpose behind some of these changes, maybe I missed something
@mikerodonnell89 Added link to related PR of ngx

.github/workflows/chromatic.yml
```diff
@@ -3,7 +3,7 @@ on: push
 
 jobs:
   test:
-    if: "contains(github.event.head_commit.message, '[ci visual]') || contains(github.event.head_commit.message, 'chore(deps-dev)')"
+    if: "contains(github.event.head_commit.message, '[ci visual]') || contains(github.event.head_commit.message, 'chore(deps-dev)') || github.event.ref == 'refs/heads/main'"
```
droshev, line 6: this should not be in this PR

Smiranin:
Cannot resolve extra changes

PR #2621, opened 2021-07-30
fix: introduce modifier class to fix nested-list-popover

DeepakSap14:
## Related Issue
Closes https://github.com/SAP/fundamental-styles/issues/2618

Breaking Change:
Added a new modifier class .fd-nested-list--popover to display nested list inside a popover.

## Description
Nested list popover has some styles issue in fundamental-ngx.
It is because `fd-popover` component in `fundamental-ngx` is using class `fd-popover__popper` and not using class `fd-popover__body`.

To fix the issue, some modifier class needs to be created here.

### Before:
image

### After:
Screenshot 2021-07-30 at 9 42 55 PM

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/_nested-list.scss
+++ b/src/_nested-list.scss
@@ -110,10 +110,6 @@ $button: #{$fd-namespace}-button;
     align-items: center;
     border-top: var(--sapList_BorderWidth) solid var(--sapList_BorderColor);
 
-    &:first-child {
-      border-top: none;
-    }
-
     @include fd-hover() {
       background: var(--sapList_Hover_Background);
 
--- a/src/_nested-list.scss
+++ b/src/_nested-list.scss
@@ -69,6 +69,10 @@ $button: #{$fd-namespace}-button;
     display: none;
   }
 
+  &__popover {
+    border-bottom: none;
+  }
+
   &__item {
     @include fd-reset();
 
@@ -79,6 +83,13 @@ $button: #{$fd-namespace}-button;
       .#{$block}__link,
       .#{$block}__content {
         border-top: none;
+      }
+    }
+
+    &--popover {
+      &:last-child {
+        border-bottom-left-radius: 0.25rem;
+        border-bottom-right-radius: 0.25rem;
       }
     }
   }
@@ -93,6 +104,10 @@ $button: #{$fd-namespace}-button;
     width: 100%;
     align-items: center;
     border-top: var(--sapList_BorderWidth) solid var(--sapList_BorderColor);
+
+    &:first-child {
+      border-top: none;
+    }
 
     @include fd-hover() {
       background: var(--sapList_Hover_Background);
@@ -242,6 +257,11 @@ $button: #{$fd-namespace}-button;
     white-space: nowrap;
     overflow: hidden;
     text-overflow: ellipsis;
+
+    &--popover {
+      border-top-left-radius: 0.25rem;
+      border-top-right-radius: 0.25rem;
+    }
   }
 
   &--no-border {
--- a/src/_nested-list.scss
+++ b/src/_nested-list.scss
@@ -32,6 +32,8 @@ $button: #{$fd-namespace}-button;
     }
   }
 
+  $border-radius: 0.25rem;
+
   @include fd-reset();
 
   list-style: none;
@@ -88,13 +90,25 @@ $button: #{$fd-namespace}-button;
 
     &--popover {
       &:first-child {
-        border-top-left-radius: 0.25rem;
-        border-top-right-radius: 0.25rem;
+        border-top-left-radius: $border-radius;
+        border-top-right-radius: $border-radius;
+
+        .#{$block}__link,
+        .#{$block}__content {
+          border-top-left-radius: $border-radius;
+          border-top-right-radius: $border-radius;
+        }
       }
 
       &:last-child {
-        border-bottom-left-radius: 0.25rem;
-        border-bottom-right-radius: 0.25rem;
+        border-bottom-left-radius: $border-radius;
+        border-bottom-right-radius: $border-radius;
+
+        .#{$block}__link,
+        .#{$block}__content {
+          border-bottom-left-radius: $border-radius;
+          border-bottom-right-radius: $border-radius;
+        }
       }
     }
   }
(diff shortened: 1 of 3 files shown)


mikerodonnell89 (dismissed):
Nice work, however it only works if the first list item is a group header:

Screen Shot 2021-07-30 at 3 12 59 PM

platon-rov (approved):

src/_nested-list.scss
```diff
@@ -81,6 +85,18 @@ $button: #{$fd-namespace}-button;
         border-top: none;
       }
     }
+
+    &--popover {
+      &:first-child {
+        border-top-left-radius: 0.25rem;
```
kavya-b, line 91: can we keep this value in a variable and reuse it?

stories/side-navigation/side-navigation.stories.js
```diff
@@ -1254,3 +1254,85 @@ nestedListWithoutLinks.parameters = {
...
+            </li>
+        </ul>
+    </div>
+    <div class="fd-popover__popper" aria-hidden="false" id="popoverA11">
+        <ul class="fd-nested-list fd-nested-list--popover fd-nested-list--compact">
+            <li class="fd-nested-list__group-header fd-nested-list__group-header--popover">
+                Group Header 1
+            </li>
+            <li class="fd-nested-list__item fd-nested-list__item--popover">
+                <a class="fd-nested-list__link"href="#/">
```
kavya-b, line 1274:
```suggestion
                <a class="fd-nested-list__link" href="#/">
```
and in other `<a>` tags in this example

kavya-b (commented):
- when i hover over the last element(or first, when header is not present), the edges are still cut off: 
image

src/_nested-list.scss
DeepakSap14: using variable now.

stories/side-navigation/side-navigation.stories.js
DeepakSap14: other places in documentation has this format. so i did not change that.
i have fixed it now.

DeepakSap14:
> * when i hover over the last element(or first, when header is not present), the edges are still cut off:
>   image

@kavya-b 
this should be working fine now.
Thanks.

DeepakSap14:
@InnaAtanasova 
do i need to keep this PR as breaking change?
i am not very sure, so asking.

stories/side-navigation/side-navigation.stories.js
kavya-b: do storyshots need updating after this change?

kavya-b (approved):
LGTM

N1XUS (approved):

src/_nested-list.scss
```diff
@@ -32,6 +32,8 @@ $button: #{$fd-namespace}-button;
     }
   }
 
+  $border-radius: 0.25rem;
```
InnaAtanasova, line 35: That's not the way to declare a variable, please check other components how it's done

src/_nested-list.scss
```diff
@@ -81,6 +87,30 @@ $button: #{$fd-namespace}-button;
...
+      }
+
+      &:last-child {
+        border-bottom-left-radius: $border-radius;
+        border-bottom-right-radius: $border-radius;
+
+        .#{$block}__link,
+        .#{$block}__content {
+          border-bottom-left-radius: $border-radius;
+          border-bottom-right-radius: $border-radius;
```
InnaAtanasova, line 110: repeatable code, should be moved to a mixin or placeholder

src/_nested-list.scss
```diff
@@ -242,6 +272,12 @@ $button: #{$fd-namespace}-button;
     white-space: nowrap;
     overflow: hidden;
     text-overflow: ellipsis;
+
+    &--popover {
+      border-top-left-radius: $border-radius;
+      border-top-right-radius: $border-radius;
```
InnaAtanasova, line 278: repeatable code

stories/side-navigation/side-navigation.stories.js
```diff
@@ -1254,3 +1254,85 @@ nestedListWithoutLinks.parameters = {
         disable: true
     }
 };
+
+export const nestedListPopover = () => `<div class="fd-popover fd-popover--right" style="margin-bottom: 200px; margin-right: 200px; margin-left: 200px;">
```
InnaAtanasova, line 1258: what are you achieving with the custom css?

stories/side-navigation/side-navigation.stories.js
```diff
@@ -1254,3 +1254,85 @@ nestedListWithoutLinks.parameters = {
...
+                </a>
+            </li>
+        </ul>
+    </div>
+    <div class="fd-popover__popper" aria-hidden="false" id="popoverA11">
+        <ul class="fd-nested-list fd-nested-list--popover fd-nested-list--compact">
+            <li class="fd-nested-list__group-header fd-nested-list__group-header--popover">
+                Group Header 1
+            </li>
+            <li class="fd-nested-list__item fd-nested-list__item--popover">
```
InnaAtanasova, line 1273: Is it necessary to have a modifier class on each element? Can't you use specificity? You have the `--popover` modifier on the parent element, you can then have something like:

```
.fd-nested-list--popover {
     .fd-nested-list__group-header { code overwrite}
     .fd-nested-list__item { code overwrite}
}
```

InnaAtanasova (changes requested):

src/_nested-list.scss
DeepakSap14: using `$fd-nested-list-popover-border-radius` now.

src/_nested-list.scss
DeepakSap14: created and moved to mixing.

src/_nested-list.scss
DeepakSap14: created and moved to mixing.

stories/side-navigation/side-navigation.stories.js
DeepakSap14: using specificity now. Thanks for the `valuable` suggestion.

stories/side-navigation/side-navigation.stories.js
DeepakSap14: to keep distance between the two examples under one doc-section, i was using `margin`.
now I have moved that outside from `fd-popover`.

InnaAtanasova (approved):

InnaAtanasova:
Please add the Breaking changes for the newly introduced class with before and after examples. Thanks

puru-hk:
Looks fine

puru-hk (approved):

PR #2634, opened 2021-08-02
feat: added missing styles for icon-tab-bar

Smiranin:
BREAKING CHANGE:

- Added new modifier `.fd-icon-tab-bar__header--left-offset` when left overflow button are shown
- Added new modifieres with semantic colors for list items inside popover `.fd-icon-tab-bar__list-item--informative`, `.fd-icon-tab-bar__list-item--positive`, `.fd-icon-tab-bar__list-item--negative`, `.fd-icon-tab-bar__list-item--critical`
- Added new html structure for list of icon type items

## Description
These changes were made during the implementation of the icon-tab-bar component of ngx repo. Related [PR](https://github.com/SAP/fundamental-ngx/pull/6193) 

- Added missing styles for sub-tabs with icon type
- Added styles for drag and drop manipulation
- Added some styles for dynamic overflow behavior.
- Added an example for icon sub-tabs

- [x] All values are in `rem`
These changes were made during the implementation of the icon-tab-bar compenent of fundamental/ngx. 
Related [PR](https://github.com/SAP/fundamental-ngx/pull/6193)

- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/icon-tab-bar.scss
+++ b/src/icon-tab-bar.scss
@@ -327,7 +327,7 @@ $fd-icon-tab-bar-semantic-values: (
       transform: translateY(-50%);
       margin: 0;
 
-      &--left {
+      &-left {
         left: 0;
 
         @include fd-rtl() {
@@ -343,10 +343,6 @@ $fd-icon-tab-bar-semantic-values: (
           margin-top: 0.375rem;
         }
       }
-    }
-
-    &--hidden {
-      display: none;
     }
   }
 
@@ -481,8 +477,6 @@ $fd-icon-tab-bar-semantic-values: (
         transform: rotate(180deg);
       }
     }
-
-    width: 2rem;
   }
 
   &__container {
@@ -752,8 +746,6 @@ $fd-icon-tab-bar-semantic-values: (
   }
 
   &__list {
-    max-height: 20rem;
-    overflow: auto;
     width: 100%;
 
     .#{$block}__list-item {
@@ -771,8 +763,7 @@ $fd-icon-tab-bar-semantic-values: (
         }
       }
 
-      .fd-list__link.is-selected,
-      .fd-list__link[aria-selected=true] {
+      @include fd-selected() {
         border-bottom: none;
       }
 
@@ -791,28 +782,26 @@ $fd-icon-tab-bar-semantic-values: (
         font-size: 1rem;
       }
 
-      &--positive {
-        @include fd-icon-tab-bar-list-item-color(var(--sapPositiveElementColor));
-      }
+      @each $set-name, $set-params in $fd-icon-tab-bar-semantic-values {
 
-      &--negative {
-        @include fd-icon-tab-bar-list-item-color(var(--sapNegativeElementColor));
-      }
-
-      &--critical {
-        @include fd-icon-tab-bar-list-item-color(var(--sapCriticalElementColor));
-      }
-
-      &--informative {
-        @include fd-icon-tab-bar-list-item-color(var(--sapInformativeElementColor));
+        &--#{$set-name} {
+          @include fd-icon-tab-bar-list-item-color(map_get($set-params, 'color'));
+        }
       }
     }
 
     .#{$block}__list-item-icon-container {
-      display: flex;
-      align-items: center;
+      @include fd-flex-vertical-center();
+      @include fd-set-margin-right(0.5rem);
+
       width: 1.25rem;
-      margin-right: 0.5rem;
+    }
+
+    // Need to rewrite styles for fd-list__icon
+    .#{$block}__list-item-icon [class*=sap-icon],
+    .#{$block}__list-item-icon [class*=sap-icon]:last-child {
+      margin: 0 !important;
+      width: auto !important;
     }
 
     .#{$block}__line-separator:first-child {
@@ -830,10 +819,6 @@ $fd-icon-tab-bar-semantic-values: (
     }
 
     @include fd-rtl() {
-      .#{$block}__list-item-icon-container {
-        margin-left: 1rem;
-        margin-right: 0;
-      }
 
       .#{$block}__list-item-icon {
         justify-content: center;
@@ -1202,23 +1187,22 @@ $fd-icon-tab-bar-semantic-values: (
   }
 
   &-dnd-preview {
-    background: var(--sapObjectHeader_Background, #fff);
-    display: flex;
+    @include fd-flex-center();
+
+    background: var(--sapObjectHeader_Background);
     flex-direction: column;
-    align-items: flex-start;
-    justify-content: center;
-    font-size: var(--sapFontSize, 0.875rem);
-    line-height: var(--sapContent_LineHeight, 1.4);
+    font-size: var(--sapFontSize);
+    line-height: var(--sapContent_LineHeight);
     opacity: 0.8;
     z-index: 9999 !important;
 
     * {
-      background: var(--sapObjectHeader_Background, #fff);
+      background: var(--sapObjectHeader_Background);
     }
   }
 
   &-dnd-hovered {
-    background: var(--sapButton_Hover_Background, #ebf5fe);
+    background: var(--sapButton_Hover_Background);
   }
 
   &-dnd-separator {
@@ -1231,7 +1215,7 @@ $fd-icon-tab-bar-semantic-values: (
       left: -0.5rem;
       width: 0.125rem;
       height: 100%;
-      background: var(--sapButton_Hover_BorderColor, #0854a0);
+      background: var(--sapButton_Hover_BorderColor);
     }
 
     &--vertical {
--- a/src/icon-tab-bar.scss
+++ b/src/icon-tab-bar.scss
@@ -242,6 +242,15 @@ $fd-icon-tab-bar-semantic-values: (
   }
 }
 
+@mixin fd-icon-tab-bar-list-item-color($color) {
+  .#{$block}__list-item-icon [class*=sap-icon],
+  .#{$block}__list-item-icon[class*=sap-icon],
+  .#{$block}__list-item-title,
+  .#{$block}__list-item-counter {
+    color: $color;
+  }
+}
+
 .#{$block} {
   @include fd-reset();
 
@@ -252,6 +261,12 @@ $fd-icon-tab-bar-semantic-values: (
     list-style: none;
     box-shadow: var(--sapContent_HeaderShadow);
     background: var(--sapObjectHeader_Background);
+    position: relative;
+    overflow-x: hidden;
+
+    &--left-offset {
+      @include fd-set-padding-left(3.5rem);
+    }
   }
 
   &__panel {
@@ -307,10 +322,19 @@ $fd-icon-tab-bar-semantic-values: (
     }
 
     &.#{$block}__item--overflow {
-      @include fd-reset-margins-x();
+      position: absolute;
+      top: 50%;
+      transform: translateY(-50%);
+      margin: 0;
 
-      flex: 1;
-      justify-content: flex-end;
+      &--left {
+        left: 0;
+
+        @include fd-rtl() {
+          left: auto;
+          right: 0;
+        }
+      }
 
       .#{$block}__popover {
         position: relative;
@@ -319,6 +343,10 @@ $fd-icon-tab-bar-semantic-values: (
           margin-top: 0.375rem;
         }
       }
+    }
+
+    &--hidden {
+      display: none;
     }
   }
 
@@ -453,6 +481,8 @@ $fd-icon-tab-bar-semantic-values: (
         transform: rotate(180deg);
       }
     }
+
+    width: 2rem;
   }
 
   &__container {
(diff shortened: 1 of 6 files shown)


mikerodonnell89 (commented):
Regarding this comment: https://github.com/SAP/fundamental-styles/pull/2616#discussion_r680949818

You can add some js in the `fundamental-styles/.storybook/static/preview-head.js` file but given the complexity of drag-n-drop let's not worry about it for now

mikerodonnell89:
e2e tests are failing, should be an easy fix: https://app.travis-ci.com/github/SAP/fundamental-styles/jobs/528530844#L1339

src/icon-tab-bar.scss
```diff
@@ -737,6 +771,11 @@ $fd-icon-tab-bar-semantic-values: (
         }
       }
 
+      .fd-list__link.is-selected,
+      .fd-list__link[aria-selected=true] {
+        border-bottom: none;
+      }
```
InnaAtanasova, line 777: You can use `fd-selected` mixin

src/icon-tab-bar.scss
```diff
@@ -1127,4 +1200,48 @@ $fd-icon-tab-bar-semantic-values: (
       @include fd-apply-responsive-paddings();
     }
   }
+
+  &-dnd-preview {
+    background: var(--sapObjectHeader_Background, #fff);
```
InnaAtanasova, line 1205: background: var(--sapObjectHeader_Background);

src/icon-tab-bar.scss
```diff
@@ -1127,4 +1200,48 @@ $fd-icon-tab-bar-semantic-values: (
...
+    display: flex;
+    flex-direction: column;
+    align-items: flex-start;
+    justify-content: center;
```
InnaAtanasova, line 1209: Use mixin `fd-flex-center`

src/icon-tab-bar.scss
```diff
@@ -1127,4 +1200,48 @@ $fd-icon-tab-bar-semantic-values: (
...
+    font-size: var(--sapFontSize, 0.875rem);
```
InnaAtanasova, line 1210: font-size: var(--sapFontSize);

src/icon-tab-bar.scss
```diff
@@ -1127,4 +1200,48 @@ $fd-icon-tab-bar-semantic-values: (
...
+    line-height: var(--sapContent_LineHeight, 1.4);
```
InnaAtanasova, line 1211: line-height: var(--sapContent_LineHeight);

src/icon-tab-bar.scss
```diff
@@ -1127,4 +1200,48 @@ $fd-icon-tab-bar-semantic-values: (
...
+    opacity: 0.8;
+    z-index: 9999 !important;
+
+    * {
+      background: var(--sapObjectHeader_Background, #fff);
```
InnaAtanasova, line 1216: background: var(--sapObjectHeader_Background);

src/icon-tab-bar.scss
```diff
@@ -1127,4 +1200,48 @@ $fd-icon-tab-bar-semantic-values: (
...
+    }
+  }
+
+  &-dnd-hovered {
+    background: var(--sapButton_Hover_Background, #ebf5fe);
```
InnaAtanasova, line 1221: please remove the fallback

src/icon-tab-bar.scss
```diff
@@ -1127,4 +1200,48 @@ $fd-icon-tab-bar-semantic-values: (
...
+    position: relative;
+
+    &::before {
+      content: '';
+      position: absolute;
+      top: 0;
+      left: -0.5rem;
+      width: 0.125rem;
+      height: 100%;
+      background: var(--sapButton_Hover_BorderColor, #0854a0);
```
InnaAtanasova, line 1234: please remove the fallback

src/icon-tab-bar.scss
```diff
@@ -1127,4 +1200,48 @@ $fd-icon-tab-bar-semantic-values: (
...
+  }
+
+  &-dnd-separator {
+    position: relative;
+
+    &::before {
+      content: '';
+      position: absolute;
+      top: 0;
+      left: -0.5rem;
```
InnaAtanasova, line 1231: shouldn't this be handled by RTL as well?

stories/icon-tab-bar/icon-tab-bar.stories.js
```diff
@@ -835,8 +835,8 @@ filter.parameters = {
 
 
 export const overflow = () => `<div class="fd-icon-tab-bar">
-    <ul role="tablist" class="fd-icon-tab-bar__header"> 
-        <li role="presentation" class="fd-icon-tab-bar__item">
+    <ul role="tablist" class="fd-icon-tab-bar__header fd-icon-tab-bar__header--left-offset"> 
+        <li role="presentation" class="fd-icon-tab-bar__item fd-icon-tab-bar__item--overflow fd-icon-tab-bar__item--overflow--left">
```
InnaAtanasova, line 839: `fd-icon-tab-bar__item--overflow--left` is not a valid BEM.
Probably `fd-icon-tab-bar__item--overflow-left`

src/icon-tab-bar.scss
```diff
@@ -320,6 +344,10 @@ $fd-icon-tab-bar-semantic-values: (
         }
       }
     }
+
+    &--hidden {
+      display: none;
+    }
```
InnaAtanasova, line 350: where is this used?

src/icon-tab-bar.scss
```diff
@@ -453,6 +481,8 @@ $fd-icon-tab-bar-semantic-values: (
         transform: rotate(180deg);
       }
     }
+
+    width: 2rem;
```
InnaAtanasova, line 485: Where did you take this from?

src/icon-tab-bar.scss
```diff
@@ -751,6 +790,29 @@ $fd-icon-tab-bar-semantic-values: (
...
+        @include fd-icon-tab-bar-list-item-color(var(--sapNegativeElementColor));
+      }
+
+      &--critical {
+        @include fd-icon-tab-bar-list-item-color(var(--sapCriticalElementColor));
+      }
+
+      &--informative {
+        @include fd-icon-tab-bar-list-item-color(var(--sapInformativeElementColor));
+      }
```
InnaAtanasova, line 808: should be documented as breaking changes as they are new classes.

InnaAtanasova: Can't you add the colors to `$fd-icon-tab-bar-semantic-values` mixin at the top of the file and then use a loop to set the color?

src/icon-tab-bar.scss
```diff
@@ -751,6 +790,29 @@ $fd-icon-tab-bar-semantic-values: (
...
+    }
+
+    .#{$block}__list-item-icon-container {
+      display: flex;
+      align-items: center;
```
InnaAtanasova, line 813: Use `fd-flex-vertical-center` mixin

src/icon-tab-bar.scss
```diff
@@ -751,6 +790,29 @@ $fd-icon-tab-bar-semantic-values: (
...
+      width: 1.25rem;
+      margin-right: 0.5rem;
```
InnaAtanasova, line 815: Use `fd-set-margin-right`. It will handle RTL

src/icon-tab-bar.scss
```diff
@@ -766,6 +828,17 @@ $fd-icon-tab-bar-semantic-values: (
 
       height: 2rem;
     }
+
+    @include fd-rtl() {
+      .#{$block}__list-item-icon-container {
+        margin-left: 1rem;
+        margin-right: 0;
```
InnaAtanasova, line 835: won't be needed if you use `fd-set-margin-right` mixin

InnaAtanasova (changes requested):

InnaAtanasova:
Some visual problems:
- The dropdown menu should have the same position as in the other examples
Screen Shot 2021-08-03 at 11 59 27 AM
- The last separator (>>>) is cut, only part of it is visible
Screen Shot 2021-08-03 at 11 59 37 AM
- overflow problem:
Screen Shot 2021-08-03 at 12 00 27 PM
- button should not be this color:
Screen Shot 2021-08-03 at 12 00 37 PM

src/icon-tab-bar.scss
Smiranin: Yes, separator styles don't depend on rtl. These styles for separator between two tabs. Rtl should be handle inside JS for drag and drop.

src/icon-tab-bar.scss
Smiranin: It is using for overflow directive. You are right, it shouldn't be in styles. Moved to ngx.

src/icon-tab-bar.scss
Smiranin: I think I added it, when was implementing overflow directive in ngx, for some version of directive and forgot to delete. 
Removed and checked ngx, works well without these styles.

src/icon-tab-bar.scss
Smiranin: Breaking changes were added to description of PR and Breaking changes page.

Smiranin:
@InnaAtanasova 

> - button should not be this color:


I didn't changes these styles. I checked it and noticed that other styles rewrite it


Also I checked pdf which you sent me and I see such color in pdf example. 


Perhaps these are the correct styles? If not, what are the requirements for this button?

src/icon-tab-bar.scss
```diff
@@ -307,10 +322,19 @@ $fd-icon-tab-bar-semantic-values: (
...
 
-      flex: 1;
-      justify-content: flex-end;
+      &-left {
+        left: 0;
+
+        @include fd-rtl() {
+          left: auto;
+          right: 0;
+        }
```
InnaAtanasova, line 336: Forgot to say during the previous review that there's a mixin for setting the position:
Screen Shot 2021-08-05 at 9 50 11 AM

Will handle the rtl

src/icon-tab-bar.scss
```diff
@@ -751,6 +781,27 @@ $fd-icon-tab-bar-semantic-values: (
...
+      @include fd-set-margin-right(0.5rem);
+
+      width: 1.25rem;
+    }
+
+    // Need to rewrite styles for fd-list__icon
+    .#{$block}__list-item-icon [class*=sap-icon],
+    .#{$block}__list-item-icon [class*=sap-icon]:last-child {
+      margin: 0 !important;
+      width: auto !important;
```
InnaAtanasova, line 804: Can't you use specificity to avoid relying on !important?

src/icon-tab-bar.scss
```diff
@@ -1127,4 +1185,47 @@ $fd-icon-tab-bar-semantic-values: (
       @include fd-apply-responsive-paddings();
     }
   }
+
+  &-dnd-preview {
+    @include fd-flex-center();
+
+    background: var(--sapObjectHeader_Background);
+    flex-direction: column;
```
InnaAtanasova, line 1193: this should be moved to the `fd-flex-center()` mixin:
```
@include fd-flex-center() {
    flex-direction: column;
}
```

stories/dnd-icon-tab-bar/dnd-icon-tab-bar.stories.js
```diff
@@ -0,0 +1,71 @@
...
+
+export const separatorHoveredTab = () => `<div class='fddocs-icon-tab-container' style="min-height: 800px;">
+<div class="fd-icon-tab-bar">
+    <ul role="tablist" class="fd-icon-tab-bar__header">  
+        <li role="presentation" class="fd-icon-tab-bar__item">
+            <a role="tab" class="fd-icon-tab-bar__tab" href="#section1" id="tab1">
+                <span class="fd-icon-tab-bar__tag">Attachments (16)</span>
+            </a>
+        </li>
+        <li role="presentation" class="fd-icon-tab-bar__item fd-icon-tab-bar-dnd-separator">
```
InnaAtanasova, line 53: `fd-icon-tab-bar-dnd-separator`, `fd-icon-tab-bar-dnd-preview` and `fd-icon-tab-bar-dnd-hovered` are not really following the BEM naming. IMO they should be named either:
`fd-icon-tab-bar__dnd-separator`, `fd-icon-tab-bar__dnd-preview` and `fd-icon-tab-bar__dnd-hovered`

OR

since they are modifying the `fd-icon-tab-bar__item` element: 
`fd-icon-tab-bar__item--dnd-separator`, `fd-icon-tab-bar__item--dnd-preview` and `fd-icon-tab-bar__item--dnd-hovered`

InnaAtanasova: Me favouring more the second one as these classes are modifier classes :)

InnaAtanasova:
Screen Shot 2021-08-05 at 10 18 45 AM
Screen Shot 2021-08-05 at 10 18 35 AM

The icons and the counter do not change the color on active state. Should have the same color as the text, o.w. will fail the a11y tests for contrast.

src/icon-tab-bar.scss
Smiranin: We have mixins for any cases of life)

src/icon-tab-bar.scss
Smiranin: Yes. Strange, I tested it yesterday and fd-list__icon has higher priority. But today I have recheck and it works well without important.
Removed important

InnaAtanasova (approved):

PR #2669, opened 2021-08-12
[WIP] feat: introduce new Timeline component

InnaAtanasova:
!!! DO NOT MERGE !!!
This component needs to be finished in Fundamental-ngx first and then reflect the css changes back here.

The current PR contains ONLY the css for the building blocks anatomy, such as group headers, posts/items, nodes and connecting lines. The positioning of the posts and the nodes requires JavaScript and the implementation of the component will be continued in Fundamental-ngx. Further changes (additional css classes, wrappers, refactoring) will follow from the ngx implementation and will be reflected here at a later stage. Once this is done we will officially add the component to the fund-styles documentation.


Smiranin (commented):
1) I don't see styles and examples for header(filter, sorting, search).

2) "More" button should have less size and I think the same color as color for actions.
Docs: 

Current implementation:

PR #2680, opened 2021-08-17
docs: update storybook version 6.3.7

droshev:
## Description
migrate to storybook version 6.3.7

1. Testing
- [X] Updated tests
2. Documentation


stories/action-bar/action-bar.stories.js
```diff
@@ -43,8 +43,9 @@ export const backBtn = () => `<div class="fd-action-bar">
 backBtn.storyName = 'Default';
 backBtn.parameters = {
     docs: {
-        iframeHeight: 400,
-        storyDescription: 'The default action bar displays a page title, description, back button and action buttons. To display a back button, add the <code>fd-action-bar\\_\\_description--back</code> modifier class to the description element.'
+        description: {
+            story: 'The default action bar displays a page title, description, back button and action buttons. To display a back button, add the `fd-action-bar__description--back` modifier class to the description element.'
```
platon-rov, line 47: Here is the only place where we have `docs => description => story` instead of `docs => storyDescription` i guess we should go with a unified approach across all the stories.

Btw, you have removed `iframeHeight: 400` i guess it should be reverted.

stories/object-list/object-list.stories.js
```diff
@@ -33,30 +33,30 @@ Below are examples of how to add semantic statuses to the components:
 
 | **Status** | <span style="margin-left: 2rem;">**Object Number**</span> | <span style="margin-left: 2rem;">**Object Status**</span> |
 | :---- | :-------------- | :-------------- |
-| Positive|<code class="docs-code" style="margin-left: 2rem;">fd-object-number--positive</code> |<code class="docs-code" style="margin-left: 2rem;">fd-object-status--positive</code> |
-| Negative|<code class="docs-code" style="margin-left: 2rem;">fd-object-number--negative</code> |<code class="docs-code" style="margin-left: 2rem;">fd-object-status--negative</code> |
-| Critical|<code class="docs-code" style="margin-left: 2rem;">fd-object-number--critical</code> |<code class="docs-code" style="margin-left: 2rem;">fd-object-status--critical</code> |
-| Informative|<code class="docs-code" style="margin-left: 2rem;">fd-object-number--informative</code> |<code class="docs-code" style="margin-left: 2rem;">fd-object-status--informative</code> |
+| Positive| <code style="margin-left: 2rem;">d-object-number--positive</code> | <code style="margin-left: 2rem;">d-object-status--positive</code> |
```
platon-rov, line 36: Here and below `d` instead of `fd`.

platon-rov (changes requested):

stories/object-list/object-list.stories.js
droshev: good catch! updated the PR

stories/action-bar/action-bar.stories.js
droshev: 1. `iframeHeight` doesn't work here. I think we need to. get rid of it.
2. i did that as an example how we should be moving toward. in Version 7.x `storyDescription` will be gone. Let's do that progressively.

platon-rov (approved):

PR #2681, opened 2021-08-18
fix: added paddings for fd-icon-tab-bar__list-item-counter

Smiranin:
## Description
Added missed styles for icon tab bar component.

**Before:** 

**After:**

### DIFF
--- a/src/icon-tab-bar.scss
+++ b/src/icon-tab-bar.scss
@@ -832,6 +832,10 @@ $fd-icon-tab-bar-semantic-values: (
       width: 1.25rem;
     }
 
+    .#{$block}__list-item-counter {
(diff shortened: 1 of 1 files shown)


N1XUS (approved):

olshansky (approved):

src/icon-tab-bar.scss
```diff
@@ -832,6 +832,10 @@ $fd-icon-tab-bar-semantic-values: (
       width: 1.25rem;
     }
 
+    .#{$block}__list-item-counter {
+      @include fd-set-padding-left(0.25rem);
+    }
```
dimamarksman, line 837: do we handle rtl as well?

PR #2682, opened 2021-08-18
fix: form grid layout

### LINKED ISSUE #2612: Negative margin values can result in form layout grid overflowing parent container
Negative margin values can result in form layout grid overflowing parent container.  This can cause problems for example when using the form layout in a dialog - the form layout is actually bigger than the dialog causing the dialog to have an unnecessary scroll bar.  We need to find an alternative to the negative values

Screen Shot 2021-07-27 at 4 18 36 PM

platon-rov:
## Related Issue

Closes #2612.

## Description

Negative margin fixes for the grid.

Negative margins were set for the container what is the wrong approach which causes bugs.

The container should have padding then the row should have negative margins and columns should have padding.
Negative margins are applied to neutralize columns padding.

## Screenshots

Screenshot 2021-08-18 at 17 50 53

- [X] All values are in `rem`
- [X] RTL support
3. Testing
- [n/a] Updated tests
4. Documentation

### DIFF
--- a/src/form-layout-grid.scss
+++ b/src/form-layout-grid.scss
@@ -19,7 +19,7 @@ $cols: 11;
 
 @media (min-width: 0) {
   .#{$blockContainer}.#{$block}-container {
-    margin: 0 -0.25rem;
+    padding: 0 0.25rem;
 
     & > .fd-row {
       & > :first-child & :not(.fd-col--wrap) {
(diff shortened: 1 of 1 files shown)


platon-rov:
Further investigation is needed. Will be discussed on the call and then will continue working on it.

Issue #2684, opened 2021-08-18
Input theming design changes


droshev:
The input component design has been amended with some theming parameters.
the goal of this issue is to identify the changes and reflect them in fundamental styles.

PR #2688, opened 2021-08-20
fix: update input theming

hanjunt:
## Related Issue
Closes SAP/fundamental-styles#2684

## Description
Update input theming parameters.

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/mixins/_forms.scss
+++ b/src/mixins/_forms.scss
@@ -15,8 +15,12 @@
 
   &.is-success {
     background-color: var(--sapField_SuccessBackground);
-    border-color: var(--sapField_SuccessColor);
-    border-width: $fd-form-border-thin-width;
+
+    @include applyBorder(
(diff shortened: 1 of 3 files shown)


src/mixins/_forms.scss
```diff
@@ -164,3 +183,9 @@
     margin-right: 0;
   }
 }
+
+@mixin applyBorder($borderColor, $borderWidth, $borderStyle) {
```
mikerodonnell89, line 187: Nice work - i think let's put this mixin in `mixins.scss` as it can be useful with other components. Otherwise looks good to go

mikerodonnell89 (commented):
One small thing, see above, otherwise looks good

mikerodonnell89 (approved):

Issue #2691, opened 2021-08-26
put nested list popover header outside list to follow Ngx implementation


DeepakSap14:
**Describe the bug**
Currently header for nested list popover is inside the list. But [Ngx](https://sap.github.io/fundamental-ngx/#/core/side-navigation#condensed-mode) implementation shows that, it need to be outside the list.
So modified example for nested list popover.
Also fixed border issue because of this change.

I am also raising a [WIP] PR in Ngx for removing nested list popover for removing the css and with changes for fixing nested list popover border cutting issue. To show the final working in Ngx .

> Is this issue related to a specific component?
Nested list popover

> What version of the Fundamental Library Styles are you using?
latest

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

> What front-end framework are you implementing in, e.g., Angular, React, etc.?
Angular

**Expected behavior**
List header should be outside list.

**Screenshots**
Ngx implementation:
<img width="1005" alt="Screenshot 2021-08-26 at 12 09 31 PM" src="https://user-images.githubusercontent.com/57661487/130913705-ee305da4-e89c-4111-871a-2372fcc07c34.png">

PR #2692, opened 2021-08-26
fix: put nested list popover header outside list items

DeepakSap14:
## Related Issue
Closes https://github.com/SAP/fundamental-styles/issues/2691

## Description
Currently header for nested list popover is inside the list(**edit**: this header is list header, not popover header). But Ngx implementation shows that, it need to be outside the list(**edit**: popover header will be outside).
So created class "__popover-header" which will apply on popover header, outside list.
Also fixed border issue because of this change.

### Before:
Screenshot 2021-08-26 at 12 18 12 PM

### After:
Screenshot 2021-08-26 at 12 18 59 PM

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/_nested-list.scss
+++ b/src/_nested-list.scss
@@ -86,8 +86,6 @@ $button: #{$fd-namespace}-button;
 
     .#{$block}__group-header {
       @include fd-nested-list-popover-border-top-radius();
-
-      border-bottom: none;
     }
 
     .#{$block}__item {
--- a/src/_nested-list.scss
+++ b/src/_nested-list.scss
@@ -84,41 +84,11 @@ $button: #{$fd-namespace}-button;
   &--popover {
     border-bottom: none;
 
-    .#{$block}__popover-header {
-      @include fd-nested-list-popover-border-top-radius();
-
-      height: 2.75rem;
-      background: var(--sapList_GroupHeaderBackground);
-      border-bottom: var(--sapList_BorderWidth) solid var(--sapList_GroupHeaderBorderColor);
-      color: var(--sapList_GroupHeaderTextColor);
-      display: flex;
-      align-items: flex-end;
-      font-size: var(--sapFontSize);
-      font-weight: bold;
-      line-height: 2rem;
-      padding-right: 1rem;
-      padding-left: 1rem;
-      white-space: nowrap;
-      overflow: hidden;
-      text-overflow: ellipsis;
-    }
-
     .#{$block}__group-header {
-      @include fd-nested-list-popover-border-top-radius();
-
       border-bottom: none;
     }
 
     .#{$block}__item {
-      &:first-child {
-        @include fd-nested-list-popover-border-top-radius();
-
-        .#{$block}__link,
-        .#{$block}__content {
-          @include fd-nested-list-popover-border-top-radius();
-        }
-      }
-
       &:last-child {
         @include fd-nested-list-popover-border-bottom-radius();
 
--- a/src/_nested-list.scss
+++ b/src/_nested-list.scss
@@ -84,8 +84,29 @@ $button: #{$fd-namespace}-button;
   &--popover {
     border-bottom: none;
 
+    .#{$block}__popover-header {
+      @include fd-nested-list-popover-border-top-radius();
+
+      height: 2.75rem;
+      background: var(--sapList_GroupHeaderBackground);
+      border-bottom: var(--sapList_BorderWidth) solid var(--sapList_GroupHeaderBorderColor);
+      color: var(--sapList_GroupHeaderTextColor);
+      display: flex;
+      align-items: flex-end;
+      font-size: var(--sapFontSize);
+      font-weight: bold;
+      line-height: 2rem;
+      padding-right: 1rem;
+      padding-left: 1rem;
+      white-space: nowrap;
+      overflow: hidden;
+      text-overflow: ellipsis;
+    }
+
     .#{$block}__group-header {
       @include fd-nested-list-popover-border-top-radius();
+
+      border-bottom: none;
     }
 
     .#{$block}__item {
--- a/src/_nested-list.scss
+++ b/src/_nested-list.scss
@@ -93,6 +93,8 @@ $button: #{$fd-namespace}-button;
       height: 2.75rem;
       padding-right: 1rem;
       padding-left: 1rem;
+      font-size: var(--sapFontLargeSize);
+      color: var(--sapList_TextColor);
       display: flex;
       align-items: center;
     }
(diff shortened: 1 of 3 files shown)


stories/side-navigation/side-navigation.stories.js
```diff
@@ -1267,10 +1267,10 @@ export const nestedListPopover = () => `<div class="fddocs-container" style="mar
             </ul>
         </div>
         <div class="fd-popover__popper" aria-hidden="false" id="popoverA11">
+            <div class="fd-nested-list fd-nested-list--popover fd-nested-list--compact">
```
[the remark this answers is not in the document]
DeepakSap14: @mikerodonnell89,
please have a look into this.

InnaAtanasova (changes requested):
Please reconsider your implementation. In one case it's about `fd-nested-list__popover-header` and in the other it's about `fd-nested-list__group-header`. Two totally different things.

DeepakSap14:
> Please reconsider your implementation. In one case it's about `fd-nested-list__popover-header` and in the other it's about `fd-nested-list__group-header`. Two totally different things.

@InnaAtanasova 
I have created a class "fd-nested-list__popover-header" for **popover header**, earlier it was not there.
I have kept **group header** as it is.
I have created examples in Styles here and Ngx [PR](https://github.com/SAP/fundamental-ngx/pull/6387).
please review once again.

stories/side-navigation/side-navigation.stories.js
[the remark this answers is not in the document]
DeepakSap14: @platon-rov. now the structure is like below. Only popover header will be outside list.
`.fd-nested-list__popover-header
.fd-nested-list
    > .fd-nested-list__header
    > .fd-nested-list__body
        > .fd-nested-list__item
    > .fd-nested-list__footer`

src/_nested-list.scss
```diff
@@ -84,6 +84,25 @@ $button: #{$fd-namespace}-button;
   &--popover {
     border-bottom: none;
 
+    .#{$block}__popover-header {
```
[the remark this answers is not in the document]
InnaAtanasova: The problem is that the popover header in ngx does not look like the group header:
Screen Shot 2021-08-31 at 10 59 39 AM

The first item in the popover is the popover header.
It's totally different from this:
Screen Shot 2021-08-31 at 11 00 51 AM


Have you checked the design before implementing it?

InnaAtanasova (changes requested):

src/_nested-list.scss
```diff
@@ -84,6 +84,25 @@ $button: #{$fd-namespace}-button;
...
+      @include fd-nested-list-popover-border-top-radius();
```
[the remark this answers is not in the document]
DeepakSap14: #{$block}__popover-header is not needed.
Popover header design is created using fd-bar element. so fd-bar styles can be used here.

src/_nested-list.scss
```diff
@@ -85,21 +85,10 @@ $button: #{$fd-namespace}-button;
     border-bottom: none;
 
     .#{$block}__group-header {
-      @include fd-nested-list-popover-border-top-radius();
```
DeepakSap14, line 88: assuming, nested list popover will always have popover header. so border rounding is not needed on group header,

src/_nested-list.scss
```diff
@@ -85,21 +85,10 @@ $button: #{$fd-namespace}-button;
...
-
       border-bottom: none;
     }
 
     .#{$block}__item {
-      &:first-child {
```
DeepakSap14, line 94: assuming, nested list popover will always have popover header. so border rounding for first list item will not be needed.

InnaAtanasova (changes requested):
Please follow the fund-ngx structure. You can't use the Popover header here.

DeepakSap14:
> Please follow the fund-ngx structure. You can't use the Popover header here.

@InnaAtanasova 
i did not find design for nested-list/nested-list-popover in wiki.
I will be helpful, if you can provide the link if it is there.

for now, i have created `fd-nested-list__popover-header` class, and used `fd-nested-list--text-only` & `fd-nested-list__title` for matching with fundamental-ngx.
let me know if this looks correct otherwise design spec needed for this.

InnaAtanasova (changes requested):
You can't have `fd-nested-list fd-nested-list--text-only fd-nested-list--popover` as the ` fd-nested-list--text-only` class is intended for the items inside the side nav (nested list) which have no icons. 
Also, your popover header is not in a compact mode as the rest of the elements. 
Screen Shot 2021-09-03 at 10 40 52 AM

DeepakSap14:
> You can't have `fd-nested-list fd-nested-list--text-only fd-nested-list--popover` as the ` fd-nested-list--text-only` class is intended for the items inside the side nav (nested list) which have no icons (all items of the list should not have icons).
> Also, your popover header is not in a compact mode as the rest of the elements.
> Screen Shot 2021-09-03 at 10 40 52 AM

Thanks Inna.
removed `fd-nested-list--text-only` from popover-header.
introduced compact behaviour for popover-header.

stories/side-navigation/side-navigation.stories.js
```diff
@@ -1303,7 +1307,13 @@ export const nestedListPopover = () => `<div class="fddocs-container" style="mar
             </ul>
         </div>
         <div class="fd-popover__popper" aria-hidden="false" id="popoverA12">
-            <ul class="fd-nested-list fd-nested-list--popover fd-nested-list--compact">
+            <div class="fd-nested-list fd-nested-list--popover">
+                <div class="fd-nested-list__popover-header">
+                    Popover Header
+                </div>
```
InnaAtanasova, line 1313: Please follow the markup from fund-ngx.

InnaAtanasova (changes requested):

stories/side-navigation/side-navigation.stories.js
DeepakSap14: Can you please give more info/describe more.
In last change, when i tried to follow the Ngx markup [here](https://github.com/SAP/fundamental-styles/pull/2692#pullrequestreview-746182875). there was concern that we can not use `fd-nested-list--text-only`.
I am not sure, what changes i need to make now.

DeepakSap14: @InnaAtanasova 
I am not sure about what markup i need to take from Ngx. what i have i already done is taken from Ngx markup.
It looks fine to me.
for comparing purpose i had created https://github.com/SAP/fundamental-ngx/pull/6387 PR, so comparision will be better. which is already linked in this PR.

Also as there is no design available and i am not clarified on exactly what i need to make change.
I am closing this PR.

DeepakSap14:
Also as there is no design available and i am not clarified on exactly what i need to make change.
So, I am closing this PR.

Issue #2690, opened 2021-08-25
Update margin value for popover no-arrow


aberikashvili:
The issue https://github.com/SAP/fundamental-ngx/issues/5985 requires popover width to be adjusted as popover control

PR #2693, opened 2021-08-26
fix: update margin value for popover no-arrow

aberikashvili:
## Related Issue
Closes SAP/fundamental-styles#2690

## Description
margin value, for popover no-arrow, need to be updated, to fit combobox, multi-select, etc. dropdown popover

### Before:
image

### After:
image

- [x] All values are in `rem`
- [N/A ] focus state of the element follow design spec
- [ N/A] active state of the element follow design spec
- [ N/A] selected state of the element follow design spec
- [ N/A] selected hover state of the element follow design spec
- [ N/A] pressed state of the element follow design spec
- [ N/A] Responsiveness rules - the component has modifier classes for all breakpoints
- [ N/A] Includes Compact/Cosy/Tablet design
- [ N/A] RTL support
- [ N/A] only one top level `fd-*` class is used in the file
- [ N/A] BEM naming convention is used
- [ N/A] Mixins are used for repeatable code (`fd-rtl`, `fd-ellipsis`, `fd-flex`, `fd-selected`, `fd-focus`, ect.)
- [ N/A] A11y support - keyboard support, screenreader support, proper ARIA attributes, etc.
- [ N/A] `fd-reset()` mixin is applied to all elements
- [ N/A] Variables are used, if some value is used more than twice.
- [ N/A] Checked if current components can be reused, instead of having new code.
3. Testing
- [ N/A] tested Storybook examples with "CSS Resources" `normalize` option 
- [ N/A] tested Storybook examples with "CSS Resources" `unnormalize` option 
- [ N/A] Verified all styles in IE11
- [ N/A] Updated tests
4. Documentation
- [ N/A] Storybook documentation has been created/updated
- [ N/A] Breaking Changes [wiki](https://github.com/SAP/fundamental-styles/wiki/Breaking-Changes) has been updated in case of breaking changes.

### DIFF
--- a/src/popover.scss
+++ b/src/popover.scss
@@ -252,7 +252,7 @@ $block: #{$fd-namespace}-popover;
     opacity: 1;
 
     &--no-arrow {
-      margin: 0 !important;
+      margin: 0 0.05rem !important;
 
       .#{$block}__arrow {
         display: none;


Smiranin:
I don't quite understand what you are fixing here, the link you shared says about problems with corners and some errors in the console in the combo-box component, in your changes I see a change in the width of the popover. Are you sure that you sent correct link on the issue?

aberikashvili:
@Smiranin here's how combobox dropdown looks with default `margin: 0` 
image

And here's how it looks after this fix is applied:
image

dimamarksman:
@InnaAtanasova @droshev should we create the fix here for an issue that started appearing after some changes in ngx PR that still isn't approved and merged?

InnaAtanasova:
> @InnaAtanasova @droshev should we create the fix here for an issue that started appearing after some changes in ngx PR that still isn't approved and merged?

nope, we shouldn't. The margin should be 0. With custom margin we are applying band-aids on a bigger problem.

InnaAtanasova (changes requested):
We should further investigate the problem. I really doubt we should fix the problem with custom margin.

Issue #2697, opened 2021-08-27
In high contrast white theme, focus indication provided for Inverted Link does not satisfy the minimum contrast ratio of 7:1 


Lokanathan-k:
Observed Behavior:
In high contrast white theme, focus indication provided for Inverted Link does not satisfy the minimum contrast ratio of 7:1 

Expected Behavior:
Ideally, in high contrast white theme, the contrast ratio should be equal to or more than 7:1

Screenshot:
![image](https://user-images.githubusercontent.com/88035260/127949713-1b8480ca-12c5-4941-8e27-c41dda53442a.png)

droshev:
@Lokanathan-k this issue is only for ngx doc site. there the example has a hardcoded color:
```
<div style="background-color:#314a5e;padding:10px">
    <a fd-link [routerLink]="['./']" fragment="inverted" [inverted]="true">Inverted Link</a>
</div>
```
the example has to be updated

PR #2698, opened 2021-08-27
fix: focus indication provided for Inverted Link.

Lokanathan-k:
fix: focus indication provided for Inverted Link [ci visual]

## Related Issue
Closes: https://github.com/SAP/fundamental-styles/issues/2697

## Description
In high contrast white theme, focus indication provided for Inverted Link does not satisfy the minimum contrast ratio of 7:1

### Before:
Screenshot 2021-08-27 at 7 07 57 PM

### After:
Screenshot 2021-08-27 at 7 11 15 PM

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/stories/link/link.stories.js
+++ b/stories/link/link.stories.js
@@ -21,7 +21,7 @@ Use a meaningful link text that indicates what will happen when the user interac
 Avoid texts such as *Click Here* or *Link*, as these do not make it clear to the user what the purpose of the link is.
 `,
         tags: ['f3', 'a11y', 'theme'],
-        components: ['icon', 'link', 'shellbar']
+        components: ['icon', 'link']
     }
 };
 
@@ -34,7 +34,7 @@ export const primary = () => `
 <br><br>
 <a class="fd-link" aria-disabled="true">Disabled link</a>
 <br><br>
-<div class="fd-shellbar">
+<div  style="background-color:var(--sapShellColor);padding:10px">
     <a href="#" class="fd-link fd-link--inverted">Inverted link</a>
 </div>
 <br><br>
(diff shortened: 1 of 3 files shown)


stories/link/link.stories.js
```diff
@@ -34,7 +34,7 @@ export const primary = () => `
 <br><br>
 <a class="fd-link" aria-disabled="true">Disabled link</a>
 <br><br>
-<div style="background-color:#314a5e;padding:10px">
+<div class="fd-shellbar">
```
droshev, line 37: you can just use the variable in the example, instead of bringing the shellbar in the link

droshev (changes requested):
see my comment in the original ticket

stories/link/link.stories.js
Lokanathan-k: @droshev if we use the variable in the documentation in the style, can we use the same variable in NGX Documentaiton too?

InnaAtanasova: You need to use var(--sapShellColor)

InnaAtanasova: > @droshev if we use the variable in the documentation in the style, can we use the same variable in NGX Documentaiton too?

yes, you can

InnaAtanasova: Screen Shot 2021-08-30 at 11 49 57 AM

Lokanathan-k: @InnaAtanasova but issue is not on the background, its on the focus by default for the link outline-color: var(--sapContent_FocusColor, #000);outline-style: var(--sapContent_FocusStyle, dotted); but the focus has to be differentiate with the light clour like hows its done for the button.

InnaAtanasova (approved):

droshev (approved):

Issue #2668, opened 2021-08-12
Horizontal navigation latest design updates


InnaAtanasova:
Following a design review meeting with the designers it became clear that we haven't followed the correct design for horizontal navigation (the Icon Tab Bar deltas). We need to refactor the component with the links sent by Carmen.

PR #2701, opened 2021-09-02
fix: introduce Flat Navigation, refactor Horizontal Navigation

InnaAtanasova:
## Related Issue
Closes SAP/fundamental-styles#2668

## Description
In the current PR:
- introduced Flat Navigation (Part of ShellNavigation)
- refactored Horizontal Navigation to match the correct deltas from IconTabBar
- introduced a wrapper class (.fd-horizontal-navigation) for IcontTabBar and ToolHeader that applies box-shadow around the two elements.

BREAKING CHANGE:
- introduced a wrapper element with class ".fd-horizontal-navigation" for Tool Header and Navigation IconTabBar
- introduced an additional modifier class ".fd-icon-tab-bar--navigation-horizontal" to IconTabBar when used in Horizontal Navigation

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/fundamental-styles.scss
+++ b/src/fundamental-styles.scss
@@ -34,6 +34,7 @@
 @import "./form-label";
 @import "./form-message";
 @import "./helpers";
+@import "./horizontal-navigation";
 @import "./icon-tab-bar.scss";
 @import "./info-label";
 @import "./input";
--- a/src/horizontal-navigation.scss
+++ b/src/horizontal-navigation.scss
@@ -1,0 +1,9 @@
+@import './mixins';
+
+$block: fd-horizontal-navigation;
+
+.#{$block} {
+  @include fd-reset();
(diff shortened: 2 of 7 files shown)


Smiranin (approved):
I haven't noticed any issues. But It would be great if next time you will share link to requirements or examples with expected result.
Because related issue doesn't have such information.

platon-rov (approved):

InnaAtanasova:
> I haven't noticed any issues. But It would be great if next time you will share link to requirements or examples with expected result.
> Because related issue doesn't have such information.

This is an open source repo and we are not supposed to post internal for SAP links or screenshots with component specs. If anything is missing or unclear you will have to contact the developer in PM, unfortunately.

PR #2717, opened 2021-09-06
fix: layout grid bug fixes, improvements, refactoring

### LINKED ISSUE #2612: Negative margin values can result in form layout grid overflowing parent container

platon-rov:
## Related Issue

Closes #2612.

## Description

Negative margins were applied to the inappropriate elements that make rows wider than container and cause appearing the scrollbars. 

Gutter between rows was set by padding for the column not by a margin for the row.

### Before + After:

Screenshot 2021-09-09 at 21 03 43

- [X] All values are in `rem`
- [X] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/layout-grid.scss
+++ b/src/layout-grid.scss
@@ -1,13 +1,15 @@
 @import "./mixins";
 
-/*!
-.fd-col
-*/
-
-$block: #{$fd-namespace}-col;
 $blockContainer: #{$fd-namespace}-container;
 $blockRow: #{$fd-namespace}-row;
-$size: (600:"md", 1024: "lg", 1440: "xl");
+$blockCol: #{$fd-namespace}-col;
+
+$vertical-gutter: 1rem;
+$vertical-gutter-small: 0.5rem;
+$horizontal-gutter: 0.5rem;
+$horizontal-gutter-small: 0.25rem;
+
+$size: (600 :"md", 1024: "lg", 1440: "xl");
 $md: 600;
 $lg: 1024;
 $xl: 1440;
@@ -20,10 +22,10 @@ $cols: 12;
 }
 
 @mixin responsive-column($sizePrefix: null) {
-  $classPrefix: #{$block};
+  $classPrefix: #{$blockCol};
 
   @if ($sizePrefix) {
-    $classPrefix: $block + '-' + $sizePrefix;
+    $classPrefix: $blockCol + '-' + $sizePrefix;
   }
 
   @for $n from 0 through $cols {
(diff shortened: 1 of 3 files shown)


InnaAtanasova:
@platon-rov please add `before` and `after` screenshots of a component affected by the changes.

src/layout-grid.scss
```diff
@@ -52,31 +49,37 @@ $cols: 12;
...
-  margin: 0 -0.25rem 0 -0.25rem;
+  padding: 0 0.25rem;
 
-  &--no-gap {
-    margin: 0;
+  &.#{$blockContainer}--no-horizontal-gap,
+  &.#{$blockContainer}--no-gap {
+    .#{$blockRow} {
+      margin-left: 0;
+      margin-right: 0;
```
InnaAtanasova, line 69: There's a mixin for resetting the horizontal margin:
```
@mixin fd-reset-margins-x() {
  margin-left: 0;
  margin-right: 0;
}
```

src/layout-grid.scss
```diff
@@ -52,31 +49,37 @@ $cols: 12;
...
 
-    .#{$block} {
-      padding: 0;
+      .#{$blockCol} {
+        padding-left: 0;
+        padding-right: 0;
```
InnaAtanasova, line 73: Please use the mixin:
```
@mixin fd-reset-paddings-x() {
  padding-left: 0;
  padding-right: 0;
}
```

src/layout-grid.scss
```diff
@@ -85,129 +88,76 @@ $cols: 12;
...
+  &:first-child {
+    margin-top: -0.5rem;
   }
 
-  .#{$block} {
-    .#{$blockRow} + .#{$blockRow} {
-      margin-top: -0.5rem;
-    }
+  .#{$blockCol} {
+    padding-top: 0.5rem;
```
InnaAtanasova, line 103: could you please put values like `0.5rem`, `0.25rem`, etc. in variables. This way we can easily change and maintain them.

InnaAtanasova (changes requested):

InnaAtanasova (approved):

Smiranin (approved):

PR #2718, opened 2021-09-09
fix: form layout grid, remove unnecessary styles

Betrozov:
## Related Issue
Closes https://github.com/sap/fundamental-ngx/issues/6389

## Description
Removed `form layout grid` unnecessary styles

- [X] RTL support
3. Testing
- [n/a] Updated tests
4. Documentation

### DIFF
--- a/src/form-layout-grid.scss
+++ b/src/form-layout-grid.scss
@@ -136,8 +136,6 @@ $cols: 11;
 
             @include fd-rtl() {
               @include placement('left');
-
-              text-align: left;
             }
 
             margin-bottom: 0;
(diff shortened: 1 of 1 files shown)


InnaAtanasova (approved):

droshev:
@Betrozov 
having this comment
Screen Shot 2021-09-10 at 7 12 10 AM
will close the issue in ngx, but it won't fix it(until you bring the version there)

N1XUS (approved):

Smiranin (approved):

Betrozov:
> @Betrozov
> having this comment
> Screen Shot 2021-09-10 at 7 12 10 AM
> will close the issue in ngx, but it won't fix it(until you bring the version there)

Hi @droshev 
Sure it will be closed but I reopen it with a comment that needs to update fundamental-styles version

PR #2737, opened 2021-09-21
fix: table semantic column styles

platon-rov:
## Related Issues

Relates to https://github.com/SAP/fundamental-ngx/issues/6360

## Description

Styles fixed, refactored before implementing Semantic Highlighting in Fundamental-NGX for Platform Table.

### Before:

image

Screenshot 2021-09-21 at 18 19 20

- [X] All values are in `rem`
- [X] RTL support
3. Testing
- [X] Updated tests
4. Documentation

### DIFF
--- a/src/table.scss
+++ b/src/table.scss
@@ -246,6 +246,24 @@ $block: #{$fd-namespace}-table;
       @include fd-table-reset-spacing();
     }
 
+    &--fixed {
+      left: 0;
+      position: absolute;
+      background-color: inherit;
+      border-bottom: $fd-table-border;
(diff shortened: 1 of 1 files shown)


src/table.scss
```diff
@@ -246,6 +246,24 @@ $block: #{$fd-namespace}-table;
       @include fd-table-reset-spacing();
     }
 
+    &--fixed {
+      left: 0;
```
InnaAtanasova, line 250: Just FYI, there's a mixin for setting the positions and it handles the RTL :)

InnaAtanasova (approved):

PR #2738, opened 2021-09-22
Feat/fn new

droshev:
## Related Issue

## Description

- [ ] All values are in `rem`
- [ ] Text elements follow the truncation rules
- [ ] hover state of the element follow design spec
- [ ] focus state of the element follow design spec
- [ ] active state of the element follow design spec
- [ ] selected state of the element follow design spec
- [ ] selected hover state of the element follow design spec
- [ ] pressed state of the element follow design spec
- [ ] Responsiveness rules - the component has modifier classes for all breakpoints
- [ ] RTL support
- [ ] only one top level `fd-*` class is used in the file
- [ ] BEM naming convention is used
- [ ] Mixins are used for repeatable code (`fd-rtl`, `fd-ellipsis`, `fd-flex`, `fd-selected`, `fd-focus`, ect.)
- [ ] A11y support - keyboard support, screenreader support, proper ARIA attributes, etc.
- [ ] `fd-reset()` mixin is applied to all elements
- [ ] Variables are used, if some value is used more than twice.
- [ ] Checked if current components can be reused, instead of having new code.
3. Testing
- [ ] Updated tests
4. Documentation
- [ ] Storybook documentation has been created/updated
- [ ] Breaking Changes [wiki](https://github.com/SAP/fundamental-styles/wiki/Breaking-Changes) has been updated in case of breaking changes.


InnaAtanasova (approved):

Smiranin (approved):

.storybook/custom/themeProvider.js
```diff
@@ -37,7 +37,12 @@ export const changeDocumentTheme = (newTheme, forComponents) => {
     };
 
     forComponents.forEach(component => {
-        let stylePath = `${component}-${newTheme}.css`;
+        let stylePath = `${component}`;
```
droshev, line 40: this is in case FN doesn't have other themes

PR #2747, opened 2021-09-24
fix: add support for multiple icons in table header [ci visual]

dmitry-stepanenko:
## Related Issue
Related to SAP/fundamental-ngx#6546

## Description
Added styles to handle multiple icons in the table column header

### Before:
image

### After:
image

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [n/a] Updated tests
4. Documentation

### DIFF
--- a/src/table.scss
+++ b/src/table.scss
@@ -418,6 +418,24 @@ $block: #{$fd-namespace}-table;
     justify-content: space-between;
   }
 
+  &__column-header-icons {
+    display: flex;
+
+    .#{$block}__icon:not(:last-child) {
+      margin-right: 0.25rem;
+    }
+
+    @include fd-rtl() {
+      .#{$block}__icon {
+        margin-right: 0;
+      }
+
+      .#{$block}__icon:not(:last-child) {
+        margin-left: 0.25rem;
+      }
+    }
+  }
+
   &__caption {
     @include fd-screen-reader-only();
   }


src/table.scss
```diff
@@ -418,6 +418,24 @@ $block: #{$fd-namespace}-table;
     justify-content: space-between;
   }
 
+  &__column-header-icons {
+    display: flex;
+
+    fd-icon:not(:last-child) {
```
InnaAtanasova, line 424: I believe you are targeting an Angular component `fd-icon`. This can't be done in Fund-styles. If the css is specific for the Angular implementation it should be in fund-ngx.

src/table.scss
```diff
@@ -418,6 +418,24 @@ $block: #{$fd-namespace}-table;
...
+      margin-right: 0.25rem;
```
InnaAtanasova, line 425: Could point to the specs from where you took this value?

InnaAtanasova (changes requested):

src/table.scss
dmitry-stepanenko: Inna, this spacing is not documented anywhere, I've applied the one that I see fit: icons look too stacked, if no margin is used.

Interestingly, in [SAPUI5 SDK](https://sapui5.hana.ondemand.com/#/entity/sap.ui.comp.smarttable.SmartTable/sample/sap.ui.comp.sample.smarttable) icons are separated with space 
image

Should I just keep it as-is or what will be the best here?

src/table.scss
dmitry-stepanenko: fixed

src/table.scss
mikerodonnell89: I am also not seeing this documented in the specs so I think it is fine to make a general assumption until we hear otherwise

InnaAtanasova (approved):

mikerodonnell89 (approved):

PR #2774, opened 2021-10-06
fix: dialog resize handle alignment [ci visual] (#2773)

Betrozov:
## Related Issue
Closes https://github.com/SAP/fundamental-ngx/issues/6332

## Description
Fixed alignment of the resize handle 

### Before:

before

### After:
after

- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/dialog.scss
+++ b/src/dialog.scss
@@ -159,15 +159,15 @@ $button: #{$fd-namespace}-button;
     @include fd-reset();
 
     position: absolute;
-    right: 0;
-    bottom: -4px;
+    right: 1px;
+    bottom: -5px;
     font-size: 1rem;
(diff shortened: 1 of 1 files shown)


src/dialog.scss
```diff
@@ -159,15 +159,15 @@ $button: #{$fd-namespace}-button;
     @include fd-reset();
 
     position: absolute;
-    right: 0;
-    bottom: -4px;
+    right: 1px;
```
InnaAtanasova, line 162: Screen Shot 2021-10-06 at 3 07 51 PM

Please use mixin for setting absolute position, it will handle RTL and you won't need to put the value in a variable.

Issue #2726, opened 2021-09-15
Style cleanup for Table Component


aberikashvili:
It looks like the technical debt with hardcoded styling in ngx has augmented. We need to address this asap.
There are few options:

1. the hardcoded styles in ngx has been already address in fundamental-styles repo
2. the hardcoded styles in ngx is not reflected in fundamental-styles repo
In case of the first option then the developer needs to remove the hardcoded styling; use the latest from styles; test and open a PR
In case of the second, then the developer needs to bring it there, open a PR, test it, merge it, then do the steps above.

https://github.com/SAP/fundamental-ngx/issues/5025

droshev:
@nickiovenko any update here?

nickiovenko:
@droshev in the corresponding PR Arkadii made the edits but didn't have a chance to check them due to the high load

PR #2876, opened 2021-10-31
fix(styles): Style cleanup for Table Component

aberikashvili:
## Related Issue
Closes SAP/fundamental-styles#2726

## Description
This PR delivers, styling cleanup changes from fundamental-ngx repo

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [N/A] Updated tests
4. Documentation


droshev (changes requested):
@aberikashvili do we have examples showing those changes?

src/styles/table.scss
```diff
@@ -11,6 +11,9 @@
 */
 $block: #{$fd-namespace}-table;
 
+$fd-table-border: 0.0625rem solid #e4e4e4;
```
InnaAtanasova, line 14: You don't need this part, it's a fallback generated with a script to make sure variables work in all browsers. 
All you need is:
```
$fd-table-border: var(--sapList_BorderWidth) solid var(--sapList_BorderColor);
```

src/styles/table.scss
```diff
@@ -11,6 +11,9 @@
...
+$fd-table-border-var: var(--sapList_BorderWidth, 0.0625rem) solid var(--sapList_BorderColor, #e4e4e4);
```
InnaAtanasova, line 15: see comment above

src/styles/table.scss
```diff
@@ -137,6 +140,39 @@ $block: #{$fd-namespace}-table;
 
   /****** Elements ******/
 
+  &__empty {
+    &-table-message {
+      padding: 1rem;
+      text-align: center;
+      display: flex;
+      align-items: center;
+      justify-content: center;
```
InnaAtanasova, line 149: There's a mixin you can use for this block of code. Please check `fd-flex-center` mixin.

src/styles/table.scss
```diff
@@ -137,6 +140,39 @@ $block: #{$fd-namespace}-table;
...
+      align-content: center;
+      width: 100%;
+    }
+
+    &:not(.fixed-height) .fd-table__row:last-of-type {
```
InnaAtanasova, line 154: I don't know where the class `.fixed-height` is coming from but it doesn't have the correct naming. It should follow BEM and be prefixed with `fd-`

src/styles/table.scss
```diff
@@ -137,6 +140,39 @@ $block: #{$fd-namespace}-table;
...
+      .fd-table__cell {
```
InnaAtanasova, line 155: You can't use `.fd-tabele__something`, you need to use `.#{$block}__something`.

src/styles/table.scss
```diff
@@ -137,6 +140,39 @@ $block: #{$fd-namespace}-table;
...
+        border-bottom: 0;
+      }
+    }
+
+    .fd-table__body {
+      border-left: none;
+
+      @include fd-rtl() {
+        border-left: $fd-table-border;
+        border-left: $fd-table-border-var;
```
InnaAtanasova, line 165: The second line is not needed.

src/styles/table.scss
```diff
@@ -137,6 +140,39 @@ $block: #{$fd-namespace}-table;
...
+        border-right: none;
+      }
+    }
+
+    &.fixed-height {
+      background-color: #fff;
+      background-color: var(--sapList_Background, #fff);
```
InnaAtanasova, line 172: First line is not needed, the second doesn't need the fallback. You can just have: background-color: var(--sapList_Background);

src/styles/table.scss
```diff
@@ -240,6 +276,26 @@ $block: #{$fd-namespace}-table;
       @include fd-table-reset-spacing();
     }
 
+    &__group {
+      min-width: 100%;
+      border-left: none;
+
+      /**
+        Due to the changed table-layout group row cell may looe it's top border.
```
InnaAtanasova, line 284: Typo in the word `looe`

src/styles/table.scss
```diff
@@ -240,6 +276,26 @@ $block: #{$fd-namespace}-table;
...
+        For example when previous (relatively to the group row) row columns tolal width < 100% of the table.
+      */
+      &::before {
+        content: '';
+        display: block;
+        position: absolute;
+        top: -0.0625rem;
+        left: 0;
+        right: 0;
```
InnaAtanasova, line 293: Shouldn't you handle RTL?

src/styles/table.scss
```diff
@@ -240,6 +276,26 @@ $block: #{$fd-namespace}-table;
...
+        border-top: $fd-table-border;
+        border-top: $fd-table-border-var;
```
InnaAtanasova, line 295: same comment as above

InnaAtanasova (changes requested):

src/styles/table.scss
aberikashvili: done

src/styles/table.scss
aberikashvili: done

src/styles/table.scss
aberikashvili: done

src/styles/table.scss
aberikashvili: done

src/styles/table.scss
aberikashvili: done

src/styles/table.scss
aberikashvili: done

src/styles/table.scss
aberikashvili: done

src/styles/table.scss
aberikashvili: done

src/styles/table.scss
aberikashvili: what to handle here? `left` and `right` both are `0`

src/styles/table.scss
```diff
@@ -137,6 +139,38 @@ $block: #{$fd-namespace}-table;
...
+      padding: 1rem;
+      text-align: center;
+      width: 100%;
+
+      @include fd-flex-center() {
+        align-content: center;
+      }
+    }
+
+    &:not(.fd-fixed-height) .fd-table__row:last-of-type {
```
platon-rov, line 153: `.fd-table__row` => `.#{$block}__row`

src/styles/table.scss
```diff
@@ -137,6 +139,38 @@ $block: #{$fd-namespace}-table;
...
+      .#{block}__cell {
+        border-bottom: 0;
+      }
+    }
+
+    .fd-table__body {
```
platon-rov, line 159: `.fd-table__body` => `.#{$block}__body`

src/styles/table.scss
platon-rov: I think as it relates to the table's empty message item it shouldn't be a separate class, let it be modifier for base class, what do you think?

src/styles/table.scss
aberikashvili: Done

src/styles/table.scss
aberikashvili: Done

aberikashvili:
> @aberikashvili do we have examples showing those changes?

@droshev I added empty table example, please check again

src/styles/table.scss
platon-rov: Bump for this one, `.fd-fixed-height` not the correct class name.

platon-rov (commented):
Good job, please resolve all conversations :)

PR #2877, opened 2021-11-01
fix: static dialog nobackdrop position fix [ci-visual]

dmitry-stepanenko:
## Related Issue
Is being done as part of the fix for https://github.com/SAP/fundamental-ngx/issues/5286

## Description
Added styles for dialog without backdrop

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [n/a] Updated tests
4. Documentation

### DIFF
--- a/src/styles/dialog.scss
+++ b/src/styles/dialog.scss
@@ -175,4 +175,10 @@ $button: #{$fd-namespace}-button;
       color: $fd-dialog-resize-handle-color;
     }
   }
+
+  &--no-backdrop:not(.#{$block}--targeted) {
+    @include fd-position-center($position: fixed);
+
+    z-index: 1000;
+  }
 }
--- a/src/styles/flexible-column-layout.scss
+++ b/src/styles/flexible-column-layout.scss
@@ -58,7 +58,7 @@ $block: #{$fd-namespace}-flexible-column-layout;
     padding: 0;
     z-index: 2;
 
-    @include fd-absolute-center();
+    @include fd-position-center();
   }
 
   &__column {
--- a/src/styles/mixins/_mixins.scss
+++ b/src/styles/mixins/_mixins.scss
@@ -365,8 +365,8 @@
   transform: translateX(-50%);
 }
 
-@mixin fd-absolute-center {
-  position: absolute;
+@mixin fd-position-center($position: absolute) {
+  position: $position;
   left: 50%;
   top: 50%;
   transform: translate(-50%, -50%);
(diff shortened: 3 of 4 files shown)


dmitry-stepanenko:
@InnaAtanasova I'll add breaking changes once PR is approved

src/styles/dialog.scss
```diff
@@ -175,4 +175,12 @@ $button: #{$fd-namespace}-button;
       color: $fd-dialog-resize-handle-color;
     }
   }
+
+  &__no-backdrop:not(.#{$block}__targeted) {
```
InnaAtanasova, line 179: Is it a modifier class or an element? Because from the naming it looks like you are introducing a new html element with the class `.fd-dialog__no-backdrop`. If it's a modifier class, added on the same level as `.fd-dialog`, then you should have a different naming, like `.fd-dialog--no-backdrop`.

src/styles/dialog.scss
```diff
@@ -175,4 +175,12 @@ $button: #{$fd-namespace}-button;
...
+    position: fixed;
```
InnaAtanasova, line 180: why the position is fixed and not absolute?

InnaAtanasova: If it's absolute you can use the mixin we already have for this:
```
@mixin fd-absolute-center {
  position: absolute;
  left: 50%;
  top: 50%;
  transform: translate(-50%, -50%);
}
```

dmitry-stepanenko: Well, I decided to keep fixed because it centers the element in the view, not in the parent. Although in fundamental-ngx both fixed and absolute will work (because parent `fd-dialog-container` is attached directly to body), it feels reasonable to provide unified solution, that works well in all scenarios

src/styles/dialog.scss
dmitry-stepanenko: done

dmitry-stepanenko:
> > @InnaAtanasova I'll add breaking changes once PR is approved
> 
> Usually you add the Breaking changes before opening the PR as the reviewer's job is to make sure they are added. If you get the approval nothing stops you from merging the PR and there's no guarantee that you will add the Breaking changes.

Updated the wiki

src/styles/dialog.scss
InnaAtanasova: Awesome :) In this case please adjust the `fd-absolute-center` mixin to accept as parameter the position and reuse it in your code. 
You can do something like:
```
@mixin fd-absolute-center($position: absolute) {
  position: $position; 
  left: 50%;
  top: 50%; 
  transform: translate(-50%, -50%); 
}
```

dmitry-stepanenko: sure, done

mikerodonnell89 (approved):
Works well 👍

InnaAtanasova (approved):

InnaAtanasova:
@dmitry-stepanenko you need to trigger the visual checks

InnaAtanasova:
🚀

PR #2878, opened 2021-11-01
fix(styles): add code inline case

droshev:
add code inline case

### DIFF
--- a/stories/code/code.stories.js
+++ b/stories/code/code.stories.js
@@ -1,7 +1,7 @@
 export default {
     title: 'Components/Code',
     parameters: {
-        description: 'A component for styling code',
+        description: 'A component for styling code snippet',
         tags: ['development'],
         components: ['code']
     }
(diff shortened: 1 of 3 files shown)


InnaAtanasova:
Screen Shot 2021-11-01 at 2 07 00 PM

Out of the scope of this PR but why are these values in variables? We put in variables values that are repeated 2+ times. In this case all values are used only once.

stories/code/code.stories.js
```diff
@@ -24,13 +24,21 @@ spec:
...
 
-primary.storyName = 'Code';
+primary.storyName = 'Code Block';
 primary.parameters = {
     docs: {
         storyDescription: 'Code example'
     }
 };
+
+export const inlineCode = () => 'Loren ipsum: <code class="fd-code fd-code--inline">FundamentalExtension</code>';
```
InnaAtanasova, line 38: Why do you need an arrow function here?

stories/code/code.stories.js
```diff
@@ -24,13 +24,21 @@ spec:
...
-    extensionManagerServiceUrl: https://extension-manager.{{ .Values.landscapeDomain }}/graphql
+    extensionManagerServiceUrl: https://extension-manager-name.{{ .Values.landscapeDomainName }}/graphql
 </code></pre>
 `;
 
-primary.storyName = 'Code';
+primary.storyName = 'Code Block';
 primary.parameters = {
     docs: {
         storyDescription: 'Code example'
```
InnaAtanasova, line 34: Maybe add more info here. It doesn't become clear what's the purpose of this modifier class.

InnaAtanasova (approved):

PR #2881, opened 2021-11-02
fix(styles): checkbox broken outline voiceover [ci visual]

genereg1:
## Related Issue
as part of https://github.com/SAP/fundamental-ngx/issues/5791
## Description
set `border-style: initial` to prevent broken outine when it selects checkbox via VoiceOver

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [n/a] Updated tests
4. Documentation

### DIFF
--- a/src/styles/checkbox.scss
+++ b/src/styles/checkbox.scss
@@ -48,6 +48,15 @@ $block: #{$fd-namespace}-checkbox;
 
   @extend %fd-checkbox-input-hidden;
 
+  border-style: initial;
+
+  &.is-warning,
+  &.is-error,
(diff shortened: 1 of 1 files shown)


Betrozov (approved):

N1XUS (approved):

InnaAtanasova:
@genereg1 please rebase main. I see in UI tests 6 files being affected by your change, which shouldn't be the case. Usually a rebase fixes this.
Screen Shot 2021-11-05 at 12 03 20 PM

mikerodonnell89 (approved):
lgtm

PR #2886, opened 2021-11-04
feat: themeable scrollbar

platon-rov:
## Related Issue

Part of DXP Issue 446

## Description

Theme-able scrollbar component.

## Screenshots

Screenshot 2021-11-04 at 19 58 20

image

- [X] All values are in `rem`
- [X] RTL support
3. Testing
- [X] Updated tests
4. Documentation


N1XUS (approved):

src/styles/theming/sap_fiori_3.scss
```diff
@@ -263,4 +263,10 @@
 
   /* Micro Process Flow */
   --fdMicro_Process_Flow_Icon_Color: var(--sapContent_ContrastIconColor);
+
+  /* Scrollbar */
+  --fdScrollbarTrackColor: var(--sapScrollBar_TrackColor);
+  --fdScrollbarThumbColor: var(--sapScrollBar_FaceColor);
+  --fdScrollbarThumbHoverColor: var(--sapScrollBar_Hover_FaceColor);
+  --fdScrollbarDimension: var(--sapScrollBar_Dimension);
```
InnaAtanasova, line 271: That's not the proper naming we have adopted. Please use:
--fdScrollbar_Track_Color: var(--sapScrollBar_TrackColor);
--fdScrollbar_Thumb_Color: var(--sapScrollBar_FaceColor);
--fdScrollbar_Thumb_Hover_Color: var(--sapScrollBar_Hover_FaceColor);
--fdScrollbar_Dimension: var(--sapScrollBar_Dimension);

src/styles/helpers/_scrollbar.scss
```diff
@@ -0,0 +1,30 @@
+@mixin fd-themed-scrollbar() {
```
InnaAtanasova, line 1: Please move this mixin in the `scrollbar.scss` file. Let's see in future if it will be reused, then we can move it to a separate file, but for now there's no value of keeping it in a separate file.

src/styles/scrollbar.scss
```diff
@@ -0,0 +1,9 @@
+@import "mixins/mixins";
+@import "helpers/scrollbar";
+
+$block: fd-scrollbar;
+
+.#{$block} {
+  @include fd-reset();
+  @include fd-themed-scrollbar();
```
InnaAtanasova, line 8: Please bring directly the code here, no need to have it as a mixin.

stories/scrollbar/scrollbar.stories.js
```diff
@@ -0,0 +1,24 @@
+export default {
```
droshev, line 1: I think it would be nice to explain why we have this component and when to use it. I guess it doesn't have to be mandatory?

platon-rov: Updated the docs, please take a look.

InnaAtanasova:
A small issue with the documentation, the description looks like this:
Screen Shot 2021-11-08 at 1 54 37 PM

InnaAtanasova:
UI Reviews and Tests are failing.

InnaAtanasova (approved):

mikerodonnell89 (approved):
LGTM

Issue #2902, opened 2021-11-08
Progress Indicator - implementation


InnaAtanasova:
Implement the component

PR #2903, opened 2021-11-09
feat: introduce Progress Indicator

genereg1:
## Related Issue
Closes SAP/fundamental-styles#2902

## Description
Adds new Fiori 3 component

**IMPORTANT NOTE**

The button inside the popover body is represented by a span as the a11y checks fail with the following error:

Screen Shot 2021-11-10 at 6 31 56 PM

Screen Shot 2021-11-10 at 6 51 18 PM

In the implementation libraries we will use a standard <button> tag and control the focus and the screen reader programatically.

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation


src/styles/progress-indicator.scss
```diff
@@ -0,0 +1,144 @@
...
+
+$fd-progress-indicator-states: (
+  "informative": ("background": var(--fdProgress_Indicator_Background_Color_Informative), "icon": var(--fdProgress_Icon_Informative)),
+  "positive": ("background": var(--fdProgress_Indicator_Background_Color_Positive), "icon": var(--fdProgress_Icon_Positive)),
+  "negative": ("background": var(--fdProgress_Indicator_Background_Color_Negative), "icon": var(--fdProgress_Icon_Negative)),
+  "critical": ("background": var(--fdProgress_Indicator_Background_Color_Critical), "icon": var(--fdProgress_Icon_Critical))
+);
+
+.#{$block} {
+  width: 100%;
```
platon-rov, line 17: `@include fd-reset()` please.

src/styles/theming/sap_fiori_3_dark.scss
```diff
@@ -268,4 +268,18 @@
...
+  /* Progress Indicator */
+  --fdProgress_Indicator_Background_Color_Neutral: var(--sapNeutralElementColor);
+  --fdProgress_Indicator_Background_Color_Informative: var(--sapInformativeElementColor);
+  --fdProgress_Indicator_Background_Color_Positive: var(--sapPositiveElementColor);
+  --fdProgress_Indicator_Background_Color_Negative: var(--sapNegativeElementColor);
+  --fdProgress_Indicator_Background_Color_Critical: var(--sapCriticalElementColor);
+  --fdProgress_Indicator_Border_Right: none;
+  --fdProgress_Icon_Offset_Right: 0;
+  --fdProgress_Icon_Offset_Left: 0;
+  --fdProgress_Icon_Informative: '';
```
platon-rov, line 281: In `sap_fiori_3_hcb.scss` icons are set and here is not, is it intended?

platon-rov (approved):

src/styles/progress-indicator.scss
InnaAtanasova: It should not have a reset as progress indicator in all cases have truncation behaviour which is handled by a popover and in this case the markup is:
```
fd-progress-indicator fd-popover
```

If we bring a reset it will overwrite the rules coming from `fd-popover`.

src/styles/theming/sap_fiori_3_dark.scss
InnaAtanasova: Yes. In light and dark themes the icons have colors which correspond to the state, but in HCB and HCW the we don't have state colors and the state is represented by an icon. That's why this icon is not shown in dark and light theme, but only visible in the contrast themes.

InnaAtanasova: Screen Shot 2021-11-15 at 1 45 31 PM
Screen Shot 2021-11-15 at 1 45 51 PM

mikerodonnell89 (changes requested):
Looks good! I'm not seeing details in the specs but I am surprised to see that the popover close button does not have hover / active styles, I would think it would follow the rest of the buttons.

Also, why do these divs need tabindex=-1? Aren't they already not tab indexed? Also wondering if removing this will allow us to use a button element for the close button.

InnaAtanasova:
> Looks good! I'm not seeing details in the specs but I am surprised to see that the popover close button does not have hover / active styles, I would think it would follow the rest of the buttons.
> 
> Also, why do these divs need tabindex=-1? Aren't they already not tab indexed? Also wondering if removing this will allow us to use a button element for the close button.

1. There aren't any states for the buttons in UI5, too. You can check here: https://sapui5.hana.ondemand.com/#/entity/sap.m.ProgressIndicator/sample/sap.m.sample.ProgressIndicator

You will need to go to mobile mode on your browser and click on the label with very long text

2. The tabindex is added as the div has a role of progressbar. By Fiori requirements the focus should go on the close button.

mikerodonnell89 (approved):

PR #2904, opened 2021-11-09
fix(styles): horizontal scroll remove in carousel [ci visual]

OleksiyDrozdov:
## Related Issue
part of [#5765](https://github.com/SAP/fundamental-ngx/issues/5765) 

## Description

### Before:
image
### After:
image

- [x] All values are in `rem`
- [n/a] RTL support
3. Testing
- [n/a] Updated tests
4. Documentation

### DIFF
--- a/src/styles/carousel.scss
+++ b/src/styles/carousel.scss
@@ -5,8 +5,10 @@
 $block: #{$fd-namespace}-carousel;
 $button: #{$fd-namespace}-button;
 
-#{$block}:focus {
-  outline: 0.0625rem dotted;
+#{$block} {
+  @include fd-focus() {
+    outline: 0.0625rem dotted;
+  }
 }
 
 .#{$block} {
@@ -34,7 +36,7 @@ $button: #{$fd-namespace}-button;
   }
 
   width: 100%;
-  min-width: 15.5rem;
+  min-width: 11rem;
   max-width: 100%;
   height: 100%;
 
@@ -120,8 +122,8 @@ $button: #{$fd-namespace}-button;
       flex-wrap: wrap;
     }
 
-    width: 9rem;
-    padding: 0.5rem;
+    min-width: 7rem;
+    padding: 0.5rem 0;
   }
 
   &__page-indicator {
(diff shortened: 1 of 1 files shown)


src/styles/carousel.scss
```diff
@@ -5,6 +5,10 @@
 $block: #{$fd-namespace}-carousel;
 $button: #{$fd-namespace}-button;
 
+#{$block}:focus {
+  outline: 0.0625rem dotted;
```
platon-rov, line 9: We have mixin for it, please use it instead.

src/styles/carousel.scss
```diff
@@ -201,3 +205,18 @@ $button: #{$fd-namespace}-button;
     }
   }
 }
+
+@media (max-width: 21.25rem) {
```
platon-rov, line 209: The `21.25rem` sounds like a magical value, fiori3 specs say nothing about this breakpoint, the smallest one is `< 600px`. 

There is nothing about changing paddings or anything at some breakpoint in wiki https://wiki.wdf.sap.corp/wiki/pages/viewpage.action?pageId=2059230114. 

The container is `15.5rem` in width what is `248px` and it's okay for the `320px` screen and there is no sense to support smaller devices. 

Instead, I propose at `< 600px` breakpoint hide the items' dots and only leave prev/next buttons, it's also not covered in fiori3 specs but i think it will be more suitable approach, @InnaAtanasova @droshev what do you think?

InnaAtanasova (changes requested):
There's no media query for Carousel component. Please check the UX guidelines on how the component should behave on mobile mode:
https://experience.sap.com/internal/fiori-design-web/carousel/#responsiveness
Screen Shot 2021-11-15 at 3 07 57 PM

src/styles/carousel.scss
InnaAtanasova: we can't propose or come up with media queries if they are not specified in Fiori :)
Here's information about what should be hidden/visible on mobile: https://experience.sap.com/internal/fiori-design-web/carousel/#responsiveness

src/styles/carousel.scss
```diff
@@ -5,6 +5,12 @@
 $block: #{$fd-namespace}-carousel;
 $button: #{$fd-namespace}-button;
 
+#{$block} {
+  @include fd-focus() {
```
platon-rov, line 9: We have `fd-focus` mixin for this purpose.

PR #2910, opened 2021-11-15
fix: select min-width

platon-rov:
## Description

Select text label `min-width` fix due to the Fiori 3 specifications.

Screenshot 2021-11-15 at 20 10 08

### Before:

image

### After:

image

### DIFF
--- a/src/styles/select.scss
+++ b/src/styles/select.scss
@@ -131,7 +131,7 @@ $outline-offset: 0.1875rem;
 
     font-size: var(--sapFontSize);
     padding-left: $fd-select-padding-x;
-    min-width: 5.625rem;
+    min-width: 2.125rem;
 
     @include fd-rtl() {
       padding-left: 0;


mikerodonnell89 (approved):

src/styles/select.scss
```diff
@@ -131,7 +131,7 @@ $outline-offset: 0.1875rem;
 
     font-size: var(--sapFontSize);
     padding-left: $fd-select-padding-x;
-    min-width: 5.625rem;
+    min-width: 2.125rem;
```
InnaAtanasova, line 134: how did you come up with this value? the specs says 5rem min-width.

platon-rov: I guessed that `5rem` it's the `min-width` for the whole select. I came up with this value by calculating `5rem - border - button width`. So the with of the select is `5rem`\`80px` as shown on the screenshot.

InnaAtanasova: We need to ask designers about this. I don't think we can guess on this. I will send an email and cc you.

platon-rov: Okay, thanks. **[BLOCKED]** till getting the answer from designers.

PR #2911, opened 2021-11-16
fix(fn): inconsistent text field borders across zoom levels [ci visual]

darksmo:
## Description
On some operating systems supporting sub-pixel rendering the browser may
inconsistently approximate the height/widths of boxes across zoom level. The
issue becomes apparent when multiple controls are placed together (e.g.,
table rows, tags, input boxes) as the user zooms in/out.

Specifying the border-width property is a solution that does not involve
using future CSS specs or GPU-based calculations.

## Screenshots

Zoom level 80% - but the same issue occurs across multiple zoom levels.

### Before:

image

### After:

image

- [na] All values are in `rem`
- [x] RTL support
3. Testing
- [na] Updated tests
4. Documentation

### DIFF
--- a/src/fn/fn-input.scss
+++ b/src/fn/fn-input.scss
@@ -17,7 +17,7 @@ $block: #{$fn-namespace}-text-field;
 
   & + .#{$block}__input-border {
     &::before {
-      background: $border;
+      border-bottom-color: $border;
     }
   }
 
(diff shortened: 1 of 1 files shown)


InnaAtanasova (approved):

darksmo:
So, who is going to approve the UI review?

PR #2912, opened 2021-11-16
fix(styles): pagination styles update

platon-rov:
## Related Issue

Part of https://github.com/SAP/fundamental-ngx/issues/7172

## Description

Pagination components styles update due to changed specifications.

## Screenshots

Please check the pagination component examples. There are too many changes to place here.

https://deploy-preview-2912--fundamental-styles.netlify.app/?path=/docs/components-pagination--first-page

### Breaking changes

https://github.com/SAP/fundamental-styles/wiki/Breaking-Changes#pagination-2912

### DIFF
--- a/src/styles/pagination.scss
+++ b/src/styles/pagination.scss
@@ -27,25 +27,27 @@ $block: #{$fd-namespace}-pagination;
 
 .#{$block} {
   @include fd-reset();
-  @include fd-flex-between();
 
-  font-size: var(--sapFontSize);
+  @include fd-flex-vertical-center() {
+    justify-content: space-between;
+  }
+
   padding: 0 1rem;
   height: 2.75rem;
   width: 100%;
 
-  .#{$block}__per-page {
+  &__per-page {
     @include fd-reset();
     @include fd-flex-center();
   }
 
-  .#{$block}__nav {
+  &__nav {
     @include fd-reset();
     @include fd-flex-center();
-    @include fd-set-margins-x('auto', 'auto');
+    @include fd-set-margins-x-equal('auto');
   }
 
-  .#{$block}__more {
+  &__more {
     @include fd-reset();
 
     &::before {
@@ -84,8 +86,16 @@ $block: #{$fd-namespace}-pagination;
     display: none;
   }
 
-  .#{$block}__button--mobile {
-    display: none;
+  .#{$block}__button {
+    [class*='sap-icon'] {
+      @include fd-rtl() {
+        transform: rotate(180deg);
+      }
+    }
+
+    &--mobile {
+      display: none;
+    }
   }
 
   &--mobile {
--- a/src/styles/pagination.scss
+++ b/src/styles/pagination.scss
@@ -8,93 +8,84 @@
 $block: #{$fd-namespace}-pagination;
 
 .#{$block} {
-
-  $fd-pagination-link-padding-x: 0.25rem;
-  $fd-pagination-padding-x: 0.5rem;
-
   @include fd-reset();
+  @include fd-flex-between();
 
   font-size: var(--sapFontSize);
+  padding: 0 1rem;
+  height: 2.75rem;
+  width: 100%;
 
-  &__total {
+  .#{$block}__per-page {
     @include fd-reset();
-
-    font-size: var(--sapFontSize);
-    color: var(--sapContent_LabelColor);
-    margin-right: $fd-pagination-padding-x;
-
-    @include fd-rtl() {
-      margin-right: 0;
-      margin-left: $fd-pagination-padding-x;
-    }
+    @include fd-flex-center();
   }
 
-  &__nav {
+  .#{$block}__nav {
     @include fd-reset();
-
-    display: inline-block;
-    padding-left: 0;
-    margin-left: 0;
-    list-style: none;
-    margin-bottom: 0;
+    @include fd-flex-center();
+    @include fd-set-margins-x('auto', 'auto');
   }
 
-  &__more {
+  .#{$block}__more {
     @include fd-reset();
 
     &::before {
       content: "...";
-      color: var(--sapLinkColor);
-      font-weight: bold;
+      display: block;
+      width: 1.5rem;
+      text-align: center;
     }
   }
 
-  &__link {
-    @include fd-reset();
-    @include fd-link();
+  .#{$block}__input {
+    min-width: 2.5rem;
+    text-align: center;
 
-    font-size: var(--sapFontSize);
-    background-color: transparent;
-    padding-left: $fd-pagination-link-padding-x;
-    padding-right: $fd-pagination-link-padding-x;
-    position: relative;
-    text-decoration: none;
+    /** Hide arrows so the text will be centered in the input. */
 
-    @include fd-selected() {
-      color: var(--sapContent_LabelColor);
-      text-decoration: none;
-      cursor: default;
-      font-weight: bold;
+    /** Firefox */
+    -moz-appearance: textfield;
+
+    /** WebKit/Blink */
+    &::-webkit-outer-spin-button,
+    &::-webkit-inner-spin-button {
+      -webkit-appearance: none;
+      margin: 0;
     }
 
-    &--previous,
-    &--next {
-      @include fd-reset();
-      @include fd-link();
+    &-label {
+      display: none;
+    }
+  }
 
-      font-size: var(--sapFontSize);
-      padding-left: $fd-pagination-padding-x;
-      padding-right: $fd-pagination-padding-x;
+  .#{$block}__input-label,
+  .#{$block}__per-page-label {
+    flex-shrink: 0;
+    align-self: center;
+  }
+
+  .#{$block}__button--mobile {
+    display: none;
+  }
+}
+
+@media (max-width: 599px) {
+  .#{$block} {
+    .#{$block}__per-page + .#{$block}__nav {
+      @include fd-set-margins-x('auto', 0);
     }
 
-    &--previous {
-      @include fd-icon("navigation-left-arrow");
-
-      font-size: var(--sapFontSmallSize);
-
-      @include fd-rtl() {
-        @include fd-icon("navigation-right-arrow");
-      }
+    .#{$block}__total,
+    .#{$block}__per-page-label,
+    .#{$block}__link,
+    .#{$block}__more {
+      display: none;
     }
 
-    &--next {
-      @include fd-icon("navigation-right-arrow");
-
-      font-size: var(--sapFontSmallSize);
-
-      @include fd-rtl() {
-        @include fd-icon("navigation-left-arrow");
-      }
+    .#{$block}__button--mobile,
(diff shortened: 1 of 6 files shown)


InnaAtanasova:
Screen Shot 2021-11-17 at 11 47 56 AM
Screen Shot 2021-11-17 at 11 48 06 AM

Why is the width of the input different in these cases?

InnaAtanasova:
Screen Shot 2021-11-17 at 11 52 24 AM

The space between 2 and the ... is different from the space between the ... and 500, aka the dots are not in the middle of the page numbers.

InnaAtanasova:
Why does the input have such big width?
Screen Shot 2021-11-17 at 11 54 44 AM

I think you need to use: `width: auto`, not `width:100%`. This will result in this
Screen Shot 2021-11-17 at 11 54 29 AM
:

InnaAtanasova:
You have Cozy and Mobile modes but the specs says Desktop and Tablet are the same. 
Fundamentals:
Screen Shot 2021-11-17 at 11 58 08 AM

Specs:
Screen Shot 2021-11-17 at 11 58 16 AM

stories/pagination/pagination.stories.js
```diff
@@ -18,30 +18,60 @@ Element | Modifier/class | Description
...
-  <nav class="fd-pagination__nav">
-      <a href="#" class="fd-pagination__link fd-pagination__link--previous" aria-label="Previous"aria-disabled="true"></a>
-      <a href="#" class="fd-pagination__link is-selected" >1</a>
-      <a href="#" class="fd-pagination__link">2</a>
-      <a href="#" class="fd-pagination__link">3</a>
-      <a href="#" class="fd-pagination__link fd-pagination__link--next" aria-label="Next"
-      aria-disabled="false"></a>
-  </nav>
+export const firstPage = () => `<style>*[dir='rtl'] .fd-pagination__button { transform: rotate(180deg) !important; }</style>
+<div class='fd-pagination'>
```
InnaAtanasova, line 42: The `dir` should not be here, it should be handles in the css code

stories/pagination/pagination.stories.js
```diff
@@ -18,30 +18,60 @@ Element | Modifier/class | Description
...
+    <nav class='fd-pagination__nav'>
+        <div style=''>
+        
+        </div>
```
InnaAtanasova, line 46: what's the purpose of this?

stories/pagination/pagination.stories.js
```diff
@@ -18,30 +18,60 @@ Element | Modifier/class | Description
...
+        <a href='#' class='fd-button fd-button--compact fd-button--transparent fd-pagination__button fd-pagination__button--mobile' aria-label='First' aria-disabled='true'>
```
InnaAtanasova, line 47: why do you have the `fd-pagination__button--mobile` modifier in desktop example?

stories/pagination/pagination.stories.js
```diff
@@ -18,30 +18,60 @@ Element | Modifier/class | Description
...
+            <i class='sap-icon sap-icon--media-rewind'></i>
+        </a>
+        
+        <a href='#' class='fd-button fd-button--compact fd-button--transparent fd-pagination__button' aria-label='Previous' aria-disabled='true'>
+            <i class='sap-icon sap-icon--navigation-left-arrow'></i>
+        </a>
+        
+        <label class='fd-form-label fd-pagination__input-label' for='firstPageInput'>Page:</label>
```
InnaAtanasova, line 55: `fd-pagination__label` instead of `fd-pagination__input-label`

stories/pagination/pagination.stories.js
```diff
@@ -18,30 +18,60 @@ Element | Modifier/class | Description
...
+        
+        <a href='#' class='fd-button fd-button--compact fd-button--transparent fd-pagination__link'>2</a>
+        
+        <a href='#' class='fd-button fd-button--compact fd-button--transparent fd-pagination__link'>3</a>
+        
+        <a href='#' class='fd-button fd-button--compact fd-button--transparent fd-pagination__button' aria-label='Next' aria-disabled='false'>
+            <i class='sap-icon sap-icon--navigation-right-arrow'></i>
+        </a>
+        
+        <a href='#' class='fd-button fd-button--compact fd-button--transparent fd-pagination__button fd-pagination__button--mobile' aria-label='Last' aria-disabled='false'>
```
InnaAtanasova, line 67: `fd-pagination__button--mobile`?

InnaAtanasova (changes requested):

InnaAtanasova:
Screen Shot 2021-11-17 at 12 25 56 PM

Please refactor this. The user should be able to copy the markup in the code section.

InnaAtanasova:
Screen Shot 2021-11-17 at 12 28 41 PM

You are not supposed to have compact buttons in mobile mode.

droshev:
@platon-rov can you maintain the scope too?

stories/pagination/pagination.stories.js
platon-rov: It's for the docs, in ngx we will have

```
<a
  fd-button
  [glyph]="_isRtl ? 'media-rewind' : 'media-forward'"
  ...
```

And I assume to have something similar from developers who will be creating their libraries on top of Fundamental Styles.

stories/pagination/pagination.stories.js
platon-rov: Removed.

stories/pagination/pagination.stories.js
platon-rov: Actually, there is no separate mobile navigation, when the screen's size is `< 600px` we will have first & last page buttons.

stories/pagination/pagination.stories.js
platon-rov: Yea, it's for buttons that are shown only on a mobile.

platon-rov:
https://github.com/SAP/fundamental-styles/pull/2912#issuecomment-971764081

> Why is the width of the input different in these cases?

https://github.com/SAP/fundamental-styles/pull/2912#issuecomment-971770438

> Why does the input have such big width? I think you need to use: width: auto, not width:100%. This will result in this

I partially fixed it by setting `width: auto` as you recommended, thank you. But it's also depends on `min\max` attribute of the input, for example if we have 500 (three digits) pages then input will be wider then if we have 5 (one digit).

https://github.com/SAP/fundamental-styles/pull/2912#issuecomment-971768009

> The space between 2 and the ... is different from the space between the ... and 500, aka the dots are not in the middle of the page numbers.

Link always has `9px` of horizontal padding, but since content `500` is longer than `2` we have buttons of the different width. So we're seeing illusion that `...` is not centered.

https://github.com/SAP/fundamental-styles/pull/2912#issuecomment-971773373

> You have Cozy and Mobile modes but the specs says Desktop and Tablet are the same.

But below in the table described to use Cozy for tablets.

Screenshot 2021-11-18 at 15 26 40

platon-rov:
> @platon-rov can you maintain the scope too?

@droshev What do you mean, change it to `feat`?

stories/pagination/pagination.stories.js
InnaAtanasova: My question is why are you combining it with `fd-button--compact`? `--compact` and `--mobile` modifiers usually define different metrics for a component. For example, in compact (default) mode, the height of an element is 24px, but in mobile mode it changes to 32px. So it makes no sense to have a modifier class on an element to set it to compact mode and then to mobile.

platon-rov: Fixed.

stories/pagination/pagination.stories.js
InnaAtanasova: But why adding functionality in ngx when we can do this with css? We have many cases in fund-styles, exactly like this one, where we just rotate the icon in RTL.

InnaAtanasova:
> > @platon-rov can you maintain the scope too?
> 
> @droshev What do you mean, change it to `feat`?

nope, he means you are missing the (styles) part in the PR title. You have: `fix: pagination styles update` and it should be `fix(styles): pagination styles update`. I updated it :)

src/styles/pagination.scss
```diff
@@ -7,94 +7,94 @@
...
+  .#{$block}__total,
+  .#{$block}__per-page-label,
+  .#{$block}__link,
+  .#{$block}__more {
+    display: none;
+  }
 
-  $fd-pagination-link-padding-x: 0.25rem;
-  $fd-pagination-padding-x: 0.5rem;
+  .#{$block}__button--mobile,
```
InnaAtanasova, line 22: I haven't found `fd-pagination__button` class that you are modifying with `fd-pagination__button--mobile`. You can't have the second without the first :)

InnaAtanasova (commented):
Screen Shot 2021-11-19 at 2 09 18 PM

The input follows the specs for Input Field and I couldn't find anything about max-width, only for mix-width, which means the input should grow to fit the content?

platon-rov:
https://github.com/SAP/fundamental-styles/pull/2912#pullrequestreview-811539731 

> The input follows the specs for Input Field and I couldn't find anything about max-width, only for mix-width, which means the input should grow to fit the content?

`min-width: 2.75rem; width: auto` accordingly to the Fiori 3 specs.

src/styles/pagination.scss
```diff
@@ -7,94 +7,94 @@
...
   @include fd-reset();
+  @include fd-flex-between();
 
   font-size: var(--sapFontSize);
+  padding: 0 1rem;
+  height: 2.75rem;
+  width: 100%;
 
-  &__total {
+  .#{$block}__per-page {
```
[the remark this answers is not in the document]
platon-rov: Okay, i reverted for pagination-only elements but left for buttons, labels, inputs as we're using appropriate components and there may be issues like `.fd-button` styles overwrite `.fd-pagination__link` styles.

Also left in `fd-pagination-mobile` mixin as it can be applied to `.fd-pagination--mobile` and then we will have for example `.fd-pagination--mobile__link` but we don't have such element.

src/styles/pagination.scss
```diff
@@ -7,94 +7,94 @@
...
-    }
+  .#{$block}__label {
+    display: none;
+  }
 
-    &--next {
-      @include fd-icon("navigation-right-arrow");
+  .#{$block}__button--mobile {
+    display: none;
+  }
```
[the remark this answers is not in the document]
platon-rov: Don't fully understand.

We have four buttons: first, last, prev, next. 
* prev & next should be visible all the time
* first & last should be visible only in mobile mode

Because of the facts above, i've added `--mobile` to mobile-only buttons. so they're `display: none` by default and `display: block` in mobile mode.

src/styles/pagination.scss
platon-rov: Screenshot 2021-11-22 at 12 12 52

InnaAtanasova (approved):

N1XUS (approved):

PR #2916, opened 2021-11-17
fix(fn): inconsistent selected item border in Select across zoom levels

InnaAtanasova:
## Description
On some operating systems supporting sub-pixel rendering the browser may
inconsistently approximate the height/widths of boxes across zoom level. The
issue becomes apparent when multiple controls are placed together (e.g.,
table rows, tags, input boxes) as the user zooms in/out.

Specifying the border-width property is a solution that does not involve
using future CSS specs or GPU-based calculations.

-  All values are in `rem`
- Text elements follow the truncation rules
- hover state of the element follow design spec
- focus state of the element follow design spec
- active state of the element follow design spec
-  pressed state of the element follow design spec
- Responsiveness rules - the component has modifier classes for all breakpoints
- Includes Compact/Cosy/Tablet design
- [x] RTL support
3. Testing
- tested Storybook examples with "CSS Resources" `normalize` option 
- tested Storybook examples with "CSS Resources" `unnormalize` option 
- Verified all styles in IE11
- Updated tests
4. Documentation
- Storybook documentation has been created/updated
- Breaking Changes [wiki](https://github.com/SAP/fundamental-styles/wiki/Breaking-Changes) has been updated in case of breaking changes.

### DIFF
--- a/src/fn/fn-select.scss
+++ b/src/fn/fn-select.scss
@@ -7,8 +7,13 @@ $block: #{$fn-namespace}-select;
   background: $background;
 
   &::before {
-    background: $border;
+    border-left-color: $border;
   }
+
+  // @include fn-rtl() {
+  //   border-left-color: transparent;
+  //   border-right-color: $border;
+  // }
 }
 
 .#{$block} {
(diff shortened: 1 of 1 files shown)


src/fn/fn-select.scss
```diff
@@ -57,13 +57,20 @@ $block: #{$fn-namespace}-select;
...
+      left: 0;
       content: '';
       height: 100%;
-      width: 0.125rem;
       position: absolute;
+      border-left-width: 2px;
+      border-left-style: solid;
+    }
+
+    @include fn-rtl() {
```
platon-rov, line 69: Previously used mixin was doing the same:

```
@mixin fn-set-position-left($left) {
  left: $left;

  @include fn-rtl() {
    left: auto;
    right: $left;
  }
}
```

InnaAtanasova: it does not work. for before/after

droshev (approved):

darksmo (approved):

PR #2919, opened 2021-11-19
fix(fn): move border style and width to variables [ci visual]

InnaAtanasova:
## Description
Select, Input, Tabs and Tags components use the same value for border-width (0.125rem) and border-style (solid) so I created variables in `_fn-variables.scss` file for easier maintenance.


droshev (approved):

darksmo (commented):
👍

PR #2972, opened 2021-12-07
fix: rating-indicator styles [ci visual]

Betrozov:
## Description
Fixed rating-indicator styles for `Half values` state

### Before:
Screen Shot 2021-12-06 at 17 21 55

### After:
Screen Shot 2021-12-06 at 17 22 39

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/styles/rating-indicator.scss
+++ b/src/styles/rating-indicator.scss
@@ -3,11 +3,6 @@
 
 $block: #{$fd-namespace}-rating-indicator;
 
-:root {
-  --sapRating-indicator-icon-rated: none;
-  --sapRating-indicator-icon-unrated: none;
-}
-
 .#{$block} {
   $fd-rating-indicator-sizes: (
     "cozy": ("fontSize": 1.5rem, "spacingBetweenIcons": 0.1875rem, "controlMargin": 0.8125rem 0),
@@ -33,7 +28,7 @@ $block: #{$fd-namespace}-rating-indicator;
     .#{$block}__label {
       margin-right: 0;
 
-      &:nth-child(4n+4) {
+      &:nth-child(4n+6) {
         margin-right: $spacingBetweenIcons;
         flex-direction: row-reverse;
 
@@ -52,20 +47,12 @@ $block: #{$fd-namespace}-rating-indicator;
       }
 
       overflow: hidden;
-      width: $iconFontSize / 2 !important;
+      width: $iconFontSize / 2;
     }
 
-    &.#{$block}--icon .#{$block}__label {
-      @include fd-rtl() {
-        background-position: right;
-      }
-
-      &:nth-child(4n+4) {
-        background-position: right;
-
-        @include fd-rtl() {
-          background-position: left;
-        }
+    &.#{$block}--icon {
+      .#{$block}__label {
+        width: $iconFontSize / 2;
       }
     }
   }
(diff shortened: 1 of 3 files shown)


src/styles/rating-indicator.scss
```diff
@@ -228,10 +214,18 @@ $block: #{$fd-namespace}-rating-indicator;
     }
   }
 
-  &--half-star {
+  &.#{$block}--half-star {
     @include fd-rating-indicator-with-half-star($fd-rating-indicator-default-font-size, $fd-rating-indicator-default-spacing-between-icons);
   }
 
+  .#{$block}__label.#{$block}__zero-rating {
```
InnaAtanasova, line 221: This is violating BEM :) You can't have 2 elements from the same block at the same level. 
You can have:
```
<div fd-element__label fd-another-element__label>...</div>
```
to preserve the self-contained rules, but you can't have:

```
<div fd-element__label fd-element__zero-rating>...</div>
```

This can be achieved with modifier classes.

src/styles/rating-indicator.scss
```diff
@@ -228,10 +214,18 @@ $block: #{$fd-namespace}-rating-indicator;
...
+    padding: 0 !important;
+    margin: 0 !important;
+    width: 0 !important;
+    overflow: hidden !important;
+    opacity: 0.0001 !important;
```
InnaAtanasova, line 226: Please try to achieve this without !important. This element becomes impossible to be modified by end users (in case they need to).

src/styles/rating-indicator.scss
```diff
@@ -241,14 +235,27 @@ $block: #{$fd-namespace}-rating-indicator;
 
       width: $fd-rating-indicator-default-font-size;
       height: $fd-rating-indicator-default-font-size;
-      background-image: var(--sapRating-indicator-icon-rated);
-      background-size: cover;
-      background-position: left;
-      background-repeat: no-repeat;
+
+      .rated,
+      .unrated {
```
InnaAtanasova, line 240: These are class names that do not follow BEM. Please modify accordingly.

src/styles/rating-indicator.scss
```diff
@@ -241,14 +235,27 @@ $block: #{$fd-namespace}-rating-indicator;
...
+        font-size: $fd-rating-indicator-default-font-size;
+        height: 100%;
+        line-height: 100%;
```
InnaAtanasova, line 243: line-height: 1 not working the same way?

InnaAtanasova (changes requested):

src/styles/rating-indicator.scss
```diff
@@ -25,46 +20,40 @@ $block: #{$fd-namespace}-rating-indicator;
...
+    .#{$block} {
+      &__input {
+        width: $iconFontSize * 0.5;
+        height: $iconFontSize;
+      }
 
-    .#{$block}__label {
-      margin-right: 0;
+      &__label {
+        margin-right: 0;
```
InnaAtanasova, line 30: no rtl?

InnaAtanasova: you can use a mixin for setting the margins, they will handle the RTL. 
You can check `fd-set-position-right` and `fd-set-position-right-pseudo`

InnaAtanasova (approved):
Just minor comments but o.w. looks great. Awesome job!

PR #2973, opened 2021-12-08
fix(styles): pagination component update due to new designs

platon-rov:
## Description

Pagination component styles update due to the new design.

## Screenshots

Too many changes, please see the doc page.

## Breaking changes

- If > 9 pages then 9 elements should be shown - including page links, more, current page input. Otherwise, all pages should be shown, use the link instead of input with `.is-active` class or `aria-selected="true"` attribute for the current page. Input should still be present to be shown in the mobile mode, but it's hidden in desktop mode if < 9 pages.

- Add an additional label with class `.fd-pagination__label` after the input in the mobile mode with the number of pages available.

See on wiki page [breaking changes](https://github.com/SAP/fundamental-styles/wiki/Breaking-Changes#pagination-2973).

### DIFF
--- a/src/styles/pagination.scss
+++ b/src/styles/pagination.scss
@@ -51,7 +51,8 @@ $block: #{$fd-namespace}-pagination;
   &__more,
   &__button,
   &__link,
-  &__input {
+  &__input,
+  &__label {
     @include fd-set-margins-x-equal(0.125rem);
   }
 
@@ -92,8 +93,6 @@ $block: #{$fd-namespace}-pagination;
 
   .#{$block}__label {
     display: none;
-
-    @include fd-set-margins-x-equal(0.125rem);
   }
 
   .#{$block}__button {
@@ -119,7 +118,13 @@ $block: #{$fd-namespace}-pagination;
   }
 }
 
-@media (max-width: 1024px) {
+@include fd-media-sm() {
+  .#{$block} {
+    @include fd-pagination-mobile();
+  }
+}
+
+@include fd-media-md() {
   .#{$block} {
     @include fd-pagination-mobile();
   }
--- a/src/styles/pagination.scss
+++ b/src/styles/pagination.scss
@@ -48,14 +48,6 @@ $block: #{$fd-namespace}-pagination;
     @include fd-set-margins-x-equal('auto');
   }
 
-  &__more,
-  &__button,
-  &__link,
-  &__input,
-  &__label {
-    @include fd-set-margins-x-equal(0.125rem);
-  }
-
   &__more {
     @include fd-reset();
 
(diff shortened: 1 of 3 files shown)


src/styles/pagination.scss
```diff
@@ -84,6 +92,8 @@ $block: #{$fd-namespace}-pagination;
 
   .#{$block}__label {
     display: none;
+
+    @include fd-set-margins-x-equal(0.125rem);
```
InnaAtanasova, line 96: maybe move this part to the block at lines 51-56

src/styles/pagination.scss
```diff
@@ -98,12 +108,18 @@ $block: #{$fd-namespace}-pagination;
...
+    }
+  }
+
   &--mobile {
     @include fd-pagination-mobile();
   }
 }
 
-@media (max-width: 599px) {
+@media (max-width: 1024px) {
```
InnaAtanasova, line 122: we have media queries in the mixin file. If it's missing you can add it there so we reuse it in future.

InnaAtanasova:
1. You are missing the scope of the PR. Should be `fix()styles: pagination styles`. It's also a good idea to add more info in the title about the change. Remember the title is what's shown on the release notes, so they should be as descriptive as possible (and allowed by GitHub)
2. You need to mention the Breaking changes in the PR description, not only to provide a link to the wiki page.
3. See the comments on the code

InnaAtanasova:
Also, please refer to the following document about the a11y in Pagination: https://www.a11ymatters.com/pattern/pagination/

You need a structure like this:
```
<nav role="navigation" aria-label="Pagination Navigation">
    <ul>
        <li><a href="/page-1" aria-label="Goto Page 1">1</a></li>
        <li><a href="/page-2" aria-label="Goto Page 2">2</a></li>
        <li><a href="/page-3" aria-label="Goto Page 3">3</a></li>
        <li><a href="/page-4" aria-label="Goto Page 4">4</a></li>
        <li><a href="/page-5" aria-label="Goto Page 5">5</a></li>
    </ul>
</nav>
```

with role, aria-label, etc.
You will also need to indicate the current page using `aria-current`.
Additional info about the component a11y can be found here: https://designsystem.digital.gov/components/pagination/

src/styles/pagination.scss
```diff
@@ -47,6 +48,14 @@ $block: #{$fd-namespace}-pagination;
...
 
+  &__more,
+  &__button,
+  &__link,
+  &__input,
+  &__label {
+    @include fd-set-margins-x-equal(0.125rem);
+  }
+
   &__more {
```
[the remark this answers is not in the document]
platon-rov: great catch, thank you!

InnaAtanasova (approved):

PR #3009, opened 2021-12-20
fix: remove active page link when in long mode

platon-rov:
## Description

Remove from displaying active page link if pagination component in long (> 9 pages) mode. This needed to avoid any extra code logic on ngx side when showing pagination component.

### Before:

Screenshot 2021-12-20 at 19 53 31

### After:

Screenshot 2021-12-20 at 19 53 03

### DIFF
--- a/src/styles/pagination.scss
+++ b/src/styles/pagination.scss
@@ -24,6 +24,12 @@ $block: #{$fd-namespace}-pagination;
   .#{$block}__input {
     display: flex;
   }
+
+  .#{$block}__link {
+    @include fd-active() {
+      display: none;
+    }
+  }
 }
 
 .#{$block} {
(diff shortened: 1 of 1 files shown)


mikerodonnell89 (approved):

mikerodonnell89:
@platon-rov so this removes the need for the these conditional templates? [here](https://github.com/SAP/fundamental-ngx/blob/main/libs/core/src/lib/pagination/pagination.component.html#L45) and [here](https://github.com/SAP/fundamental-ngx/blob/main/libs/core/src/lib/pagination/pagination.component.html#L116)

Do we need corresponding changes in ngx?

platon-rov:
> @platon-rov so this removes the need for the these conditional templates? [here](https://github.com/SAP/fundamental-ngx/blob/main/libs/core/src/lib/pagination/pagination.component.html#L45) and [here](https://github.com/SAP/fundamental-ngx/blob/main/libs/core/src/lib/pagination/pagination.component.html#L116)
> 
> Do we need corresponding changes in ngx?

Yea, we need to adopt these changes in ngx and i will do it.

InnaAtanasova (approved):

PR #3031, opened 2021-12-27
fix: chromatic [ci visual]

### LINKED ISSUE #3023: chromatic is broken
At the moment running chromatic fails with this error:
here is an [example](https://github.com/SAP/fundamental-styles/runs/4634610021?check_suite_focus=true)

is it worth exploring migration to `yarn`?

this is the feedback that we got:
```
It seems that the issue is not actually with Chromatic, but with node-gyp package instead. To whoever sees this message it should check what is happening in the said package.
```

```
It looks like you are using yarn in CI, but have a package-lock.json rather than a yarn.lock file. That doesn't seem right. You should either update your CI script to use npm or check in your yarn.lock.
My usual approach to fixing node-gyp issues is to completely trash node_modules and reinstall it all. Sometimes it helps to remove the lockfile and regenerate it, but you may also get into more trouble that way. Could be worth a try though.
```

platon-rov:
## Related Issue

Closes #3023.

## Description

Chromatic GA fix.

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -8,9 +8,9 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - run: |
-        yarn && yarn build
+        npm i && npm run build
     - uses: chromaui/action@v1
-      with: 
+      with:
         projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
         token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


droshev (approved):

droshev:
let me review the visual changes and i will merge it. Thanks!

platon-rov:
> let me review the visual changes and i will merge it. Thanks!

Yea, here is a lot of changes since we didn't run chromatic for the last weeks, I will also take a look

PR #3037, opened 2021-12-30
feat(fn): introduce fn-card and fn-info-label

mikerodonnell89:
Waiting on some additional specifications but this might be good for an experimental component


src/fn/fn-card.scss
```diff
@@ -0,0 +1,81 @@
...
+  &__title-icon-button-container {
+    @include fn-flex();
+    @include fn-flex-vertical-center();
+
+    .#{$block}__title-button {
+      @include fn-set-margin-left(auto);
+    }
+    .#{$block}__title-icon {
+      @include fn-reset();
+      @include fn-set-rectangle(1rem, 1rem);
```
InnaAtanasova, line 46: Just for your information, you could use `fn-set-square`but this one is ok, I accept it

InnaAtanasova (approved):
Awesome! 🚀

Issue #2928, opened 2021-11-24
FN Next Next


droshev:
Tasks related to the component design updates:

- [ ] update the colors

update existing components
- [ ] ...

enhance existing components
- [ ] ...

create new components
- [ ] ...

PR #3069, opened 2022-01-13
feat(fn): implement fn-form

mikerodonnell89:
part of #2928 

implements fn-form, fn-form__item, fn-form__group, fn-form__label etc

### DIFF
--- a/src/fn/fn-form.scss
+++ b/src/fn/fn-form.scss
@@ -5,7 +5,7 @@ $block: #{$fn-namespace}-form;
 .#{$block} {
   @include fn-reset();
 
-  &-item {
+  &__item {
     @include fn-reset();
     @include fn-flex(column);
 
--- a/src/fn/fn-form.scss
+++ b/src/fn/fn-form.scss
@@ -1,0 +1,65 @@
+@import "./fn-settings";
+
+$block: #{$fn-namespace}-form;
+
+.#{$block} {
+  @include fn-reset();
+
+  &-item {
+    @include fn-reset();
+    @include fn-flex(column);
+
+    &:not(:last-child) {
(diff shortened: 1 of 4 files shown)


src/fn/fn-form.scss
```diff
@@ -0,0 +1,65 @@
+@import "./fn-settings";
+
+$block: #{$fn-namespace}-form;
+
+.#{$block} {
+  @include fn-reset();
+
+  &-item {
```
platon-rov, line 8: You wrote `fn-form__item` in description but here it's `fn-form-item`, which one is correct?

mikerodonnell89: good catch, should be `fn-form__item`

InnaAtanasova:
Screen Shot 2022-01-14 at 3 50 12 PM
There's 2px spacing between the label and the input

src/fn/fn-form.scss
```diff
@@ -0,0 +1,65 @@
...
+
+  &__item {
+    @include fn-reset();
+    @include fn-flex(column);
+
+    &:not(:last-child) {
+      margin-bottom: 0.75rem;
+    }
+
+    .#{$block}__label {
```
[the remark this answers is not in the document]
InnaAtanasova: Maybe needs a discussion with the team, but in my opinion the label should be a separate component, as we had many cases in Fund-styles where a label was not necessarily part os a form. So I think having it as a separate component could be a good idea.

InnaAtanasova (changes requested):

mikerodonnell89:
@InnaAtanasova addressed your comments, thank you

InnaAtanasova (approved):
Awesome!

PR #3094, opened 2022-01-19
feat(fn): implement message strip

mikerodonnell89:
part of #2928 

Implements the fn-message-strip

Screen Shot 2022-01-19 at 4 12 09 PM

### DIFF
--- a/src/fn/fn-message-strip.scss
+++ b/src/fn/fn-message-strip.scss
@@ -2,12 +2,21 @@
 
 $block: #{$fn-namespace}-message-strip;
 
+@mixin fn-message-strip-states($background, $iconColor) {
+  background-color: $background;
+
+  .#{$block}__icon {
+    &::before {
+      color: $iconColor;
+    }
+  }
+}
+
 .#{$block} {
   @include fn-reset();
-  @include fn-flex();
   @include fn-set-paddings-x(0.5rem, 0.5rem);
+  @include fn-flex-vertical-center();
 
-  align-items: center;
   background-color: $fn-color-blue-50;
   color: $fn-color-grey-700;
   min-height: 1.75rem;
@@ -22,14 +31,10 @@ $block: #{$fn-namespace}-message-strip;
     max-width: 95%;
   }
 
-  &__icon {
-    background-color: transparent;
-    margin-right: 0.25rem !important;
+  .#{$block}__icon {
+    @include fn-set-margin-right(0.25rem);
 
-    @include fn-rtl() {
-      margin-right: 0 !important;
-      margin-left: 0.25rem !important;
-    }
+    background-color: transparent;
 
     &::before {
       color: $fn-color-blue-500;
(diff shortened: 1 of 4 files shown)


src/fn/fn-message-strip.scss
```diff
@@ -0,0 +1,101 @@
+@import "./fn-settings";
+
+$block: #{$fn-namespace}-message-strip;
+
+.#{$block} {
+  @include fn-reset();
+  @include fn-flex();
+  @include fn-set-paddings-x(0.5rem, 0.5rem);
+
+  align-items: center;
```
InnaAtanasova, line 10: There's a mixin that combines display: flex with align-items: center
```
@mixin fn-flex-vertical-center {
  display: flex;
  align-items: center;
  @content;
}
```

src/fn/fn-message-strip.scss
```diff
@@ -0,0 +1,101 @@
...
+  border-radius: $fn-border-radius-4;
+
+  &__text {
+    @include fn-reset();
+
+    color: inherit;
+    max-width: 95%;
+  }
+
+  &__icon {
```
InnaAtanasova, line 25: If you want to avoid the usage of !important you can use `.#{$block}__icon` instead of `&__icon`. This will enforce the specificity. And then you can use the mixin for setting the margin-right `@include fn-set-margin-right(0.25rem)`

src/fn/fn-message-strip.scss
```diff
@@ -0,0 +1,101 @@
...
+
+    &::before {
+      color: $fn-color-blue-500;
+      display: flex;
+      margin-top: 0.125rem;
+    }
+  }
+
+  &--dismissible {
+    @include fn-flex();
```
InnaAtanasova, line 42: You already have flex coming from the main class

src/fn/fn-message-strip.scss
```diff
@@ -0,0 +1,101 @@
...
+  &--error {
+    background-color: $fn-color-red-50;
+
+    .#{$block}__icon {
+      &::before {
+        color: $fn-color-red-500;
+      }
+    }
+  }
+}
```
InnaAtanasova, line 101: There are 2 ways to avoid this code repetition.
First:
create a css variable at the beginning holding the stare values:
```
$color-states: (
  "success": ("background": $fn-color-green-50, "iconColor": $fn-color-green-500),
  "warning": ("background": $fn-color-orange-50, "iconColor": $fn-color-orange-400),
  "error": ("background": $fn-color-red-50, "iconColor": $fn-color-red-500)
)
```

Then you just loop through this object:
```
  @each $set-name, $color-set in $color-states {
    &--#{$set-name} {
      background-color: map_get($color-set, "background");
      
      .#{$block}__icon {
          &::before {
            color: map_get($color-set, "iconColor");
          }
       }
    }
  }
```

InnaAtanasova: Second option is with a mixin:
```
@mixin fn-message-strip-states($background, $iconColor) {
    background-color: $background;

    .#{$block}__icon {
      &::before {
        color: $iconColor;
      }
    }
}
```

Then you just need to do:
```
&--success {
     @include fn-message-strip-states($fn-color-green-50, $fn-color-green-500);
}
&--warning {
     @include fn-message-strip-states($fn-color-orange-50, $fn-color-orange-400);
}
&--error {
     @include fn-message-strip-states($fn-color-redd-50, $fn-color-red-500);
}
```

src/fn/fn-message-strip.scss
```diff
@@ -0,0 +1,101 @@
...
+      display: flex;
+      margin-top: 0.125rem;
+    }
+  }
+
+  &--dismissible {
+    @include fn-flex();
+
+    .#{$block}__close-button {
+      @include fn-reset();
```
InnaAtanasova, line 45: should not have a reset as it will reset the rules coming from fn-button

src/fn/fn-message-strip.scss
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    &::before {
+      color: $fn-color-blue-500;
+      display: flex;
+      margin-top: 0.125rem;
+    }
+  }
+
+  &--dismissible {
```
InnaAtanasova, line 41: Why do you need a modifier class? If it's a dismissible message strip we can just add button to the markup.

src/fn/fn-message-strip.scss
mikerodonnell89: It does not reset all the rules from button and still brings some use

InnaAtanasova (approved):

PR #3121, opened 2022-01-26
docs(deps): synchronization of toolbar addons

g-cheishvili:
## Related Issue
None

## Description

- Previously when you'd switch theme or directionality on one component page, it would remain default on another component page. Now changes persist.
- Refactored small amount of code to allow better readability of custom stuff for storybooks.
- Previously when you had app run in development mode you'd see two controllers for theme, one on toolbar and one on header. Now in development mode only toolbars are visible and in built version only header switcher is left.
- Changed the way directionality is applied, now it just wraps story source code into `div` with directionality, instead of going through all nodes on page.
- Now directionality works on Canvas too and not only on Docs.


.storybook/custom/components/IfBlock.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+
+export const IfBlock = ({condition, children}) => {
+    if (condition) {
+        return <>{children}</>
+    }
+    return <></>;
```
platon-rov, line 7: we can return `null` if we don't want to render anything.

platon-rov (approved):

InnaAtanasova (approved):

.storybook/custom/components/IfBlock.js
g-cheishvili: ha, thanks for catch, I did not knew that

droshev (approved):

PR #3123, opened 2022-01-26
feat(styles): focus outline is visible now [ci visual]

g-cheishvili:
## Related Issue
Part of SAP/fundamental-ngx#7587

## Description
Focus outline was not visible on tab headers in case of icon tabs. This was because offset was negative exactly same as width of outline. I removed negativity and now it is visible

### Before:
before

### After:
after

### DIFF
--- a/src/styles/tabs.scss
+++ b/src/styles/tabs.scss
@@ -12,7 +12,7 @@ $block: #{$fd-namespace}-tabs;
 
 .#{$block} {
   %fd-tabs-focus {
-    outline-offset: 0.0625rem;
+    outline-offset: var(--sapContent_FocusWidth);
     outline-style: var(--sapContent_FocusStyle);
     outline-width: var(--sapContent_FocusWidth);
     outline-color: var(--sapContent_FocusColor);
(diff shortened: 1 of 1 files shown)


platon-rov (commented):
It looks got however I didn't find any information about how it should looks in wiki. 

In sapui5 it's totally different, [example](https://sapui5.hana.ondemand.com/#/entity/sap.m.IconTabBar/sample/sap.m.sample.IconTabBarMulti)

In icon tab bar component it's also a bit different, [example](https://fundamental-ngx.netlify.app/#/platform/icon-tab-bar#icon-type).

I would say let's make it's the same as in the icon tab bar component and send an email to confirm how exactly it should looks like.

InnaAtanasova (changes requested):
I sent you and @platon-rov link to the visual specs for the focus

g-cheishvili:
> It looks got however I didn't find any information about how it should looks in wiki.
> 
> In sapui5 it's totally different, [example](https://sapui5.hana.ondemand.com/#/entity/sap.m.IconTabBar/sample/sap.m.sample.IconTabBarMulti)
> 
> In icon tab bar component it's also a bit different, [example](https://fundamental-ngx.netlify.app/#/platform/icon-tab-bar#icon-type).
> 
> I would say let's make it's the same as in the icon tab bar component and send an email to confirm how exactly it should looks like.

I made it look like sapui5 one

InnaAtanasova (approved):

PR #3145, opened 2022-02-04
docs(fn): replace experimental with FN

droshev:
replace Experimental with Fiori Next


g-cheishvili (approved):
Everything works as expected

InnaAtanasova (approved):

PR #3159, opened 2022-02-07
fix(fn): adjust slider handle position [ci visual]

N1XUS:
## Related Issue
none

## Description
Previously, slider handle was "tweaked" based on aria-valuenow property, which can be different in real-case scenario. Added wrapper which compensates handle offset so at the edge positions of the slider, handle would fit into the size of the slider itself.

- [x] All values are in `rem`
- n/a Includes Compact/Cosy/Tablet design
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/fn/fn-slider.scss
+++ b/src/fn/fn-slider.scss
@@ -3,6 +3,7 @@
 $block: #{$fn-namespace}-slider;
 $tag-border-width: 0.125rem;
 $slider-handle-width: 2.25rem;
+$slider-wrapper-offset: $slider-handle-width * 0.5;
 $slider-handle-height: 1.625rem;
 $slider-dot-size: 0.25rem;
 $slider-height: 0.5rem;
@@ -48,6 +49,14 @@ $slider-height: 0.5rem;
     opacity: $fn-disabled-opacity;
   }
 
+  &__wrapper {
+    @include fn-reset();
+
+    position: relative;
+    margin-left: $slider-wrapper-offset;
+    margin-right: $slider-wrapper-offset;
+  }
+
   &__track {
     @include fn-reset();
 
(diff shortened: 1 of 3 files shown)


src/fn/fn-slider.scss
```diff
@@ -48,6 +49,14 @@ $slider-height: 0.5rem;
     opacity: $fn-disabled-opacity;
   }
 
+  &__wrapper {
+    @include fn-reset();
+
+    position: relative;
+    margin-left: $slider-wrapper-offset;
+    margin-right: $slider-wrapper-offset;
```
InnaAtanasova, line 57: You can use the `fn-set-margins-x` mixin here

src/fn/fn-slider.scss
```diff
@@ -59,10 +68,10 @@ $slider-height: 0.5rem;
 
   &__track-range {
     @include fn-reset();
-    @include fn-set-position-left(0);
 
     z-index: 2;
-    width: 100%;
+    left: -$slider-wrapper-offset;
+    right: -$slider-wrapper-offset;
```
InnaAtanasova, line 74: for consistency, please use the mixins `fn-set-position-right` and `fn-set-position-left`. Thank you!

InnaAtanasova (approved):

src/fn/fn-slider.scss
InnaAtanasova: Discussed with Denis, ignore my comment

N1XUS: No need since values are equal

src/fn/fn-slider.scss
N1XUS: No need since values are equal

PR #3165, opened 2022-02-08
feat(fn): introduce fn-table [ci visual]

mikerodonnell89:
This PR introduces the fn-table component, and makes some minor tweaks to the list component as the button had also been updated.


src/fn/fn-table.scss
```diff
@@ -0,0 +1,175 @@
...
+      @include fn-reset();
+      @include fn-flex-vertical-center();
+
+      display: block;
+      color: $fn-color-grey-9;
+    }
+
+    &--byline {
+      display: flex;
+      align-items: center;
```
InnaAtanasova, line 135: `fn-flex-vertical-center`  mixin

src/fn/fn-table.scss
```diff
@@ -0,0 +1,175 @@
...
+
+  &__end-container {
+    @include fn-reset();
+    @include fn-flex();
+
+    :not(:last-child) {
+      @include fn-set-margin-right($fn-table-item-margin);
+    }
+
+    align-items: center;
```
InnaAtanasova, line 169: you have `fn-flex` at line 163, so you can directly use `fn-flex-vertical-center` and get rid of this line

src/fn/fn-table.scss
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    @include fn-disabled() {
+      pointer-events: none;
+      opacity: $fn-disabled-opacity;
+    }
+
+    .#{$block}__avatar {
+      @include fn-set-margin-right($fn-table-item-margin);
+    }
```
InnaAtanasova, line 123: you can avoid this additional css class by relying of first-child.
With byline list you always have `.fn-table__start-container` element. When you have an avatar in front of it, this element is never the first child, so in this case you can add margin-left. Something like:
```
.fn-table__start-container:not(:first-child) {
    @include fn-set-margin-left($fn-table-item-margin);
}
```

This will add the spacing only when the element has a sibling before it and the good part is that it will be valid for any element not just avatar

src/fn/fn-table.scss
```diff
@@ -0,0 +1,175 @@
...
+      box-shadow: inset 0 -0.063rem 0 #0070f1;
+    }
+
+    padding: $fn-table-item-margin 0.75rem;
+
+    .#{$block}__icon {
+      @include fn-set-margins-x(0.25rem, 0.625rem);
+
+      color: $fn-color-grey-7;
+    }
```
InnaAtanasova, line 114: Are you using this? I couldn't find any occurrences of this class in the examples

src/fn/fn-table.scss
```diff
@@ -0,0 +1,175 @@
...
+
+    @include fn-not-focus-visible() {
+      @include fn-hover() {
+        background: $fn-color-grey-1;
+      }
+    }
+
+    @include fn-selected() {
+      background: $fn-color-blue-2;
+      box-shadow: inset 0 -0.063rem 0 #0070f1;
```
InnaAtanasova, line 105: This is a repeated value for List component. Could you move it to the variables file, please.

src/fn/fn-table.scss
```diff
@@ -0,0 +1,175 @@
...
+      white-space: nowrap;
+    }
+  }
+
+  &__cell,
+  &__group-header,
+  &__header {
+    @include fn-reset();
+
+    box-shadow: inset 0 -0.063rem 0 #e4e5e7;
```
InnaAtanasova, line 43: This is a repeated value for List component. Could you move it to the variables file, please.

src/fn/fn-table.scss
```diff
@@ -0,0 +1,175 @@
...
+  &__cell {
+    &:not(.fn-table__cell--no-hover) {
+      @include fn-hover() {
+        background: $fn-color-grey-1;
+      }
+    }
+
+    @include fn-focus-visible() {
+      outline: none;
+      border-radius: $fn-table-item-focus-margin;
```
InnaAtanasova, line 93: could you use the variable `$fn-border-radius-2`? It will help us if borders are updated in the design.

src/fn/fn-table.scss
```diff
@@ -0,0 +1,175 @@
...
+  max-width: 100%;
+  border-collapse: collapse;
+
+  &__row {
+    @include fn-reset();
+
+    @include fn-selected() {
+      .#{$block}__cell {
+        background: $fn-color-blue-2;
+        box-shadow: inset 0 -0.063rem 0 #0070f1;
```
InnaAtanasova, line 21: Same as the other comment. It should be a variable for easier maintenance.

InnaAtanasova:
It's good to add some transition for the hover. You can use `fn-transition` mixin

InnaAtanasova (approved):

PR #3171, opened 2022-02-10
fix: card header non-interactive outline

platon-rov:
## Related Issue

None.

## Description

Outline removed from the card header if it has non-interactive class.

### Before:

image

### After:

image


InnaAtanasova:
Please trigger Chromatic checks so we can merge it

PR #3215, opened 2022-03-03
fix: popover styles refactoring

platon-rov:
## Related Issue

Closes ngx https://github.com/SAP/fundamental-ngx/issues/7706

Follow up pr in ngx https://github.com/SAP/fundamental-ngx/pull/7832

## Description

Core popover styles refactoring.

### Before:

image

### After:

image

### DIFF
--- a/src/styles/popover.scss
+++ b/src/styles/popover.scss
@@ -240,6 +240,11 @@ $block: #{$fd-namespace}-popover;
     width: 100%;
   }
 
+  &__inline-help {
+    padding: 0.75rem;
+    font-size: var(--sapFontSmallSize);
+  }
+
   &__popper {
     @include fd-reset();
 
@@ -252,11 +257,6 @@ $block: #{$fd-namespace}-popover;
     border-radius: $fd-popover-border-radius;
     background: $fd-popover-background-color;
     opacity: 1;
-
-    &.fd-inline-help__content {
-      padding: 0.75rem;
-      font-size: var(--sapFontSmallSize);
-    }
 
     &--no-arrow {
       margin: 0 !important;
--- a/src/styles/popover.scss
+++ b/src/styles/popover.scss
@@ -20,11 +20,6 @@ $block: #{$fd-namespace}-popover;
 
   $fd-popover-transition-params: $fd-animation-speed !default;
 
-  @mixin set-list-border-radius($position, $radius: $fd-popover-border-radius) {
-    border-#{$position}-right-radius: $radius;
-    border-#{$position}-left-radius: $radius;
-  }
-
   @include fd-reset();
 
   position: relative;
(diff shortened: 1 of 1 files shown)


src/styles/popover.scss
```diff
@@ -236,14 +244,20 @@ $block: #{$fd-namespace}-popover;
...
+    position: relative;
+    width: 100%;
     white-space: nowrap;
     box-shadow: $fd-popover-box-shadow;
     z-index: $fd-popover-z-index;
     border-radius: $fd-popover-border-radius;
     background: $fd-popover-background-color;
     opacity: 1;
 
+    &.fd-inline-help__content {
```
InnaAtanasova, line 256: this is not allowed in styles. It's breaking the self-contained rule, which means that you can't modify the properties of one component inside another component. In this case you can't modify Inline Help content inside the block code for Popover.

src/styles/popover.scss
```diff
@@ -279,82 +293,151 @@ $block: #{$fd-namespace}-popover;
...
         &::after {
           left: 1px;
-          margin: -$fd-popover-arrow-width-half 0;
           border-width: $fd-popover-arrow-width-half $fd-popover-arrow-height $fd-popover-arrow-width-half 0;
           border-right-color: $fd-popover-background-color;
         }
       }
+
+      &-x--start {
+        left: $fd-popover-arrow-offset;
```
InnaAtanasova, line 360: shouldn't this have RTL?

src/styles/popover.scss
```diff
@@ -279,82 +293,151 @@ $block: #{$fd-namespace}-popover;
...
+      }
+
+      &-x--center {
+        left: 50%;
+        transform: translateX(-50%);
+      }
+
+      &-x--end {
+        right: $fd-popover-arrow-offset;
```
InnaAtanasova, line 369: shouldn't this have RTL?

src/styles/popover.scss
```diff
@@ -279,82 +293,151 @@ $block: #{$fd-namespace}-popover;
...
+        top: 50%;
+        transform: translateY(-50%);
+      }
+
+      &-y--bottom {
+        bottom: $fd-popover-arrow-offset;
+      }
+    }
+
+    .fd-list {
```
InnaAtanasova, line 386: Please check my comment about the self-contained elements.

src/styles/popover.scss
```diff
@@ -279,82 +293,151 @@ $block: #{$fd-namespace}-popover;
...
+          @include set-list-border-radius(top);
+        }
+
+        &:last-child {
+          @include set-list-border-radius(bottom);
+        }
+      }
+    }
+
+    .fd-nested-list {
```
InnaAtanasova, line 400: Same as above.

src/styles/popover.scss
```diff
@@ -279,82 +293,151 @@ $block: #{$fd-namespace}-popover;
...
+                @include set-list-border-radius(bottom, 0);
+              }
+            }
+          }
+        }
+      }
+    }
+
+    .fd-side-nav,
+    .fd-nested-list__item {
```
InnaAtanasova, line 431: Same as above. If it's needed you can rely on modifier classes.

InnaAtanasova (changes requested):

src/styles/popover.scss
platon-rov: No, it shouldn't. See the example.

image

src/styles/popover.scss
platon-rov: https://github.com/SAP/fundamental-styles/pull/3215#discussion_r820681683

src/styles/popover.scss
platon-rov: Agree. But i'm not sure about changing it.

```
// Case 1
<fd-popover>
   <fd-list class="fd-list__popover"> // class can be easily added, refactoring is easy, however it's still the breaking change
   
// Case 2
<ng-template #template>
  // this template is for the popover, but there no information about it, it's handled explicitly
  // it's very hard to refactor
  <fd-list> 
```

src/styles/popover.scss
platon-rov: https://github.com/SAP/fundamental-styles/pull/3215#discussion_r820692761

src/styles/popover.scss
platon-rov: Appropriate css classes will be generated when initializing popover, then `start` will be `end` in RTL and vice versa.

mikerodonnell89 (approved):
works great when compiled and paired with https://github.com/SAP/fundamental-ngx/pull/7832

nice work 👍

InnaAtanasova (approved):

InnaAtanasova (commented):
I approved the code but there are some visual issues you need to check:
Screen Shot 2022-03-11 at 11 25 46 AM
Screen Shot 2022-03-11 at 11 26 09 AM

Please see https://www.chromatic.com/pullrequest?appId=5f85e3904b19d600221872af&number=3215&view=changes

PR #3241, opened 2022-03-14
fix(fn): update the position of the icon on multi line for Message Strip

InnaAtanasova:
## Related Issue
none, reported bug by Alex

## Description
- use position `absolute` to ensure that the icon and the close button always stay aligned with the first line of the text.

### Before:
Screen Shot 2022-03-14 at 4 15 29 PM
Screen Shot 2022-03-14 at 4 15 08 PM

### After:
Screen Shot 2022-03-14 at 4 14 59 PM
Screen Shot 2022-03-14 at 4 14 54 PM

### DIFF
--- a/src/fn/fn-message-strip.scss
+++ b/src/fn/fn-message-strip.scss
@@ -14,35 +14,49 @@ $block: #{$fn-namespace}-message-strip;
 
 .#{$block} {
   @include fn-reset();
-  @include fn-set-paddings-x-equal(0.5rem);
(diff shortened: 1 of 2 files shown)


g-cheishvili:
image
When there is one line of the message, icon's distance from top is bigger than from bottom. Looks not that good

g-cheishvili (approved):
Good for me

droshev (approved):

Issue #3213, opened 2022-03-02
Horizon themes parameters adaptation


droshev:
The goal of this issue is to bring the current fundamental-styles to consume the `horizon` theme family as much as possible without making changes on the markup. (It is accepted that some components won't much the theming design 100%).

PR #3263, opened 2022-03-25
feat(styles): horizon theming for Switch component

N1XUS:
## Related Issue
Relates #3213

## Description
Updated switch theme to reflect horizon theming.

### After:
image
image
image
image
image
image
image
image

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/styles/theming/common/switch/_sap_horizon.scss
+++ b/src/styles/theming/common/switch/_sap_horizon.scss
@@ -1,0 +1,79 @@
+:root {
+  // Compact mode
+  --fdSwitch_Compact_Width: 2rem;
+  --fdSwitch_Compact_Height: 1.25rem;
+  --fdSwitch_Compact_Handle_Width: 1.25rem;
+  --fdSwitch_Compact_Handle_Height: 1rem;
+  --fdSwitch_Compact_Handle_Offset: 0.3075rem;
+  --fdSwitch_Compact_Active_Handle_Offset: var(--fdSwitch_Compact_Handle_Offset);
+  --fdSwitch_Compact_Padding: 0.375rem 0;
+
+  // Standard
+  --fdSwitch_Handle_Border_Width: 0.125rem;
+  --fdSwitch_On_Hover_Slider_Background: var(--sapButton_Track_Selected_Background);
+  --fdSwitch_Border_Width: 0;
+  --fdSwitch_Width: 2.5rem;
+  --fdSwitch_Height: 1.5rem;
+  --fdSwitch_Border_Radius: 0.75rem;
+  --fdSwitch_Handle_Width: 1.5rem;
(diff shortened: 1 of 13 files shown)


src/styles/theming/common/switch/_sap_horizon.scss
```diff
@@ -0,0 +1,74 @@
+:root {
```
platon-rov, line 1: i like this approach but we need to unify the chosen solution over the project. if we're going with separate files for each component - let's do it for all components. if we're placing all the variables in one file - let's do this for all components.

let's raise that question on the call so we can discuss it and agree with the team.

N1XUS: > i like this approach but we need to unify the chosen solution over the project. if we're going with separate files for each component - let's do it for all components. if we're placing all the variables in one file - let's do this for all components.
> 
> let's raise that question on the call so we can discuss it and agree with the team.

I chose this approach because amount of changed variables is enormous. It would be unreadable in case of storing such variables in one file per theme. Another thing is that for all fiori variables are the same, thus, copy-pasting this variables into each theme file doesn't make much sense.

InnaAtanasova: You can't just choose approach without discussing it with the team. There are 30+ components in this library. If you do it for this one you need to update the rest. Fundamental-styles will not become the mess fundamental-ngx is, where each dev is influencing their way of coding. And I totally +1 what Platon says, it's not about the approach, it's about the consistency. If we do it for one, we do it for all components.

g-cheishvili:
image
Semantic switch is a bit off, on Horizon

g-cheishvili (changes requested):
On horizon, semantic compact switch seems to be overflowing. See previous comment. Other than that, very good job!

mikerodonnell89 (commented):
The docs mention that the "touchable height" for cozy switches is 2.75rem (44px) but here I am seeing 36px. Compact looks good

src/styles/theming/common/switch/_sap_horizon.scss
platon-rov: Discussed, agreed.

platon-rov (commented):
* Horizon, no icon

Current:
image

Specs:
image

* Horizon, Circle should not be colored in HCW/HCB

Current:
image

Specs:
image

N1XUS:
@platon-rov that's agreed with @droshev that we do not change html markup, and since for fiori icons are rendered with html markup, we skip it for horizon

platon-rov (approved):

mikerodonnell89 (approved):
Looks good. Can you remove the "code needs fixing" text from the label example before merge?

PR #3267, opened 2022-03-28
feat(styles): add delta horizon theming support for Token [ci visual]

InnaAtanasova:
## Related Issue

## Description
add delta horizon theming support for Token 
Tokenizer will be implemented separately with the Inputs

### DIFF
--- a/src/styles/theming/sap_fiori_3.scss
+++ b/src/styles/theming/sap_fiori_3.scss
@@ -121,6 +121,7 @@
 
   /* Token */
   --fdToken_Text_Shadow: none;
+  --fdToken_Border_Radius: 0.25rem;
   --fdToken_Text_Color: var(--sapTextColor);
   --fdToken_Padding_Left: 0.3125rem;
   --fdToken_Background_Color_Hover: var(--sapButton_Hover_Background);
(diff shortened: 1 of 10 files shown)


mikerodonnell89 (commented):
Here I am seeing read-only tokens get a shadow when hovered, but it looks like the general input specs say nothing should happen when read-only inputs get hovered, should that be the case here as well? Nothing is specifically mentioned for read-only hovered tokens but I'm wondering if we can just assume the same as inputs for tokens

platon-rov (commented):
* Horizon, Focus kinda different from the specs

Current:

image

Specs:

image

* Horizon, Read-Only tokens incorrect in HCB/HCW

Current:

image

Specs:

image

* Horizon, Hover styles present only for Toggled tokens in HCW/HCB

* Horizon, Tokens in HCB/HCW should have smaller border-radius

image

InnaAtanasova:
> * Horizon, Focus kinda different from the specs
> 
> Current:
> 
> image
> 
> Specs:
> 
> image
> 
> * Horizon, Read-Only tokens incorrect in HCB/HCW
> 
> Current:
> 
> image
> 
> Specs:
> 
> image
> 
> * Horizon, Hover styles present only for Toggled tokens in HCW/HCB
> * Horizon, Tokens in HCB/HCW should have smaller border-radius
> 
> image

Point 1 about the focus: I disagree. The first image is Cosy Token and per specs the focus has an offset:
Specs:
Screen Shot 2022-03-31 at 4 43 25 PM

Fund-styles:
Screen Shot 2022-03-31 at 4 43 49 PM

Specs compact:
Screen Shot 2022-03-31 at 4 44 16 PM

Fund-styles compact:
Screen Shot 2022-03-31 at 4 44 23 PM

Point 2: agreed, the variables have been updated. After discussing with designers today I also removed the hover state for read-only


Point 3: I don't understand what you mean, please provide more info or let's discuss it during the daily

Point 4: I am not sure what you mean.The radius is a variable with doesn't have a delta for HCW/HCB. It's coming from the theming css variables and we can't control this.

InnaAtanasova:
> Here I am seeing read-only tokens get a shadow when hovered, but it looks like the general input specs say nothing should happen when read-only inputs get hovered, should that be the case here as well? Nothing is specifically mentioned for read-only hovered tokens but I'm wondering if we can just assume the same as inputs for tokens

Mike, it's now fixed.

platon-rov (approved):

PR #3268, opened 2022-03-29
feat(styles): add delta horizon theming support for Bar [ci visual]

InnaAtanasova:
## Related Issue
pat of https://github.com/SAP/fundamental-styles/issues/3213

## Description
- add delta horizon theming support for Bar component


mikerodonnell89 (commented):
Here I am seeing read-only tokens get a shadow when hovered, but it looks like the general input specs say nothing should happen when read-only inputs get hovered, should that be the case here as well? Nothing is specifically mentioned for read-only hovered tokens but I'm wondering if we can just assume the same as inputs for tokens

PR #3269, opened 2022-03-29
feat(styles): add delta horizon theming for Toolbar [ci visual]

InnaAtanasova:
## Related Issue
part of https://github.com/SAP/fundamental-styles/issues/3213

## Description
 - add delta horizon theming for Toolbar 
 - updated Toolbar title

BREAKING CHANGE:
- updated the markup for the title element in the Toolbar with Title 
- added a new class for Title used in Toolbar: fd-toolbar__title

**Before:**
```
<h4>Producs (104)</h4>
```

**After:**
```
<h4 class="fd-title fd-title--h4 fd-toolbar__title">Products (104)</h4>
```

### DIFF
--- a/src/styles/theming/sap_fiori_3.scss
+++ b/src/styles/theming/sap_fiori_3.scss
@@ -290,6 +290,7 @@
 
   /* Toolbar */
   --fdToolbar_Auto_Background: inherit;
+  --fdToolbar_Solid_Background: var(--sapBackgroundColor);
 
   /* Tool Header */
   --fdTool_Header_Icon_Color_Default: var(--sapShell_TextColor);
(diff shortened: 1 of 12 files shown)


src/styles/theming/sap_fiori_3.scss
```diff
@@ -290,6 +290,7 @@
 
   /* Toolbar */
   --fdToolbar_Auto_Background: inherit;
+  --fdToolbar_Solid_Background: var(--sapBackgroundColor);
```
mikerodonnell89, line 293: Why do we need a variable for this if it's the same value for each theme?

InnaAtanasova: nice catch, thank you! I updated the value for Horizon. It should be `--sapToolbar_Background` for Horizon and `--sapBackgroundColor` for Quartz in Solid design.

platon-rov (commented):
* ~~Horizon, Transparent buttons should have blue color~~ _Button Related_

Current:
image

Specs:
image

platon-rov (approved):

PR #3271, opened 2022-03-30
feat(styles): add delta theming for Bar [ci visual]

InnaAtanasova:
## Related Issue
part of https://github.com/SAP/fundamental-styles/issues/3213

## Description
- add delta theming for Bar
- removed unused class `fd-bar__element--title`

BREAKING CHANGE:
removed unused class `fd-bar__element--title`

Before:
```
<div class="fd-bar__element fd-bar__element--title">
      <h6 class="fd-title fd-title--h6" aria-label="text">TEXT</h6>
</div>
```

After:
```
<div class="fd-bar__element">
      <h6 class="fd-title fd-title--h6" aria-label="text">TEXT</h6>
</div>
```

### DIFF
--- a/src/styles/bar.scss
+++ b/src/styles/bar.scss
@@ -4,7 +4,7 @@
 $block: #{$fd-namespace}-bar;
 
 .#{$block} {
-  $fd-bar-padding-x: 0.5rem !default;
+  $fd-bar-padding-x: 1rem !default;
   $fd-bar-element-spacing: 0.5rem !default;
   $fd-bar-floating-footer-z-index: map-get($fd-z-index-levels, "top") !default;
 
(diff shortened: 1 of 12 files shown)


platon-rov (commented):
* Fiori, Items sticked to bottom border in the Default example, but not in the Header example, HCW/HCB

image
image

* Horizon, Transparent buttons should have blue color of the icons

image

P.S. I noticed that segmented buttons have incorrect background when toggled in HCW/HCB but it's related to the button specifications and will be resolved separately.

InnaAtanasova:
> https://deploy-preview-3271--fundamental-styles.netlify.app/

Well, it's because they are different :)
In one case you have compact bar with cosy elements (the height of the Bar is 40px and that's why the elements "touch" the bottom border) and in the other case you have a cosy bar (height 44px) and cosy elements. 
Screen Shot 2022-03-31 at 1 47 33 PM
Screen Shot 2022-03-31 at 1 47 12 PM

It's not something related to the theming part

platon-rov (approved):

PR #3295, opened 2022-04-05
fix: select min-width fix, overflow fix

platon-rov:
## Related Issue

Relates to https://github.com/SAP/fundamental-ngx/issues/7929

## Description

Select component:
* Min-width fix,
* Selected item overflow.

### Before:

image

image

### After:

image

image

### DIFF
--- a/src/styles/select.scss
+++ b/src/styles/select.scss
@@ -29,6 +29,10 @@ $block: #{$fd-namespace}-select;
     }
   }
 
+  @mixin fd-select-set-min-width($width) {
+    min-width: calc(#{$width} - var(--sapField_BorderWidth) * 2);
+  }
+
   @include fd-reset();
 
   text-shadow: var(--fdSelect_Text_Shadow);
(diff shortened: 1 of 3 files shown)


aberikashvili (approved):
LGTM

src/styles/select.scss
```diff
@@ -107,7 +111,7 @@ $block: #{$fd-namespace}-select;
       .#{$block}__text-content {
         @include fd-set-padding-left($fd-select-padding-x-compact);
 
-        min-width: 2rem;
+        min-width: calc(2.75rem - var(--sapField_BorderWidth) * 2);
```
InnaAtanasova, line 114: repeatable code that can be avoided and easily maintained with a mixin:
```
@mixin fd-select-set-min-width($width) {
     min-width: calc($width - var(--sapField_BorderWidth) * 2);
}
```

and then call it with 2.5rem, 2.75rem and 5rem:
```
@include fd-select-set-min-width(5rem);
```

platon-rov: Good catch, thank you!

InnaAtanasova (approved):

PR #3297, opened 2022-04-05
feat(styles): add delta theming for Dynamic Page [ci visual]

InnaAtanasova:
## Related Issue

## Description
add delta theming for Dynamic Page

### DIFF
--- a/src/styles/theming/common/switch/_sap_fiori.scss
+++ b/src/styles/theming/common/switch/_sap_fiori.scss
@@ -70,4 +70,13 @@
   --fdSwitch_Compact_Semantic_Handle_Offset: -1.9375rem;
   --fdSwitch_Compact_Semantic_Active_Handle_Offset: 0.0625rem;
   --fdSwitch_Compact_Semantic_Active_Handle_Offset_Rtl: var(--fdSwitch_Active_Handle_Offset);
+
+  // Dynamic Page Layout
+  --fdDynamicPage_Title_Color: var(--sapGroup_TitleTextColor);
+  --fdDynamicPage_Title_Font_Size: var(--sapFontHeader3Size);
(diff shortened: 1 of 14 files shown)


src/styles/theming/common/switch/_sap_fiori.scss
```diff
@@ -70,4 +70,13 @@
   --fdSwitch_Compact_Semantic_Handle_Offset: -1.9375rem;
   --fdSwitch_Compact_Semantic_Active_Handle_Offset: 0.0625rem;
   --fdSwitch_Compact_Semantic_Active_Handle_Offset_Rtl: var(--fdSwitch_Active_Handle_Offset);
+
+  // Dynamic Page Layout
```
platon-rov, line 74: It's probably shouldn't be here, in switch styles.

InnaAtanasova: oh, nice catch! I didn't see the `switch` folder, just the `common` folder.  :D

platon-rov (commented):
* font sizes are different, but should be the same, 0.875rem

Screenshot 2022-04-14 at 14 51 13

* I didn't find any example like this in wiki, but i assume that we're missing gap between items

image

mikerodonnell89 (changes requested):
In quartz light, these buttons have lost their hover and down states

Screen Shot 2022-04-14 at 1 29 17 PM

InnaAtanasova:
> * font sizes are different, but should be the same, 0.875rem
> 
> Screenshot 2022-04-14 at 14 51 13
> 
> * I didn't find any example like this in wiki, but i assume that we're missing gap between items
> 
> image

I don't think this is the correct PR :)

InnaAtanasova:
> https://deploy-preview-3297--fundamental-styles.netlify.app

fixed

mikerodonnell89 (approved):

PR #3299, opened 2022-04-06
fix: table text shadow removal

platon-rov:
## Related Issue

Relates to https://github.com/SAP/fundamental-ngx/issues/7930

## Description

Text shadow moved from the table to the table text element. 

Previously it was added to the cells in pr https://github.com/SAP/fundamental-styles/pull/1571 but in that way it breaks other components styles, see the mentioned issue. 

### Before

image

### After

image

## BREAKING CHANGE

* Now you should wrap text in element with `fd-table__text` class to apply text shadow to it.

### DIFF
--- a/src/styles/table.scss
+++ b/src/styles/table.scss
@@ -15,15 +15,17 @@ $block: #{$fd-namespace}-table;
   @mixin fd-table-active() {
     background-color: var(--sapList_Active_Background);
     color: var(--sapList_Active_TextColor);
-    text-shadow: none;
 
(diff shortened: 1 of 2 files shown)


InnaAtanasova (changes requested):
Sorry, approved by mistake :)
You need to fix the documentation as now the elements with the following markup `<td class="fd-table__cell">Middle Name</td>` won't have text-shadow.

InnaAtanasova (approved):

PR #3304, opened 2022-04-08
feat(styles): add delta horizon theming for Input

StevenZhao318:
## Related Issue

## Description
Horizon support for input and input group dependent components

Breaking changes: 
* fd-button—transparent is no longer used in select and input-group components. Buttons used in semantic inputs should have the class: fd-button--{type}
    * fd-button--positive
    * fd-button--negative
    * fd-button--attention
    * fd-button—information

* The expanded Select component’s dropdown button should be set to the active state to receive proper styling.


src/styles/mixins/_forms.scss
```diff
@@ -4,13 +4,14 @@
 
 @mixin fd-form-states {
 
-  $fd-form-border-thin-width: 0.0625rem;
-  $fd-form-border-thick-width: 2 * $fd-form-border-thin-width;
+  // $fd-form-border-thin-width: 0.0625rem;
+  // $fd-form-border-thick-width: 2 * $fd-form-border-thin-width;
```
platon-rov, line 8: Remove this if it's not needed anymore.

src/styles/mixins/_forms.scss
```diff
@@ -90,7 +95,7 @@
   &.is-alert,
   &.is-information,
   &.is-error {
-    border-width: $fd-form-border-thick-width;
+    // border-width: $fd-form-border-thick-width;
```
platon-rov, line 98: Remove also.

src/styles/mixins/_forms.scss
```diff
@@ -219,3 +224,198 @@
     margin-right: 0;
   }
 }
+
+@mixin fd-form-input() {
```
platon-rov, line 228: It's the same we have above, at the top of the file, in `fd-form-states` mixin. Can we remove one of these and left only one?

src/styles/mixins/_states.scss
```diff
@@ -46,6 +28,17 @@
   }
 }
 
+// FOCUS state (tab onto, click into)
+// override params $shadow-var, $shadow-value
+// `.is-focus-within` is for demo purposes
+@mixin fd-focus-within() {
```
platon-rov, line 34: Not used anywhere.

src/styles/theming/sap_fiori_3.scss
```diff
@@ -141,13 +144,46 @@
...
+  --fdInput_Outline_ErrorColor: var(--sapContent_FocusColor);
+  --fdInput_Outline_WarningColor: var(--sapContent_FocusColor);
+  --fdInput_Outline_SuccessColor: var(--sapContent_FocusColor);
+  --fdInput_Outline_InformationColor: var(--sapContent_FocusColor);
+  --fdInput_Base_Focus_Background: var(--sapField_Background);
+  --fdInput_Error_Focus_Background: var(--sapField_InvalidBackground);
+  --fdInput_Warning_Focus_Background: var(--sapField_WarningBackground);
+  --fdInput_Success_Focus_Background: var(--sapField_SuccessBackground);
+  --fdInput_Information_Focus_Background: var(--sapField_InformationBackground);
+  --fdInput_Readonly_Focus_Background: var(--sapField_ReadOnly_Background);
```
platon-rov, line 186: Seems like a lot of variables here, maybe we could move it to input's separate file?

platon-rov (commented):
Also there are a lot of breaking self-contained styles rule but i think it's okay as all these input components highly tied up, fyi @InnaAtanasova

src/styles/input-group.scss
```diff
@@ -78,19 +85,29 @@ $fd-input-border-radius: var(--sapField_BorderCornerRadius);
     }
   }
 
+  .fd-tokenizer {
```
mikerodonnell89, line 88: This should be done in tokenizer.scss

mikerodonnell89 (changes requested):
Nice work! Readonly and disabled should have no hover/active/focus styles

mikerodonnell89:
> Also there are a lot of breaking self-contained styles rule but i think it's okay as all these input components highly tied up, fyi @InnaAtanasova

On first glance I think the desired changes can be made in their respective files without breaking this rule

src/styles/input.scss
```diff
@@ -8,6 +8,8 @@ $block: #{$fd-namespace}-input;
 
 .#{$block} {
   @include fd-form-text();
+  // @include fd-form-input();
```
N1XUS, line 11: Remove commented code

src/styles/input-group.scss
StevenZhao318: this is to remove the background for when the tokenizer is wrapped inside the input-group only

mikerodonnell89: This breaks our rule that you can not set properties of one component from another component's scss file. And I see that you have set the background to none in tokenizer.scss so you should be fine to delete this bit

mikerodonnell89 (changes requested):

platon-rov (changes requested):
* Read-only state for all the inputs looks ugly, is it intended? 

image

* Input group, button keeps its shadow after click

image

* Input group, button has incorrect shadow color

Current:
image

Specs:
image

* I'm not sure that disabled variants of any inputs should have any interaction states. For the disabled button no shadows/borders are applied on any interaction state, for example.

image

* On down/active state background color of the button should be different

image

* Select looks incorrect

Current:
image

Specs:
image

* Select value in Mobile mode not vertically centered

image

* Disable Step Input isn't actually disabled

image

* Tokenizr should have background, bottom border, rounded corners

image

* Tokenizr, is it intended to have field visually separated? See nothing in the docs

image

mikerodonnell89:
> * Read-only state for all the inputs looks ugly, is it intended?
> 
> image

setting border-bottom-width to 0 fixes this

mikerodonnell89:
The specs indicate that there are no hover/active states for readonly and disabled inputs. So I think nothing should happen on hover and these should not receive focus outlines

src/styles/input.scss
```diff
@@ -8,6 +8,8 @@ $block: #{$fd-namespace}-input;
 
 .#{$block} {
   @include fd-form-text();
+  @include fd-form-states();
```
droshev, line 11: mixin `fd-form-text` in doing the same thing, so you can remove it

StevenZhao318: the fd-form-states gives the fiori semantic inputs border color.
it also adds the box shadow around the input for horizon styles when you hover over it. 

These 2 were separated because the standonline tokenizer does not seem do have any interactions involving hover/focus as per the spec. (It uses the parent component's interactions when used within input-group)

droshev:
It has been moved into #3386

Issue #3300, opened 2022-04-07
NPM package, unpkg and Message page issue


StanZGenchev:
**Describe the bug**

When I use the npm package (either locally of through unpkg) and try to render `fd-message-page`, I get an empty page with a missing icon (screenshot 1). However, if I build fundamental-styles locally and copy the dist folder over the dist folder from the npm package, it starts working (screenshot 2).
The code I use is 100% identical from the documentation.
A quick comparison between the local dist folder and the npm package dist folder reveals that in the local dist folder, all the css widgets also have a specific theme css (for example `bar-sap_fiori_3.css`) along with the generic one (`bar.css`).
The unpkg link from the "Introduction" documentation page does not work:
```
<link href='https://unpkg.com/fundamental-styles@0.23.0/dist/fundamental-styles-sap_fiori_3.css' rel='stylesheet'>
```

It says that the file does not exist. When I install the npm package locally and inspect its contents, I don't find the theme files either.

This is the html code for the message page:

```html
<html>
  <head>
    <meta charset="utf-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1" />
    <title>Test</title>
    <link type="text/css" rel="stylesheet" href="/node_modules/fundamental-styles/dist/fundamental-styles.css" />
    <link type="text/css" rel="stylesheet" href="/node_modules/fundamental-styles/dist/theming/sap_fiori_3.css" />
    <link type="text/css" rel="stylesheet" href="/node_modules/@sap-theming/theming-base-content/content/Base/baseLib/sap_fiori_3/css_variables.css" />
    <!-- For SAP fonts and icons: -->
    <link type="text/css" rel="stylesheet" href="theme.css" />
  </head>
  <body>
    <div class="fd-message-page">
      <div class="fd-message-page__container">
        <div class="fd-message-page__icon-container">
          <i role="presentation" class="sap-icon--search fd-message-page__icon"></i>
        </div>
        <div role="status" aria-live="polite" class="fd-message-page__content">
          <div class="fd-message-page__title">
            No matching items found.
          </div>
        </div>
      </div>
    </div>
  </body>
</html>
```

[...]

PR #3325, opened 2022-04-12
fix(styles,fn,deps,ci): removed third party postcss-clean and updated cssnano

g-cheishvili:
## Related Issue
Closes SAP/fundamental-styles#3300

## Description
This PR updates cssnano and replaces third-party library, which was wrapping clean-css library(postcss-clean) with custom sync plugin. Also replaced clean-css library's level "1" with level "2", for more aggressive removal of duplicated css properties


g-cheishvili:
To reviewers: Please compare before and after outputs

platon-rov (commented):
dist got bigger, is it intended? 

image

I've compared old table.css and new, you can check it here https://www.diffchecker.com/O6UiQ1Az and seems like new adds more tabulation, semicolons, leading zeros to the values (was .5rem, now 0.5rem etc), should we do that? but in new the selectors one-lined, that's okay.

g-cheishvili:
Few bytes of difference is acceptable, but it should not be 100kb 😄 Few bytes of difference comes from `convertValues` plugin, which had bug before, it was not transforming values correctly

g-cheishvili:
Oh, @platon-rov , versions are not same in your test. you are comparing rc29 with rc24. can you use only rc24 in both cases?

g-cheishvili:
Now it can be checked with actual main branch. My comparisons 
Old vs New in bytes
Gzip fundamental-styles.css: 107 037 vs 106 680
Size on disk of whole dist folder: 3 182 592 vs  3 125 248

Gains are not substantial but as far as I could check works as expected

droshev (approved):

PR #3328, opened 2022-04-14
fix(style): button refactoring

platon-rov:
## Related Issue

Closes none.

## Description

Button components refactoring.

## Screenshots

Chromatic detects many changes but they're actually not relevant.


mikerodonnell89 (approved):

src/styles/button.scss
```diff
@@ -17,34 +11,26 @@ $fd-button-height: 2.25rem;
...
 $fd-button-compact-clickable-height: 2rem;
 $fd-button-compact-touchable-area: ($fd-button-compact-clickable-height - $fd-button-compact-height) * 0.5;
 $fd-button-compact-padding-x: calc(0.5rem - #{$fd-button-border-width});
-$fd-button-compact-min-width: 2rem;
-$fd-button-compact-line-height: calc(#{$fd-button-compact-height} - 2 * #{$fd-button-border-width});
 
-$fd-menu-button-max-width: 12rem;
-
-$block: #{$fd-namespace}-button;
+.#{$block-button} {
```
N1XUS, line 20: I think we use just `$block` naming for the root selector

platon-rov: file where this variable is defined is imported by the many other files, it's done to avoid any collisions.

N1XUS (approved):

InnaAtanasova (changes requested):
I'd suggest you revisit the focus concept:
1. In the specs I don't see focus on `down` state, but I might be wrong. In case I am wrong, then the color used in this case for sure is wrong (should be contrast).
Screen Shot 2022-04-25 at 3 50 45 PM
Screen Shot 2022-04-25 at 3 50 14 PM
 
2. This is the case with Horizon (styles vs specs)
Screen Shot 2022-04-25 at 3 52 16 PM
Screen Shot 2022-04-25 at 3 51 56 PM

3. The offset should be double-checked. In h
Screen Shot 2022-04-25 at 3 35 40 PM
Screen Shot 2022-04-25 at 3 33 39 PM
Screen Shot 2022-04-25 at 3 33 33 PM
igh contrast themes it's really off:

platon-rov:
> In the specs I don't see focus on down state, but I might be wrong. In case I am wrong, then the color used in this case for sure is wrong (should be contrast).

It's the same as for the toggle button, fixed.

> This is the case with Horizon

When we're pressing down the mouse button on the button element it gets both states, focus & active, it's intended behavior. That's why we see that focus ring. To change it a big amount of work should be done, also for many components, cause we have the same behavior everywhere. I'm not sure we should fix it in scope of this pr.

> The offset should be double-checked. In high contrast themes it's really off

We have `outline-offset: -0.1875rem;` for both - fiori & horizon, although we changed this value in horizon to `0.25rem` to get our visual closer to the screenshots. Probably we should raise that topic with the ui team, why when we're using variables correctly we end up with the visuals that don't match the screenshots in the wiki. 

About this exact example & split button

image

It won't be ideal unfortunately. You can see that we have single border between the buttons, thou there should be double border (border-right of the left button and border-left of the right button) and we cannot set `outline-offset-left` for the right button, we just can manage the whole `outline-offset` and if we change it - it would be broken on the other three sides. Probably, as an option we can draw outline via the pseudo elements (before/after) in this exact case, however it's not the correct solution and a lot of hardcode will come. What do you think?

src/styles/button.scss
```diff
@@ -60,215 +46,32 @@ $block: #{$fd-namespace}-button;
...
-    &.#{$block}--toggled {
-      @include standardToggledHoverPressed();
-    }
+    font-weight: inherit;
+    color: inherit;
   }
 
-  &.#{$block}--transparent {
-    @include transparentHover();
+  * + * {
```
g-cheishvili, line 58:
```suggestion
  *:not(:first-child) {
```
It's a longer, but easier to grasp

g-cheishvili (approved):
Once @InnaAtanasova 's notes are applied I do not have anything against this PR. Code is VERY good!

droshev:
> https://deploy-preview-3328--fundamental-styles.netlify.app

We are not changing the design and behaviour but the implementation, so one of the expectations is we won't have regression with the current visuals/chromatic(excluding when we fix bugs)

InnaAtanasova (approved):

PR #3330, opened 2022-04-15
feat(styles): add delta horizon theming for Grid List

platon-rov:
## Related Issue

Part of #3213.

## Description

Horizon support for the Grid List component.

### Before:

image

image

image

### After:

image

image

image

## BREAKING CHANGE
* `fd-grid-list__btn-navigation` to `fd-grid-list__item-navigation-indicator` class change, element shouldn't be a button anymore
* Grid List Item body markup changed a lot

https://github.com/SAP/fundamental-styles/wiki/Breaking-Changes#grid-list-3330


mikerodonnell89 (changes requested):
In Fiori mode, I see the alignment of the content is improved to better match the designs. But seeing this checkbox is no longer centered with the avatar

Screen Shot 2022-04-19 at 1 19 38 PM
 
Also the specs say that all action buttons including the right navigation arrow should simply be standard transparent buttons, but here the navigation button is no longer a button

mikerodonnell89:
Per the specs, the highlight should be z-index'd under the focus outline

Screen Shot 2022-04-19 at 1 41 48 PM

platon-rov:
> In Fiori mode, I see the alignment of the content is improved to better match the designs. But seeing this checkbox is no longer centered with the avatar

Fixed

> Also the specs say that all action buttons including the right navigation arrow should simply be standard transparent buttons, but here the navigation button is no longer a button

This is intended.

image
image
image

> Per the specs, the highlight should be z-index'd under the focus outline

Probably here is the best we could reach. As outline isn't a pseudo element we cannot force it to be z-indexed above than highlight element.

image

src/styles/theming/sap_fiori_3_light_dark.scss
```diff
@@ -228,6 +228,9 @@
 
   /* Grid List */
   --fdGrid_List_Item_Hover_Border_Color: var(--sapTile_Interactive_BorderColor);
+  --fdGrid_List_Item_Border_Radius: var(--sapElement_BorderCornerRadius);
+  --fdGrid_List_Item_Toolbar_Shadow: var(--sapContent_HeaderShadow);
+  --fdGrid_List_Highlight_Offset: 0.1875rem;
```
N1XUS, line 233: From what I've noticed, css variables for all fiori-related themes are the same. Maybe move it to a single separate file and import it per theme?

N1XUS (approved):

src/styles/theming/sap_fiori_3_light_dark.scss
platon-rov: I thought we can left it as it is because there are no many variables, but okay, created separate files and placed it there.

InnaAtanasova (approved):

mikerodonnell89 (approved):

Issue #3303, opened 2022-04-07
Chromatic optimizations


droshev:
Chromatic testing optimizations/improvements:
- [ ] remove testing the FM components
- [ ] add testing Morning Horizon for now

PR #3343, opened 2022-04-18
chore(e2e,fn): removed fn tests and storyshots

g-cheishvili:
## Related Issue
Closes SAP/fundamental-styles#3303

## Description

- Moved fiori next components to separate folder in src to exclude them from visual stories generation easier.
- Added Horizon into the visual stories generation.
- Removed fiori next components from jest tests.
- Updated storyshots


scripts/build-visual-stories.js
```diff
@@ -13,8 +13,8 @@ const themes = [
     { value: 'sap_fiori_3', title: 'Quartz Light' },
     // { value: 'sap_fiori_3_dark', title: 'Quartz Dark' },
     // { value: 'sap_fiori_3_hcw', title: 'High Contrast White' },
-    { value: 'sap_fiori_3_hcb', title: 'High Contrast Black' }
-    //{ value: 'sap_horizon', title: 'Horizon' }
+    { value: 'sap_fiori_3_hcb', title: 'High Contrast Black' },
+    { value: 'sap_horizon', title: 'Morning Horizon' }
```
platon-rov, line 17: We for sure will be overexciting limit very ofter with this new setting applied.

We should increase the amount of storyshots available per month or specify which stories will be built for certain pr, that's probably possible as we have `components` array in each storyshot file, that explicitly points to which components are used in the stories.

platon-rov (commented):
We should separate

scripts/build-visual-stories.js
g-cheishvili: FN components were removed from tests and one theme was added instead. I think this will more or less balance each other

platon-rov (approved):

droshev (approved):

droshev (changes requested):
when i run `npm run test:update` there are bunch of new files created
Screen Shot 2022-04-18 at 18 48 16

droshev (approved):

Issue #3344, opened 2022-04-18
bug: list missing features or examples


platon-rov:
**Describe the bug**

There are some features missing from the list component:

* Selected items count info bar
* Wrapping feature for list with Byline (see details below)
* No data

These features might be present in styles but isn't shown in the docs. 

Review & document them.

### Wrapping feature for list with Byline

We have this feature partially present in the docs, see `List with long Title and Byline` example. But there are some differences between standard list with byline & wrapping one:

* Height: 5rem fixes for standard, not limited for wrapping variant.
* Byline, padding top: 0.5rem for standard, 0.75rem for wrapping variant.

> Is this issue related to a specific component?

List

> What version of the Fundamental Library Styles are you using?

0.24.0-rc.35

PR #3345, opened 2022-04-18
feat(styles): add delta horizon theming for List

platon-rov:
## Related Issue

Part of #3213.

## Description

List & vertical nav styles refactoring, horizon support implemented for the List component.

Refactoring were done to reduce code coupling and code repetition, improve naming and code styling, make code easier to read. Also it reduced bundle size for around 30kb (3 679Kb vs 3 644Kb).

_Note: Please note that some features are missing\not represented in docs, here is the ticket https://github.com/SAP/fundamental-styles/issues/3344, nothing were done for them as these features should be implemented\shown in docs first._

### Before:

image

### After:

image


g-cheishvili:
I think distance between focus outline and select underline is too big compared to the specification. This also applies to other cases with focus outline

image
image

g-cheishvili:
image
This action button should not have outline and in general I think this needs revision, because margin should not be here too

g-cheishvili (changes requested):
Read my comments

platon-rov:
> I think distance between focus outline and select underline is too big compared to the specification. This also applies to other cases with focus outline

It might looks different, but here is the specs

image

And we have this offset set exactly to -0.125rem.

> This action button should not have outline and in general I think this needs revision, because margin should not be here too 

It doesn't have, seems like you see it because themes still not changing completely and we see some elements' styles from different themes.

> By specs height of the byline list item should be 80px, but we have 89 and 81, depending on it's content. If there is description it makes it higher

Yea, but there is also wrapping feature should be present so we cannot limit height to 5rem, check description of the https://github.com/SAP/fundamental-styles/issues/3344, i've described this item there so we can finish implementation for this one separately.

mikerodonnell89 (changes requested):
When an item is clicked it is also rendering a white focus outline, but there's nothing in the specs that say we should have that

Screen Shot 2022-04-19 at 1 55 54 PM

platon-rov:
> If you do bottom border with ::after instead of border-bottom it will have correct distance. In design they mean inset border

In my opinion we shouldn't use pseudo elements to achieve this styling. We also can set inset box shadow, that is less hacky but i'm not sure, that property also should be used for different purposes. Let's agree on this one on the call.

> When an item is clicked it is also rendering a white focus outline, but there's nothing in the specs that say we should have that

Outline color changed to blue. But we cannot avoid applying it as focus state applies automatically on mouse up.

mikerodonnell89 (approved):

src/styles/mixins/vertical-nav/_vertical-nav-navigation.scss
```diff
@@ -0,0 +1,374 @@
...
+          margin-left: $fd-list-navigation-item-indicator-spacing;
+        }
+      }
+    }
+
+    &__navigation-item {
+      @include fd-reset();
+      @include fd-fiori-focus();
+      @include fd-set-paddings-x(0.5rem, 1rem);
+      @include _fd-vertical-nav-selected-color();
```
InnaAtanasova, line 43: why is this prefixed with `_`?

platon-rov: I tried to achieve "private" mixins, i.e. that will be used only within this file.

g-cheishvili (approved):

g-cheishvili (changes requested):
Check out the diff on chromatic, few odd changes

platon-rov:
> Check out the diff on chromatic, few odd changes

Some of the changes are relevant, fixed the issues. Some others come from the another prs, that's strange. Looks like baseline calculated not correctly, maybe because of the rebasing. I think we can investigate this after we're done with horizon stuff and not hurrying anymore.

g-cheishvili (approved):

PR #3350, opened 2022-04-19
feat(styles): add delta horizon theming for Object List

platon-rov:
## Related Issue

Part of #3213.

## Description

Object list styles refactoring, horizon support implementation.

_Note: In Horizon specs colors of the Object Statuses in active state are remain the same and in Horizon they're made be one color. Unfortunately, we cannot make them one color in Fiori and don't change in Horizon, there are no any options to conditionally apply the styling._

Current:
image

Specs:
image

### Before:

image

### After:

image

### DIFF
--- a/src/styles/object-list.scss
+++ b/src/styles/object-list.scss
@@ -4,9 +4,8 @@
 
 $block: #{$fd-namespace}-object-list;
 
-$fd-object-list-with-padding-value-half: 0.5rem;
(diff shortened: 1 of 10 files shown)


N1XUS (changes requested):
Good job. One issue found:

* Focused list item has outline. Design specs does not show any outline for such state.

InnaAtanasova (approved):

PR #3273, opened 2022-03-31
feat(styles): add delta theming for Calendar [ci visual]

mikerodonnell89:
part of #3213

PR #3357, opened 2022-04-25
fix(styles): calendar year/month buttons

platon-rov:
## Related Issue

Closes none.
Follow-up https://github.com/SAP/fundamental-styles/pull/3273.

## Description

Year & Month buttons markup refactoring inside the Calendar component.

## Screenshots

Shouldn't be any changes.

## BREAKING CHANGE

* markup changed for the year/month items in the calendar. 

_Breaking change described in section https://github.com/SAP/fundamental-styles/wiki/Breaking-Changes#calendar-3273_

### DIFF
--- a/stories/calendar/calendar.stories.js
+++ b/stories/calendar/calendar.stories.js
@@ -343,68 +343,68 @@ export const Months = () => `<div style="display: flex; justify-content: space-a
                         <tr class="fd-calendar__row">
                             <td role="gridcell" class="fd-calendar__item fd-calendar__item--current">
                                 <button class="fd-button fd-button--transparent fd-calendar__item-button" type="button">
-                                    <span class="fd-calendar__text" role="button">January</span>
+                                    <span class="fd-calendar__text">January</span>
                                 </button>
                             </td>
                             <td role="gridcell" class="fd-calendar__item">
                                 <button class="fd-button fd-button--transparent fd-calendar__item-button" type="button">
-                                    <span class="fd-calendar__text" role="button">February</span>
+                                    <span class="fd-calendar__text">February</span>
                                 </button>
                             </td>
                             <td aria-selected="true" role="gridcell" class="fd-calendar__item">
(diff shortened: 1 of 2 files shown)


stories/calendar/calendar.stories.js
```diff
@@ -341,47 +341,71 @@ export const Months = () => `<div style="display: flex; justify-content: space-a
                 <table class="fd-calendar__table" role="grid">
                     <tbody class="fd-calendar__group">
                         <tr class="fd-calendar__row">
-                            <td role="gridcell" class="fd-calendar__item fd-calendar__item--current fd-button fd-button--transparent">
-                                <span class="fd-calendar__text" role="button">January</span>
+                            <td role="gridcell" class="fd-calendar__item fd-calendar__item--current">
+                                <button class="fd-button fd-button--transparent fd-calendar__item-button" type="button">
```
InnaAtanasova, line 345: why do you need `type="button"` when you have a native button element?

stories/calendar/calendar.stories.js
```diff
@@ -341,47 +341,71 @@ export const Months = () => `<div style="display: flex; justify-content: space-a
...
+                                    <span class="fd-calendar__text" role="button">January</span>
```
InnaAtanasova, line 346: should this have a role of button? isn't it just text now?

InnaAtanasova (changes requested):
The focus is no longer working for Calendar.

InnaAtanasova:
Screen Shot 2022-04-25 at 12 18 05 PM

Hover on selected state is not correct.

stories/calendar/calendar.stories.js
InnaAtanasova: You marked this as resolved but neither changed it nor responded :) 
`<button type="button>` is different from a simple `<button>` with a blank or missing type attribute only in a `form`. There the default type is submit so it makes sense to specify the type, but in the Calendar case I don't see why it's needed.

InnaAtanasova (approved):

PR #3358, opened 2022-04-25
feat(styles): popover body compact, arrow position fix

platon-rov:
## Related Issue

Closes none.

## Description

Popover's arrow placement fix.

### Before:

image

### After:

image

### DIFF
--- a/src/styles/popover.scss
+++ b/src/styles/popover.scss
@@ -15,8 +15,8 @@ $block: #{$fd-namespace}-popover;
   $fd-popover-arrow-width: 1rem !default;
   $fd-popover-arrow-height: 0.5rem !default;
   $fd-popover-arrow-width-half: $fd-popover-arrow-width * 0.5 !default;
-  $fd-popover-arrow-offset: 0.625rem;
-  $fd-popover-arrow-offset-compact: 0.5rem;
+  $fd-popover-arrow-offset: 0.575rem;
+  $fd-popover-arrow-offset-compact: 0.44rem;
   $fd-popover-inline-help-arrow-offset: 0.25rem !default;
   $fd-popover-arrow-color: var(--sapContent_ForegroundBorderColor) !default;
 
--- a/src/styles/popover.scss
+++ b/src/styles/popover.scss
@@ -15,7 +15,8 @@ $block: #{$fd-namespace}-popover;
   $fd-popover-arrow-width: 1rem !default;
   $fd-popover-arrow-height: 0.5rem !default;
   $fd-popover-arrow-width-half: $fd-popover-arrow-width * 0.5 !default;
-  $fd-popover-arrow-offset: var(--fdPopover_Arrow_Offset);
+  $fd-popover-arrow-offset: 0.625rem;
+  $fd-popover-arrow-offset-compact: 0.5rem;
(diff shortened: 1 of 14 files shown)


InnaAtanasova:
Screen Shot 2022-04-25 at 10 13 07 AM
Screen Shot 2022-04-25 at 10 12 49 AM
Screen Shot 2022-04-25 at 10 13 11 AM

In Action Bar and Action Sheet the arrow should be in the middle of the control. Is is possible to add 1 example in the Popover where the arrow is placed in the middle of the control? This way we can detect future problems. Thanks.

InnaAtanasova (changes requested):
The comments from yesterday are still relevant. The arrow. is not placed in the middle of the component. Here's how it is in Action Bar:
Screen Shot 2022-04-26 at 3 11 32 PM

Made the button with text so it's more clear:
Screen Shot 2022-04-26 at 3 18 28 PM

Like we discussed, can we add an example in the Popover so we have visual tests there as well?

platon-rov:
> The comments from yesterday are still relevant. The arrow. is not placed in the middle of the component.

Fixed.

> Made the button with text so it's more clear:

We cannot achieve without the breaking changes that arrow will be placed in the middle of the button of a random width. Arrow is the part of the popover body and the button is the part of the popover control, they're totally independent by their widths. When we're setting left/right offset of the arrow it's calculated relatively to the popover body, not the popover control.  

We only want to position the arrow so it will be in center if the popover control is a button with icon (without text).

> Like we discussed, can we add an example in the Popover so we have visual tests there as well?

Updated "Alignment" examples.

image

> Also, from what I see you introduced Breaking Changes. You need to document them.

It's not the breaking changes, it's a new features we didn't have previously. "Body variants" examples updated.

platon-rov:
Here is the issue, i was talking about.

image

I didn't find such examples in the sapui5 docs. But in ngx we're providing this feature, so we can set that the body can be center aligned (1, 2) or by one of its sides - left, right, top, bottom (3, 4). Here is the example in [ngx](https://fundamental-ngx.netlify.app/#/core/popover#placement). **_In the Align mode popover trigger may be bigger by its side than the body so you'll see something like on the 4th example if you try to center the arrow on the trigger, in case of centering it on the body - it won't be centered on the trigger :D_**

So our current behavior seems fine for me and it's simple. If we have popover in Align mode and the trigger is the icon button - arrow will be just in the center of that button. If the button is way bigger - arrow won't be in the center of the button.

image

For sure we may enhance this. I think an approach here may be the next:
* if trigger smaller than body then place the arrow in the center of trigger (example 3 on the picture).
* else (trigger is bigger than the body) place the arrow in the center of the body.
This manipulations can be done via javascript, so for the styles things remain the same. We cannot rely on body **and** trigger, when setting the position of the arrow, because it's tied to only the one element - body (current behavior) **or** trigger (we can change our implementation to this variant but still no profit).

cc @InnaAtanasova @droshev

platon-rov:
Here is the discussion in ngx https://github.com/SAP/fundamental-ngx/discussions/8038. No matter which decision we will make it will be handled in ngx via javascript. In styles we only can have fixed arrow position to avoid any issues. And this pr makes arrow centered if the trigger is icon button what's better than nothing.

InnaAtanasova (approved):

PR #3362, opened 2022-04-27
fix(styles): load styles once

N1XUS:
## Related Issue
Closes none

## Description
Previously, componentStylesManager loaded styles for every example. This created an overhead of loaded identical style files, which could cause hang of the developer tools of the browser. This behaviour was changed that only unique hrefs will be loaded now.

### Before:
image

### After:
image


platon-rov (approved):
good job!

InnaAtanasova (approved):

PR #3365, opened 2022-04-27
feat(styles): card footer implementation

g-cheishvili:
## Related Issue
part of #3213

## Description
Card footer was missing from the styles. Some of the functionalities should be done with javascript. When action items overflow footer, there should be overflow button which will(probably, not clear from specs) open up dropdown for additional actions

### DIFF
--- a/src/styles/card.scss
+++ b/src/styles/card.scss
@@ -121,12 +121,31 @@ $fd-card-header-outline-offset: 0.0625rem !default;
 
   &__footer {
     @include fd-reset();
-    @include fd-flex();
+    @include fd-flex-vertical-center();
 
-    padding: 1rem;
+    padding: var(--fdCard_Footer_Padding);
     background: $fd-card-default-body-background-color;
     border-bottom: $fd-card-header-border;
     border-radius: 0 0 $fd-card-border-radius $fd-card-border-radius;
+
+    &__actions {
+      @include fd-flex();
+
+      justify-content: flex-end;
+      width: 100%;
+
+      &__item {
+        &:not(:last-child) {
+          @include fd-set-margin-right(var(--fdCard_Footer_Actions_Item_Spacing));
+        }
+      }
+    }
+
+    &__link {
+      @include fd-ellipsis();
+
+      text-overflow: clip;
+    }
   }
 
   &__header-text {
--- a/src/styles/card.scss
+++ b/src/styles/card.scss
@@ -146,8 +146,9 @@ $fd-card-header-outline-offset: 0.0625rem !default;
       }
     }
 
-    &-link {
+    .#{$block}__footer-link {
       @include fd-ellipsis();
+      @include fd-set-margin-right(var(--fdCard_Footer_Actions_Item_Spacing));
 
       text-overflow: clip;
     }
--- a/src/styles/card.scss
+++ b/src/styles/card.scss
@@ -6,7 +6,7 @@ $object-status: #{$fd-namespace}-object-status;
 $numeric-content: #{$fd-namespace}-numeric-content;
 
 $fd-card-header-border: var(--fdCard_Header_Border_Bottom) !default;
-$fd-card-header-border-radius: var(--fdCard_Border_Radius) !default;
+$fd-card-border-radius: var(--fdCard_Border_Radius) !default;
 $fd-card-avatar-text-spacing: 0.75rem !default;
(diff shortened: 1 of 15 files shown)


stories/card/card.stories.js
```diff
@@ -53,6 +54,61 @@ export const CardAnatomy = () => `<div style="display:flex; justify-content:spac
                 </div>
             </a>
             <div class="fd-card__content" role="group" aria-label="Card Content"></div>
+            <div class="fd-card__footer">
+                <div class="fd-card__footer__actions">
```
N1XUS, line 58: I don't like too much of block selectors. Since we don't have actions in card, maybe we can simplify the selectors?
Something like: `fd-card__footer__actions` -> `fd-card__actions` and `fd-card__footer__actions__item` -> `fd-card__action-item`?

g-cheishvili: Yes we could and I like it that way too, but this way in CSS side everything is inside __footer without using any .#{$block} or something like it. Since we only want them inside footer and outside of it to not have any effect

N1XUS: I'd still prefer more simpler naming due to readability reasons. Anyway, I'd like to ask @InnaAtanasova her opinion too

N1XUS (approved):

src/styles/card.scss
```diff
@@ -104,20 +104,50 @@ $fd-card-header-outline-offset: 0.0625rem !default;
...
+
+    padding: var(--fdCard_Footer_Padding);
+    background: $fd-card-default-body-background-color;
+    border-bottom: $fd-card-header-border;
+    border-radius: 0 0 $fd-card-border-radius $fd-card-border-radius;
+
+    &__actions {
+      @include fd-flex();
+
+      justify-content: flex-end;
```
InnaAtanasova, line 134: this should go inside `fd-flex`
```
@include fd-flex() {
    justify-content: flex-end;
}
```

src/styles/card.scss
```diff
@@ -104,20 +104,56 @@ $fd-card-header-outline-offset: 0.0625rem !default;
...
+      @include fd-reset();
+
+      @include fd-flex() {
+        justify-content: flex-end;
+      }
+
+      width: 100%;
+
+      &-item {
+        @include fd-reset();
```
InnaAtanasova, line 141: in this case you don't need a reset as you are using the class like this:
```
<button class="fd-button fd-button--positive fd-card__footer-actions-item">
```

if you add a reset for `fd-card__footer-actions-item` it might reset some of the rules set by `fd-button` and `fd-button--positive`

src/styles/card.scss
```diff
@@ -104,20 +104,56 @@ $fd-card-header-outline-offset: 0.0625rem !default;
...
+
+        &:not(:last-child) {
+          @include fd-set-margin-right(var(--fdCard_Footer_Actions_Item_Spacing));
+        }
+      }
+    }
+
+    &-link {
+      @include fd-reset();
```
InnaAtanasova, line 150: same for this element. The reset is already coming from `fd-link`

InnaAtanasova (dismissed):

src/styles/card.scss
g-cheishvili: Actually that should not have been that way. Button should be separate instance from the elements that are wrapping it

InnaAtanasova (approved):
nice work!

mikerodonnell89 (commented):
This looks great. I did notice the table card had some overflowing border here
Screen Shot 2022-05-03 at 4 40 21 PM
:

In fundamental-styles, this is avoided by adding some extra padding bottom. But this style is only added to the compact card, perhaps it should be added to both. This is not an issue if you use the `fd-table--no-vertical-borders` class
Screen Shot 2022-05-03 at 4 43 22 PM

g-cheishvili:
> This looks great. I did notice the table card had some overflowing border here
> Screen Shot 2022-05-03 at 4 40 21 PM
> :
> 
> In fundamental-styles, this is avoided by adding some extra padding bottom. But this style is only added to the compact card, perhaps it should be added to both. This is not an issue if you use the `fd-table--no-vertical-borders` class
> Screen Shot 2022-05-03 at 4 43 22 PM
> 
> 

By specs padding should not be added if it is not in compact, soo this needs to be fixed

PR #3375, opened 2022-05-02
feat(styles): horizon family delta for link

droshev:
horizon family delta for link

### DIFF
--- a/src/styles/theming/sap_fiori_3.scss
+++ b/src/styles/theming/sap_fiori_3.scss
@@ -260,6 +260,14 @@
   /* Link */
   --fdLink_Text_Decoration_Opposite: underline;
   --fdLink_Text_Decoration: none;
+  --fdLink_Text_Focus_Color: var(--sapLinkColor);
+  --fdLink_Focus_Background: transparent;
+  --fdLink_Focus_Border_Radius: 0;
+  --fdLink_Focus_Text_Shadow: none;
+  --fdLink_Hover_Outline_Offset: -0.0625rem;
+  --fdLink_Hover_Outline_Width: var(--sapContent_FocusWidth);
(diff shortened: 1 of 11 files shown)


src/styles/theming/sap_fiori_3.scss
```diff
@@ -264,6 +264,14 @@
   /* Link */
   --fdLink_Text_Decoration_Opposite: underline;
   --fdLink_Text_Decoration: none;
+  --fdLink_Text_Focus: var(--sapLinkColor);
```
mikerodonnell89, line 267: It would be preferable to move this out each theming file and into its own `src/styles/theming/common/link` directory to avoid duplication - but of course this needs to be done for a few other components, so I am ok allowing this PR and maybe the refactor can be handled for every component in a new PR

platon-rov: Probably should be named `--fdLink_Text_Focus_Color`.

platon-rov (commented):
Text shadow shouldn't be present in fiori but should be present in horizon, in `fd-link` mixin it's set to be present in both.

Screenshot 2022-05-05 at 10 50 32

droshev:
> Text shadow shouldn't be present in fiori but should be present in horizon, in `fd-link` mixin it's set to be present in both.
> 
> Screenshot 2022-05-05 at 10 50 32

@platon-rov actually text shadow is for all the themes:
Fiori 3 themes:
Screen Shot 2022-05-06 at 09 29 00

Horizon themes
Screen Shot 2022-05-06 at 09 29 18
:

src/styles/theming/sap_fiori_3.scss
droshev: @platon-rov renamed

platon-rov (approved):

Issue #3354, opened 2022-04-22
[Horizon theme] Additional components 


N1XUS:
List of additional components that were introduced with `delta` theming for Horizon Family
- [x] Carousel - Platon #3380
- [x] Link - Deno
- [x] Progress Indicator - https://github.com/SAP/fundamental-styles/pull/3429
- [x] Rating Indicator - https://github.com/SAP/fundamental-styles/pull/3447
- [x] Scrollbar - @N1XUS #3367
- [x] Slider - @N1XUS #3381
- [x] Wizard - Mike #3382 
- [x] Panel - Mike
- [x] Splitter - Platon #3377
- [x] Feed list item - Denis #3384

**We apply only the color/parameters as part of this ticket. Any other improvements/fixes should be done separately - log issues for those**

PR #3381, opened 2022-05-05
feat(styles): slider horizon theme

N1XUS:
## Related Issue
Relates SAP/fundamental-styles#3354

## Description
Added delta theming for Horizon theme for the Slider component

**BREAKING CHANGE:**
Slider handle is now moved outside `.fd-slider__track` element.

_Before:_

```
<div class="fd-slider__inner">
    <div class="fd-slider__track">
        <div class="fd-slider__track-range" style="width: 50%;"></div>
        <div class="fd-slider__handle" tabindex="0" role="slider" aria-label="slider" aria-valuemin="1" aria-valuemax="100" aria-valuenow="50" style="left: 50%;"></div>
     </div>
</div>
```

_After:_

```
<div class="fd-slider__inner">
    <div class="fd-slider__handle" tabindex="0" role="slider" aria-label="slider" aria-valuemin="1" aria-valuemax="100" aria-valuenow="50" style="left: 50%;">
    </div>
    <div class="fd-slider__track">
        <div class="fd-slider__track-range" style="width: 50%;"></div>
    </div>
</div>
```

Range slider now should have modifier class `.fd-slider--range`

## Screenshots
image
image
image
image

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/styles/slider.scss
+++ b/src/styles/slider.scss
@@ -1,15 +1,19 @@
 @import "./new-settings";
 @import "./mixins";
 
+@mixin fd-slider-dot {
+  background-color: var(--fdSlider_Endpoint_Background);
+  border: var(--fdSlider_Endpoint_Border);
+}
+
+@mixin fd-slider-active-dot {
+  border: var(--fdSlider_Endpoint_Active_Border);
+  background-color: var(--fdSlider_Endpoint_Active_Background);
+}
+
 $block: #{$fd-namespace}-slider;
-$handle-desktop-dimensions: 1.25rem;
-$handle-hit-area-desktop-dimensions: 2rem;
-$handle-mobile-dimensions: 1.625rem;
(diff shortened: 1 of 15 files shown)


platon-rov (commented):
* Horizon, Border width of the grid & track should be the same in hc themes
Specs: 
image
Current:
image

* Horizon, Grip should be transparent when in focus
Specs:
image
Current:
image

* Fiori, Filled track part should be blue 
Specs:
image
Current:
image

By the way, why we cannot achieve start/end dots in Horizon?
image
It can be done using pseudo elements attached to the `.fd-slider__handle` element, isn't it? Sorry, maybe i forgot that you spoke on that on the call, but i cannot remember.

platon-rov (approved):

InnaAtanasova (approved):

g-cheishvili (approved):

PR #3384, opened 2022-05-06
feat(styles): feed list item horizon theme

N1XUS:
## Related Issue
Relates https://github.com/SAP/fundamental-styles/issues/3354

## Description
Added delta theming for feed list item component. No breaking changes.

## Screenshots
image
image
image

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation
- n/a Breaking Changes [wiki](https://github.com/SAP/fundamental-styles/wiki/Breaking-Changes) has been updated in case of breaking changes.

### DIFF
--- a/src/styles/feed-list.scss
+++ b/src/styles/feed-list.scss
@@ -15,7 +15,7 @@ $block: #{$fd-namespace}-feed-list;
   @include fd-reset();
   @include fd-flex();
 
-  background-color: var(--sapField_Background);
+  background-color: var(--fdFeed_List_Item_Background);
   max-width: 40rem;
 
   .#{$block}__thumb {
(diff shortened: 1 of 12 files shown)


platon-rov (commented):
* Fiori, bottom border not present for last item (it's ok in horizon):
image

Also, i see we're setting border-top for `item + item` so only between the items, but i didn't find that logic described in the wiki so probably we should just go with `border-bottom` for item every time except it's in the borderless state.

N1XUS:
> i so probably we should just go with

I was preserving previous behaviour for the Fiori theme, where's there no bottom border for the last item

N1XUS:
> * Fiori, bottom border not present for last item (it's ok in horizon):
> 
> image
> 
> Also, i see we're setting border-top for `item + item` so only between the items, but i didn't find that logic described in the wiki so probably we should just go with `border-bottom` for item every time except it's in the borderless state.

I was preserving previous behaviour for the Fiori theme, where's there no bottom border for the last item

N1XUS:
@platon-rov Regarding different border widths of the track and handle:
Design specs states that track should have 0.0625rem border width, and handle should have 0.125rem

platon-rov (approved):

PR #3385, opened 2022-05-07
[WIP] feat(styles): horizon family for ToolHeader

droshev:
horizon family for ToolHeader

### DIFF
--- a/src/styles/theming/sap_fiori_3.scss
+++ b/src/styles/theming/sap_fiori_3.scss
@@ -424,4 +424,13 @@
 
   /** Nested List */
   --fdNested_List_Border_Corner_Radius: var(--sapElement_BorderCornerRadius);
+
+  /** Tool Header **/
+  --fdTool_Header_Height: 2.75rem;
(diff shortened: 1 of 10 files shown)


platon-rov (commented):
* Horizon, Button interactive states not correct
Overflow issue:
image
No border no mouse press (active state):
image

* Horizon, object status color probably should be light. It's not mentioned anywhere in docs but when color is dark on dark background the contract-ratio is insufficient.
image

droshev:
> * Horizon, Button interactive states not correct
>   Overflow issue:
> 
> image
> 
> No border no mouse press (active state):
> image
> 
> * Horizon, object status color probably should be light. It's not mentioned anywhere in docs but when color is dark on dark background the contract-ratio is insufficient.
> 
> image

@platon-rov for the object status in Tool Header - I asked the designers for support

InnaAtanasova (changes requested):
Lots of repeatable variables for Horizon and Quartz. Please follow the approach we adopted for parameters that are the same within Quartz's themes but different for Horizon. You can check `theming/common` folder.

Issue #3327, opened 2022-04-13
Before release Defect Hunting + Bug fixes


InnaAtanasova:
- [ ] Test the components in all themes before the Horizon release.
- [ ] Distinguish the release blockers
- [ ] Bug fixes

Browsers for testing:
- [ ] Firefox - @g-cheishvili 
- [ ] Edge - @platon-rov 
- [ ] Chrome - @InnaAtanasova 
- [ ] Safari - @N1XUS 

Themes for testing:
- [ ] Quartz Light or Dark
- [ ] Horizon Morning
- [ ] Horizon Evening
- [ ] HCB for Horizon or HCW for Horizon

RTL and LTR testing too

Components for testing:
- [ ] Avatar
- [ ] Icons
- [ ] Busy Indicator
- [ ] Button Components - Badge on Button 
- [ ] Button Components - Button 
- [ ] Button Components - Button Shortcuts
- [ ] Button Components - Menu Button
- [ ] Button Components - Segmented Button
- [ ] Button Components - Toggle Button 
- [ ] Calendar Components - Date Picker 
- [ ] Checkbox
- [ ] Dialog and Popover Components - Dialog 
- [ ] Dialog and Popover Components - Popover
- [ ] Dialog and Popover Components - Message Box 
- [ ] Toolbar and Bar
- [ ] Input Components - Combobox 
- [ ] Input Components - Input Field
    - [ ] Step Input 
    - [ ] Input Group
- [ ] Input Components - Multi Input and Multi Combo
- [ ] Message Strip 
- [ ] Page and Layout
    - [ ] Dynamic Page
    - [ ] Dynamic Side Content
    - [ ] Flexible Column Layout 
    - [ ] Message Page 
    - [ ] Page 
- [ ] Radio Button 
- [ ] Switch
- [ ] Tab Bar
- [ ] Text Components 
- [ ] Cards and Tiles 
- [ ] Shellbar
- [ ] Input Components - Dropdown
- [ ] List View
    - [ ] List 
    - [ ] GridList 
    - [ ] ObjectList 
    - [ ] Table

mikerodonnell89:
- [ ] table - bottom border of the first column in "fixed header" example appears as two lines
![Screen Shot 2022-04-15 at 10 51 25 AM](https://user-images.githubusercontent.com/2471874/163598329-dbac9be8-72f6-4289-baaf-98ba27f38974.png)

platon-rov:
[...]

PR #3392, opened 2022-05-09
fix(styles): calendar horizon issues

platon-rov:
## Related Issue

## Description

Various calendar issues fixes.

### Before:
image
image
image

### After:
image
image
image


InnaAtanasova (changes requested):
Hover state on range element is not correct

InnaAtanasova (approved):

PR #3402, opened 2022-05-10
fix(styles): add fix for focus in active state

InnaAtanasova:
## Related Issue
part of https://github.com/SAP/fundamental-styles/issues/3327

## Description
change the focus outline color on active state

### Before:
Screen Shot 2022-05-10 at 5 01 08 PM

### After:
Screen Shot 2022-05-10 at 5 01 32 PM

### DIFF
--- a/src/styles/product-switch.scss
+++ b/src/styles/product-switch.scss
@@ -107,6 +107,10 @@ $block: #{$fd-namespace}-product-switch;
     }
 
     @include fd-active() {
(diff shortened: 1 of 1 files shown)


g-cheishvili (approved):
You could wrap that fd-fiori-focus call into local mixin to not reference same variable twice, similarly to function enhancers in JS. But you know, this works too

N1XUS (approved):

PR #3411, opened 2022-05-11
fix(styles): add fix for focus color on active state for Horizon

InnaAtanasova:
## Related Issue

## Description
fix the focus color of menu item on active state  for Horizon

### Before:
horiz

### After:
Screen Shot 2022-05-11 at 4 28 29 PM

### DIFF
--- a/src/styles/menu.scss
+++ b/src/styles/menu.scss
@@ -164,7 +164,7 @@ $block: #{$fd-namespace}-menu;
       }
 
       @include fd-focus() {
-        outline-color: var(--sapContent_ContrastFocusColor);
+        outline-color: var(--fdMenu_Active_State_Focus);
       }
     }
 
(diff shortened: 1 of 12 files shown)


src/styles/menu.scss
```diff
@@ -164,7 +164,7 @@ $block: #{$fd-namespace}-menu;
       }
 
       @include fd-focus() {
-        outline-color: var(--sapContent_ContrastFocusColor);
+        outline-color: var(--fdMenuActiveStateFocus);
```
N1XUS, line 167: From what I can see from other variable names, we use camelCase for defining the component name, and then we use `_` as a separator for the property name, so it should be like `fdMenu_ActiveState_Focus`

InnaAtanasova: fixed

PR #3413, opened 2022-05-11
fix(styles): add fix for text color for Counter and theming

InnaAtanasova:
## Related Issue
part of https://github.com/SAP/fundamental-styles/issues/3327

## Description
fix the Counter color in different theming

### Before:
counter

### After:
Screen Shot 2022-05-11 at 5 44 17 PM


N1XUS (commented):
image
Is it correct that for horizon hcw there's no border for the counter?

InnaAtanasova:
> image
> 
> Is it correct that for horizon hcw there's no border for the counter?

no, it's not correct! I just didn't save the changes on the HCW file :D Thanks for noticing! It should be now fixed.

N1XUS (approved):

PR #3416, opened 2022-05-12
fix(styles): remove the width limitation from the label of time column

dmitry-stepanenko:
## Related Issue
part of https://github.com/SAP/fundamental-ngx/issues/4669

## Description
removing width limitation from the label of time column in order to be able to use longer text in labels

### Before:
image

### After:
image

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [n/a] Updated tests
4. Documentation

### DIFF
--- a/src/styles/time.scss
+++ b/src/styles/time.scss
@@ -163,7 +163,6 @@ $block: #{$fd-namespace}-time;
     line-height: $fd-time-item-compact-height;
     height: $fd-time-item-compact-height;
     font-size: var(--sapFontSmallSize);
-    max-width: $fd-time-item-width;
     overflow: hidden;
     white-space: nowrap;
     text-overflow: clip;


droshev (approved):

droshev:
@dmitry-stepanenko can you re-check the CLA?

dmitry-stepanenko:
> @dmitry-stepanenko can you re-check the CLA?

@droshev looks like chromatic job cannot be completed due to exceeding the plan limit

Karthickc1994:
Hello Colleague,

Retest has been taken reported issue has been fixed.

Retest Environment: 
URL:https://deploy-preview-3416--fundamental-styles.netlify.app/
Browser: Chrome: Version 102.0.5005.63 (Official Build) (64-bit)

Regards,
Karthikeyan.c

PR #3417, opened 2022-05-12
fix(styles): shellbar css specificity & styles from ngx

platon-rov:
## Description
Shellbar css specificity & styles from ngx

### DIFF
--- a/src/styles/shellbar.scss
+++ b/src/styles/shellbar.scss
@@ -105,61 +105,61 @@ $block: #{$fd-namespace}-shellbar;
     }
   }
 
-  &__action {
+  .#{$block}__group .#{$block}__action {
     @include fd-reset();
     @include fd-set-margin-left($fd-shellbar-item-spacing);
 
     > * {
       vertical-align: middle;
     }
+  }
 
-    .#{$block}__input-group {
-      background: transparent;
-      background-color: var(--fdShellbar_Input_Background);
-      border: var(--fdShellbar_Input_Border);
-      border-radius: var(--fdShellbar_Input_Border_Radius);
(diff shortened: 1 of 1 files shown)


src/styles/shellbar.scss
```diff
@@ -105,61 +105,61 @@ $block: #{$fd-namespace}-shellbar;
...
+      border-radius: var(--fdShellbar_Input_Border_Radius);
+      border-width: var(--fdShellbar_Input_Addon_Border_Width);
+    }
   }
 
-  .#{$block}__button {
+  .#{$block}__group .#{$block}__button {
     outline-color: var(--fdShellbar_Button_Outline_Color);
-    color: var(--sapShell_InteractiveTextColor);
+    color: var(--sapShell_InteractiveTextColor) !important;
```
InnaAtanasova, line 162: That's not ok. You have 3 levels of nesting and still needing `!important`?

InnaAtanasova: That basically make this element impossible to customize by the client.

InnaAtanasova (approved):

src/styles/shellbar.scss
platon-rov: I'm fully agree with your arguments. However, to remove importants from this file we should remove them from some other files too. Yea, a total mess. 

Unfortunately, if we remove them somewhere it may lead to unexpected visual changes. And now we cannot catch them as we're out of limit in Chromatic.

I've opened a ticket https://github.com/SAP/fundamental-styles/issues/3420 so let's process this one once we have Chromatic to ensure we won't break anything.

PR #3421, opened 2022-05-13
fix(styles): fixed focus outlines and examples for Grid List

g-cheishvili:
## Related Issue
Part of SAP/fundamental-styles#3327

## Description
Replaced regular outline focus with pseudo element, because it should have hidden highlight element partially and also for better inset sizing, also fixed border radius for HC themes

### Before:
image
image
image

### After:
image
image
image

### DIFF
--- a/src/styles/grid-list.scss
+++ b/src/styles/grid-list.scss
@@ -80,12 +80,17 @@ $fd-grid-list-highlight: (
     position: relative;
     overflow: hidden;
 
-    @include fd-fiori-focus() {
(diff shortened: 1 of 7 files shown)


InnaAtanasova (changes requested):
Screen Shot 2022-05-13 at 11 55 31 AM

Unfortunately it's not an acceptable solution. You can't mix cozy and compact like this. Cozy is meant to be for mobile devices. In desktop apps it must be compact.

InnaAtanasova (approved):

PR #3429, opened 2022-05-15
fix(styles): horizon progress indicator

g-cheishvili:
## Related Issue

Part of #3354

## Description
Added Horizon theming support as good as it was possible. A lot of diff, had to hack way around to make things work.

### Before:
image
image

### After:
image
image

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [ n/a ] Verified all styles in IE11
- [ n/a ] Updated tests
4. Documentation

[Breaking Changes](https://github.com/SAP/fundamental-styles/wiki/Breaking-Changes#progress-indicator-3429)


N1XUS (commented):
please add [ci visual] to last commit message

src/styles/progress-indicator.scss
```diff
@@ -77,18 +158,47 @@ $fd-progress-indicator-states: (
...
     @include fd-reset();
     @include fd-flex();
-    @include fd-ellipsis();
+
+    @include fd-progress-indicator-popover-control() {
+      flex-grow: 1;
+    }
+  }
+
+  &__icon {
```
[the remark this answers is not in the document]
g-cheishvili: I have to move both label and icon in different directions when Horizon is applied, so I had to extract them

platon-rov (commented):
* Fiori, No gap before icon

Screenshot 2022-05-19 at 14 06 00

* Fiori, HC, Icons not centered vertically

image

g-cheishvili:
> * Fiori, No gap before icon
> 
> Screenshot 2022-05-19 at 14 06 00
> 
> * Fiori, HC, Icons not centered vertically
> 
> image

Fixed again. Now it's according to specs

platon-rov (approved):

mikerodonnell89 (changes requested):
In fiori mode, this rating indicator overflow text appears to have lost its ellipses: 
Screen Shot 2022-05-24 at 8 49 01 AM

src/styles/progress-indicator.scss
```diff
@@ -2,93 +2,232 @@
...
+      border-radius: 50%;
+      top: 0.0005rem;
+    }
+
+    &::before {
+      left: 0.0005rem;
+    }
+
+    &::after {
+      right: 0.0005rem;
```
mikerodonnell89, line 125: The specs call for 0.063rem here, giving 1px between the dots and the edge of the indicator

mikerodonnell89 (changes requested):
The top one is missing the white outline around the progress bar
Screen Shot 2022-05-24 at 9 35 04 AM

src/styles/progress-indicator.scss
g-cheishvili: Yes, you are right, that should be 0, because border makes that 1px spacing

Issue #3434, opened 2022-05-16
Select and Pagination Chromatic differences


InnaAtanasova:
Select needs some changes to match the Horizon design for when the dropdown list is expanded. Pagination should automatically be fixed when Select is fixed but to double check if more fixes are needed.

PR #3438, opened 2022-05-17
fix(styles): fix the button states in expanded Select [ci visual]

InnaAtanasova:
## Related Issue
Closes SAP/fundamental-styles#](https://github.com/SAP/fundamental-styles/issues/3434)

## Description
This PR fixes the button in Select control when it's in expanded state. The fix automatically solves the issue in Pagination as it uses Select. 

### Before:
Screen Shot 2022-05-17 at 3 21 22 PM
Screen Shot 2022-05-17 at 3 21 40 PM
Screen Shot 2022-05-17 at 3 20 49 PM

### After:
Screen Shot 2022-05-17 at 3 20 00 PM

Screen Shot 2022-05-17 at 3 19 50 PM
Screen Shot 2022-05-17 at 3 19 21 PM
Screen Shot 2022-05-17 at 3 19 05 PM

### DIFF
--- a/src/styles/mixins/_forms.scss
+++ b/src/styles/mixins/_forms.scss
@@ -241,7 +241,6 @@
 
   .#{$block}__button {
     @include fd-ellipsis();
-    @include fd-set-border-left(var(--sapButton_BorderWidth) solid transparent);
 
     border: none;
     z-index: auto !important;
@@ -249,6 +248,8 @@
     border-radius: var(--fdInput_Group_Button_Border_Raduis);
     background: var(--fdInput_Group_Button_Background);
     color: var(--fdInput_Group_Button_Text_Color);
+
+    @include fd-set-border-left(var(--sapButton_BorderWidth) solid transparent);
 
     &::before {
       display: none !important;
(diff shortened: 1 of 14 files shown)


src/styles/mixins/_forms.scss
```diff
@@ -250,6 +249,8 @@
     background: var(--fdInput_Group_Button_Background);
     color: var(--fdInput_Group_Button_Text_Color);
 
+    @include fd-set-border-left(var(--sapButton_BorderWidth) solid transparent);
```
mikerodonnell89, line 252: What is the point of this anyways? The icon looks just slightly out-of-center in the button and I think this may be the culprit

N1XUS (approved):

platon-rov (approved):

src/styles/select.scss
```diff
@@ -9,6 +9,67 @@ $fd-select-padding-x-compact: 0.5rem;
   min-width: calc(#{$width} - var(--sapField_BorderWidth) * 2);
 }
 
+@mixin fd-select-button-settings($bkgdColor, $borderLeft) {
```
g-cheishvili, line 12: We should use dash-case in scss variables

g-cheishvili (changes requested):
We could use CSS variable scopes to reduce size. It will not break anything, just will improve

src/styles/mixins/_forms.scss
mikerodonnell89: So I think we can have a border width of 0 for non-HC modes

PR #3380, opened 2022-05-05
feat(styles): add delta horizon theming for Carousel

platon-rov:
## Related Issue

Part of #3354.

## Description

Horizon support implemented for the Carousel component.

### Before:

image

image

### After:

image

image

PR #3439, opened 2022-05-18
feat(styles): carousel full-size image, classes & styles improvements

platon-rov:
## Related Issue

Part of https://github.com/SAP/fundamental-ngx/pull/8104#pullrequestreview-975799058

## Description

Carousel classes refactoring, styles improvements, examples update.

**As updated/changed classes in scope of this pr weren't mentioned in the docs there is no breaking changes.**

### Before:
image

### After:
image

## BREAKING CHANGE

* Carousel content markup change, from now element with class fd-carousel__slides for carousel items.

[WIKI](https://github.com/SAP/fundamental-styles/wiki/Breaking-Changes#carousel-3439)

### DIFF
--- a/src/styles/carousel.scss
+++ b/src/styles/carousel.scss
@@ -18,10 +18,8 @@ $fd-carousel-dot-active-size: 0.5rem !default;
   max-width: 100%;
 
   @include fd-hover() {
-    .#{$block}__content {
-      .#{$block}__button {
-        display: block;
-      }
+    .#{$block}__content .#{$block}__button {
+      display: block;
     }
   }
 
@@ -33,6 +31,7 @@ $fd-carousel-dot-active-size: 0.5rem !default;
     @include fd-reset();
 
     background: var(--fdCarousel_Content_Background);
+    flex-basis: 100%;
     position: relative;
     overflow: hidden;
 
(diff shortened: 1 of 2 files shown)


N1XUS (approved):

InnaAtanasova (changes requested):
Please document the Breaking Changes. Adding the div element with class `.fd-carousel__slides` is a Breaking Change and should have been done in https://github.com/SAP/fundamental-styles/pull/3380

src/styles/carousel.scss
```diff
@@ -18,10 +18,8 @@ $fd-carousel-dot-active-size: 0.5rem !default;
   max-width: 100%;
 
   @include fd-hover() {
-    .#{$block}__content {
-      .#{$block}__button {
-        display: block;
-      }
+    .#{$block}__content .#{$block}__button {
+      display: block;
```
InnaAtanasova, line 22: what do you achieve with this change? We usually nest the children elements like this because it's easier (visually) to see the number of nesting levels. BEM suggests we don't use more than 3 levels, that's why nesting them like it was done before allows you to easily determine these levels.

platon-rov: oh, accidentally added this. reverted.

PR #3442, opened 2022-05-18
fix(styles): remove product switch control class [ci visual]

mikerodonnell89:
Part of https://github.com/SAP/fundamental-ngx/issues/7984

**Breaking change** removes an old class that is no longer required because the styles are covered by `fd-button--transparent` class.

### DIFF
--- a/stories/product-switch/product-switch.stories.js
+++ b/stories/product-switch/product-switch.stories.js
@@ -14,7 +14,7 @@ export default {
 -	You want your users to navigate within the current product. In this case, use a product menu (see **Shellbar**).
         `,
         tags: ['f3', 'a11y', 'theme'],
-        components: ['product-switch', 'popover', 'button', 'icon']
+        components: ['product-switch', 'popover', 'button', 'icon', 'shellbar']
     }
 };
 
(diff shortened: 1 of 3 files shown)


platon-rov (commented):
Isn't it a breaking change? As we don't need to apply this class anymore.

mikerodonnell89:
> 3\. Maybe use the code for `fd-shellbar__button` class by moving it to a mixin or placeholder and reuse it for Product Switch:

Product switch doesn't need this anymore if all the styles are coming from button and fd-shellbar__button

stories/product-switch/product-switch.stories.js
```diff
@@ -34,7 +34,7 @@ export const Shellbar = () => `${localStyles}
     <div class="fd-product-switch">
         <div class="fd-popover fd-popover--right">
             <div class="fd-popover__control">
-                <button class="fd-button fd-button--transparent fd-popover__control fd-product-switch__control"
+                <button class="fd-button fd-button--transparent fd-popover__control"
```
InnaAtanasova, line 37: You have a div on line 36 with `fd-popover__control` class and the same class applied on the button. Maybe remove the parent div.

stories/shellbar/shellbar.stories.js
```diff
@@ -614,7 +614,7 @@ export const ProductSwitch = () => `<div style="height:600px">
                     <div class="fd-popover fd-popover--right">
                         <div class="fd-popover__control">
                             <button
-                                class="fd-button fd-button--transparent fd-shellbar__button fd-popover__control fd-product-switch__control"
+                                class="fd-button fd-button--transparent fd-shellbar__button fd-popover__control"
```
InnaAtanasova, line 617: same as above

InnaAtanasova (changes requested):

InnaAtanasova (changes requested):
Screen Shot 2022-05-26 at 9 25 05 AM


The example in Product Switch should be updated. We either use ShellBar in this example or remove it from there, o.w. users can report it as a bug.

mikerodonnell89:
> Screen Shot 2022-05-26 at 9 25 05 AM
> 
> The example in Product Switch should be updated. We either use ShellBar in this example or remove it from there, o.w. users can report it as a bug.

Thanks! updated to use shellbar

InnaAtanasova (approved):

PR #3448, opened 2022-05-19
fix(styles): layout grid nested rows

platon-rov:
## Related Issue

## Description

Nested rows horizontal gutter fix.

### Before:
image

### After:

image


g-cheishvili (approved):
I approve this change, but code would use some fine refactoring in your style :)

PR #3465, opened 2022-05-22
fix(styles): input field readonly state fixes and adaptations

g-cheishvili:
## Related Issue
part of #3327

## Description
Readonly fields became unfocusable and unactionable with cursor and placeholder colors became visible in Horizon

### DIFF
--- a/src/styles/mixins/_forms.scss
+++ b/src/styles/mixins/_forms.scss
@@ -190,7 +190,7 @@
     opacity: var(--sapContent_DisabledOpacity);
 
     &::placeholder {
-      color: transparent;
+      color: var(--fdInput_Non_Interactive_State_Placeholder_Color);
     }
   }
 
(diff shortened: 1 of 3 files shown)


mikerodonnell89 (commented):
This is better, but we can still focus the readonly textarea via keyboard, should this be possible at all?

InnaAtanasova (approved):

InnaAtanasova:
> This is better, but we can still focus the readonly textarea via keyboard, should this be possible at all?

hi @mikerodonnell89 readonly should be focusable. The users should be able to click and copy text if they needed to. The only thing they can't do is to modify the content.

platon-rov (approved):

g-cheishvili:
> This is better, but we can still focus the readonly textarea via keyboard, should this be possible at all?

Yeah, via keyboard readonly should be focusable, but with mouse it should not have any interaction. Text selection is allowed at this moment, so I think it's the way it should be now

Issue #3454, opened 2022-05-20
Package version 0.23.0 doesn't load the progress-indicator.css file


rchrdchn:
**Describe the bug**
Package version `0.23.0` doesn't load the `progress-indicator.css` file

> Is this issue related to a specific component?
Progress Indicator

> What version of the Fundamental Library Styles are you using?
`0.23.0`

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?
SF

> What front-end framework are you implementing in, e.g., Angular, React, etc.?
JUIC

**Expected behavior**
Package version `0.23.0` loads the `progress-indicator.css` file -- see here (**Control/Command + F**, search for _fd-progress-indicator_) --> https://unpkg.com/fundamental-styles@0.23.0/dist/fundamental-styles.css

PR #3476, opened 2022-05-24
fix: automatically add missing components

N1XUS:
## Related Issue
Closes SAP/fundamental-styles#3454

## Description
Added automatic check that all main components are imported into the root css file.


g-cheishvili (approved):
Cool thing ! :)

platon-rov (approved):

rchrdchn:
Thanks for pushing this fix! When is the next release so our team can upgrade the package and use the newly-added components styling?

rchrdchn:
Following up on the above question.

Issue #3453, opened 2022-05-20
Switch


vmasl-s:
Semantic switch Cozy icon size should be var --sapFontLargeSize; which is 16px
Compact is correct with --sapFontSize; which is 14px

Found it somewhat confusing to read the Fund. Lib. implemented metrics for track Spacing and Touchable size, but Handle properties seem correct. 

**Default state should have icons**
Active:
icon: sap-icon://accept;
color: --sapButton_Handle_Selected_TextColor;

Inactive:
icon: sap-icon://less;
color: --sapButton_Handle_TextColor;

**Track Hover state for Default look**
Track should also change its background and border colour on Hover. Currently only the handle has Hover state change.

PR #3494, opened 2022-05-25
fix: add missing switch variables

N1XUS:
## Related Issue
Closes SAP/fundamental-styles#3453

## Description
* Added missing variables and styles
* Optimized css structure
* Added sync script to reformat theming content files to same structure and detect missing variables across theming files

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/src/styles/switch.scss
+++ b/src/styles/switch.scss
@@ -9,14 +9,10 @@
 $block: #{$fd-namespace}-switch;
 
 @mixin fd-slider-track-position($off-position, $on-position) {
-  .#{$block}__track {
-    transform: translate($off-position, -50%);
-  }
+  --fdSwitch_Track_Offset: translate(#{$off-position}, -50%);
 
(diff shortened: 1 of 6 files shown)


platon-rov (approved):

scripts/sync-css-variables.js
```diff
@@ -0,0 +1,74 @@
+/**
```
InnaAtanasova, line 1: Is this file part of the scope of the PR? 
We avoid having different scopes in 1 PR as it will be hard to revert changes if needed.

N1XUS: It's not, but was needed for me to reformat theming files (for switch they are huge and to control all of the variables is becoming very hard). Since it's an optional tool which is not called anywhere (only manual execution), I don't think it can affect future development activities

InnaAtanasova (approved):

Issue #3495, opened 2022-05-25
Horizon/Quartz Smoke testing


droshev:
- [ ] Test the components in all themes before the Horizon release.
- [ ] Distinguish the release blockers
- [ ] Bug fixes

Browsers for testing:
- [ ] Firefox - @g-cheishvili 
- [ ] Edge - @platon-rov 
- [ ] Chrome - @InnaAtanasova 
- [ ] Safari - @N1XUS 

Themes for testing:
- [ ] Quartz Light or Dark
- [ ] Horizon Morning
- [ ] Horizon Evening
- [ ] HCB for Horizon or HCW for Horizon

RTL and LTR testing too

Components for testing:
- [ ] Avatar
- [ ] Icons
- [ ] Busy Indicator
- [ ] Button Components - Badge on Button 
- [ ] Button Components - Button 
- [ ] Button Components - Button Shortcuts
- [ ] Button Components - Menu Button
- [ ] Button Components - Segmented Button
- [ ] Button Components - Toggle Button 
- [ ] Calendar Components - Date Picker 
- [ ] Checkbox
- [ ] Dialog and Popover Components - Dialog 
- [ ] Dialog and Popover Components - Popover
- [ ] Dialog and Popover Components - Message Box 
- [ ] Toolbar and Bar
- [ ] Input Components - Combobox 
- [ ] Input Components - Input Field
    - [ ] Step Input 
    - [ ] Input Group
- [ ] Input Components - Multi Input and Multi Combo
- [ ] Message Strip 
- [ ] Page and Layout
    - [ ] Dynamic Page
    - [ ] Dynamic Side Content
    - [ ] Flexible Column Layout 
    - [ ] Message Page 
    - [ ] Page 
- [ ] Radio Button 
- [ ] Switch
- [ ] Tab Bar
- [ ] Text Components 
- [ ] Cards and Tiles 
- [ ] Shellbar
- [ ] Input Components - Dropdown
- [ ] List View
    - [ ] List 
    - [ ] GridList 
    - [ ] ObjectList 
    - [ ] Table

droshev:
Carryover from #3327
- [ ] #3477 
- [ ] #3465
- [ ] #3474
- [ ] Tree, Fiori/Horizon, HC. Focus on the expanders look quite different from all other elements. Also state active is missing. However i didn't find how it should look like, probably we should ask designers about it.
[...]

PR #3510, opened 2022-06-02
fix(styles): focus border radius on list items in popovers [ci visual]

mikerodonnell89:
part of #3495 

FIxes an issue with the focus outline on list items in a popover. This does unfortunately break our self-contained style rule, but as discussed we will address in a later refactor

before:
image

after:
Screen Shot 2022-06-02 at 10 16 44 AM

### DIFF
--- a/src/styles/dialog.scss
+++ b/src/styles/dialog.scss
@@ -17,6 +17,7 @@
 $block: #{$fd-namespace}-dialog;
 $bar: #{$fd-namespace}-bar;
 $button: #{$fd-namespace}-button;
+$menu: #{$fd-namespace}-menu;
 
 .#{$block} {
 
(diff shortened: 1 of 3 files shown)


g-cheishvili:
We can have other way around. So in list component for first and last elements we say that border radius of those two corners should be var(--listContainerBorderRadius) and in popover context we can just rewrite that value. This way I think it is more acceptable thing

InnaAtanasova (approved):

mikerodonnell89:
> We can have other way around. So in list component for first and last elements we say that border radius of those two corners should be var(--listContainerBorderRadius) and in popover context we can just rewrite that value. This way I think it is more acceptable thing. This way popover just says that "for me as a container for list, border radius should be this" and list just says "I will comply to the container, whoever it is"

I don't think we can just rewrite the variable value in popover because of instances such as below where the first item in the fd-list actually should not be receiving rounded borders due to its sibling element, the list group header. Or maybe I am misunderstanding your suggestion, could you provide some code snippets?

Screen Shot 2022-06-02 at 11 09 55 AM

Screen Shot 2022-06-02 at 11 10 55 AM

PR #3517, opened 2022-06-03
fix(styles): add fix for double focus in single click area for IconTabBar

InnaAtanasova:
## Related Issue
Part of https://github.com/SAP/fundamental-styles/issues/3495

## Description
fixes the issue with the double focus for Universal Icon Tab Bar Single Click Area and the wrong position of the badge on focus

### Before:

171494104-98f24575-12ac-4fbe-be7e-d98a6058916f
171494099-c062dfa0-5054-41bd-bc9e-b68ac7fc4085

### After:

Screen Shot 2022-06-03 at 1 40 54 PM
Screen Shot 2022-06-03 at 2 12 07 PM
Screen Shot 2022-06-03 at 1 40 28 PM

### DIFF
--- a/src/styles/icon-tab-bar.scss
+++ b/src/styles/icon-tab-bar.scss
@@ -157,6 +157,16 @@ $fd-icon-tab-bar-semantic-values: (
   }
 }
 
+@mixin fd-tabs-remove-focus-from-tag() {
+  .#{$block}__tag {
+    outline: none;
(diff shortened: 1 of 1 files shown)


mikerodonnell89 (approved):
Looks great. Should the circular icons have circular focus outlines? Can be addressed in another pr

PR #3546, opened 2022-06-07
docs(styles): flexible column layout background

platon-rov:
## Related Issue

Part of https://github.com/SAP/fundamental-ngx/issues/7984#issuecomment-1128609647

## Description

Flexible Column Layout background to support various themes.

### Before:
image

### After:

image


mikerodonnell89 (changes requested):
Do we want to be adding the background to the component itself, or rather just the documentation? I think the latter

mikerodonnell89 (approved):

InnaAtanasova (approved):

InnaAtanasova:
I renamed the PR to docs as it's not fixing the code, but rather the documentation.

PR #3556, opened 2022-06-08
fix(styles): form message margin

platon-rov:
## Related Issue

Closes https://github.com/SAP/fundamental-ngx/issues/8111

## Description

Form Message margin fix.

### Before:
image

### After:
image


InnaAtanasova (changes requested):
Screen Shot 2022-06-10 at 1 34 09 PM
Screen Shot 2022-06-10 at 1 34 02 PM

It's now broken in fund-styles.

PR #3708, opened 2022-07-01
fix(styles): some fixes for fd-bar [ci visual]

mikerodonnell89:
part of https://github.com/SAP/fundamental-styles/issues/3600

Some minor fixes for bar based on spec review

### DIFF
--- a/src/styles/bar.scss
+++ b/src/styles/bar.scss
@@ -28,23 +28,6 @@ $block: #{$fd-namespace}-bar;
     right: $margin-x;
   }
 
-  @mixin fd-bar-responsive-paddings() {
-    @media (max-width: 599px) {
-      left: 1rem;
-      right: 1rem;
-    }
-
-    @media (min-width: 600px) and (max-width: 1439px) {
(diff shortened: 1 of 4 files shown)


src/styles/bar.scss
```diff
@@ -74,14 +81,10 @@ $block: #{$fd-namespace}-bar;
     display: flex;
     align-items: center;
     justify-content: center;
```
InnaAtanasova, line 83: this can also be a mixin

InnaAtanasova (approved):

InnaAtanasova:
@mikerodonnell89 in the description you mention `part of 3600`. This means than more work will be done for this task, correct?

mikerodonnell89:
> @mikerodonnell89 in the description you mention `part of 3600`. This means that more work will be done for this task, correct?

I am fairly certain this is all that needs to be done but wondering if anything else would come up in the review process, so I wasn't sure if I should say this PR will fully close it yet

platon-rov (approved):

PR #3745, opened 2022-07-21
fix(ci): fixed chromatic workflow

g-cheishvili:
## Description
- fixed chromatic workflow. Now it uses yarn instead of npm, which was issue
- updated actions:
  - checkout


droshev (approved):

g-cheishvili:
Updated version to be [0.24.3-rc.0](https://www.npmjs.com/package/fundamental-styles/v/0.24.3-rc.0), because that is actual version right now. Previous CI workflow failed on that moment and did not update in package.json

PR #3816, opened 2022-08-17
feat(styles): add flex to common css (helpers)

InnaAtanasova:
## Related Issue
common css

## Description
- flex


src/styles/flex.scss
```diff
@@ -0,0 +1,212 @@
...
+  // ----------- FLEX ALIGN CONTENT -----------
+  @mixin fd-flex-align-content($type: normal) {
+    display: flex;
+    flex-wrap: wrap;
+    align-content: $type;
+
+    @content;
+  }
+
+  @each $type in (flex-start flex-end center space-between space-around space-evenly stretch start end baseline) {
```
g-cheishvili, line 32: I think it would be better if you moved these values to some variable above and then used it here, much easier read

src/styles/flex.scss
```diff
@@ -0,0 +1,212 @@
...
+    flex-direction: $direction;
+
+    @content;
+  }
+
+  @include fd-flex();
+
+  // ----------- FLEX ALIGN CONTENT -----------
+  @mixin fd-flex-align-content($type: normal) {
+    display: flex;
```
g-cheishvili, line 25: Why not use fd-flex mixin here?

src/styles/flex.scss
```diff
@@ -0,0 +1,212 @@
...
+
+  @each $type in (flex-start flex-end center space-between space-around space-evenly stretch start end baseline) {
+    &--align-content-#{$type} {
+      @include fd-flex-align-content(#{$type});
+    }
+  }
+
+  // ----------- FLEX ALIGN ITEMS -----------
+  @mixin fd-flex-align-items($type: stretch) {
+    display: flex;
```
g-cheishvili, line 40: fd-flex?

g-cheishvili (changes requested):

g-cheishvili:
This would be much more helpful to end user if we ship raw scss to the npm repo, this way they would be able to use our mixins in case they have scss on their end

src/styles/flex.scss
```diff
@@ -0,0 +1,233 @@
+@import './new-settings';
```
platon-rov, line 1: shouldn't be common CSS in a different folder? in case we will have 10+ common stylesheets there will be a mess.

InnaAtanasova: yes, it should be! It's even with a different prefix: `sap-` not `fd-`. I will open soon a new PR for `display` and after it's merged I will rebase main and move this in the proper folder.

mikerodonnell89:
I don't think the examples are working, they are all displaying the same thing (zoomed out)
Screen Shot 2022-08-30 at 11 32 10 AM

InnaAtanasova:
> https://deploy-preview-3816--fundamental-styles.netlify.app

they should be working now

stories/Helpers/flex/flex.stories.js
```diff
@@ -0,0 +1,955 @@
+export default {
```
platon-rov, line 1: empty space below stories

зображення

InnaAtanasova: I don't understand. Where do you see this in the documentation?

PR #3835, opened 2022-08-29
fix(styles): add shellbar title focus

hardl:
fixes shellbar title focus, similar to https://github.com/SAP/fundamental-styles/pull/3730

### DIFF
--- a/src/styles/shellbar.scss
+++ b/src/styles/shellbar.scss
@@ -75,8 +75,13 @@ $block: #{$fd-namespace}-shellbar;
     vertical-align: middle;
   }
 
+  &__logo, &__title {
+    outline-color: var(--fdShellbar_Button_Outline_Color);
+    outline-width: var(--sapContent_FocusWidth);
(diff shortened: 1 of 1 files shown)


mikerodonnell89:
I think the focus outline may be the incorrect color, would think it should be white both for the title and for the logo

hardl:
IDK, I just did copy&paste :D

platon-rov (approved):

g-cheishvili (approved):

mikerodonnell89 (changes requested):
Then I think instead let's do:

```
  &__logo, &__title {
    outline-color: var(--fdShellbar_Button_Outline_Color);
    outline-width: var(--sapContent_FocusWidth);
    outline-style: var(--sapContent_FocusStyle);
}
```

And remove `@include fd-fiori-focus(0);` from `&__title` and `&__logo`

InnaAtanasova (approved):

mikerodonnell89 (approved):

PR #3838, opened 2022-08-30
fix(styles): form layout grid

platon-rov:
## Related Issue

Refers to https://github.com/SAP/fundamental-ngx/issues/7854

## Description

Minor Form Grid fixes.

### Before:
Screenshot 2022-08-30 at 20 19 56
Screenshot 2022-08-30 at 20 18 16
зображення
image

### After:
Screenshot 2022-08-30 at 20 20 02
Screenshot 2022-08-30 at 20 34 26
зображення
image

### DIFF
--- a/src/styles/form-layout-grid.scss
+++ b/src/styles/form-layout-grid.scss
@@ -101,6 +101,8 @@ $gridCol: #{$fd-namespace}-col;
   }
 
   .#{$gridRow} {
+    flex-direction: row;
+
     .#{$gridCol} {
       @include fd-ellipsis();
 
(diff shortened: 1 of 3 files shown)


InnaAtanasova:
this is introducing Breaking changes, isn't it?

mikerodonnell89 (approved):
Thanks for fixing these. I'm not sure we need the black 1px borders in the examples

platon-rov:
> Thanks for fixing these. I'm not sure we need the black 1px borders in the examples

I didn't find a way to change the examples' iframe width and the reason to have the borders is to show which space relates to the form and which isn't.

PR #3843, opened 2022-09-01
fix(styles): full height to shellbar select

shrvr:
## Related Issue
Part of SAP/fundamental-ngx#7966

## Related NGX PR
SAP/fundamental-ngx#8347

## Description
Changed shellbar-select style to fix mouse hover effect.

### Before:
Screen Shot 2022-09-01 at 11 54 19 AM

### After:
Screen Shot 2022-09-01 at 2 26 44 PM

### DIFF
--- a/src/styles/shellbar.scss
+++ b/src/styles/shellbar.scss
@@ -146,9 +146,9 @@ $block: #{$fd-namespace}-shellbar;
       }
 
       &--select {
+        @include fd-flex-vertical-center();
+
         font-style: normal;
-        display: flex;
-        align-items: center;
         height: 100%;
       }
     }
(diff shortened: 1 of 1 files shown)


src/styles/shellbar.scss
```diff
@@ -147,6 +147,9 @@ $block: #{$fd-namespace}-shellbar;
 
       &--select {
         font-style: normal;
+        display: flex;
+        align-items: center;
```
InnaAtanasova, line 151: please use mixin here:
```
@mixin fd-flex-vertical-center {
  display: flex;
  align-items: center;
  @content;
}
```

InnaAtanasova (changes requested):

src/styles/shellbar.scss
InnaAtanasova: it's existing in fd-styles, just use it in your code

InnaAtanasova (approved):

Issue #3804, opened 2022-08-12
Icon missing in Form Message for input validation (Morning Horizon)


obc514:
According to the [specs](https://wiki.one.int.sap/wiki/pages/viewpage.action?pageId=2698914037), there should be an icon in the Form Message for input validation (Morning Horizon)

PR #3858, opened 2022-09-06
fix(styles): form message icon in horizon themes

platon-rov:
## Related Issue

Closes https://github.com/SAP/fundamental-styles/issues/3804

## Description

* form message icon in horizon themes

### Before:

image

### After:

Screenshot 2022-09-06 at 11 51 58

### DIFF
--- a/src/styles/form-message.scss
+++ b/src/styles/form-message.scss
@@ -9,6 +9,7 @@ $block: #{$fd-namespace}-form-message;
 .#{$block} {
   @include fd-reset();
   @include fd-ellipsis();
+  @include fd-flex-vertical-center();
 
   padding: 0.5rem;
   min-width: 6rem;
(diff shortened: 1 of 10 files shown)


src/styles/form-message.scss
```diff
@@ -35,17 +49,37 @@ $block: #{$fd-namespace}-form-message;
...
+    }
   }
 
   &--information {
     background-color: var(--sapInformationBackground);
+
+    &::before {
+      content: "\e202";
+      color: var(--sapInformativeElementColor);
+    }
```
InnaAtanasova, line 83: Lots of repeating code. Could you make a mixin that sets the state. You can pass the background-color, content and color as params. 
Another option could be to put all these values in an object and loop through it, like we do for many other components.

src/styles/theming/sap_fiori_3.scss
```diff
@@ -175,6 +175,7 @@
   --fdMessage_Box_Shadow: var(--sapContent_Shadow1);
   --fdMessage_Border: none;
   --fdMessage_Border_Radius: 0;
+  --fdMessage_Icon_Display: none;
```
InnaAtanasova, line 178: These values should go in the common folder/file as they are the same across Quartz themes and Horizon themes

InnaAtanasova (changes requested):

InnaAtanasova (approved):
Looks great!

PR #3913, opened 2022-10-05
feat(cx): introduce Side Nav

InnaAtanasova:
## Description
The current PR is a refactor for Side Nav which has a new design

!!! IMPORTANT !!! Note:
When reviewing please switch to Horizon theme or use this link: https://deploy-preview-3913--fundamental-styles.netlify.app/?path=/docs/components-side-navigation--cozy-wide-icons&globals=theme:sap_horizon

### DIFF
--- a/.storybook/static/preview-head.js
+++ b/.storybook/static/preview-head.js
@@ -376,10 +376,12 @@ function toggleNestedListSubmenu(event) {
     let button = event.target;
     let icon = button.children[0];
 
-    if(button && button.classList.contains('is-expanded')) {
-        icon.classList = 'sap-icon--navigation-right-arrow';
-    } else {
+    if(button && (button.getAttribute('aria-expanded')=='false')) {
         icon.classList = 'sap-icon--navigation-down-arrow';
+        button.setAttribute('aria-expanded', 'true');
+    } else {
+        icon.classList = 'sap-icon--navigation-right-arrow';
+        button.setAttribute('aria-expanded', 'false');
     }
 }
 
--- a/packages/cx/project.json
+++ b/packages/cx/project.json
@@ -18,6 +18,18 @@
                 }
             }
         },
+        "build-visual": {
+            "executor": "workspace-plugins:visual-stories",
+            "options": {
+                "themes": [
+                    "sap_horizon"
+                ],
+                "excludedStoriesKinds": [
+                    "^(Helpers|Dev)"
+                ]
+            },
+            "outputs": ["stories/Visuals/styles"]
+        },
         "lint": {
             "executor": "@nrwl/linter:eslint",
             "outputs": ["{options.outputFile}"],
--- a/packages/cx/src/_nested-list.scss
+++ b/packages/cx/src/_nested-list.scss
@@ -1,0 +1,257 @@
+@import "../../styles/src/new-settings";
+@import "../../styles/src/mixins";
+
+$block: #{$fd-namespace}-nested-list;
+$button: #{$fd-namespace}-button;
+
+.#{$block} {
+  $fd-nested-list-item-height: 2.75rem;
+  $fd-nested-list-item-height-compact: 1.5rem;
+  $fd-nested-list-item-border-radius: 0.5rem;
+  $fd-nested-list-item-spacing: 0.25rem;
+  $fd-nested-list-item-spacing-compact: 0.125rem;
+
+  @mixin fd-nested-list-states-colors($textColor, $backgroundColor) {
+    color: $textColor;
+    background: $backgroundColor;
+  }
+
+  @mixin fd-nested-list-focus($offset: 0.125rem, $width: var(--sapContent_FocusWidth)) {
+    &:focus,
+    &.is-focus {
+      outline-offset: $offset;
+      outline-width: $width;
+      outline-color: var(--sapContent_FocusColor);
+      outline-style: var(--sapContent_FocusStyle);
+      @content;
+    }
+  }
+
+  @include fd-reset();
+
+  width: 100%;
+  list-style: none;
+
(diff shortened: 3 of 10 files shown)


InnaAtanasova:
When reviewing please use this link or switch to Horizon theme
https://deploy-preview-3913--fundamental-styles.netlify.app/?path=/docs/components-side-navigation--cozy-wide-icons&globals=theme:sap_horizon

src/styles/nested-list.scss
```diff
@@ -5,378 +5,253 @@ $block: #{$fd-namespace}-nested-list;
...
   @include fd-reset();
 
-  list-style: none;
   width: 100%;
-  border-bottom: var(--sapList_BorderWidth) solid var(--sapList_BorderColor);
+  list-style: none;
 
-  & [class*="level"] {
-    border-bottom: none;
+  &.level-1 {
```
platon-rov, line 35: is it correct according to the bem?

src/styles/nested-list.scss
```diff
@@ -5,378 +5,253 @@ $block: #{$fd-namespace}-nested-list;
...
-        padding-left: 0;
-      }
     }
   }
 
-  &__group-header {
+  &__arrow {
     @include fd-reset();
+    @include fd-flex-center();
+    @include set-width(1rem);
```
platon-rov, line 202: prefix?

platon-rov (approved):
Wow, nice!

mikerodonnell89:
The popovers have popover arrows in the designs, and I think it'd be beneficial to show the whole side nav for the examples where the popover is shown rather than just showing the popover

mikerodonnell89:
In the specs there are 8px from the top of the nav to the top of the first item, and 8px from the bottom of the bottom item to the bottom of the nav. In the side nav in this PR we have 8px on the top then 12px on the bottom, I think due to the whole sidenav having different paddings for top/bottom.

mikerodonnell89:
Docs are missing the vertical overflow and filter examples but that can come later and I don't think that may be totally agnostic from this component

src/styles/nested-list.scss
mikerodonnell89: These class names are being added programmatically in ngx in `nested-list.directive.ts`. Maybe time for a rename but it'd have to be done on both ngx and styles

mikerodonnell89: These class names are being added programmatically in ngx in `nested-list.directive.ts`. Maybe time for a rename, it'd have to be done on both ngx and styles

InnaAtanasova: we have in fundamental-styles similar cases with the classes `is-open`, `has-child`, etc. 
`level-1`, `level-2`, etc are not modifier classes in the sense of BEM. They don't modify the properties of the element they are attached to, but rather to its child elements.
In BEM we have something like: `fd-nested-list  fd-nested-list--transparent`. This modifies the background of ``fd-nested-list`, but in our case `level-2` would modify the item and the title, which are child and grandchild of the element. In the past we wanted to distinguish such behaviour and that's why classes like `is-open`, `has-child`, `level-*` appeared. I am open to discussion about that. Not a problem at all to rename the classes.

InnaAtanasova:
> More of a docs thing, and just a suggestion - The popovers have popover arrows in the designs, and I think it'd be beneficial to show the whole side nav for the examples where the popover is shown rather than just showing the popover

In fundamental-styles we don't have popover body with arrow placed like this. We can do this in fund-ngx
Screen Shot 2022-10-11 at 12 45 28 PM

InnaAtanasova:
> Designs state that the first level always shows an icon (I guess that cannot be enforced on our end)
> 
> Screen Shot 2022-10-06 at 10 30 25 AM

Yes, I addressed this question to the designers as well, because it's written that Level 1 should always have icons but in the specs there's this one:
Screen Shot 2022-10-11 at 12 48 27 PM

InnaAtanasova:
> Docs are missing the vertical overflow and filter examples but that can come later and I don't think that may be totally agnostic from this component

I am not sure if they are part of the final design. There's no design for the vertical overflow buttons and in the "Backstage Design Specs" the overflow is shown with a scrollbar:
Screen Shot 2022-10-11 at 1 02 50 PM

src/styles/nested-list.scss
InnaAtanasova: I opened a PR for renaming the mixins missing the prefix. Depending on which PR is merged first I will rebase and update. 
https://github.com/SAP/fundamental-styles/pull/3921

src/styles/nested-list.scss
mikerodonnell89: Thanks for the explanation, I think just leave them as-is then. If we do want to change, let's make a separate issue and discuss

mikerodonnell89 (changes requested):
Apologies for nit-picking but in the specs looks like there should be 12px of padding on top of the cozy divider and 16px of the bottom
Screen Shot 2022-10-12 at 12 11 50 PM

mikerodonnell89 (approved):

mikerodonnell89 (changes requested):
If this gets merged, and we update the fundamental-styles version in fd-ngx, anyone using the quartz/fiori theme will have a broken side nav, even if we make the required API updates. I think we may need a new component entirely with a new name

mikerodonnell89 (approved):

droshev:
We need to keep the old implementation in styles and introduce the new one in CX package

droshev (changes requested):
who will move this to the `cx` packages. BTW the previous version of the side nav should stay in `fundamental-styles` deprecated and pointing to the new implementation

droshev (approved):

Issue #3720, opened 2022-07-08
Common CSS next steps


droshev:
- [x] publish `SCSS` with the mixins @droshev 
- [x] explore the infrastructure
- [x] Base
- - [x] normalize (incl. font)
- - [x] reset (global and comp)
- [x] colors? (it requires some investigation which colors should be exposed)
- [x] spacing
- [x] font-size
- [x] focus (outline/ fake focus)
- [x] border radius 
- [x] borders
- [x] shadows
- [x] typography
- [x] flex
- [x] elevation
- [x] overflow
- [x] display
- [x] screen readers
- [x] media queries (break points)
- [x] modifiers for responsive paddings ( e.g. `--sm`, `--md`)
- [x] responsive paddings
- [x] documentation - on the main styles nav

droshev:
Let's advertise how ppl can use "their" normalize from any other library

PR #3925, opened 2022-10-13
feat(styles): add focus to common css

InnaAtanasova:
## Related Issue
part of https://github.com/SAP/fundamental-styles/issues/3720

## Description
add Visual Focus Concept to common css

### DIFF
--- a/src/common-css/_common-mixins.scss
+++ b/src/common-css/_common-mixins.scss
@@ -971,3 +971,94 @@ $sap-font-sizes: (
     border: $_border;
   }
 }
+
+// FOCUS
+@mixin sap-fiori-focus($offset: -0.1875rem, $alternative: false) {
+  &:focus,
+  &.is-focus {
+    outline-offset: $offset;
+    outline-width: var(--sapContent_FocusWidth);
+    outline-style: var(--sapContent_FocusStyle);
+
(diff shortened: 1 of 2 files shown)


src/common-css/_common-mixins.scss
```diff
@@ -936,3 +936,100 @@ $sap-font-sizes: (
   }
   @content;
 }
+
+// FOCUS
+@mixin sap-fiori-focus($offset: -0.1875rem, $alternative: false) {
+  &:focus,
+  &.is-focus {
+    @if $offset != 0 {
+      outline-offset: $offset;
```
mikerodonnell89, line 945: If the dev passes 0 here, I would think we want to set the outline offset to 0? If they deliberately pass 0 and this `if` is skipped over, the outline-offset could be unintentionally set by some other selector

mikerodonnell89 (changes requested):

src/common-css/_common-mixins.scss
InnaAtanasova: We always had this mixin like this. It makes sense what you said so I removed the check

platon-rov (approved):

mikerodonnell89 (approved):

PR #3926, opened 2022-10-13
feat(styles): add border mixin to common-css

InnaAtanasova:
## Related Issue
part of https://github.com/SAP/fundamental-styles/issues/3720

## Description
add border helper mixin to common-css

### DIFF
--- a/src/common-css/_common-mixins.scss
+++ b/src/common-css/_common-mixins.scss
@@ -936,3 +936,38 @@ $sap-font-sizes: (
   }
   @content;
 }
+
+// BORDER
+@mixin sap-border($width, $style, $color, $pos: all) {
+  $_border: $width $style $color;
+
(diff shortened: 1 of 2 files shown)


mikerodonnell89 (commented):
Don't want to sound harsh, but this one seems a bit extra to me. The only use cases in which it actually cuts down on code is if pos is X or Y. And even then, just one line. Personally I'd rather just see native CSS than these custom functions that don't add much value

InnaAtanasova:
> Don't want to sound harsh, but this one seems a bit extra to me. The only use cases in which it actually cuts down on code is if pos is X or Y. And even then, just one line. Personally I'd rather just see native CSS than these custom functions that don't add much value

it's mostly for border-left and border-right because it handles RTL

platon-rov (approved):

PR #3956, opened 2022-10-25
fix(styles): add flex-center to list-title-text

shrvr:
## Related Issue
Support for SAP/fundamental-ngx#8805

## Description
Added class `fd-list__title-text` with flex-center

### Before:
image

image

### After:
image

image

#### Please check whether the PR fulfills the following requirements

- [x] Mixins are used for repeatable code (`fd-rtl`, `fd-ellipsis`, `fd-flex`, `fd-selected`, `fd-focus`, ect.)

PR #3988, opened 2022-11-01
fix(styles): need only vertical alignment to list-title-text

shrvr:
## Related Issue
Produced from SAP/fundamental-styles#3956

## Description
Needed only vertical alignment to list-title-text

### Before:
image
### After:
image


packages/styles/src/mixins/list/_list-base.scss
```diff
@@ -128,7 +128,8 @@
     }
 
     .#{$block }__title-text {
-      @include fd-flex-center();
+      display: flex;
+      align-items: center;
```
InnaAtanasova, line 132: Please use the `fd-flex-vertical-center` mixin

InnaAtanasova (changes requested):

mikerodonnell89 (approved):

InnaAtanasova (approved):

PR #3995, opened 2022-11-04
fix(styles): add changes to Slider related to version 1 [ci visual]

InnaAtanasova:
## Related Issue
Part of  SAP/fundamental-styles#](https://github.com/SAP/fundamental-styles/issues/3652)

## Description
add code changes related to the latest design, preparation for version 1


mikerodonnell89 (changes requested):
Reading the specs, I think all sliders need to become transparent when focused, not just the range slider. I think ui5 web components needs to fix this as well

InnaAtanasova:
> Reading the specs, I think all sliders need to become transparent when focused, not just the range slider. I think ui5 web components needs to fix this as well

I don't know... these are the images from the specs. In all images the pattern is consistent:
Screenshot 2022-11-09 at 3 02 07 PM
Screenshot 2022-11-09 at 3 01 49 PM

And from the specs:
Screenshot 2022-11-09 at 3 08 45 PM
Screenshot 2022-11-09 at 3 07 06 PM

mikerodonnell89:
Hmm.... I saw this, which says it should be transparent for slider value or range. So I guess there is some discrepancy in the specs. I'm fine to move forward with this PR until the designers ask us otherwise

Screen Shot 2022-11-09 at 1 12 57 PM

mikerodonnell89 (approved):

platon-rov (approved):

N1XUS (approved):

PR #4017, opened 2022-11-11
chore(styles): info label v1

g-cheishvili:
## Related Issue
Closes SAP/fundamental-styles#3685

## Description
Fixed one issue with the misalignment of the icon. Moved variables to the common variables files. Made sure that every variable matches spec.


InnaAtanasova (commented):
Please try to add more info/description to the examples, like what modifier classes are applied, etc. Part of the v1 task is improvement of our docs/API

mikerodonnell89 (approved):

PR #4025, opened 2022-11-15
fix(cx): add fixes after design review [ci visual]

InnaAtanasova:
## Related Issue
none

## Description
- fixed the arrow alignment
- changed `box-shadow: var(--sapContent_Shadow1);` to `box-shadow: var(--sapContent_Shadow0);`
- added functionality and examples for the case where a list item can have a link (text and icon) and expand/collapse functionality (button)
- added example to the popover with the case where the first level item with button is a link

### DIFF
--- a/packages/cx/src/_nested-list.scss
+++ b/packages/cx/src/_nested-list.scss
@@ -78,12 +78,27 @@ $button: #{$fd-cx-namespace}-button;
     &--header {
       margin-top: 0.25rem;
 
-      .#{$block}__link {
-        cursor: auto;
+      h1,
+      h2,
+      h3,
+      h4,
+      h5,
+      h6 {
+        &.#{$block}__link {
+          cursor: auto;
 
-        @include fd-hover() {
-          @include fd-nested-list-states-colors(var(--sapList_TextColor), var(--sapList_Background));
+          @include fd-hover() {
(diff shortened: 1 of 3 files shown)


packages/cx/src/_nested-list.scss
```diff
@@ -217,6 +233,35 @@ $button: #{$fd-cx-namespace}-button;
...
+
+    [class*=sap-icon] {
+      @include fd-flex-center();
+      @include fd-set-width(1rem);
+      @include fd-set-height(1rem);
+
+      line-height: 1;
+
+      @include fd-rtl() {
+        -moz-transform: scale(-1, 1);
```
mikerodonnell89, line 256: I'm surprised you haven't made a mixin for this!

mikerodonnell89 (approved):
LGTM

packages/cx/src/_nested-list.scss
N1XUS, line 256: Do we need all those vendor prefixes? isn't Autoprefixer does that by default?

InnaAtanasova: Autoprefixer does indeed that by default but in the past we faced issues with Nested List and had to manually add these (this was added as a fix).

InnaAtanasova: I removed them :)

N1XUS (approved):

PR #4026, opened 2022-11-15
fix(styles): popover horizon and v1 refactoring

platon-rov:
## Related Issue

Refers to 

## Description

* Popover styles refactoring
* Popover docs refactoring
* Popover placement feature
* Popover minor issues fixes

*Only one misalignment with specs is the arrow position, it should be `0.5rem` accordingly to the designs but in that case arrow won't be centered to the button with icon.*

CleanShot 2022-11-16 at 10 38 36@2x

### Before:

CleanShot 2022-11-15 at 22 25 20@2x

### After:

CleanShot 2022-11-15 at 22 25 43@2x

CleanShot 2022-11-15 at 22 26 04@2x


N1XUS (approved):
Nice job!

InnaAtanasova:
Screenshot 2022-11-16 at 9 51 06 AM

Hey @platon-rov please add the project and milestone to your PRs. It helps us to monitor what gets done during a sprint and also I don't have to go and add these to the PRs as a primary school teacher :D

InnaAtanasova:
@platon-rov I see changes in the markup (adding classes). This should be considered a Breaking Change, shouldn't it?

platon-rov:
> @platon-rov I see changes in the markup (adding classes). This should be considered a Breaking Change, shouldn't it?

Interesting question, accordingly to the semver.org any new feature or deprecation should bump minor (0.X.0) version number and in our case minor is a major because the major is always 0 because we're in beta. But we don't mark new features as breaking changes, because they don't break anything for sure. The same with deprecation, it doesn't brake anything by itself, only when we then remove this deprecation we introduce breaking changes, it makes sense. Having breaking changes twice (when deprecating (1) and then when removing (2)) for the same thing is a bit too much.

So in this exact case i deprecated some stuff and updated examples to don't use show deprecated things there.

packages/styles/stories/Components/popover/popover.stories.js
```diff
@@ -34,176 +34,6 @@ export default {
     }
 };
 
-export const Alignment = () => `<div class="fddocs-container" style="margin-bottom: 200px">
```
g-cheishvili, line 37: FYI, You can extract HTML files into the separate HTML files if you want to for proper HTML formating or something

g-cheishvili:
> > @platon-rov I see changes in the markup (adding classes). This should be considered a Breaking Change, shouldn't it?
> 
> Interesting question, accordingly to the semver.org any new feature or deprecation should bump minor (0.X.0) version number and in our case minor is a major because the major is always 0 because we're in beta. But we don't mark new features as breaking changes, because they don't break anything for sure. The same with deprecation, it doesn't brake anything by itself, only when we then remove this deprecation we introduce breaking changes, it makes sense. Having breaking changes twice (when deprecating (1) and then when removing (2)) for the same thing is a bit too much.
> 
> So in this exact case i deprecated some stuff and updated examples to don't use show deprecated things there.

I get your point, related to double breaking changes, but what about `fd-popover--input-message-group` class relocations?

g-cheishvili (approved):
Accept new base on chromatic

platon-rov:
> I get your point, related to double breaking changes, but what about `fd-popover--input-message-group` class relocations?

I've added a new class `fd-popover__body--input-message-group` and changed usages in docs to use this way, but also left the old one `fd-popover--input-message-group` so it's still just a deprecation, not removing.

Issue #3868, opened 2022-09-13
Time Picker new design updates


obc514:
Time Picker got a new design(not only Horizon). Let's review and plan it
(https://wiki.one.int.sap/wiki/pages/viewpage.action?pageId=2698915874) and the [UI5 Implementation](https://sapui5nightly.int.sap.eu2.hana.ondemand.com/#/entity/sap.m.TimePicker/sample/sap.m.sample.TimePicker)

PR #4030, opened 2022-11-18
feat(styles): add new Time Picker component

InnaAtanasova:
## Related Issue
Closes https://github.com/SAP/fundamental-styles/issues/3868

## Description
- deprecated the Pattern Time Picker
- added a new component Time Picker
- added a modifier class for buttons when used in Dialog footer `fd-button--decisive`

### DIFF
--- a/packages/styles/src/timepicker.scss
+++ b/packages/styles/src/timepicker.scss
@@ -2,16 +2,67 @@
 @import './mixins';
 
 $block: #{$fd-namespace}-time-picker;
+$fd-time-picker-element-size: 2.75rem;
+$fd-time-picker-element-size-compact: 2rem;
+
+$fd-time-picker-clock-face: 18rem;
+$fd-time-picker-clock-face-compact: 16.5rem;
 
 .#{$block} {
-  @include fd-reset();
+  &__dropdown {
+    @include fd-reset();
 
+    padding: 1rem;
+  }
+
+  &__header {
+    @include fd-reset();
+
+    @include fd-flex-center() {
+      gap: 0.75rem;
+    }
+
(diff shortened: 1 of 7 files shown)


packages/styles/src/timepicker.scss
```diff
@@ -0,0 +1,188 @@
...
+@while $i < 361 {
+  .#{$block}__item--deg#{$i} {
+    transform: translate(-50%) rotate(#{$i}deg);
+
+    .#{$block}__number {
+      transform: rotate(-#{$i}deg);
+    }
+  }
+  $i: $i + 6;
+}
```
g-cheishvili, line 188: CSS variables would take a lot less space for this one

g-cheishvili: But I would go with [Data attributes](https://css-tricks.com/a-complete-guide-to-data-attributes/#aa-using-data-attributes-visually)

g-cheishvili (changes requested):

packages/styles/src/timepicker.scss
InnaAtanasova: could you be more specific?

InnaAtanasova:
@g-cheishvili your comments have been addressed. Please check again. Thanks

g-cheishvili:
> @g-cheishvili your comments have been addressed. Please check again. Thanks

```12:29:00 AM: Could not resolve '../../../src/popover.scss' from packages/styles/stories/Deprecated/Components/time-picker/time-picker.stories.js```
It is not building
Also I could not see in HTML how degrees are passed

g-cheishvili (approved):

PR #4063, opened 2022-12-01
fix(styles): popover's visibility for vertical nav

platon-rov:
## Related Issue

Closes none

## Description

Popover's visibility for vertical nav issue fix

### Before:

CleanShot 2022-12-01 at 13 51 58@2x

### After:

CleanShot 2022-12-01 at 13 52 07@2x


g-cheishvili (approved):
Nice

Issue #3589, opened 2022-06-16
Some input components default margin-top can cause problems in some cases


mikerodonnell89:
See this example in fundamental-ngx:

![image](https://user-images.githubusercontent.com/2471874/174104925-5ab41df6-6b43-47ce-8f7a-3b772a5ae452.png)

The select has a margin-top by default that causes the select to become misaligned from other components. I am not sure if there should be a new modifier class or some other solution to resolve this

PR #4079, opened 2022-12-05
fix(styles): add fix for margin-bottom of Select [ci visual]

InnaAtanasova:
## Related Issue
Closes https://github.com/SAP/fundamental-styles/issues/3589

## Description
The inputs (by design) have margins top and bottom of 0.25rem. When the inputs are controls, aka open/close a menu, which is the case of Select component, the bottom margin should disappear. In recent fix the margin-bottom of the Select control was set to 0, which makes the control not vertically aligned when positioned with other controls. 
The current fix sets the margin-bottom to 0 only when the input has aria-expanded set to true or the class `is-expanded` is added to the element

### DIFF
--- a/packages/styles/src/select.scss
+++ b/packages/styles/src/select.scss
@@ -88,8 +88,11 @@ $fd-select-padding-x-compact: 0.5rem;
       pointer-events: none;
     }
 
+    @include fd-expanded() {
+      margin-bottom: 0;
+    }
+
     padding: 0;
-    margin-bottom: 0;
 
     .#{$block}__button {
       @include fd-set-margin-left(0.25rem);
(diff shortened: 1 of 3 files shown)


droshev (approved):

packages/styles/src/select.scss
```diff
@@ -89,7 +89,14 @@ $fd-select-padding-x-compact: 0.5rem;
     }
 
     padding: 0;
-    margin-bottom: 0;
+
+    &[aria-expanded="true"] {
```
N1XUS, line 93: You can use `fd-expanded` mixin here :)

N1XUS (changes requested):

packages/styles/src/select.scss
InnaAtanasova: true! Thank you. It's now fixed

platon-rov (approved):

PR #4103, opened 2022-12-14
feat(styles): list unread indicator

N1XUS:
## Related Issue
relates https://github.com/SAP/fundamental-ngx/issues/8874

## Description
Changed unread dot indicator styling to support different types of byline list.

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation

### DIFF
--- a/packages/styles/src/mixins/list/_list-byline.scss
+++ b/packages/styles/src/mixins/list/_list-byline.scss
@@ -51,8 +51,6 @@ $fd-list-byline-section-text-color: var(--sapContent_LabelColor) !default;
     }
 
     .#{$block}__notification {
-      @include fd-reset();
-
       position: absolute;
       top: 50%;
 
@@ -152,7 +150,7 @@ $fd-list-byline-section-text-color: var(--sapContent_LabelColor) !default;
 
     &.#{$block}--unread-indicator {
       .#{$block}__item {
-        @include fd-set-paddings-x(var(--fdList_Unread_Indicator_Spacing), $fd-list-byline-item-padding-x);
+        @include fd-set-paddings-x($fd-list-unread-indicator-spacing, $fd-list-byline-item-padding-x);
       }
     }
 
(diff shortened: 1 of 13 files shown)


g-cheishvili (approved):

packages/styles/src/mixins/list/_list-byline.scss
```diff
@@ -51,8 +51,14 @@ $fd-list-byline-section-text-color: var(--sapContent_LabelColor) !default;
     }
 
     .#{$block}__notification {
-      @include fd-set-paddings-x(1rem, 0.75rem);
+      @include fd-reset();
```
InnaAtanasova, line 54: Screenshot 2022-12-14 at 9 44 05 AM

You don't need reset. The reset is coming from the `sap-icon-*` class

packages/styles/src/theming/sap_fiori_3.scss
```diff
@@ -161,6 +161,7 @@
   --fdList_Button_Spacing: 0.125rem;
   --fdList_Growing_Font_Weight: normal;
   --fdList_Active_Outline_Color: var(--sapContent_ContrastFocusColor);
+  --fdList_Unread_Indicator_Spacing: 2.125rem;
```
InnaAtanasova, line 164: why is this a theming parameter when its value is the same in all themes?

InnaAtanasova (changes requested):

packages/styles/src/theming/sap_fiori_3.scss
N1XUS: It is used across multiple files. Like for byline list, navigation list item, no border list item

InnaAtanasova: that's not a good reason. If it's used in multiple files, it should be placed somewhere where all files have access to it. Imagine the value needs to be updated. You have to go to all these files and update it.

InnaAtanasova: in the theming files we should have only variables that concern the theming delta

InnaAtanasova (approved):
Thanks!

Issue #3955, opened 2022-10-24
Defect Hunting Sprint 103


InnaAtanasova:
Please list the issues you've found during the Defect Hunting. Don't forget to mention if it's browser specific or theme specific.

platon-rov:
* Input Group, No focus styles, Horizon
<img width="225" alt="CleanShot 2022-12-16 at 19 05 37@2x" src="https://user-images.githubusercontent.com/20265336/208161128-6fd91e0c-ebf6-4d1d-8556-5055f2078ae0.png">

platon-rov:
* Textarea, No blue outline on focus, Horizon
<img width="168" alt="CleanShot 2022-12-16 at 19 08 23@2x" src="https://user-images.githubusercontent.com/20265336/208161555-3896af08-7b97-4faf-8676-60e9180ac7ab.png">

platon-rov:
* Breadcrumbs, Insufficient paddings, Horizon
<img width="48" alt="CleanShot 2022-12-16 at 19 11 46@2x" src="https://user-images.githubusercontent.com/20265336/208162147-3fa67104-db51-4b60-9965-1252dbcda9b5.png">

Specs:
<img width="114" alt="CleanShot 2022-12-16 at 19 12 08@2x" src="https://user-images.githubusercontent.com/20265336/208162178-fca8e46d-2f62-4c66-953a-e575944439dc.png">

platon-rov:
* Object Marker, Wrong icon color in active state if used as Link, Horizon
<img width="81" alt="CleanShot 2022-12-16 at 19 22 25@2x" src="https://user-images.githubusercontent.com/20265336/208164015-58c32ff0-79cb-4555-80a9-31e33a14c113.png">

platon-rov:
* Object status, broken example, Horizon

<img width="566" alt="CleanShot 2022-12-16 at 19 23 43@2x" src="https://user-images.githubusercontent.com/20265336/208164154-8e3099c4-e435-4d15-b2f7-d6adbf96cd14.png">

platon-rov:
* Quick View, Broken examples, span tags used with Input classes, Horizon

<img width="283" alt="CleanShot 2022-12-16 at 19 25 32@2x" src="https://user-images.githubusercontent.com/20265336/208164499-c313ff9c-297f-4779-a839-c924baaabb1f.png">

platon-rov:
* Shellbar, Wrong outline shape, should be round, Horizon
[...]

PR #4124, opened 2022-12-19
fix(styles): object marker link & refactoring

platon-rov:
## Related Issue

Relates to https://github.com/SAP/fundamental-styles/issues/3955#issuecomment-1355371525

## Description

Object Marker link mode style fixes & component refactoring.

### Before:
CleanShot 2022-12-19 at 15 56 44@2x
CleanShot 2022-12-19 at 15 57 42@2x

### After:
CleanShot 2022-12-19 at 15 48 15@2x
CleanShot 2022-12-19 at 15 58 35@2x

### DIFF
--- a/packages/styles/src/mixins/_mixins.scss
+++ b/packages/styles/src/mixins/_mixins.scss
@@ -129,11 +129,17 @@
   }
 }
 
-@mixin fd-link-underline() {
+@mixin fd-link-underline($content-selector: null) {
   text-decoration: var(--fdLink_Text_Decoration, none);
 
   &__content {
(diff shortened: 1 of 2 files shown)


mikerodonnell89 (changes requested):
Looking at the specs, it says to follow link, and the icon in link changes color with the text as well when focused

platon-rov:
> Looking at the specs, it says to follow link, and the icon in link changes color with the text as well when focused

Spec says about the text, not the icon

CleanShot 2022-12-20 at 18 08 09@2x

And if we look at the SAPUI5 example it behaves like that - text changes color, icon doesn't

PR #4154, opened 2023-01-05
fix(styles): form message with icon layout, styles & docs refactoring

platon-rov:
## Related Issue

Closes https://github.com/SAP/fundamental-styles/issues/4202.

## Description

Fixes issue when icon placed on the separate line in Form Message.

### Before:
https://user-images.githubusercontent.com/20265336/217062688-ec41c0c9-f256-46ce-bd76-7e8da0d9a78e.mov

### After:
https://user-images.githubusercontent.com/20265336/217062546-ea706ed3-0869-4a94-bf3c-cf95362a91c0.mov

### DIFF
--- a/packages/styles/src/form-message.scss
+++ b/packages/styles/src/form-message.scss
@@ -66,7 +66,7 @@ $form-message-states: (
 
   &--static {
     margin-top: 0;
-    display: table;
+    display: flex;
     width: auto;
   }
 
--- a/packages/styles/src/form-message.scss
+++ b/packages/styles/src/form-message.scss
@@ -65,9 +65,12 @@ $form-message-states: (
   }
 
   &--static {
-    margin-top: 0;
-    display: flex;
     width: auto;
+  }
+
+  &--embedded {
+    box-shadow: none;
+    max-width: 100%;
   }
 
   @each $set-name, $state-set in $form-message-states {
--- a/packages/styles/src/input.scss
+++ b/packages/styles/src/input.scss
@@ -15,10 +15,6 @@ $block: #{$fd-namespace}-input;
     z-index: 4;
   }
 
-  &:last-child {
-    margin-bottom: $fd-form-input-margin;
-  }
-
   &--compact {
     @include fd-input-field-base-compact();
   }
--- a/packages/styles/src/popover.scss
+++ b/packages/styles/src/popover.scss
@@ -58,6 +58,10 @@ $fd-popover-arrow-offset-x: calc(var(--fdIcon_Button_Width) * 0.5 - #{$fd-popove
       > * {
         margin-bottom: 0;
       }
+
+      & + .#{$block}__body--input-message-group {
+        margin-top: 0;
+      }
     }
 
     @include fd-disabled() {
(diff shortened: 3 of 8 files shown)


InnaAtanasova (changes requested):
In which component is this issue? I can't reproduce it in Form Message component. Also, your fix is introducing Breaking Changes and we should give it a try to fix the problem without introducing them.

g-cheishvili:
Even though position is fixed now, `/docs/components-forms-form-message--success` this does not show up now at all now

platon-rov:
> In which component is this issue? I can't reproduce it in Form Message component. Also, your fix is introducing Breaking Changes and we should give it a try to fix the problem without introducing them.

Here is the current workflow

1. Default state ⇒ margin (yellow)
image
2. Expanded state ⇒ margin removed (see no yellow at bottom) with `fd-expanded()`
image
3. Default state with form-message ⇒ popover lifted with `.fd-popover__body--input-message-group` class
image

So, we're in the situation when we should apply and keep both approaches to avoid big breaking changes, as moving to one of them bring us A LOT of breaking changes.

Only one thing I've done in select is that I applied (3) to the select to remove margin in default state when there is form message.

platon-rov:
> Even though position is fixed now, `/docs/components-forms-form-message--success` this does not show up now at all now

It's fragile docs, not related to form message itself, see for more https://github.com/SAP/fundamental-styles/issues/4166

g-cheishvili:
> Even though position is fixed now, `/docs/components-forms-form-message--success` this does not show up now at all now

Screenshot 2023-01-09 at 21 42 17

It appeared, but shows like this

platon-rov:
> It appeared, but shows like this

Still the same reason, fragile js in docs, https://github.com/SAP/fundamental-styles/issues/4166

mikerodonnell89:
Some additional margin is displayed when I'm clicking in and out of the input, pushing the code example down. Maybe related to #4166 ?
https://user-images.githubusercontent.com/2471874/214439020-d9a7f9c7-1df5-43a7-93c2-dde7440e2c89.mov

platon-rov:
> In which component is this issue? I can't reproduce it in Form Message component. Also, your fix is introducing Breaking Changes and we should give it a try to fix the problem without introducing them.

https://github.com/SAP/fundamental-styles/pull/4154#issuecomment-1375886407

InnaAtanasova:
What's the progress on this issue? The messages still appear broken. If the issue is JS, please make them static (to always show).
Screenshot 2023-02-14 at 2 50 47 PM
Screenshot 2023-02-14 at 2 50 38 PM

InnaAtanasova:
Screenshot 2023-02-15 at 3 02 05 PM

Please make the example area bigger so that the message is not cut. Thank you!

platon-rov:
> Please make the example area bigger so that the message is not cut. Thank you!

raised ticket https://github.com/SAP/fundamental-styles/issues/4259, looks like `iframeHeight` doesn't change anything at the moment.

InnaAtanasova (approved):

PR #4155, opened 2023-01-06
feat(styles): introduced Dot type in illustrated-message

g-cheishvili:
## Related Issue
Part of SAP/fundamental-ngx#9114

## Description
Styles implementation of "dot" type illustrated message

### DIFF
--- a/packages/styles/src/illustrated-message.scss
+++ b/packages/styles/src/illustrated-message.scss
@@ -3,6 +3,9 @@
 $block: fd-illustrated-message;
 
 .#{$block} {
+  --illustrationW: 20rem;
+  --illustrationH: 15rem;
+
   @include fd-reset();
 
   @include fd-flex(column) {
@@ -19,8 +22,8 @@ $block: fd-illustrated-message;
   &__illustration {
     @include fd-reset();
 
-    height: 15rem;
-    width: 20rem;
+    width: var(--illustrationW);
+    height: var(--illustrationH);
   }
 
   &__figcaption {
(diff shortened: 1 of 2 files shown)


packages/styles/src/illustrated-message.scss
```diff
@@ -163,4 +163,43 @@ $block: fd-illustrated-message;
...
+      height: 2.813rem;
+      width: 2.813rem;
+      margin-right: 0.25rem;
+      &+.#{$block}__figcaption {
+        width: calc(100% - 2.813rem - 0.25rem);
+      }
+    }
+
+    .#{$block}__figcaption {
+      .#{$block}__title {
```
InnaAtanasova, line 183: do you need this specificity? Can't you just use `.#{$block}__title {...}`, `.#{$block}text {...}`?

packages/styles/src/illustrated-message.scss
```diff
@@ -163,4 +163,43 @@ $block: fd-illustrated-message;
...
+
+  &--dot {
+    padding: 0.25rem;
+    flex-direction: row;
+    flex-wrap: wrap;
+    margin: 0;
+
+    .#{$block}__illustration {
+      height: 2.813rem;
+      width: 2.813rem;
```
InnaAtanasova, line 175: this value should go in a variable as it's used 3 times.

InnaAtanasova: you can also use a mixin here:
```
@mixin fd-square($size) {
  height: $size;
  width: $size;
  min-width: $size;
}
```

packages/styles/src/illustrated-message.scss
```diff
@@ -163,4 +163,43 @@ $block: fd-illustrated-message;
...
+      margin-right: 0.25rem;
```
InnaAtanasova, line 176: you need to handle RTL, better use a mixin which does this automatically:
```
@mixin fd-set-margin-right($right: 0) {
  margin-right: $right;

  @include fd-rtl() {
    margin-right: 0;
    margin-left: $right;
  }
}
```

InnaAtanasova (approved):

Issue #4091, opened 2022-12-09
Horizon Delta for User Menu


InnaAtanasova:
The User Menu should be updated to match the Horizon specs

PR #4168, opened 2023-01-10
fix(styles): add Horizon delta for User Menu [ci visual]

InnaAtanasova:
## Related Issue
Closes https://github.com/SAP/fundamental-styles/issues/4091

### DIFF
--- a/packages/styles/src/user-menu.scss
+++ b/packages/styles/src/user-menu.scss
@@ -25,6 +25,12 @@ $block: #{$fd-namespace}-user-menu;
     &.#{$block}__body--default {
       width: var(--fdUserMenu_Body_Width);
     }
+
+    ul:last-child li:last-child::before,
+    ul li:last-child::before,
+    ul li:last-child a::before {
+      border-radius: 0 !important;
+    }
   }
 
   &__header {
--- a/packages/styles/src/user-menu.scss
+++ b/packages/styles/src/user-menu.scss
@@ -28,7 +28,10 @@ $block: #{$fd-namespace}-user-menu;
 
     ul:last-child li:last-child::before,
     ul li:last-child::before,
-    ul li:last-child a::before {
+    ul li:last-child a::before,
+    ul:first-child li:first-child::before,
+    ul li:first-child::before,
+    ul li:first-child a::before {
       border-radius: 0 !important;
     }
   }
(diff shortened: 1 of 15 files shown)


mikerodonnell89 (changes requested):
I would think when there is a footer present, the last list item should not have bottom border radii when focused:

Screenshot 2023-01-11 at 10 17 46 AM

platon-rov (commented):
Why do we have such big spacing?

image

Looks like in wiki it's 8px always
image

Plus, rounding of focus outline seems redundant when there are paddings.

InnaAtanasova:
> https://deploy-preview-4168--fundamental-styles.netlify.app



> I would think when there is a footer present, the last list item should not have bottom border radii when focused:
> 
> Screenshot 2023-01-11 at 10 17 46 AM

This is List component used inside User Menu. The fix should be coming from List, not this component (out of this PR scope)

InnaAtanasova:
> Why do we have such big spacing?
> 
> image
> 
> Looks like in wiki it's 8px always image
> 
> Plus, rounding of focus outline seems redundant when there are paddings.

For the rounded corners: This is List component used inside User Menu. The fix should be coming from List, not this component (out of this PR scope). If we fix it only here, other components that have similar case will also have to fix it.

As for the spacing, the Popover paddings are these:
Screenshot 2023-01-13 at 10 19 01 AM

InnaAtanasova:
created an issue for the rounded corners: https://github.com/SAP/fundamental-styles/issues/4173

InnaAtanasova:
> I would think when there is a footer present, the last list item should not have bottom border radii when focused:
> 
> Screenshot 2023-01-11 at 10 17 46 AM

fixed

mikerodonnell89 (approved):

packages/styles/src/user-menu.scss
```diff
@@ -25,6 +25,12 @@ $block: #{$fd-namespace}-user-menu;
     &.#{$block}__body--default {
       width: var(--fdUserMenu_Body_Width);
     }
+
+    ul:last-child li:last-child::before,
+    ul li:last-child::before,
+    ul li:last-child a::before {
+      border-radius: 0 !important;
```
mikerodonnell89, line 32: Probably first-child too?
Screenshot 2023-01-24 at 4 01 24 PM

mikerodonnell89 (changes requested):

packages/styles/src/user-menu.scss
InnaAtanasova: fixed

mikerodonnell89 (approved):

Issue #4197, opened 2023-01-19
Review chromatic delta


droshev:
Review chromatic delta

- [ ] list all issues
- [ ] fix the issues

PR #4198, opened 2023-01-19
fix(styles): select's, step-input's button border styles in hc [ci visual]

platon-rov:
## Related Issue

Closes https://github.com/SAP/fundamental-styles/issues/4197

## Description

Fixes Select's and StepInput's button border styles in HC themes.

### Before:
image
image

### After:
image
image

### DIFF
--- a/packages/styles/src/mixins/_forms.scss
+++ b/packages/styles/src/mixins/_forms.scss
@@ -233,6 +233,34 @@ $fd-input-field-height--compact: 1.625rem;
   }
 }
 
+@mixin _fd-input-group-button-first-child($selector) {
+  #{$selector} {
+    @include fd-set-border-right(var(--fdInputGroup_ControlButton_SideBorder));
+
+    @include fd-hover() {
+      @include fd-set-border-right(var(--fdInputGroup_ControlButton_SideBorder_Active));
+    }
+
+    @include fd-active() {
+      @include fd-set-border-right(var(--fdInputGroup_ControlButton_SideBorder_Active));
+    }
+  }
+}
+
+@mixin _fd-input-group-button-last-child($selector) {
+  #{$selector} {
+    @include fd-set-border-left(var(--fdInputGroup_ControlButton_SideBorder));
+
+    @include fd-hover() {
+      @include fd-set-border-left(var(--fdInputGroup_ControlButton_SideBorder_Active));
+    }
+
+    @include fd-active() {
+      @include fd-set-border-left(var(--fdInputGroup_ControlButton_SideBorder_Active));
+    }
+  }
+}
+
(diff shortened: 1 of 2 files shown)


packages/styles/src/mixins/_forms.scss
```diff
@@ -233,6 +233,34 @@ $fd-input-field-height--compact: 1.625rem;
   }
 }
 
+@mixin _fd-input-group-button-first-child($selector) {
```
InnaAtanasova, line 236: Please don't prefix the mixins. It should be `@mixin fd-input....`

packages/styles/src/mixins/_forms.scss
```diff
@@ -233,6 +233,34 @@ $fd-input-field-height--compact: 1.625rem;
...
+    @include fd-hover() {
+      @include fd-set-border-left(var(--fdInputGroup_ControlButton_SideBorder_Active));
+    }
+
+    @include fd-active() {
+      @include fd-set-border-left(var(--fdInputGroup_ControlButton_SideBorder_Active));
+    }
+  }
+}
+
```
InnaAtanasova, line 263: The 2 mixins can be combined into one and pass first/last as a parameter, together with the selector. Something like:
```
@mixin fd-input-group-button($selector, $child) {
    #{$selector} {
        @if($child === last) {...}
        @else {....}
    }
}

```

packages/styles/src/mixins/_forms.scss
```diff
@@ -257,37 +285,14 @@ $fd-input-field-height--compact: 1.625rem;
...
-        }
-      }
-    }
-  }
+  // To support `fd-input-group__addon` and avoid breaking changes
+  @include _fd-input-group-button-first-child(".#{$block}__addon:first-child .#{$fd-namespace}-button");
+  @include _fd-input-group-button-last-child(".#{$block}__addon:not(:first-child) .#{$fd-namespace}-button");
 
-  .#{$block}__button {
+  .#{$fd-namespace}-button {
```
InnaAtanasova, line 295: with this you are breaking the self-contained rule. You are directly modifying button inside Select component.

packages/styles/src/mixins/_forms.scss
```diff
@@ -317,7 +322,7 @@ $fd-input-field-height--compact: 1.625rem;
   @each $set-name, $set-props in $fd-input-group-button-states {
     &.is-#{$set-name} {
 
-      .#{$block}__button {
+      .#{$fd-namespace}-button {
```
InnaAtanasova, line 325: same as above

InnaAtanasova (changes requested):

InnaAtanasova (approved):

PR #4237, opened 2023-02-06
feat: added js to publish

g-cheishvili:
## Description
This change adds JS module export and also it's typing to published packages. This is for using in cases when you might want to import sources in environment, where you do not have access to bundlers, for example when you are writing package for angular.


N1XUS:
@g-cheishvili shouldn't final package.json be also updated for paths of import?

N1XUS (approved):

Issue #4259, opened 2023-02-16
bug(docs): iframeHeight not being respected


platon-rov:
**Describe the bug**

`iframeHeight` setting not being respected.

<img width="510" alt="SCR-20230216-iuf" src="https://user-images.githubusercontent.com/20265336/219366241-ddd1e471-1e2c-4194-82c6-fe7fce84bfef.png">

IB002:
Due to new update in storybook iframeHeigh is no longer supported.

PR #4277, opened 2023-02-22
feat(styles): introduce compact mixin [ci visual]

### LINKED ISSUE #4273: Enhancement: revisit 'compact' mode selectors.
Current overview of handing 'compact' state:
- Unique class based on BEM technique is needed for every component;
- Some components uses 'compact' selector for inner elements and not for the root one;
- Libraries that uses fundamental-styles (e.g. fundamental-ngx) required to adjust their library code to follow this logic.

On the other hand, we have 'state' modifier, which are not following this technique:
- we have `is-selected`, `is-focus`, `is-hover`, `is-warning`, etc. modifier classes that do not follow BEM logic, and easy to use with almost any component since they are **generic**.

The proposition is to: introduce `.is-compact`, `is-cozy` and `is-condensed` classes which is assigned to the root of the component if needed.
Three of them make sense for cases when root component is compact, inner is cozy and inner's child is condensed. This way we would still keep the unique content density state for each component, yet, if we follow the same logic we have with `fd-rtl()` mixin, we would need to specify `.is-compact` only **once** per root component, and all child elements would follow it, if no `is-cozy` or `fd-condensed` classes applied to child elements.

N1XUS:
Closes #4273

### DIFF
--- a/packages/cx/src/_nested-list.scss
+++ b/packages/cx/src/_nested-list.scss
@@ -265,7 +265,7 @@ $button: #{$fd-cx-namespace}-button;
     }
   }
 
-  @include fd-is-compact() {
+  @include fd-compact() {
     .#{$block}__item {
       margin: 0.25rem 0;
 
--- a/packages/cx/src/_nested-list.scss
+++ b/packages/cx/src/_nested-list.scss
@@ -10,6 +10,15 @@ $button: #{$fd-cx-namespace}-button;
   $fd-nested-list-item-height: 2.75rem;
   $fd-nested-list-item-height-compact: 1.5rem;
   $fd-nested-list-item-border-radius: 0.5rem;
+
+  --fdNestedList_Item_Height: 2.75rem;
+  --fdNestedList_Item_Spacing: 0.25rem;
+  --fdNestedList_Item_Spacing_Compact: 0.125rem;
+  --fdNestedList_Item_Margin: 0.5rem 0;
+  --fdNestedList_Item_Header_Margin_Top: 0.25rem;
+  --fdNestedList_Focus_Offset: 0.125rem;
+  --fdNestedList_Focus_Offset_Width: var(--sapContent_FocusWidth);
+  --fdNestedList_Button_Margin_Left: 0.5rem;
   $fd-nested-list-item-spacing: 0.25rem;
   $fd-nested-list-item-spacing-compact: 0.125rem;
 
@@ -18,7 +27,7 @@ $button: #{$fd-cx-namespace}-button;
     background: $backgroundColor;
   }
 
-  @mixin fd-nested-list-focus($offset: 0.125rem, $width: var(--sapContent_FocusWidth)) {
+  @mixin fd-nested-list-focus($offset: var(--fdNestedList_Focus_Offset), $width: var(--fdNestedList_Focus_Offset_Width)) {
     &:focus,
     &.is-focus {
       outline-offset: $offset;
@@ -39,11 +48,11 @@ $button: #{$fd-cx-namespace}-button;
       @include fd-set-margins-x-equal(0.5rem);
 
       &:first-child {
-        margin-top: $fd-nested-list-item-spacing;
+        margin-top: var(--fdNestedList_Item_Spacing);
       }
 
       &:last-child {
-        margin-bottom: $fd-nested-list-item-spacing;
+        margin-bottom: var(--fdNestedList_Item_Spacing);
       }
     }
 
@@ -73,10 +82,10 @@ $button: #{$fd-cx-namespace}-button;
     @include fd-reset();
 
     display: block;
-    margin: 0.5rem 0;
+    margin: var(--fdNestedList_Item_Margin);
 
     &--header {
-      margin-top: 0.25rem;
+      margin-top: var(--fdNestedList_Item_Header_Margin_Top);
 
       h1,
       h2,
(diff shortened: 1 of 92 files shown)


packages/cx/src/_nested-list.scss
```diff
@@ -265,7 +265,7 @@ $button: #{$fd-cx-namespace}-button;
     }
   }
 
-  &--compact {
+  @include fd-is-compact() {
```
InnaAtanasova, line 268: not a big deal, but I think `fd-compact` is a better name :D

packages/styles/src/mixins/_mixins.scss
```diff
@@ -757,3 +757,29 @@
...
+  @at-root {
+    [class*="--condensed"] &,
+    &[class*="--condensed"] {
+      :not([class*="--cozy"]),
+      :not([class*="--compact"]) {
+        @content;
+      }
+    }
+  }
+}
```
InnaAtanasova, line 785: mixin for cozy state? I don't remember what we decided on this? Do we avoid it because the default state is cozy?

InnaAtanasova:
Looking great so far! I like this approach, it's even better than what we discussed about `is-compact`.

packages/styles/src/mixins/_mixins.scss
N1XUS: @InnaAtanasova I think we should avoid it. `--cozy` should be used only in one case: if parent element is `--compact` or `--condensed`, and inner element should be `--cozy`

InnaAtanasova: yes, that's exactly the case I have in mind, because in the past one of the reasons we avoided setting the `compact` modifier on the parent is that it prevents the user from mixing compact and cozy nested elements. So for example, if you set the parent to be in `compact` mode all children "inherit" it and there's no way to set a child element to `cozy` as it's the default state and doesn't have a modifier.

packages/styles/src/mixins/list/_list-base.scss
```diff
@@ -47,7 +49,9 @@
       padding: 0;
       text-align: center;
       border-bottom: var(--fdList_Item_Action_Border);
-      height: var(--fdList_Item_Action_Height);
+
+      --fdList_Item_Height: var(--fdList_Item_Action_Height);
+      // height: var(--fdList_Item_Action_Height);
```
platon-rov, line 54: forgot to remove

packages/styles/src/token.scss
```diff
@@ -130,14 +130,22 @@ $block: #{$fd-namespace}-token;
...
-    &.#{$block}--compact {
+    @include fd-compact() {
       padding: $fd-token-compact-padding-vertical $fd-token-compact-readonly-padding-horizontal;
 
       @include fd-rtl() {
         padding-left: $fd-token-compact-readonly-padding-horizontal;
       }
     }
 
+    // &.#{$block}--compact {
```
platon-rov, line 141: forgot to remove

packages/styles/stories/Components/Forms/form-message/form-message.stories.js
```diff
@@ -26,7 +26,7 @@ Read-only&nbsp;&nbsp;&nbsp; | \`is-readonly\`&nbsp;&nbsp;&nbsp; | Used to displa
     }
 };
 
-export const Success = () => `<div class="fd-form-item">
+export const Success = () => `<div class="fd-form-item" style="min-height: 100px;">
```
platon-rov, line 29: I prefer to fix the `iframeHeight` instead of doing this, here is the ticket https://github.com/SAP/fundamental-styles/issues/4259

platon-rov (commented):
LGTM.

1. Please create ticket about adding content density switcher to header.

2. How we process with the Bar & Toolbar components, which are compact by default, do we change their behavior or something? I also notice that we don't have fd-cozy mixin so these two won't be affected by global density, isn't it? If they should be processed separately - please create a ticket.

g-cheishvili (changes requested):
- I think `fd-compact-condensed` should be used everywhere instead of compact, because if condensed is chosen and component does not support it, it should fallback to the compact.
I mean, components inside condensed should be compact, if they do not support otherwise

- I believe, that we will need cozy, just for cases when we will need to modify the condensed or compact's child.

g-cheishvili:
> Looking great so far! I like this approach, it's even better than what we discussed about `is-compact`.

I think final destination should be `is-compact`, this is just replacing those fixed `--compact` modifiers with more dynamic one and adds context to them.
`is-compact` is much better than having `--compact` in any form, because you could stick single class name anywhere and it would just work

N1XUS:
@InnaAtanasova please take a look at @g-cheishvili comment regarding class name. I think that the final decision should be made by you and @droshev

InnaAtanasova:
getting rid of `--compact` modifier and introducing `.is-compact` will result in a lot of Breaking Changes. I'd suggest the following 2 options:
- keep the `--compact` modifier on the parent level
- have both options, will be doubled code in fund-styles (I think we can even avoid this with proper setup of the mixins) but in fund-ngx we can handle the breaking changes behind the scene so the users won't be affected.

InnaAtanasova:
> * I think `fd-compact-condensed` should be used everywhere instead of compact, because if condensed is chosen and component does not support it, it should fallback to the compact.
>   I mean, components inside condensed should be compact, if they do not support otherwise
> * I believe, that we will need cozy, just for cases when we will need to modify the condensed or compact's child.

`compact` and `condensed` modes are not alike and should not be mixed. There are 2 components with this mode:
1. Table - and this is the only component that has the notion of cozy, compact and condensed
2. Side Nav - in condensed mode only icons are shown, aka the text is hidden.

platon-rov (approved):

g-cheishvili (approved):
Like the new mixin names

InnaAtanasova (approved):

PR #4282, opened 2023-02-23
fix: popover wrapper styles overriding custom body class styles

g-cheishvili:
## Related Issue
Closes SAP/fundamental-ngx#9228

## Description
Popover wrapper had a reset styles, which were overriding styles that were assigned to the popover body if the user used custom classes


platon-rov (approved):

InnaAtanasova:
> We're using `fd-reset` for all the elements, so i would say it's intended to have it.

we discussed this with Giorgi. In this particular case we can omit it as it's introducing problems with custom classes.

InnaAtanasova (approved):

Issue #4369, opened 2023-03-26
Component Design Update: Calendar


droshev:
- [ ] Horizon specs(visual and interaction) updates
- [ ] Delta theming
- [ ] identify and document breaking changes

PR #4514, opened 2023-05-08
feat(styles): update Calendar component - Horizon 2023

droshev:
## Related Issue
Closes SAP/fundamental-styles#4369

## Description
Update Calendar and Date Picker with the latest Horizon specs

BREAKING CHANGE:
complete calendar markup refactoring(item, navigation, legend). Many classes were removed.

**Navigation**
Before:
```
<header class="fd-calendar__header">
    <div class="fd-calendar__navigation">
...
    </div>
</header>
```

After:
```
<header class="fd-calendar__navigation">
...
</header>
```

**Calendar Day**
Before:
```
<td role="gridcell" class="fd-calendar__item">
    <span class="fd-calendar__text" role="button">8</span>
</td>
```

After:
```
<td role="gridcell" class="fd-calendar__item">
    <div class="fd-calendar__text-wrapper">
        <span class="fd-calendar__text" role="button">8</span>
    </div>
</td>
```

**Calendar Container**
Before:
```
<div class="fd-calendar__content fd-calendar__content--dates" id="IO0cp341">
    <table class="fd-calendar__table" role="grid">
        <thead class="fd-calendar__group">
...
    </table>
</div>
```

After:
```
<table class="fd-calendar__content" id="IO0cp341" role="grid">
    <thead class="fd-calendar__group">
...
</table
```


packages/styles/src/calendar.scss
```diff
@@ -340,16 +205,102 @@ $fd-calendar-special-days: (
...
+    min-width: var(--fdCalendar_Item_Width);
+    border-radius: var(--fdCalendar_Border_Radius);
+    background-color: var(--fdCalendar_Item_Background);
+    flex: 1 1 calc(var(--fdCalendar_Item_Current_Border_Width) * 2 + var(--fdCalendar_Item_Spacing) * 2);
+    cursor: pointer;
+    display: inline-block;
+    border: var(--fdCalendar_Item_Border);
+
+    @include fd-hover() {
+      background-color: var(--sapList_Hover_Background);
```
InnaAtanasova, line 224: you can use a CSS variable for the hover background and then instead of every time writing
```
@include fd-hover() {
    --fdCalendar_Item_Background: var(--sapList_Hover_Background);
}
```

you can just change the var value. 
For example:
`--fdCalendar_Item_Background_Hover: var(--sapList_Hover_Background);`

Here on line 224 you do:
```
@include fd-hover() {
    background-color: var(--fdCalendar_Item_Background_Hover);
}
```

and then on lines 246, 266, etc you just do:
```
--fdCalendar_Item_Background: new value
```

packages/styles/src/calendar.scss
```diff
@@ -126,178 +40,129 @@ $fd-calendar-special-days: (
 );
 
 @mixin special-days($nameColorMap) {
-  @each $name, $color in $nameColorMap {
-    &__special-day--#{$name} {
+  &__special-day {
+    .#{$block}__text::after {
```
InnaAtanasova, line 44: `.#{$block}__text::after` is repeated many times. Maybe put it in a variable where `.#{$block}` is defined.

```
$block: #{$fd-namespace}-calendar;
$specialDay: #{$fd-namespace}-calendar__text::after;
```

then you can use it like this: `.#{$specialDay}`

InnaAtanasova (changes requested):
1. The focus outline should be on top of the special day indicator. Maybe use z-index.
Screenshot 2023-05-08 at 4 13 10 PM

2. Wrong BEM in my opinion in this case:
```
fd-calendar__item fd-calendar__item--weekend fd-calendar__item--current is-active fd-calendar__special-day--1
```
The block is `fd-calendar__item` and the following modifier classes modify the default behaviour: `fd-calendar__item--weekend` and `fd-calendar__item--current`, but the class `fd-calendar__special-day--1` is not used properly. Here are the problems:
- `fd-calendar__special-day--1` is a block and can't be used together with another block (in this case `fd-calendar__item`
- you need `fd-calendar__special-day` to have the `fd-calendar__special-day--1`
But in general, I wouldn't use 2 block classes on the same element

The code is in "Days" example
Screenshot 2023-05-08 at 4 19 56 PM


3. Related to the above comment:
```
fd-calendar__item fd-calendar__special-day fd-calendar__special-day--2
```

the workaround could be to use a modifier class for `fd-calendar__item`, like `fd-calendar__item--special-day-2`

packages/styles/src/calendar.scss
```diff
@@ -125,36 +39,91 @@ $fd-calendar-special-days: (
   "20": var(--sapLegendColor20),
 );
 
-@mixin special-days($nameColorMap) {
-  @each $name, $color in $nameColorMap {
-    &__special-day--#{$name} {
+@mixin legend-days($nameColorMap) {
```
InnaAtanasova, line 42: should be prefixed with `fd-`

packages/styles/src/calendar.scss
```diff
@@ -125,36 +39,91 @@ $fd-calendar-special-days: (
...
 
-      @include fd-active() {
+    .#{$block}__text-wrapper::after {
+      content: "";
+      position: absolute;
+      width: calc(100% - 0.125rem);
+      height: 0.5rem;
+      bottom: 0.0625rem;
+      right: 0;
+      left: 0.0625rem;
```
InnaAtanasova, line 87: RTL for these?

packages/styles/src/calendar.scss
```diff
@@ -125,36 +39,91 @@ $fd-calendar-special-days: (
...
-          }
-        }
+      .#{$block}__text-wrapper::after {
+        content: "";
+        position: absolute;
+        width: calc(100% - 0.125rem);
+        height: var(--fdCalendar_Item_Text_Wrapper_Height);
+        bottom: 0.0625rem;
+        right: 0;
+        left: 0.0625rem;
```
InnaAtanasova, line 120: RTL?

packages/styles/src/calendar.scss
```diff
@@ -340,98 +223,190 @@ $fd-calendar-special-days: (
...
+      left: 0.0625rem;
+      border: solid 0.0625rem var(--sapList_SelectionBorderColor);
+      border-radius: var(--fdCalendar_Item_Today_Selected_Bottom_Radius);
+    }
+
+    &--today.#{$block}__item--range::after {
+      content: "";
+      position: absolute;
+      width: calc(100% - 2px);
+      height: calc(100% - 2px);
```
InnaAtanasova, line 315: 2px should be 0.125rem

packages/styles/src/calendar.scss
```diff
@@ -340,98 +223,190 @@ $fd-calendar-special-days: (
...
+      width: calc(100% - 0.125rem);
+      height: calc(100% - 0.125rem);
+      top: 0.0625rem;
+      bottom: 0.0625rem;
+      left: 0.0625rem;
+      border: solid 0.0625rem var(--sapList_SelectionBorderColor);
+      border-radius: var(--fdCalendar_Item_Today_Selected_Bottom_Radius);
+    }
+
+    &--today.#{$block}__item--range::after {
```
InnaAtanasova, line 311: I don't see any difference between this one and `&--today.#{$block}__item--selected::after`. If they are the same the code should be combined.

packages/styles/src/calendar.scss
```diff
@@ -340,98 +223,190 @@ $fd-calendar-special-days: (
...
-            @extend %fd-calendar-item-current;
-          }
+      &::after {
+        content: "";
+        position: absolute;
+        width: calc(100% - 0.125rem);
+        height: calc(100% - 0.125rem);
+        top: 0.0625rem;
+        bottom: 0.0625rem;
+        left: 0.0625rem;
```
InnaAtanasova, line 336: I see this block of content repeated already 3 times. Maybe extract it to a mixin and reuse it. Will be easier to maintain in future.

packages/styles/src/calendar.scss
```diff
@@ -340,98 +223,190 @@ $fd-calendar-special-days: (
...
-        }
+    @include fd-disabled() {
+      pointer-events: none;
+      opacity: 0.4;
 
-        .#{$block}__text {
-          color: inherit;
-        }
+      &:hover {
+        cursor: not-allowed;
```
InnaAtanasova, line 245: this will not work if you have the code on line 241 (pointer-events: none;)

InnaAtanasova:
RTL issues
Screenshot 2023-06-06 at 12 33 58 PM
Screenshot 2023-06-06 at 12 33 50 PM

packages/styles/src/calendar.scss
```diff
@@ -340,98 +223,190 @@ $fd-calendar-special-days: (
...
+      --fdCalendar_Item_Border: var(--fdCalendar_Active_Item_Border);
+      --fdCalendar_Item_Background: var(--sapList_SelectionBackgroundColor);
+
+      @include fd-hover() {
+        --fdCalendar_Item_Background: var(--sapList_Hover_SelectionBackground);
+      }
+
+      // &.#{$block}__item--current { //TODO REVIEW!!!
+      //   @extend %fd-calendar-item-current;
+      // }
```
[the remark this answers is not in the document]
droshev: removed

InnaAtanasova (approved):

Issue #4502, opened 2023-05-03
Tile Header is using wrong styling.


ShirSAP:
Bug in the tile header.
Header should use "font-weight: bold;"

PR #4516, opened 2023-05-08
fix(styles): update tile title font size [ci visual]

droshev:
## Related Issue
Closes SAP/fundamental-styles#4502

update tile title font size

### DIFF
--- a/packages/styles/src/tile.scss
+++ b/packages/styles/src/tile.scss
@@ -209,7 +209,8 @@ $fd-line-tile-badge-horizontal-spacing: 0.375rem !default;
     @include fd-line-clamp();
 
     color: var(--sapTile_TitleTextColor);
-    font-size: var(--sapFontHeader5Size);
+    font-size: var(--fdTile_Title_Font_Size);
+    font-weight: var(--fdTile_Title_Font_Weight);
   }
 
   &__subtitle,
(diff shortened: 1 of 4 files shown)


g-cheishvili (approved):

packages/styles/src/tile.scss
```diff
@@ -209,7 +209,8 @@ $fd-line-tile-badge-horizontal-spacing: 0.375rem !default;
     @include fd-line-clamp();
 
     color: var(--sapTile_TitleTextColor);
-    font-size: var(--sapFontHeader5Size);
+    font-size: var(--sapFontHeader6Size);
```
InnaAtanasova, line 212: that's the font-size for small Tile

InnaAtanasova (changes requested):

packages/styles/src/tile.scss
droshev: updated

InnaAtanasova (changes requested):
The values are the same for all themes in Horizon and all themes in Quartz. You need to put them in the `common` folder and have only 2 files, 1 for Horizon and 1 for Quartz, instead of 9 files

droshev:
> The values are the same for all themes in Horizon and all themes in Quartz. You need to put them in the `common` folder and have only 2 files, 1 for Horizon and 1 for Quartz, instead of 9 files

done

InnaAtanasova (approved):

PR #4518, opened 2023-05-10
fix(styles): screenreaders attempting to read icons from content [ci visual]

mikerodonnell89:
part of https://github.com/SAP/fundamental-ngx/issues/9556

Screenreaders try to read the "content" property of CSS pseudoelements, so we are seeing this type of thing on checkboxes and icons:

Screenshot 2023-05-10 at 12 51 59 PM

This PR provides an alternate (empty) value so that the content property is not read.


N1XUS:
@mikerodonnell89 please fix commit message

InnaAtanasova:
@mikerodonnell89 can you enable the visual tests? We need to check if this broke something in the other components.

mikerodonnell89:
> @mikerodonnell89 please fix commit message

@N1XUS i think this was a CI hiccup

mikerodonnell89:
> @mikerodonnell89 can you enable the visual tests? We need to check if this broke something in the other components.

@InnaAtanasova done

InnaAtanasova (approved):

PR #4537, opened 2023-05-18
fix: fixed generic tile compact mode

g-cheishvili:
## Related Issue
Closes none

## Description
Generic tile was not using `is-compact` modifiers

### DIFF
--- a/packages/styles/src/tile.scss
+++ b/packages/styles/src/tile.scss
@@ -122,6 +122,18 @@ $fd-line-tile-badge-horizontal-spacing: 0.375rem !default;
   @include fd-set-height($fd-tile-height);
   @include fd-set-width($fd-tile-width);
 
+  --fd-line-tile-height: $fd-line-tile-height;
+  --fd-line-tile-max-height: $fd-line-tile-max-height;
+  --fd-line-tile-line-height: 1.625rem;
+  --fd-line-tile-title-horizontal-spacing: $fd-line-tile-title-horizontal-spacing;
+
+  @include fd-compact-and-condensed() {
+    --fd-line-tile-height: $fd-line-tile-height-compact;
+    --fd-line-tile-max-height: $fd-line-tile-max-height-compact;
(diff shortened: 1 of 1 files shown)


packages/styles/src/tile.scss
```diff
@@ -837,40 +839,21 @@ $fd-line-tile-badge-horizontal-spacing: 0.375rem !default;
...
-      line-height: 1.625rem;
+      line-height: var(--fdGenericTileLineHeight);
       min-width: $fd-line-tile-min-width;
     }
 
     .#{$block}__title {
       @include fd-ellipsis();
 
-      margin-right: $fd-line-tile-title-horizontal-spacing;
+      margin-right: var(--fdGenericTileTitleHorizontalSpacing);
```
InnaAtanasova, line 849: Please use `fd-set-margin-right` mixin

packages/styles/src/tile.scss
```diff
@@ -719,11 +725,7 @@ $fd-line-tile-badge-horizontal-spacing: 0.375rem !default;
         flex-wrap: wrap;
         margin-right: 0;
         margin-left: 0;
```
N1XUS, line 727: `@include fd-set-margins-x-equal(0);`

packages/styles/src/tile.scss
```diff
@@ -837,40 +839,21 @@ $fd-line-tile-badge-horizontal-spacing: 0.375rem !default;
...
     .#{$block}__title {
       @include fd-ellipsis();
 
-      margin-right: $fd-line-tile-title-horizontal-spacing;
+      margin-right: var(--fdGenericTileTitleHorizontalSpacing);
       color: var(--sapButton_TextColor);
       display: inline;
       -webkit-line-clamp: initial;
 
       @include fd-rtl() {
```
N1XUS, line 854: by using `fd-set-margin-right` this rtl block is not needed

N1XUS (approved):

PR #4557, opened 2023-05-25
fix(styles): remove popover --static class, no longer needed

mikerodonnell89:
fixes none, part of https://github.com/SAP/fundamental-ngx/pull/9924

Removes a class that is no longer needed in NGX and was unused in styles


InnaAtanasova:
@mikerodonnell89 could you trigger the visual checks?

g-cheishvili (approved):
It can be a breaking change though and should be documented as such

PR #4634, opened 2023-06-23
feat(styles): add skeleton examples and styling [ci visual]

droshev:
## Related Issue
Closes none

## Description
add skeleton examples and styling

### DIFF
--- a/packages/styles/src/fundamental-styles.scss
+++ b/packages/styles/src/fundamental-styles.scss
@@ -80,6 +80,7 @@
 @import "select";
 @import "shellbar";
 @import "side-nav";
+@import "skeleton";
 @import "slider";
 @import "status-indicator";
 @import "step-input";
--- a/packages/styles/src/skeleton.scss
+++ b/packages/styles/src/skeleton.scss
@@ -1,0 +1,46 @@
+
+@import "./new-settings";
+@import "./mixins";
+
+$block: #{$fd-namespace}-skeleton;
+
+.#{$block} {
+  @include fd-reset();
+
+  display: inline-block;
+  cursor: wait;
+
+  &__canvas {
+    display: block;
+    fill: url(#skeletonGradient);
+
+    --fdSkeletonGradientMiddleColor: #999;
+  }
+
+  &__outer {
+    stop-color: var(--sapContent_Placeholderloading_Background);
+  }
+
+  &__middle {
+    stop-color: var(--fdSkeletonGradientMiddleColor);
+    animation: fd-skeleton-shimmer 2s linear infinite;
+  }
+}
+
+@keyframes fd-skeleton-shimmer {
+  0% {
+    stop-color: var(--sapContent_Placeholderloading_Background);
+  }
+
+  35% {
+    stop-color: var(--fdSkeletonGradientMiddleColor);
+  }
+
+  65% {
+    stop-color: var(--fdSkeletonGradientMiddleColor);
+  }
+
+  100% {
+    stop-color: var(--sapContent_Placeholderloading_Background);
+  }
+}
--- a/packages/styles/stories/Components/skeleton/circle.example.html
+++ b/packages/styles/stories/Components/skeleton/circle.example.html
@@ -1,0 +1,17 @@
+<div class="fd-skeleton" style="width: 50px; height: 50px;">
+    <svg width="100%" height="100%" class="fd-skeleton__canvas">
+        <defs>
+            <linearGradient id="skeletonGradient">
+                <stop offset="0" class="fd-skeleton__outer"></stop>
+                <stop class="fd-skeleton__middle">
+                    <animate attributeName="offset" values="0; 0; 0.5; 1; 1" dur="2s" repeatCount="indefinite"></animate>
+                </stop>
+                <stop offset="1" class="fd-skeleton__outer"></stop>
+            </linearGradient>
+            <mask id="fd-skeleton-2">
+                <circle cx="50%" cy="50%" r="50%" class="ng-star-inserted"></circle>
+            </mask>
+        </defs>
+        <rect x="0" y="0" width="100%" height="100%" mask="url(#fd-skeleton-2)"></rect>
+    </svg>
+</div>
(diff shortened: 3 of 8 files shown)


packages/styles/src/fundamental-styles.scss
```diff
@@ -108,3 +108,4 @@
 
 // BTP Specific components/overwrites
 @import "./btp/avatar.scss";
+@import "skeleton";
```
InnaAtanasova, line 111: why is this here? Under line 109 are only BTP components overwrites aka have the same names as Fiori

packages/styles/src/skeleton.scss
```diff
@@ -0,0 +1,48 @@
...
+    fill: var(--sapContent_Placeholderloading_Background);
+  }
+
+  &__middle {
+    stop-color: var(--fdSkeletonGradientMiddleColor);
+    animation: fd-skeleton-shimmer 2s linear infinite;
+  }
+
+  &--animated {
+    .#{$block}__canvas {
```
InnaAtanasova, line 24: I don't get this specificity, `--fdSkeletonGradientMiddleColor` is part of `fd-skeleton__middle` element where you specify the stop color and the animation. Please check.

packages/styles/src/skeleton.scss
```diff
@@ -0,0 +1,48 @@
...
+    display: block;
+    fill: var(--sapContent_Placeholderloading_Background);
+  }
+
+  &__middle {
+    stop-color: var(--fdSkeletonGradientMiddleColor);
+    animation: fd-skeleton-shimmer 2s linear infinite;
+  }
+
+  &--animated {
```
InnaAtanasova, line 23: I don't get this modifier class. Shouldn't the `fd-skeleton` be always animated? I couldn't find an example with just `fd-skeleton` class. It's always `<div class="fd-skeleton fd-skeleton--animated" style="...">`.
In this case do we need this extra modifier?

packages/styles/stories/Components/skeleton/skeleton.example.html
```diff
@@ -0,0 +1,24 @@
...
+            </linearGradient>
+            <mask id="fd-skeleton-0">
+                <rect x="0" y="0" width="100%" rx="4" height="100%" class="ng-star-inserted"></rect>
+            </mask>
+        </defs>
+        <rect x="0" y="0" width="100%" height="100%" mask="url(#fd-skeleton-0)">
+    
+        </rect>
+    </svg>
+</div>
```
InnaAtanasova, line 24: please add empty line at the end of file

InnaAtanasova (changes requested):

packages/styles/stories/Components/skeleton/circle.example.html
```diff
@@ -0,0 +1,17 @@
+<div class="fd-skeleton fd-skeleton--animated" style="width: 50px; height: 50px;">
+    <svg width="100%" height="100%" class="fd-skeleton__canvas">
+        <defs>
+            <linearGradient id="skeletonGradient">
+                <stop offset="0" stop-color="var(--sapContent_Placeholderloading_Background)"></stop>
+                <stop class="fd-skeleton__middle">
+                    <animate attributeName="offset" values="0; 0; 0.5; 1; 1" dur="2s" repeatCount="indefinite"></animate>
+                </stop>
+                <stop offset="1" stop-color="var(--sapContent_Placeholderloading_Background)"></stop>
```
InnaAtanasova, line 9: for me this is very weird. Why do you set the `stop-color` for the first and last stop in the markup and for the middle you use a class?

packages/styles/src/skeleton.scss
InnaAtanasova: ok, I got it :) It's because you have `<stop offset="0" stop-color="var(--sapContent_Placeholderloading_Background)"></stop>` which is a child element of `fd-skeleton__canvas`. Weird approach but I understand now. I would still use a class for the forst and last stop (like for the middle).

packages/styles/src/fundamental-styles.scss
droshev: this was autogenerated

packages/styles/src/skeleton.scss
droshev: merged it with the block

packages/styles/stories/Components/skeleton/circle.example.html
droshev: added a class for it

packages/styles/src/skeleton.scss
```diff
@@ -0,0 +1,46 @@
...
+  cursor: wait;
+
+  &__canvas {
+    display: block;
+    fill: url(#skeletonGradient);
+
+    --fdSkeletonGradientMiddleColor: #999;
+  }
+
+  &__begin-end {
```
InnaAtanasova, line 20: maybe rename to `&__outer` ?

packages/styles/stories/Components/skeleton/circle.example.html
```diff
@@ -0,0 +1,17 @@
+<div class="fd-skeleton" style="width: 50px; height: 50px;">
+    <svg width="100%" height="100%" class="fd-skeleton__canvas">
+        <defs>
+            <linearGradient id="skeletonGradient">
+                <stop offset="0" stop-color="var(--sapContent_Placeholderloading_Background)"></stop>
```
InnaAtanasova, line 5: shouldn't this be a class now?

packages/styles/stories/Components/skeleton/circle.example.html
```diff
@@ -0,0 +1,17 @@
...
+                <stop class="fd-skeleton__middle">
+                    <animate attributeName="offset" values="0; 0; 0.5; 1; 1" dur="2s" repeatCount="indefinite"></animate>
+                </stop>
+                <stop offset="1" stop-color="var(--sapContent_Placeholderloading_Background)"></stop>
```
InnaAtanasova, line 9: same. Shouldn't this be a class now?

InnaAtanasova (changes requested):
Please update all examples to use classes not inline var for the stop color

InnaAtanasova (approved):

Issue #3179, opened 2022-02-13
explore the option of showing/support font awesome


droshev:
create storybook examples with the usage of font awesome and document what the applications would need to do to use it(as steps).

0. research using font icons
1. https://cdnjs.com/libraries/font-awesome
2. https://fonts.google.com/icons?icon.set=Material+Icons
3. https://flowbite.com/docs/customize/icons/
4. https://heroicons.com/

There was some work done with the [PR](https://github.com/SAP/fundamental-styles/pull/4724)

droshev:
@kalo2711 any progress?

kalo2711:
StoryBook Icons examples added using font-awesome, google fonts, Flowbite and Heroicons SVGs. 
Docs/font awesome #5209

kalo2711:
Issue #3179 closed

kalo2711:
Can I close this issue?

droshev:
> @droshev Can I close this issue?

we haven't done it, have we?

PR #4724, opened 2023-07-26
Docs(styles): font awesome example [ci visual]

IB002:
## Related Issue
Relates to https://github.com/SAP/fundamental-styles/issues/3179

## Description

Added examples for Font Awesome

## Screenshots
N/A


N1XUS (commented):
@IB002 Nice job. One thing to consider: we often use `class*=sap-icon` selector to style inner icons. As you may see, it aims only at sap-icon font family and will not have any effect on any other third-party icons. @InnaAtanasova any thoughts?

g-cheishvili (approved):

PR #4784, opened 2023-08-20
chore(deps): update nx

droshev:
update nx


g-cheishvili (approved):

g-cheishvili:
Something is not working in sidebar. The list of the stories is not right, they should be grouped by the package

PR #4826, opened 2023-09-11
chore(ci): fix nx caches

g-cheishvili:
## Description
Fixed nx caches and reverted commit which increased version before

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fundamental-styles",
-  "version": "0.30.2-rc.28",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fundamental-styles",
-  "version": "0.30.2-rc.28",
+  "version": "0.30.2-rc.27",
```
N1XUS, line 3: Is it correct that the version is downgraded?

N1XUS (approved):

PR #4925, opened 2023-10-25
chore: moved btp button examples into the BTP subfolder

g-cheishvili:
## Related Issue

## Description
Moved BTP button examples to the separate sub-section for the BTP components and added restrictions on tool-header badge contents


packages/styles/src/button.scss
```diff
@@ -5,6 +5,55 @@
 
 $fd-button-icon-font-size: 1rem;
 $fd-button-badge-spacing: 0.25rem;
+$btp-button-block: #{$block-button}--btp;
+
+@mixin base-badge() {
```
N1XUS, line 10: image
I see that there's a lot of collision between `.fd-button .fd-button__badge` and `.fd-button.fd-button--btp .fd-button__badge`. Maybe it's a good idea to introduce css var overrides rather than directly assigning new properties?

N1XUS (approved):

InnaAtanasova (changes requested):
This type of badge (red solid circle) can't be used with other types of buttons. It can be used only with Tool Header button

InnaAtanasova (approved):

Issue #4912, opened 2023-10-19
Design Review Findings


droshev:
Components with issues:
- [x] Calendar - Mike https://github.com/SAP/fundamental-styles/pull/4930
- [ ] Card - Mike https://github.com/SAP/fundamental-styles/pull/4933
- [ ] Carousel - Mike https://github.com/SAP/fundamental-styles/pull/4935
- [ ] Dialog - Mike https://github.com/SAP/fundamental-styles/pull/4937
- [x] Generic Tag - Giorgi https://github.com/SAP/fundamental-styles/pull/4968
- [ ] Link - N1XUS 
- [ ] Message Popover - N1XUS
- [ ] Message Strip - Mike https://github.com/SAP/fundamental-styles/pull/4948
- [ ] Object List - Giorgi
- [x] Object Marker - text color checked and confirmed  ![image](https://github.com/SAP/fundamental-styles/assets/6586561/4188a93a-ef5a-4c3e-a1b1-b06451d76cc3)

- [x] Object Status- Giorgi https://github.com/SAP/fundamental-styles/pull/4968
- [ ] Panel - Mike https://github.com/SAP/fundamental-styles/pull/4949
- [ ] Popover - Mike https://github.com/SAP/fundamental-styles/pull/4950
- [ ] Progress Indicator- Giorgi
- [x] Select - N1XUS https://github.com/SAP/fundamental-styles/pull/4932
- [ ] Splitter - Mike

g-cheishvili:
Progress indicator popover note:
> This popover comes from fiori specifications, the Horizon does not have any mentioning of the truncation. On UI5, they just truncate, without any popover. In general they should not share the same HTML structure, but we're forced to make them the same

Object list note:
> The spacing between the attributes is 0.5 rem already. When they say "inner padding" they mean the spacing between the attribute and the status.
Visually they might not appear distanced from each other with 0.5rem, because there is also a `line-height` specification, which states that it should have 1.5 line height and naturally the height of the text is increased

PR #4933, opened 2023-10-27
fix(styles): card updates

mikerodonnell89:
part of #4912


packages/styles/src/card.scss
```diff
@@ -193,8 +193,6 @@ $fd-card-header-outline-offset: 0.0625rem !default;
     @include fd-flex() {
       align-items: flex-end;
     }
-
-    margin-top: 0.25rem;
```
InnaAtanasova, line 197: Screenshot 2023-10-31 at 12 04 51 PM

I don't think you should remove this. There's 0.25rem spacing between the title and subtitle. I think the issue is coming from the line height of the title text and subtitle text. Our line-height is 1.5 and according to the visual specs it should be 1rem (16px). 

Design specs:
Screenshot 2023-10-31 at 12 03 48 PM

packages/styles/src/card.scss
```diff
@@ -323,6 +322,10 @@ $fd-card-header-outline-offset: 0.0625rem !default;
       overflow: hidden;
     }
 
+    .#{$list} {
+      @include fd-set-paddings-x(0.125rem);
+    }
```
InnaAtanasova, line 327: I don't think this is correct! The card content should have additional 1rem left and right paddings:
Screenshot 2023-10-31 at 12 21 28 PM

InnaAtanasova (changes requested):

packages/styles/src/card.scss
```diff
@@ -212,6 +210,7 @@ $fd-card-header-outline-offset: 0.0625rem !default;
     text-align: right;
     margin: var(--fdCard_Counter_Margin);
     white-space: nowrap;
+    color: var(--sapTile_TitleTextColor);
```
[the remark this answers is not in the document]
mikerodonnell89: This variable is coming directly from the figma notes

packages/styles/src/card.scss
mikerodonnell89: The card content does have 1rem of padding, and I think the figma note is saying there should be 4px more of padding for the list, so I think I need to bump this up to 0.25rem

InnaAtanasova: Screenshot 2023-10-31 at 3 04 10 PM

The card content by default doesn't have paddings. It's left to the applications to add, if needed. In this case 1rem left and right paddings should be applied.

packages/styles/src/card.scss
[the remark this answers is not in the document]
InnaAtanasova: I know, but it's not correct

InnaAtanasova:
@mikerodonnell89 could you trigger the visual updates with [ci visual]?

packages/styles/src/card.scss
```diff
@@ -388,6 +390,10 @@ $fd-card-header-outline-offset: 0.0625rem !default;
     @include fd-reset();
 
     padding-bottom: 0.5rem;
+
+    h3 {
+      font-size: var(--sapFontSize);
+    }
```
InnaAtanasova, line 396: I don't think this is correct. Using `h3` brings additional margins top and bottom to the element, which completely deviates from the visual design:
Screenshot 2023-11-07 at 12 11 37 PM

InnaAtanasova: The size of the title is something the user should specify. What comes from us is only the spacing below the container. In my opinion, the comment in the figma should be ignored.

Screenshot 2023-11-07 at 12 15 02 PM

packages/styles/stories/Components/card/object-card.example.html
```diff
@@ -19,38 +19,38 @@
             <div class="fd-card__content" role="group" aria-label="Card Content">
                 <div class="fd-card__content-container">
                     <div class="fd-card__content-section">
-                        <div class="fd-card__content-title-container">Group Title 1</div>
+                        <div class="fd-card__content-title-container"><h3>Group Title 1</h3></div>
```
InnaAtanasova, line 22: this should be reverted

packages/styles/stories/Components/card/object-card.example.html
```diff
@@ -19,38 +19,38 @@
...
                         <div class="fd-card__content-group">
                             <div class="fd-card__content-group-text">
-                                <span>Label</span>
-                                <span>Text</span>
+                                <span>Label:</span>
+                                <span class="fd-link">Text</span>
```
InnaAtanasova, line 26: you can't style a span as link. This should be reverted for the entire file.
What goes in this container is up to the user. We only provide the spacing for the container.

InnaAtanasova (changes requested):

InnaAtanasova (approved):

InnaAtanasova:
@mikerodonnell89 please trigger the visual checks

droshev (changes requested):
some vertical misalignments

Screenshot 2023-11-15 at 5 10 16 PM
Screenshot 2023-11-15 at 5 09 07 PM
Screenshot 2023-11-15 at 5 08 21 PM

PR #4935, opened 2023-10-27
fix(styles): carousel updates

mikerodonnell89:
part of #4912


packages/styles/src/carousel.scss
```diff
@@ -214,6 +213,7 @@ $fd-carousel-button-content-offset: 0.5rem !default;
 
     text-align: center;
     color: var(--sapPageFooter_TextColor);
+    font-size: var(--sapFontSmallSize);
```
InnaAtanasova, line 216: Screenshot 2023-10-31 at 3 28 29 PM
Screenshot 2023-10-31 at 3 28 40 PM

According to the specs we should have `--sapFontSize`

InnaAtanasova: I see in one design Figma that it's 12px so I will reach out the designers

InnaAtanasova (changes requested):
blocking until we get a response from the designers

packages/styles/src/carousel.scss
InnaAtanasova: it should --sapFontSize (like it was)

InnaAtanasova:
@mikerodonnell89 can you resolve the conflict and trigger the visual checks with [ci visual]?

packages/styles/src/carousel.scss
```diff
@@ -152,8 +152,7 @@ $fd-carousel-button-content-offset: 0.5rem !default;
     @include fd-reset();
     @include fd-flex-center();
 
-    min-height: 2.75rem;
-    max-height: 3.5rem;
```
InnaAtanasova, line 156: I think this is not right. Now all page indicators, with and without buttons, have the same height of 2.75rem. 
Screenshot 2023-11-07 at 12 02 34 PM

**Per specs, the height of the page indicator with buttons should be 3.5rem**
Screenshot 2023-11-07 at 12 02 51 PM

InnaAtanasova (changes requested):

InnaAtanasova (approved):

InnaAtanasova:
@mikerodonnell89 you need to trigger the visual checks

PR #4937, opened 2023-10-27
docs(styles): dialog updates

mikerodonnell89:
part of #4912

### DIFF
--- a/packages/styles/stories/Components/dialog/default-dialog.example.html
+++ b/packages/styles/stories/Components/dialog/default-dialog.example.html
@@ -1,12 +1,9 @@
 
 <section class="fd-dialog-docs-static fd-dialog fd-dialog--active">
-    <div class="fd-dialog__content" role="dialog" aria-modal="true" aria-labelledby="dialog-title-1">
+    <div class="fd-dialog__content fd-dialog__content--s" role="dialog" aria-modal="true" aria-labelledby="dialog-title-1">
         <span class="fd-dialog__resize-handle"></span>
-        <header class="fd-dialog__header fd-bar fd-bar--header-with-subheader">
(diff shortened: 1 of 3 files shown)


InnaAtanasova (approved):

packages/styles/stories/Components/dialog/default-dialog.example.html
```diff
@@ -1,37 +1,28 @@
 
 <section class="fd-dialog-docs-static fd-dialog fd-dialog--active">
-    <div class="fd-dialog__content" role="dialog" aria-modal="true" aria-labelledby="dialog-title-1">
+    <div class="fd-dialog__content fd-dialog__content--s" role="dialog" aria-modal="true" aria-labelledby="dialog-title-1">
```
InnaAtanasova, line 3: adding `fd-dialog__content--s` is not the default case. In my opinion it should be as before. The content of the body is specified by the case. 

This is what's in the specs. Only the vertical paddings are set by default. The horizontal can be responsive, or set for all screens with modifier classes.
Screenshot 2023-10-31 at 3 18 48 PM

packages/styles/stories/Components/dialog/default-dialog.example.html
```diff
@@ -1,37 +1,28 @@
...
                 </div>
             </div>
         </header>
-        <div class="fd-dialog__subheader fd-bar fd-bar--subheader">
-            <div class="fd-bar__left">
-                <div class="fd-bar__element">
-                    Dialog subheader
-                </div>
-            </div>
-        </div>
```
InnaAtanasova, line 23: I don't see why the subheader is removed. It's an optional case and users can opt for it. It's good to have it as an example so people know they can have a subheader.

PR #4949, opened 2023-10-30
docs(styles): make panel button work, rename function [ci visual]

mikerodonnell89:
part of #4912


InnaAtanasova (approved):
it should be docs, not fix as it only concerns the documentation

PR #4950, opened 2023-10-30
fix(styles): popover updates, and do not use menu in popover examples

mikerodonnell89:
part of #4912 

In the fundamental-ngx repo, we decided against using menus inside the popover examples, since menus have become their own thing with some quirks that make them not work well when inside popover bodies. So, like with the fundamental-ngx docs, I replaced the menus with avatars

### DIFF
--- a/packages/styles/src/popover.scss
+++ b/packages/styles/src/popover.scss
@@ -3,13 +3,8 @@
 
 $block: #{$fd-namespace}-popover;
 $blockBody: #{$fd-namespace}-popover__body;
-$bar: #{$fd-namespace}-bar;
 
 .#{$block} {
-  .#{$bar} {
-    @include fd-set-paddings-x-equal(0.5rem);
-  }
-
   @include fd-reset();
 
   position: relative;
--- a/packages/styles/src/popover.scss
+++ b/packages/styles/src/popover.scss
@@ -3,8 +3,13 @@
 
 $block: #{$fd-namespace}-popover;
 $blockBody: #{$fd-namespace}-popover__body;
+$bar: #{$fd-namespace}-bar;
 
 .#{$block} {
+  .#{$bar} {
+    @include fd-set-paddings-x-equal(0.5rem);
+  }
+
   @include fd-reset();
 
   position: relative;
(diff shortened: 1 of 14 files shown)


packages/styles/src/popover.scss
```diff
@@ -3,8 +3,13 @@
 
 $block: #{$fd-namespace}-popover;
 $blockBody: #{$fd-namespace}-popover__body;
+$bar: #{$fd-namespace}-bar;
 
 .#{$block} {
+  .#{$bar} {
+    @include fd-set-paddings-x-equal(0.5rem);
+  }
```
InnaAtanasova, line 11: The Figma comment is not correct. You can see that it's written that the padding left should be 1rem. 
The current implementation has padding 1 rem
Screenshot 2023-10-31 at 2 25 43 PM


Fundamental-styles:
Screenshot 2023-10-31 at 2 25 38 PM

No changes are needed. Not to mention the paddings of the Bar are responsive and they adjust based on the screen size.

InnaAtanasova (changes requested):

packages/styles/src/popover.scss
mikerodonnell89: Ok reverting this bit with the bar, but i think we should keep the other changes

InnaAtanasova (approved):

packages/styles/src/theming/sap_horizon.scss
```diff
@@ -95,7 +95,7 @@
   /* Popover */
   --fdPopover_Border_Radius: var(--sapPopover_BorderCornerRadius);
   --fdPopover_Border_Width: 0.0625rem;
-  --fdPopover_Border_Color: #acb8c1;
+  --fdPopover_Border_Color: var(--sapNeutralBorderColor);
```
droshev, line 98: this should be done also for horizon_dark; horizon_hcb and horizon_hcw.
for all SAP components this variable looks better, but looks worse for Vertical Nav and User Menu.
@mikerodonnell89  can you check with Paige what variable to use for those 2 BTP components?

InnaAtanasova:
will be fixed as part of the v1 ticket for Popover

PR #4951, opened 2023-10-31
fix(styles): message popover adjustments

N1XUS:
relates #4912

- Adjusted message popover parameters for arrow icon font size;
- List item paddings;
- List item title font metrics

### DIFF
--- a/packages/styles/src/mixins/list/_list-message-view.scss
+++ b/packages/styles/src/mixins/list/_list-message-view.scss
@@ -2,6 +2,7 @@
 
 .#{$block} {
   &--message-view {
+    --faListNavigationIndicatorFontSize: 0.75rem;
     --fdList_Item_Height: calc(2.75rem + var(--sapList_BorderWidth));
     --fdList_Message_View_Item_Byline_Height: calc(3.325rem + var(--sapList_BorderWidth));
     --fdList_Message_View_Title_Size: var(--sapFontHeader5Size);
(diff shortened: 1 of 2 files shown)


packages/styles/src/mixins/list/_list-message-view.scss
```diff
@@ -2,6 +2,7 @@
 
 .#{$block} {
   &--message-view {
+    --faListNavigationIndicatorFontSize: 0.75rem;
```
mikerodonnell89, line 5: `--fa`? Also there's a variable, not sure how important it is to use but `--sapFontSmallSize`

mikerodonnell89 (changes requested):
see my comment above

packages/styles/src/mixins/list/_list-message-view.scss
```diff
@@ -40,7 +41,14 @@
       @include fd-ellipsis();
     }
 
+    .fd-list__item {
+      height: auto;
+    }
```
InnaAtanasova, line 46: `.fd-list` should be `.#{$block}`

packages/styles/src/mixins/list/_list-message-view.scss
InnaAtanasova, line 5: typo, should be `--fd` not `--fa`

packages/styles/src/mixins/list/_list-message-view.scss
```diff
@@ -40,7 +41,14 @@
...
+
     .#{$block}__link {
+      padding-top: 0.5rem;
+      padding-bottom: 0.5rem;
```
InnaAtanasova, line 50: you can use the `fd-set-paddings-y-equal` mixin

InnaAtanasova (changes requested):

PR #4997, opened 2023-11-20
feat(styles): btp horizontal nav update

N1XUS:
## Related Issue
Relates https://github.com/SAP/fundamental-ngx/issues/10670

## Description
- Added horizontal navigation example with Two-Click Area Variant

### After:
image

- [x] All values are in `rem`
- [x] RTL support
3. Testing
- [x] Updated tests
4. Documentation


packages/styles/src/navigation.scss
```diff
@@ -101,9 +101,29 @@ $navListContainerSubmenu: #{$block}__list-container--submenu;
...
+  }
+
+  @mixin _fd-horizontal-nav() {
+    &--horizontal {
+      @content;
+    }
+  }
+
+  @mixin _fd-vertical-nav-with-expander() {
+    &.#{$block}--vertical {
```
InnaAtanasova, line 117: there's already a SCSS variable for this that you can use:
$navHorizontal: #{$block}--horizontal;

packages/styles/src/navigation.scss
```diff
@@ -101,9 +101,29 @@ $navListContainerSubmenu: #{$block}__list-container--submenu;
...
+    &--vertical {
+      @content;
+    }
+  }
+
+  @mixin _fd-horizontal-nav() {
+    &--horizontal {
+      @content;
+    }
+  }
```
InnaAtanasova, line 114: I am against these mixins. The code refactoring later is not very fun. Please use SCSS variables

packages/styles/src/navigation.scss
```diff
@@ -378,7 +400,7 @@ $navListContainerSubmenu: #{$block}__list-container--submenu;
       }
     }
 
-    .#{$navItemWithExpander} {
+    @include _fd-vertical-nav-with-expander() {
```
InnaAtanasova, line 403: how is this a better approach?

packages/styles/src/navigation.scss
```diff
@@ -274,27 +294,29 @@ $navListContainerSubmenu: #{$block}__list-container--submenu;
...
     .#{$navItem} {
-      @include fd-expanded() {
-        .#{$navLink} {
-          --fdNavigation_Link_Child_Indicator_Opacity: 1;
-          --fdNavigation_Link_Background: var(--sapMenu_Active_Background);
-          --fdNavigation_Link_Border_Color: var(--fdNavigation_Link_Border_Color_Active);
-          --fdNavigation_Link_Border_Bottom_Color: var(--fdNavigation_Link_Border_Color_Active);
-
-          @include fd-focus() {
+      &:not(.#{$block}__item--with-expander) {
```
InnaAtanasova, line 297: there's a scss variable:
$navItemWithExpander: #{$block}__item--with-expander;

packages/styles/src/navigation.scss
```diff
@@ -1023,4 +1030,56 @@ $navListContainerSubmenu: #{$block}__list-container--submenu;
       --fdNavigation_Link_Back_Indicator_Icon: "\e1ed";
     }
   }
+
+  @include _fd-vertical-nav() {
+    .#{$block} {
+      &__item {
+        &--with-expander {
```
InnaAtanasova, line 1037: why is this nesting needed? Isn't this just  #{$block}__item--with-expander or $navItemWithExpander?

packages/styles/src/navigation.scss
```diff
@@ -1023,4 +1030,56 @@ $navListContainerSubmenu: #{$block}__list-container--submenu;
...
+          }
+        }
+      }
+    }
+  }
+
+  @include _fd-horizontal-nav() {
+    .#{$block} {
+      &__item {
+        &--with-expander {
```
InnaAtanasova, line 1058: same comment as above

packages/styles/src/navigation.scss
```diff
@@ -1023,4 +1030,56 @@ $navListContainerSubmenu: #{$block}__list-container--submenu;
...
+
+            &:active {
+              &:focus {
+                --fdButton_Nested_Border_Radius: var(--fdNavigation_Link_Border_Radius);
+              }
+            }
+          }
+
+          @include fd-expanded() {
+            .fd-navigation__expander {
```
InnaAtanasova, line 1077: please use `$navExpander: #{$block}__expander;`

packages/styles/src/navigation.scss
```diff
@@ -1023,4 +1030,56 @@ $navListContainerSubmenu: #{$block}__list-container--submenu;
...
+        }
+      }
+    }
+  }
+
+  @include _fd-horizontal-nav() {
+    .#{$block} {
+      &__item {
+        &--with-expander {
+          .fd-navigation__expander {
```
InnaAtanasova, line 1059: please use `$navExpander: #{$block}__expander;`

packages/styles/src/navigation.scss
```diff
@@ -1023,4 +1030,56 @@ $navListContainerSubmenu: #{$block}__list-container--submenu;
...
+            margin: 0 0.0625rem;
+            transition: var(--fdNavigation_Link_Transition);
+
+            --fdButton_Nested_Border_Radius: var(--fdNavigation_Link_Border_Radius);
+
+            @include fd-selected() {
+              background-color: var(--fdNavigation_Item_Selected_Background);
```
InnaAtanasova, line 1066: since the background is changing depending on the state maybe make the background a css variable

packages/styles/src/navigation.scss
```diff
@@ -1023,4 +1030,56 @@ $navListContainerSubmenu: #{$block}__list-container--submenu;
...
+            &:active {
+              &:focus {
+                --fdButton_Nested_Border_Radius: var(--fdNavigation_Link_Border_Radius);
+              }
+            }
+          }
+
+          @include fd-expanded() {
+            .fd-navigation__expander {
+              background: var(--fdNavigation_Item_Selected_Background);
```
InnaAtanasova, line 1078: css variable?

InnaAtanasova (changes requested):

InnaAtanasova:
Screenshot 2023-11-21 at 10 51 51 AM

There's something wrong with the naming.
Per specs, Two-click area variant is the case where you have a link + expander:
Screenshot 2023-11-21 at 10 53 13 AM


The default one is where the whole item serves as an expander.

InnaAtanasova:
The underline for selected state is not at the correct position. There's 1-2px space under it. It's more noticeable in focus state:
Screenshot 2023-11-21 at 10 56 25 AM

Screenshot 2023-11-21 at 10 55 49 AM

packages/styles/src/navigation.scss
N1XUS: @InnaAtanasova previous line would apply styling for both horizontal and vertical navs. This mixin allows to apply styling only for a vertical nav. Same with `_fd-horizontal-nav-with-expander` mixin.

InnaAtanasova:
Screenshot 2023-11-22 at 2 52 52 PM
This is still not fixed. There's 1 px spacing under the selected indicator

InnaAtanasova:
Please fix the underline spacing. The code looks great so I will approve the PR to not block you tomorrow.

PR #5085, opened 2024-01-01
fix: splitter resizer overflow and positioning

g-cheishvili:
## Description
This PR does few things:
- Previously the Splitter resizer would cut the handle, because of the overflow hidden
- The splitter resizer did not have a relative positioning and as a aresult the handle would be absolute to the nearest relative or to the body and it would have a wrong placement on places outside our docs page
- Updated the nx to the latest(minor update, nothing had changed)
- Moves splitter into the BTP section in the docs, because it is a BTP component


N1XUS:
In last example, when resizer is pressed, it overlays the panel. Maybe add higher z-index?
image

N1XUS (approved):

PR #5087, opened 2024-01-03
[WIP]fix(styles): horizontal nav a11y

N1XUS:
## Related Issue
Closes none

## Description
- Updated accessibility attributes for horizontal navigation;
- Updated two-click area example with correct dropdown arrow styling

### DIFF
--- a/packages/styles/stories/BTP/Navigation/horizontal/navigation-horizontal-expander.example.html
+++ b/packages/styles/stories/BTP/Navigation/horizontal/navigation-horizontal-expander.example.html
@@ -1,19 +1,18 @@
 <div style="height: 500px">
     <nav class="fd-navigation fd-navigation--horizontal" aria-label="Horizontal Navigation">
         <div class="fd-navigation__container">
-            <ul class="fd-navigation__list" role="menubar" aria-label="Horizontal Navigation">
+            <ul class="fd-navigation__list" role="tablist" aria-orientation="horizontal" aria-label="Horizontal Navigation">
                 <li
                     role="none"
                     tabindex="-1"
                     class="fd-navigation__list-item fd-navigation__list-item--home"
-                    aria-hidden="true"
                 >
                     <div
                         class="fd-navigation__item"
                         aria-current="page"
                         aria-selected="false"
                         aria-expanded="false"
-                        role="menuitem"
+                        role="tab"
                     >
                         <a class="fd-navigation__link" role="link" href="#">
                             <span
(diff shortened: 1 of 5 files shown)


packages/styles/stories/BTP/Navigation/horizontal/navigation-horizontal-expander.example.html
```diff
@@ -1,19 +1,18 @@
 <div style="height: 500px">
     <nav class="fd-navigation fd-navigation--horizontal" aria-label="Horizontal Navigation">
         <div class="fd-navigation__container">
-            <ul class="fd-navigation__list" role="menubar" aria-label="Horizontal Navigation">
+            <ul class="fd-navigation__list" role="tablist" aria-orientation="horizontal" aria-label="Horizontal Navigation">
```
InnaAtanasova, line 4: the default aria-orientation for role="tablist" is horizontal, you don't need to specify it (just FYI)

packages/styles/stories/BTP/Navigation/horizontal/navigation-horizontal-expander.example.html
```diff
@@ -1,19 +1,18 @@
...
                     class="fd-navigation__list-item fd-navigation__list-item--home"
-                    aria-hidden="true"
                 >
                     <div
                         class="fd-navigation__item"
                         aria-current="page"
                         aria-selected="false"
                         aria-expanded="false"
-                        role="menuitem"
+                        role="tab"
```
InnaAtanasova, line 15: Role `tab` needs the following:
- aria-labelledby: this should be leading to the span that wraps the text
- aria-controls: referring to the tabpanel that the tab controls
- aria-posinset and aria-setsize
- aria-selected
- aria-expanded
- aria-haspopup: if the tab has a popup menu

InnaAtanasova: https://www.w3.org/WAI/ARIA/apg/patterns/tabs/

packages/styles/stories/BTP/Navigation/horizontal/navigation-horizontal-expander.example.html
```diff
@@ -1,19 +1,18 @@
...
                     >
                         <a class="fd-navigation__link" role="link" href="#">
```
InnaAtanasova, line 17: in this case, since we now have Tabs, I am not sure if we need the role link and the href

N1XUS: Well, technically it's still a navigation, and link can be applied to it, no?

packages/styles/stories/BTP/Navigation/horizontal/navigation-horizontal-expander.example.html
N1XUS: UI5 does

droshev:
waiting for the signed off a11y design spec

droshev:
can be reopened once the a11y design spec is finalized

PR #5130, opened 2024-01-23
chore(deps): remove unsed @babel/register

droshev:
remove unsed @babel/register


package.json
```diff
@@ -47,12 +47,9 @@
   "devDependencies": {
     "@actions/core": "^1.9.0",
     "@babel/core": "^7.19.1",
-    "@babel/generator": "^7.19.5",
-    "@babel/parser": "^7.19.1",
```
g-cheishvili, line 51: These two are used in generators

droshev: updated the Pr

PR #5143, opened 2024-01-29
fix(styles): Avatar component v1.0.0 updates

InnaAtanasova:
## Related Issue
v1.0.0 

## Description
- added new modifier classes for the Avatar sizes to unify the naming with the other components. The old ones are kept to avoid breaking changes in existing apps.
`fd-avatar--sm`, `fd-avatar--md` and `fd-avatar--lg`
- the examples were updated to use these classes so future users use them instead of the old ones
- updated the examples to use the proper icons representing user and product
- Value states badges were updated to use the correct state icons
- For interactive Avatars the active/toggled states were updated with the latest design
- improvements in the documentation were added


N1XUS (approved):
LGTM!

PR #5180, opened 2024-02-13
fix(styles): add display block to fd-title [ci visual]

mikerodonnell89:
from @droshev PR comment here: https://github.com/SAP/fundamental-ngx/pull/11329#discussion_r1472223505

BREAKING CHANGE:
facet title now uses `div` elements and the developer is expected to provide `role="heading"` and the appropriate `aria-level`


droshev (changes requested):
we need chromatic comparison too

packages/styles/stories/Dev/facets/combined.example.html
```diff
@@ -5,7 +5,7 @@
             </span>
         </div>
         <div class="fd-facet fd-facet--form fd-margin-end--md fd-margin-bottom--sm" role="group" aria-labelledby="formFacetTitle1">
-            <h1 class="fd-title fd-title--h5 fd-margin-bottom--sm" id="formFacetTitle1">Technical Data</h1>
+            <span role="heading" aria-level="3" class="fd-title fd-margin-bottom--sm" id="formFacetTitle1">Technical Data</span>
```
InnaAtanasova, line 8: you can use `div` and then you won't need `display: block;` in the CSS. Confirmed with Nikolay K.

InnaAtanasova (approved):

PR #5478, opened 2024-07-05
feat(common-css): add data attribute to common css Text [ci visual]

InnaAtanasova:
## Related Issue
Closes none, improvements suggested by @trilodge 

## Description
add data attr as alternative to modifier classes to handle hyphens, wrap, line-clamp in Text (common CSS)


trilodge (approved):
Using the attribute keeps the paragraph rendering to normal display: block instead of the modified webkit syntax. That's HTML like it should be used :)

dpavlenishvili (approved):

PR #5595, opened 2024-08-26
feat(styles): added background styling for closeOnOutsideClick [ci visual]

alexhristov14:
added background-color for popover when closeOnOutsideClick is false

### DIFF
--- a/packages/styles/src/popover.scss
+++ b/packages/styles/src/popover.scss
@@ -6,12 +6,29 @@
 $block: #{$fd-namespace}-popover;
 $blockBody: #{$fd-namespace}-popover__body;
 
+body.fd-overlay-active::before {
+  content: '';
+  position: fixed;
+  top: 0;
+  left: 0;
+  width: 100vw;
+  height: 100vh;
+  background: var(--fdModal_BackgroundColor);
+  z-index: 999;
+  pointer-events: all;
+}
+
 .#{$block} {
   @include fd-reset();
 
   position: relative;
   display: inline-block;
   max-width: 100%;
+
+  &__modal {
+    position: relative;
+    z-index: 1000;
+  }
 
   &__control {
     @include fd-reset();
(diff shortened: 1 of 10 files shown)


packages/styles/src/popover.scss
```diff
@@ -6,13 +6,31 @@
 $block: #{$fd-namespace}-popover;
 $blockBody: #{$fd-namespace}-popover__body;
 
+body.overlay-active::before {
```
InnaAtanasova, line 9: `overlay-active` is not a proper name. If the application that uses Fundamental-styles also uses another library and they have a class called `overlay-active` there will be unexpected behaviour. Your class should be prefixed with `fd-` to avoid this

packages/styles/src/popover.scss
```diff
@@ -6,13 +6,31 @@
...
+  content: '';
+  position: fixed;
+  top: 0;
+  left: 0;
+  width: 100vw;
+  height: 100vh;
+  background-color: var(--fdModal_BackgroundColor);
```
InnaAtanasova, line 16: we usually do `background: var(--fdModal_BackgroundColor);` in such cases. It does absolutely the same with a slight difference. 
`background` is a shortcut for:
 - background-color
 - background-image
 - background-repeat
 - background-attachment
 - background-position

So it will set the background-color and do exactly the same as `background-color: var(--fdModal_BackgroundColor);`. 

`background` overwrites all previous style. If you have:
```
background-image:url("..."); 
background-size:100% auto;
```

`background: red;` <- will replace all properties
`background-color: red;` <-will only replace bgcolor property

packages/styles/src/popover.scss
```diff
@@ -6,13 +6,31 @@
...
   @include fd-reset();
 
   position: relative;
   display: inline-block;
   max-width: 100%;
 
+  &__modal {
+    position: relative;
+    z-index: 1000;
+    pointer-events: auto;
```
InnaAtanasova, line 31: Why do you set explicitly the pointer-event to `auto`? This is the default value for the browsers:
Screenshot 2024-08-27 at 1 46 05 PM

info: https://www.w3schools.com/cssref/css3_pr_pointer-events.php

InnaAtanasova (changes requested):
1. Please add an example to the documentation
2. Your commit message should contain the string `[ci visual]` in order to start the visual (Chromatic) checks. It should be something like `feat(styles): added background styling for closeOnOutsideClick [ci visual]`

InnaAtanasova:
Your commit message must contain the string `[ci visual]`. Changing the title won't work

InnaAtanasova (approved):

PR #5674, opened 2024-10-02
fix(styles): inline form group margin block start issue [ci visual]

mikerodonnell89:
fixes none

Found this on ngx radio button page. Before:
Screenshot 2024-10-02 at 10 18 26 AM

After:
Screenshot 2024-10-02 at 10 20 19 AM

### DIFF
--- a/packages/styles/src/form-group.scss
+++ b/packages/styles/src/form-group.scss
@@ -13,6 +13,11 @@ $block: #{$fd-namespace}-form-group;
       justify-content: flex-start;
       flex-wrap: wrap;
     }
+
+    .#{$block}-form-item + .#{$block}-form-item {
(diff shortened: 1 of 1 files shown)


InnaAtanasova (changes requested):

InnaAtanasova (approved):

InnaAtanasova:
Actually the fix should be done in `form-item.scss` file where the spacing is set to 0.625rem. 
The top spacing is added in horizontal and vertical form group. It's necessary when we have inputs in a form. This bug needs further investigation and won't be that simple fix. Closing the issue.

Screenshot 2024-10-07 at 10 10 12 AM


Screenshot 2024-10-07 at 10 08 06 AM

PR #5707, opened 2024-10-18
feat(styles): added an optional Calendar legend section [ci visual]

alexhristov14:
The styling needed for the calendar legend: https://github.com/SAP/fundamental-ngx/issues/12403

Example:
Screenshot 2024-10-18 at 10 08 46 AM

### DIFF
--- a/packages/styles/src/calendar.scss
+++ b/packages/styles/src/calendar.scss
@@ -29,13 +29,15 @@ $fd-calendar-special-days: (
   '20': var(--sapLegendColor20)
 );
 
-@mixin legendTextBase() {
+@mixin fd-legend-text-base() {
   color: var(--sapContent_LabelColor);
   font-family: var(--sapFontFamily);
   font-size: var(--sapFontSmallSize);
   font-weight: normal;
   padding: 0.3125rem;
-  background: var(--sapList_Background);
+  padding-inline: 0.3125rem;
+  padding-block: 0.3125rem;
+  // background: var(--sapList_Background);
   column-gap: 0.5rem;
   cursor: default;
 }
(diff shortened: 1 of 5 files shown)


packages/styles/stories/Components/calendar/calendar.stories.js
```diff
@@ -96,6 +97,18 @@ These modifier classes are used to style various states of calendar grid element
...
+### Legend
+These modifier classes are used to style calendar grid elements with colors.\n
+| Modifier class | Use to... |
+| -------------:| :-------------- |
+| \`fd-calendar__row--legend\` | Create a row space for legend items 
+| \`fd-calendar--legend__item\` | Create a legend container 
+| \` fd-calendar--legend__item--circle\` | Makes the legend item representation a circle
+| \` fd-calendar--legend__item--square\` | Makes the legend item representation a square
+| \` fd-calendar--legend__item--text\` | Adds a descriptive title to the legend item
+| \` fd-calendar--legend__item--text-description\` | Adds any additional information to the legend item
```
InnaAtanasova, line 110: These class names are not following the BEM rules.
`fd-calendar` is the block element, because the entire element can be "extracted" from a project and used in another project, aka it's an independent unit. The elements inside this blocks have the double underscore in their class names. 
For example: `fd-calendar__row`, `fd-calendar__item`, `fd-calendar__date`, etc.
When you look at these classes you can immediately tell that it's a row/item/date, etc. that belong to the calendar component (block). 
The double dash notation is a modifier. If you have an item, or date, that are a variation of the common one, you use a modifier class to add specifics for it. For example, if the weekend dates and the work days have different background, then you do something like this:
`fd-calendar__date fd-calendar__date--weekend` or `fd-calendar__date fd-calendar__date--work`, `fd-calendar__date fd-calendar__date--vacation`.
When you read the class `fd-calendar__date--vacation` it basically says: "it's a vacation variation of date element that belongs to calendar". 
If you have to read this class `fd-calendar--legend__item--circle`, you can't tell what is an element and what is a modifier.

InnaAtanasova: > here's more on BEM: 
https://getbem.com/introduction/
https://en.bem.info/methodology/css/
https://css-tricks.com/bem-101/

packages/styles/stories/Components/calendar/legend.example.html
```diff
@@ -0,0 +1,275 @@
...
+                        <span class="fd-calendar__text" role="button">4</span>
+                    </div>
+                </td>
+                <td role="gridcell" class="fd-calendar__item fd-calendar__item--other">
+                    <div class="fd-calendar__text-wrapper">
+                        <span class="fd-calendar__text" role="button">5</span>
+                    </div>
+                </td>
+            </tr>
+            <tr class="fd-calendar__row--legend">
```
InnaAtanasova, line 249: you can't have a modifier class without the base class that you modify. `fd-calendar__row--legend` says it modifies `fd-calendar__row` base class, but there's no base class.
The correct way would be `fd-calendar__row fd-calendar__row--legend`
`fd-calendar__row` would bring the base CSS code and `fd-calendar__row--legend` would bring the CSS that is specific for the legend case

InnaAtanasova, line 249: Screenshot 2024-10-21 at 2 12 59 PM

Markup wise this shouldn't be a part of the Table row. The legend should be a separate unit, that can be placed on the bottom or on the right/left side of the actual calendar. 
I think the best way to construct is something like this:
```
<div class="fd-calendar__legend" role="list">
    <div class="fd-calendar__legend-item" role="listitem">
        <span class="fd-calendar__legend-marker"></span>
        <span class="fd-calendar__legend-label">Square marker</span>
    </div>
    <div class="fd-calendar__legend-item" role="listitem">
        <span class="fd-calendar__legend-marker fd-calendar__legend-marker--circle"></span>
        <span class="fd-calendar__legend-label">Circle marker</span>
    </div>
    <div class="fd-calendar__legend-item" role="listitem">
        <span class="fd-calendar__legend-marker fd-calendar__legend-marker--background"></span>
        <span class="fd-calendar__legend-label">With background marker</span>
    </div>
</div>
```

InnaAtanasova: `<div class="fd-calendar__legend" role="list">` should be a sibling element to the table element with class `fd-calendar__content`. This would make the legend "independent" and can be on the bottom or on the side of the table.

InnaAtanasova (changes requested):
https://github.com/user-attachments/assets/aee2c659-c32b-4361-b9c6-a90284accaaa


On focus elements "jump". Element's states (hover, focus, active) should not affect the element's dimensions. In this case, the focus outline should be shown without a change in the element height, which causes the "jump"

packages/styles/src/calendar.scss
```diff
@@ -411,8 +422,60 @@ $fd-calendar-special-days: (
     @include fd-reset();
 
     display: flex;
+
+    &--legend {
+      display: grid;
+      grid-template-columns: repeat(auto-fit, minmax(7.5rem, 1fr));
+      grid-gap: 0.5rem;
+      margin-top: 0.5rem;
```
InnaAtanasova, line 430: if this is not part of the table, like I suggested in another comment, you can achieve this with flex. The reason to use flex here is because with modifier classes you can easily achieve different variations of the legend. It can be the way you did it, but it can also be like this:
Screenshot 2024-10-21 at 2 50 18 PM

packages/styles/src/calendar.scss
```diff
@@ -411,8 +422,60 @@ $fd-calendar-special-days: (
...
+    }
   }
 
+  &--legend__item {
+    @include fd-flex();
+
+    align-items: center;
```
InnaAtanasova, line 437: `@include fd-inline-flex-center()`

packages/styles/src/calendar.scss
```diff
@@ -411,8 +422,60 @@ $fd-calendar-special-days: (
...
+        background-color: none;
+    }
+
+    &--square {
+        @include fd-square(0.875rem);
+
+        border: 0.0625rem solid var(--sapContent_ForegroundBorderColor); // 0.125rem if not ForegroundBorderColor
+        border-radius: 0.125rem;
+        background-color: none;
+        margin: 0;
```
InnaAtanasova, line 454: margin-inline: 0;
margin-block: 0;

packages/styles/src/calendar.scss
```diff
@@ -411,8 +422,60 @@ $fd-calendar-special-days: (
...
+    }
+
+    &--text {
+        @include legendTextBase();
+
+        &-description {
+            @include legendTextBase();
+
+            position: absolute;
```
InnaAtanasova, line 463: absolute positioning should be your last option. Try with flex.

packages/styles/src/calendar.scss
```diff
@@ -411,8 +422,60 @@ $fd-calendar-special-days: (
...
+            vertical-align: middle;
+            white-space: pre-line;
+            text-overflow: ellipsis;
```
InnaAtanasova, line 466:
```
@mixin fd-ellipsis {
  white-space: nowrap;
  overflow: hidden;
  text-overflow: ellipsis;
}
```

packages/styles/src/calendar.scss
```diff
@@ -411,8 +422,60 @@ $fd-calendar-special-days: (
...
+            overflow: hidden;
+            padding-top: 0;
```
InnaAtanasova, line 468: padding-block-start: 0;

packages/styles/src/calendar.scss
```diff
@@ -411,8 +422,60 @@ $fd-calendar-special-days: (
...
+        }
+    }
+
+    @include fd-focus(){
+        padding: 0.125rem;
```
InnaAtanasova, line 473: padding-inline: 0.125rem;
padding-block: 0.125rem;

InnaAtanasova (approved):

droshev (approved):

PR #5724, opened 2024-10-25
fix(styles): Checkbox link in  label  is not clickable

khotcholava:
fix(core): checkbox link in  label  is not clickable

closes [#12577](https://github.com/SAP/fundamental-ngx/issues/12577)

## Description
Updated the checkbox component to ensure that links within custom labels are clickable and function as intended.

### DIFF
--- a/packages/styles/src/checkbox.scss
+++ b/packages/styles/src/checkbox.scss
@@ -100,7 +100,7 @@ $blockCheckmark: #{$fd-namespace}-checkbox__checkmark;
     @include fd-ellipsis();
 
     height: 1rem;
-    pointer-events: none;
+    pointer-events: auto;
     max-width: inherit;
   }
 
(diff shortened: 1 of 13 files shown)


packages/styles/src/checkbox.scss
```diff
@@ -100,7 +100,7 @@ $blockCheckmark: #{$fd-namespace}-checkbox__checkmark;
     @include fd-ellipsis();
 
     height: 1rem;
-    pointer-events: none;
+    pointer-events: auto;
```
InnaAtanasova, line 103: you can just delete `pointer-events: none;` and it will default to auto.

InnaAtanasova (changes requested):
You have to change your commit message
```
✖   scope must be one of [styles, docs, e2e, release, deps, deps-dev, changelog, ci, common-css, cx] [scope-enum]
```
your commit message should also contain `[ci visual]` string to trigger Chromatic checks.

PR #6035, opened 2025-04-23
fix(styles): adopt fonts from theming-base

droshev:
adopt fonts from theming-base

BREAKING CHANGE:
remove our local copy of font-faces and adopt this from theming-base-content


droshev:
Issue 1: Business app icons are not loading

InnaAtanasova (approved):

PR #6294, opened 2026-05-01
chore(styles): update doc styling

InnaAtanasova:
## Related Issue
Closes none

## Description
Docs restyling:
- CSS update for the top nav and side nav
- Added section navigation on the right (desktop only, hides for smaller screens)


droshev (approved):

MariaIDineva (commented):
I pretty much like the table of contents but there's something weird going on with it when I scroll. It would not always scroll up to the topmost section, it would sometimes not scroll down to the latest one either.


https://github.com/user-attachments/assets/80b279f1-8ea6-4458-a193-dd6de9e7eaad

MariaIDineva (approved):

PR #6323, opened 2026-06-02
fix(styles): prevent overflow content from pushing siblings outside container

InnaAtanasova:
## Related Issue
Closes none

### Problem
- When `fd-bar__left` or `fd-bar__right` contained long text, the flex layout would push the other section out of view
- Previous implementation used `width: 100%` on all sections, causing each to demand full container width
- Buttons and icons would overlap with padding or disappear entirely

### Solution
- Changed sections to use `flex: 0 1 auto` (don't grow by default, can shrink, start at content size)
- Removed `flex-grow: 1` from all sections to prevent equal distribution
- Added `min-width: 0` to allow proper shrinking and text overflow
- Elements containing headings (`h1-h6`) can shrink with `flex-shrink: 1` and `min-width: 0`
- Buttons have min-width set to the button height

## Screenshots
In dialog component

### Before:
Screenshot 2026-06-02 at 2 04 04 PM

### After:
Screenshot 2026-06-02 at 1 48 06 PM

### DIFF
--- a/docs/component-catalog.json
+++ b/docs/component-catalog.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "version": "0.41.7-rc.8",
+  "version": "0.41.7-rc.15",
   "description": "Machine-readable catalog of Fundamental Library Styles components",
   "generatedBy": "scripts/generate-ai-metadata.js",
   "relatedFiles": {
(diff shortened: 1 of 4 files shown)


mikerodonnell89 (changes requested):
Definitely an improvement, but it doesn't seem to respect the paddings of the adjacent element. Not sure if this is avoidable 
Image

InnaAtanasova:
> Definitely an improvement, but it doesn't seem to respect the paddings of the adjacent element. Not sure if this is avoidable Image

Added a fix for this, which is not ideal, but we don't have many options here...

MariaIDineva (dismissed):

mikerodonnell89 (dismissed):

PR #6343, opened 2026-06-22
feat(styles, docs): update dynamic page styles and examples

### LINKED ISSUE #5633: Object Page Header design updates
Object Page Header design has been updated and the spacing was revisited

MariaIDineva:
## Related Issue
Closes [SAP/fundamental-styles#](https://github.com/SAP/fundamental-styles/issues/5633)

#### `__title-area`
- Added `box-shadow: var(--sapContent_HeaderShadow)` on `&--collapsed` — moved out of the responsive mixin into the element block

#### `__title-container`
- Added `padding-block-start: 0.3125rem`

#### `__title`
- Added `&--small` modifier — uses `--fdDynamicPage_Title_Collapsed_Font_Size` in both expanded and collapsed states, for applications that require a smaller title in all states

#### `__title-content`
- `padding-inline-end`: `0.2rem` → `0` per spec (`padding: 0 0 0 1rem`)

#### `__toolbar`
- Added `padding-inline-end: 0.25rem` to prevent hover state shadow cut-off on toolbar buttons

#### `__collapsible-header`
- Added `box-shadow: $fd-dynamic-page-box-shadow`

#### `__pin-button`
- `--fdButtonBackgroundColor`: `var(--fdDynamicPage_Button_Pin_Background)` → `var(--sapObjectHeader_Background)` per spec

#### `--responsive-paddings` (new modifier)
- Applies inline padding automatically based on viewport width via media queries, following the same spacing values as the explicit size modifiers — use instead of a fixed size modifier when the page should adapt to the viewport

---

#### `large.example.html` (new)
- Added new story for the L size (`fd-dynamic-page--lg`, 1024–1439px) — previously unrepresented

#### `responsive-paddings.example.html` (new)
- Added new story demonstrating `fd-dynamic-page--responsive-paddings`

#### `dynamic-page.stories.js`
- Registered the new `Large` and `ResponsivePaddings` stories with descriptions
- Added `fd-dynamic-page--responsive-paddings` to the modifiers table in the component docs

---

#### `sap_horizon_hcb.scss`, `sap_horizon_hcw.scss`
- `--fdDynamicPage_Hover_Border_Bottom`: changed border color from `--sapObjectHeader_Hover_Background` to `--sapObjectHeader_BorderColor` per HC spec


packages/styles/src/dynamic-page.scss
```diff
@@ -417,11 +423,11 @@ $block: #{$fd-namespace}-dynamic-page;
 
   &--md,
   &--lg {
-    @include fd-responsive-page($fd-dynamic-page-header-padding-top-bottom, 2rem);
+    @include fd-responsive-page($fd-dynamic-page-header-padding-top-bottom, var(--sapContent_Space_M));
   }
 
   &--xl {
-    @include fd-responsive-page($fd-dynamic-page-header-padding-top-bottom, 3rem);
+    @include fd-responsive-page($fd-dynamic-page-header-padding-top-bottom, var(--sapContent_Space_XL));
```
InnaAtanasova: I am not sure these variables will be available to users who did not added Common CSS package to their application.

packages/styles/src/dynamic-page.scss
```diff
@@ -176,6 +178,10 @@ $block: #{$fd-namespace}-dynamic-page;
       vertical-align: bottom;
     }
 
+    &--condensed {
```
InnaAtanasova: Could you please come up with another name for this modifier? My concern with this name is because we use condensed, compact and cozy for the content density mode and could be quite confusing.

packages/styles/src/dynamic-page.scss
MariaIDineva: Fair point, I'll revert those, thanks!

packages/styles/src/dynamic-page.scss
MariaIDineva: OK, you have a point. I suggest we call it`--small` then. Other options are `--reduced` but I don't like it, it does not feel like connected to font-size for me. Also, maybe `--small-font` or `--reduced-font`, but it is a little too verbose in my opinion. If you have any better ideas, please feel free to suggest.

InnaAtanasova: I like `--small`

InnaAtanasova (approved):
🚀

PR #6347, opened 2026-06-26
fix(styles): replace outline with fake focus in Input Group [ci visual]

### LINKED ISSUE #6346: [Accessibility]: fd-input-group with addon button shows incomplete focus indicator in High Contrast themes
### Describe your proposal

When fd-input-group is used with an addon button ([button]="true"), the focus indicator is incomplete in High Contrast themes (sap_horizon_hcb, sap_horizon_hcw). Only three sides of the focus outline are visible — the side adjacent to the addon button is occluded by the button's container, leaving a broken/partial focus ring.

This is a serious accessibility issue: WCAG 2.4.7 (Focus Visible) and SAP Fiori a11y guidelines require a clear, fully visible focus indicator on all interactive elements.

### Which versions of Angular and Fundamental Library for Angular are affected?

Angular 21.1.0 Fundamenta-ngx/core 0.62.3

### Affected Component

fd-input-group

### Steps to Reproduce (for bugs)

1. Open an Angular app using @fundamental-ngx/core with the SAP Horizon HCB or HCW theme applied (or enable Windows High Contrast mode).
2. Render an fd-input-group with an addon button:

<fd-input-group
  [button]="true"
  [glyph]="'show'"
  [type]="'password'"
  [placeholder]="'Pre-shared key'"
  [formControl]="control">
</fd-input-group>

3. Tab into the input field with the keyboard.

### StackBlitz example(s)

https://stackblitz.com/edit/zzornrel?file=src%2Fapp%2Finput-group-search-example.component.html

### Priority

None

### Declaration

- [x] I'm not disclosing any internal or sensitive information.

InnaAtanasova:
## Related Issue
Closes https://github.com/SAP/fundamental-styles/issues/6346

## Description
The focus outline in Input Group was hidden under the addon buttons which created a11y concerns. There is no way to add z-index on the outline and the solution was to create a fake focus with `::after`, a practice we used in the past to deal with focus border radius. 
While testing in HC themes noticed a bug in read-only mode and fixed that as well.

### Before:
Screenshot 2026-06-26 at 12 52 54 PM
Screenshot 2026-06-26 at 12 52 59 PM

### After:

Screenshot 2026-06-26 at 12 47 47 PM
Screenshot 2026-06-26 at 12 47 52 PM

### Before:
Screenshot 2026-06-26 at 12 55 07 PM

### After:

Screenshot 2026-06-26 at 12 47 59 PM

### Before:
We were relying on `is-focus` class to apply the correct focus color and type, but for the states, when you click on the input the wrong focus was applied (all fields get blue focus. See the example bellow called "Focus" for the correct ones)
Screenshot 2026-06-26 at 12 56 18 PM

### After:
Screenshot 2026-06-26 at 1 03 34 PM

### DIFF
--- a/packages/styles/src/input-group.scss
+++ b/packages/styles/src/input-group.scss
@@ -29,11 +29,38 @@ $fd-input-border-radius: var(--sapField_BorderCornerRadius);
   --fdInput_Height: 100%;
   --fdInput_Compact_Height: 100%;
 
+  &.is-focus,
   &:has(input:is(:focus)) {
-    background: var(--sapField_Focus_Background);
-    outline: var(--fdInput_Outline_Color) var(--sapContent_FocusStyle) var(--sapContent_FocusWidth);
-    outline-offset: var(--fdInput_Outline_Offset);
+    outline: none;
     box-shadow: none;
+    position: relative;
+    background: var(--sapField_Focus_Background);
+
+    &::after {
+      content: "";
+      position: absolute;
+      pointer-events: none;
+      inset: var(--fdInputGroup_Outline_Offset);
+      border-radius: var(--fdInput_Outline_Border_Radius);
+      border: var(--fdInput_Outline_Color) var(--sapContent_FocusStyle) var(--sapContent_FocusWidth);
+    }
+
+    &.is-success {
+      --fdInput_Outline_Color: var(--fdInput_Success_Outline_Color);
+    }
+
+    &.is-alert,
+    &.is-warning {
+      --fdInput_Outline_Color: var(--fdInput_Warning_Outline_Color);
+    }
+
+    &.is-error {
+      --fdInput_Outline_Color: var(--fdInput_Error_Outline_Color);
+    }
+
+    &.is-information {
+     --fdInput_Outline_Color: var(--fdInput_Information_Outline_Color);
+    }
   }
 
   .#{$block}__input {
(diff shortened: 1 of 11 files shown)


packages/styles/src/input-group.scss
```diff
@@ -29,11 +29,41 @@ $fd-input-border-radius: var(--sapField_BorderCornerRadius);
...
+    &.is-warning {
+      --fdInput_Outline_Color: var(--fdInput_Warning_Outline_Color);
+    }
+
+    &.is-error {
+      --fdInput_Outline_Color: var(--fdInput_Error_Outline_Color);
+    }
+
+    &.is-alert {
+      --fdInput_Outline_Color: var(--fdInput_Alert_Outline_Color);
```
MariaIDineva: Per Claude:

> Issue: is-alert maps to --fdInput_Alert_Outline_Color, but this variable does not exist in any theming file. 

Also, is this class `is-alert` really needed? I don't see it used in any of the examples?

InnaAtanasova: yes, `alert` was renamed to `warning` but we keep these for backwards compatibility. We don't put it in the doc examples as we don't want people to use it. This is the case for Checkbox, Radio, Input Groups and inputs.

MariaIDineva (approved):

PR #6350, opened 2026-07-01
fix(styles): ACC-282.1 (Level AA) Section headers truncated at 320px viewport width

### LINKED ISSUE #6348: [Core][fd-panel] ACC-282.1 (Level AA) Section headers truncated at 320px viewport width
### Describe your proposal

When using `fd-panel` to render section headers at narrow viewports (e.g., 320 CSS pixels wide, which is the WCAG 1.4.10 Reflow target), the heading text inside the panel header is truncated and not fully visible. Long heading content does not wrap, reflow, or resize appropriately, which violates WCAG 2.1 Success Criterion 1.4.10 (Reflow, Level AA).

**Reproduction steps (generic):**
1. Render an `fd-panel` with a heading/title in its header area.
2. Resize the viewport to 320 CSS pixels in width.
3. Observe that the heading text is clipped/truncated rather than wrapping or reflowing.

**Observed behavior:** Section headers inside `fd-panel` are truncated at 320px width.

**Expected behavior:** Heading text inside `fd-panel` should wrap, reflow, or resize so it remains fully visible and readable at viewports as narrow as 320 CSS pixels, per WCAG 1.4.10.

### Proposed Solution

The `fd-panel` component should ensure that heading text within the panel header wraps to multiple lines (or otherwise reflows) when horizontal space is insufficient, rather than truncating with ellipsis or being clipped. The visual presentation should match the accessible content at all supported viewport widths down to 320 CSS pixels.

### Proposed Alternatives

_No response_

### Organization

_No response_

### StackBlitz example(s)

_No response_

### Priority

Medium

### Declaration

- [x] I'm not disclosing any internal or sensitive information.

MariaIDineva:
## Related Issue
Closes [SAP/fundamental-styles#](https://github.com/SAP/fundamental-styles/issues/6348)

## Description
**panel.scss** 
- removed `height: var(--fdPanel_Header_Height)`, keeping only `min-height` so the header can grow taller when the title wraps.
-  Added title wrapping
By default, the panel title truncates with an ellipsis when it exceeds the available width. To allow the title to wrap onto multiple lines instead — for example, to comply with WCAG 1.4.10 (Reflow) at narrow viewports — add the `fd-panel__title--wrap` modifier class to the `fd-panel__title` element.

**fixed.example.html**
- updated the story title to a longer string so the reflow behavior is visible at 320px without needing to modify the HTML manually.

## Screenshots
Screenshot 2026-07-01 at 15 52 36

### DIFF
--- a/packages/styles/src/panel.scss
+++ b/packages/styles/src/panel.scss
@@ -31,7 +31,6 @@ $block: #{$fd-namespace}-panel;
     border-bottom: var(--sapGroup_TitleBorderWidth) solid var(--fdPanel_Header_Border_Color, var(--sapGroup_TitleBorderColor));
     width: var(--fdPanel_Header_Width);
     min-width: var(--fdPanel_Header_Width);
-    height: var(--fdPanel_Header_Height);
     min-height: var(--fdPanel_Header_Height);
     top: var(--fdPanel_Header_Position_Top, unset);
     background-color: var(--sapGroup_TitleBackground);
@@ -55,7 +54,6 @@ $block: #{$fd-namespace}-panel;
 
   &__title {
     @include fd-typography(var(--sapGroup_Title_FontSize));
-    @include fd-ellipsis();
 
     flex: 1;
     width: 100%;
--- a/packages/styles/src/panel.scss
+++ b/packages/styles/src/panel.scss
@@ -54,12 +54,17 @@ $block: #{$fd-namespace}-panel;
 
   &__title {
     @include fd-typography(var(--sapGroup_Title_FontSize));
+    @include fd-ellipsis();
 
     flex: 1;
     width: 100%;
     max-width: 100%;
     font-size: var(--fdPanel_Title_Font_Size);
     padding-inline-start: 0.25rem;
+
+    &--wrap {
+      @include fd-wrap();
+    }
   }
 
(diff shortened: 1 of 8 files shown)


packages/styles/src/panel.scss
```diff
@@ -55,7 +54,6 @@ $block: #{$fd-namespace}-panel;
 
   &__title {
     @include fd-typography(var(--sapGroup_Title_FontSize));
-    @include fd-ellipsis();
```
InnaAtanasova, line 58: Unfortunately, if we do this we can cause problems when applications update their projects. For this reason, we keep the truncation as a default behavior and introduce a modifier class for removing the default behavior. You can introduce a modifier class `&__title--wrap` and there reverse the ellipsis functionality. You can check `quick-view.scss`, `title.scss`, `_list-byline.scss` files how it was done there. There's also a mixin that reverses the ellipsis:
```
@mixin fd-wrap {
  white-space: normal;
  overflow: visible;
  text-overflow: clip;
}
```
You can use it if suitable.

MariaIDineva: OK, I'll add a modifier, thanks!

InnaAtanasova (approved):

Issue #6378, opened 2026-08-25
[Component Request]:  Implement Single Line Tokenizer – Read Only State


MariaIDineva:
### Component name

Tokenizer

## Background

The Tokenizer (Horizon) spec defines two distinct tokenizer variants:

**1. Input Field Tokenizer** — used inside Multi Input / Multi Combobox controls. Has a field-like appearance.
- Read Only state: retains field background (`--sapField_ReadOnly_Background`), currently implemented as `fd-tokenizer--readonly`.

**2. Single Line Tokenizer (standalone)** — described as "a transparent wrapper-container with height resulting from the touch area of the Token." Has three sub-states:
- **Regular** — editable, base `fd-tokenizer` class.
- **Read Only Delta** — transparent container, tokens in `fd-token--readonly` state, N-more visual.
- **Display Only** — transparent container, tokens in `fd-token--display` state with bullet separators, currently implemented as `fd-tokenizer--display`.

## Gap

There is no CSS modifier for the **Single Line Tokenizer – Read Only** state. The existing `fd-tokenizer--readonly` maps to the Input Field Tokenizer readonly variant (field background), which is a different thing. The "Single Line Tokenizer – Read Only" state requires:

- Transparent background (no `--sapField_ReadOnly_Background`)
- No border, no box-shadow
- Tokens aligned flush to the left (no leading inner padding)
- `fd-token--readonly` children (bordered chips, no close button, no bullet separators)
- `fd-tokenizer__indicator` for the N-more count
- Hidden text input

## What Needs to Be Done

1. **Create `fd-tokenizer--standalone`** according to the full Tokenizer spec:
   - Verify spacing in cozy and compact density.
   - Verify alignment of first token (no leading gap — spec note: "The spacing in front of the first token from the input field tokenizer does not apply here").
   - Verify N-more (`fd-tokenizer__indicator`) visual matches the Link (Horizon) spec.
   - Verify appearance in Morning Horizon, Evening Horizon, HCB, HCW themes.
   - Verify RTL layout.

2. **Clarify `fd-tokenizer--readonly` background** — the Figma IxD spec appears to show no background on the standalone readonly tokenizer. Investigate whether `fd-tokenizer--readonly` should remain as-is (Input Field context only) or whether it also needs to cover the standalone case.

3. **Add Storybook stories** for `fd-tokenizer--standalone` in the tokenizer stories, covering:
   - Regular (cozy + compact)
   - With N-more indicator
   - RTL

[...]

PR #6379, opened 2026-08-26
fix(styles): form layout fields spacing

### LINKED ISSUE #4463: Form Layout and fields spacing
The Form design got updated. (VD usual wiki:
- Edit Mode
- Display Mode
- Header
- Grouping
- VD - FIORIDESIGN-11309
- ID -  FIORIDESIGN-11310
Review the current form layout and fields spacing, margin, and padding.
The design has changed and we need to update the implementation.

MariaIDineva:
## Related Issue

Closes [SAP/fundamental-styles#](https://github.com/SAP/fundamental-styles/issues/4463)

### 1. Fix label-to-input gap in standalone horizontal form item (Spec E: 0.25rem)

**File**: [form-item.scss](packages/styles/src/form-item.scss)

`margin-inline-end` on `.fd-form-label` inside `.fd-form-item--horizontal` was `0.5rem`. The spec defines 0.25rem between the label and the input control in horizontal layout.

**Change**: `margin-inline: 0 0.5rem` → `margin-inline: 0 0.25rem`

---

### 2. Fix unused `$marginTop` parameter in `fd-form-vertical` mixin

**File**: [form-layout-grid.scss](packages/styles/src/form-layout-grid.scss) — `fd-form-vertical` mixin

The mixin accepted a `$marginTop` parameter (default `0.625rem`) but the body hardcoded `margin-top: 0.625rem`, ignoring the parameter.

**Change**: `margin-top: 0.625rem` → `margin-top: $marginTop` so callers can override the default spacing.

---

### 3. Add display mode CSS block (new)

**File**: [form-layout-grid.scss](packages/styles/src/form-layout-grid.scss)

Added a new `.fd-form-layout-grid-container--display-mode` block that implements all display mode spacing values from the spec:

| Spec item | Value | Implementation |
|-----------|-------|----------------|
| A: Label padding-bottom | 0.25rem | `padding-block-end: 0.25rem` on `fd-form-label` inside full-width columns |
| B: Text touch area (cozy) | 2.25rem | `padding-block: 0.625rem` on `fd-text` (text line-height ≈ 1rem → total ≈ 2.25rem) |
| B: Text touch area (compact) | 1.625rem | `padding-block: 0.3125rem` on `fd-text` inside `fd-compact-or-condensed()` |
| C: Spacing between pairs (vertical) | 1rem | `margin-top: 1rem` on `fd-form-item`, `margin-top: 0` on first item |

The C gap (1rem) is applied via `margin-top` on `fd-form-item` (not on the label element, to avoid displacing pseudo-element decorators such as `--colon::after`) to:
- Containers explicitly marked `--vertical`, `--md-vertical`, `--lg-vertical`, or `--xl-vertical` at any viewport width.
- All display mode containers at the S breakpoint (≤599px), where the grid collapses to full-width columns regardless of explicit modifier.

---

### 4. Add display mode Storybook stories (new)

**Files**: [form-grid.stories.js](packages/styles/stories/Components/Forms/form-grid/form-grid.stories.js), [form-group.stories.js](packages/styles/stories/Components/Forms/form-group/form-group.stories.js)

Added new stories and the missing `text.scss` import where needed:

- **Display mode – Horizontal layout** (`DisplayModeHorizontal`, form-grid): Uses `fd-col-md--4` / `fd-col-md--8` column split at MD+. Labels carry `fd-form-label--colon`. Empty fields rendered as an en dash in italic.
- **Display mode – Vertical layout** (`DisplayModeVertical`, form-grid): Uses `fd-form-layout-grid-container--vertical` with plain `fd-col` columns. Labels use `fd-form-label--colon` (no `--required` — required indicators are not shown in display mode).
- **Group header (form grid) – display mode** (`GroupHeaderInFormGridDisplayMode`, form-group): Display mode counterpart of the existing "Group header (form grid)" story. Same three-group structure; `fd-input` controls replaced with `fd-text` paragraphs, `fd-form-layout-grid-container--display-mode` added to the container, colons moved from label text to the `fd-form-label--colon` modifier.

---

### 5. Fix compact height on form group header (Spec: 2rem)

**File**: [form-group.scss](packages/styles/src/form-group.scss)

`fd-form-group__header` had `min-height: var(--sapElement_LineHeight)` (2.75rem) with no compact override. The spec defines 2rem in compact density.

**Change**: Added `fd-compact-or-condensed()` override with `min-height: var(--sapElement_Compact_LineHeight)`, consistent with the token used by the list component for compact row height.

---

## What is not in scope

The Tokenizer in display/readonly mode requires separate work. See https://github.com/SAP/fundamental-styles/issues/6378 for a full description of the gap and what needs to be done.

## Screenshots
Screenshot 2026-08-26 at 11 22 49

Screenshot 2026-08-26 at 14 29 54

### DIFF
--- a/packages/styles/src/form-layout-grid.scss
+++ b/packages/styles/src/form-layout-grid.scss
@@ -15,9 +15,11 @@ $gridCol: #{$fd-namespace}-col;
   .#{$gridContainer}.#{$formContainer}.#{$verticalSelector} {
     &.#{$formGroup},
     .#{$formGroup} {
-      .#{$formItem}:not(:first-of-type) {
-        > .#{$gridCol} > .#{$fd-namespace}-form-label {
-          padding-block-start: $marginTop;
+      .#{$formItem} {
+        margin-top: $marginTop;
+
+        &:first-of-type {
+          margin-top: 0;
         }
       }
     }
--- a/packages/styles/src/form-layout-grid.scss
+++ b/packages/styles/src/form-layout-grid.scss
@@ -16,10 +16,10 @@ $gridCol: #{$fd-namespace}-col;
     &.#{$formGroup},
     .#{$formGroup} {
       .#{$formItem} {
-        margin-top: $marginTop;
+        margin-top: var(--fdFormLayoutGridItemMarginTop, #{$marginTop});
 
         &:first-of-type {
-          margin-top: 0;
+          margin-top: var(--fdFormLayoutGridItemFirstOfTypeMarginTop, 0);
         }
       }
     }
@@ -110,6 +110,19 @@ $gridCol: #{$fd-namespace}-col;
 }
 
 .#{$gridContainer}.#{$formContainer}.#{$formContainer}--display-mode {
+  @include fd-compact-or-condensed() {
+    --fdFormLayoutGridTextPaddingBlock: 0.3125rem;
+  }
+
+  // 1rem vertical spacing between label+value pairs (spec vertical C)
+  &.#{$formContainer}--vertical,
+  &.#{$formContainer}--md-vertical,
+  &.#{$formContainer}--lg-vertical,
+  &.#{$formContainer}--xl-vertical {
+    --fdFormLayoutGridItemMarginTop: 1rem;
+    --fdFormLayoutGridItemFirstOfTypeMarginTop: 0;
+  }
+
   &.#{$formGroup},
   .#{$formGroup} {
     .#{$formItem} {
@@ -118,11 +131,7 @@ $gridCol: #{$fd-namespace}-col;
       }
 
       .#{$fd-namespace}-text {
-        padding-block: 0.625rem;
-
-        @include fd-compact-or-condensed() {
-          padding-block: 0.3125rem;
-        }
+        padding-block: var(--fdFormLayoutGridTextPaddingBlock, 0.625rem);
       }
     }
   }
@@ -132,24 +141,6 @@ $gridCol: #{$fd-namespace}-col;
     .#{$gridCol}--12 {
       > .#{$fd-namespace}-form-label {
         padding-block-end: 0.25rem;
-      }
-    }
-  }
-
-  // 1rem vertical spacing between label+value pairs (spec vertical C)
-  // Covers explicit vertical modifiers at all breakpoints
-  &.#{$formContainer}--vertical,
-  &.#{$formContainer}--md-vertical,
-  &.#{$formContainer}--lg-vertical,
-  &.#{$formContainer}--xl-vertical {
-    &.#{$formGroup},
-    .#{$formGroup} {
-      .#{$formItem} {
-        margin-top: 1rem;
-
-        &:first-of-type {
-          margin-top: 0;
-        }
       }
     }
   }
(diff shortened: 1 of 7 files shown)


packages/styles/src/form-layout-grid.scss
```diff
@@ -109,13 +109,73 @@ $gridCol: #{$fd-namespace}-col;
...
+      .#{$fd-namespace}-form-label {
+        margin-inline-end: 0.5rem;
+      }
+
+      .#{$fd-namespace}-text {
+        padding-block: 0.625rem;
+
+        @include fd-compact-or-condensed() {
+          padding-block: 0.3125rem;
+        }
```
InnaAtanasova: Such patterns we usually do with CSS variables. 
you can define a local CSS variable with default value and later overwrite the value without repeating the property. This minimizes the generated code. 
For example:
```
.#{$fd-namespace}-text {
    padding-block: var(--myLocalVariableForForm, 0.625rem);

    @include fd-compact-or-condensed() {
        --myLocalVariableForForm: 0.3125rem;
    }
}
```

In such cases you don't need the `@include fd-compact-or-condensed()` in the same block. You can have one single `@include fd-compact-or-condensed()` per component and there modify all the CSS variables. The cool thing is that in such cases you don't care about specificity:

```
.#{$fd-namespace}-text {
    padding-block: var(--myLocalVariableForForm, 0.625rem);
}

....
....
....

@include fd-compact-or-condensed() {
    --myLocalVariableForForm: 0.3125rem;
    --anotherLocalVariableForForm: 1em;
    --andAnotherLocalVariableForForm: 0;
}
}
```

packages/styles/src/form-layout-grid.scss
```diff
@@ -16,7 +16,7 @@ $gridCol: #{$fd-namespace}-col;
     &.#{$formGroup},
     .#{$formGroup} {
       .#{$formItem} {
-        margin-top: 0.625rem;
+        margin-top: $marginTop;
 
         &:first-of-type {
           margin-top: 0;
```
InnaAtanasova: following my first comment, you can put these in CSS variables:
```
margin-top: var(--fdFormLayoutGridItemMarginTop, $marginTop);

&:first-of-type {
     margin-top: var(--fdFormLayoutGridItemFirstOfTypeMarginTop, 0);
}
```

Later, instead of repeating the selectors you can have:
```
&.#{$formContainer}--vertical,
&.#{$formContainer}--md-vertical,
&.#{$formContainer}--lg-vertical,
&.#{$formContainer}--xl-vertical {
    --fdFormLayoutGridItemMarginTop: 1rem;
    --fdFormLayoutGridItemFirstOfTypeMarginTop: 0;
}
```

and 
```
.#{$gridContainer}.#{$formContainer}.#{$formContainer}--display-mode {
    --fdFormLayoutGridItemMarginTop: 1rem;
    --fdFormLayoutGridItemFirstOfTypeMarginTop: 0;
}
```

InnaAtanasova (commented):
some suggestions for code optimization.

packages/styles/src/form-layout-grid.scss
InnaAtanasova: You can check `info-label.scss` and how we handle `display` modifier for a better illustration of what I tried to explain.

InnaAtanasova: `card.scss` as well. There are even more complicated cases/examples of how we achieve the interactive states of the components with only CSS variables and not repeating the CSS attributes.

InnaAtanasova (approved):

PR #6381, opened 2026-08-28
docs(styles): product switch design updates - move some variables to the root

MariaIDineva:
## Related Issue
Closes none

## Description
 `--fdProductSwitch_Body_BorderRadius` was defined on `.fd-product-switch`, but in fundamental-ngx this component renders `.fd-product-switch__body` without that ancestor, so the variable was out of scope: 
 <pre>

Screenshot 2026-08-28 at 12 27 22

</pre>

Similar was the situation with `--fdProduct_Switch_Item_Border_Color` used for the outline of `fd-product-switch__item`:
<pre>

Screenshot 2026-08-28 at 12 28 57
</pre>

To remediate the issue, both variables are now on `:root`.

### DIFF
--- a/packages/styles/src/product-switch.scss
+++ b/packages/styles/src/product-switch.scss
@@ -7,11 +7,6 @@
 
 $block: #{$fd-namespace}-product-switch;
 
-:root {
-  --fdProduct_Switch_Item_Border_Color: transparent;
-  --fdProductSwitch_Body_BorderRadius: var(--sapPopover_BorderCornerRadius, var(--sapElement_BorderCornerRadius));
-}
-
 .#{$block} {
   @mixin fd-product-pseudo-element-background($background) {
     .#{$block}__title {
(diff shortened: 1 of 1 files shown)


InnaAtanasova:
I added some changes:
instead of declaring the variables on root level, which we avoid in general, I refactored them a bit to fix the problem.
1. There was no need of declaring `--fdProductSwitch_Body_BorderRadius` variable. It's used in 1 place only, so I directly updated the value of the border-radius: `border-radius: var(--sapPopover_BorderCornerRadius, var(--sapElement_BorderCornerRadius));`
2. Moved the declaration of `--fdProduct_Switch_Item_Border_Color` in the block level of `fd-product-switch__item` where it's first used. The overwriting of the value happens in the same scope, so no need to be declared outside of this block.

droshev (approved):

PR #6400, opened 2026-09-11
fix(styles): update List spacings in Notifications view of Settings dialog

InnaAtanasova:
## Related Issue
Closes none, design updates in Figma

## Description
- Updated the vertical paddings of Byline list in the context of Notifications
- added 0.75rem spacing between the content container and the switch
- added 0.125rem spacing between the switch and the button
- updated the spacing between the title and the byline


packages/styles/src/mixins/list/_list-base.scss
```diff
@@ -474,6 +474,10 @@
         align-self: center;
       }
 
+      button {
+        margin-inline-start: 0.125rem;
```
MariaIDineva: Can we use `--fdList_Button_Spacing` instead of hardcoding `0.125rem`?

packages/styles/src/mixins/list/_list-base.scss
```diff
@@ -474,6 +474,10 @@
         align-self: center;
       }
 
+      button {
```
MariaIDineva: Calude claims that 

> SCSS style guide explicitly prohibits tag selectors,

 so should that be `.fd-button` instead?

InnaAtanasova: We have a rule that we cannot modify another component's properties. This is breaking the self-contained rule. But I see we have an existing modifier class `fd-list__button`. This would also solve your other comment.

packages/styles/src/mixins/list/_list-base.scss
InnaAtanasova: this is no longer needed as I am using the modifier class (see the other comment).

MariaIDineva (approved):

PR #6371, opened 2026-08-12
feat(styles): new truncation rules for standard and byline lists

MariaIDineva:
## Related Issue
Closes [SAP/fundamental-styles#](https://github.com/SAP/fundamental-styles/issues/6319)

## Description

Implements updated truncation rules across Standard List, Byline List, and Subline List per the agreed design decision: **truncation is the default behavior; wrapping is opt-in via `--wrap` modifier classes.**

- **Subline List** — behavior inversion (breaking change): default changed from wrap → truncate; `--wrap` modifier added; `--truncate` kept as deprecated no-ops for backward compatibility
- **Byline List** — bug fix: existing `--wrap` modifiers were not resetting `overflow`, so text was still clipped even with the modifier applied; cleaned up a stale combined rule; fixed a second pre-existing bug in borderless mode where `max-height: 5rem; height: 100%` on items caused wrapped content to overflow out of the item boundary
- **Standard List** — additive: `--wrap` modifiers added to `__title`, `__secondary`, and group-header `__title`; new story added; checkbox alignment in wrapped selection items fixed

### SCSS

- [packages/styles/src/mixins/list/_list-subline.scss](packages/styles/src/mixins/list/_list-subline.scss) — default behavior changed to truncate; `--wrap` modifier added; `--truncate` kept as deprecated no-ops; split combined `__item--wrap, :has()` rule so only `__item--wrap` wraps all children; removed stale `overflow: visible` from `__content:has()`; avatar kept vertically centered in all wrap scenarios
- [packages/styles/src/mixins/list/_list-byline.scss](packages/styles/src/mixins/list/_list-byline.scss) — `--wrap` on `__title` and `__byline` now correctly calls `@include fd-wrap()`; removed stale combined rule; removed `max-height`/`height: 100%` from borderless item override; fixed thumbnail/icon top-alignment in wrapped link items (`align-items: flex-start` on `fd-list__link`); navigation arrow kept centered via `align-self: center` on `::after`; updated padding-block-start to 0.75 rem to match the specs.
- [packages/styles/src/mixins/list/_list-base.scss](packages/styles/src/mixins/list/_list-base.scss) — `--wrap` added to `__title`, `__secondary`, and group-header `__title`; split combined `__item--wrap, :has()` rule; `__secondary` gets `align-self: center` in `:has()` rule; `__item-counter`, `__icon`, and `__button` get `align-self: center` in all wrap rules; breathing room via `margin-block: 0.75rem` scoped to `:not(.fd-list--byline)`
[...]

PR #6401, opened 2026-09-14
feat(styles): extend new truncation rules for standard and byline lists to remaining list components

### LINKED ISSUE #6319: Implement new truncation rules for Standard and Byline Lists
### Description

Implement the updated truncation behavior for the List component according to the latest designs for both Standard List and Byline List variants.

The current implementation relies on truncation modifiers applied to individual content elements. The new implementation should align with the updated truncation requirements while minimizing breaking changes for existing consumers.

### Requirements

- Implement the new truncation rules for:
    -   Standard List
    -   Byline List

- Preserve existing behavior by default whenever possible.
- Introduce new functionality through modifier classes rather than changing existing APIs or default behavior.
- Ensure long content can be configured to either truncate or wrap according to the new design specifications.
- Support truncation behavior at the list and/or list item level where appropriate.
- Update documentation and examples to reflect the new behavior.

### Acceptance Criteria

- New truncation rules are implemented according to design specifications.
- Existing list implementations remain unaffected unless new modifier classes are applied.
- Modifier classes provide a clear and scalable way to opt into the new truncation behavior.
- Documentation and examples are updated.
- Accessibility considerations are verified for wrapped and truncated content.

MariaIDineva:
## Related Issue
Closes [SAP/fundamental-styles#](https://github.com/SAP/fundamental-styles/issues/6319)

## Description

This PR extends the truncation-by-default / `--wrap` opt-in pattern — already [established](https://github.com/SAP/fundamental-styles/pull/6371) for Standard, Byline, Subline, Selection lists, and User Menu — to all remaining list-like components.

**Design contract:** text truncates (ellipsis) by default; consumers add a `--wrap` modifier to enable wrapping.

---

### `navigation-list.scss`
Added `fd-navigation-list--wrap` modifier.
- `__content`: switches from fixed `height` to `min-height` so items grow with wrapped text; `overflow: visible`; adds `padding-block: 0.875rem` (cozy) / `0.5rem` (compact) per spec — replacing the vertical-centering-by-fixed-height trick with explicit padding
- `__text`: removes `fd-ellipsis()`, sets `white-space: normal`, `line-height: normal` (spec-specified)
- Added `--fdNavigationList_Content_Wrap_Padding_Block: 0.875rem` CSS variable for the cozy block padding; compact padding extracted to `--fdNavigationList_Content_Wrap_Compact_Padding_Block: 0.5rem`

### `mixins/vertical-nav/_vertical-nav-navigation.scss`
Updated `fd-list--wrap` modifier inside `fd-vertical-nav-navigation()` mixin.
- `__navigation-item`: switches from fixed height to `min-height`; adds `padding-block: 0.875rem` per spec — replacing the vertical-centering-by-fixed-height trick with explicit padding
- `__navigation-item-text`: removes `white-space: nowrap`, `overflow: hidden`, `text-overflow: ellipsis`; sets `line-height: normal` (spec-specified, was incorrectly `1.4`)

### `_message-view.scss`
Fixed `fd-message-view__list` width behaviour.
- Base rule kept as `max-width: 26rem` (preserves details-view layout); `width: 26rem; max-width: 100%` scoped to `.fd-list--wrap` inside `__list` — prevents the popover from shrinking below 26rem only when wrapping is active

### `mixins/list/_list-message-view.scss`
Added `&.#{$block}--wrap` scoped inside `--message-view`.
- `__content`: `overflow: visible` (unblocks the `fd-flex(column) + fd-ellipsis()` clip)
- `__title`, `__subtitle`: removes `fd-ellipsis()`

### `mixins/list/_list-dropdown.scss`
Added `fd-list__label--wrap` element modifier inside `--multi-input`.
- Removes `overflow: hidden` and `text-overflow: ellipsis` from the full-width checkbox label
- Also overrides `fd-checkbox__label-container` (`height: auto`, `overflow: visible`, `white-space: normal`) and `fd-checkbox__text` (`overflow: visible`, `white-space: normal`) — the actual truncation points inside the checkbox markup

### `upload-collection.scss`
Added `&.fd-list--wrap` override.
- `__description` and `__status-group-item-text`: removes `fd-ellipsis()` — both have their own `@include fd-ellipsis()` in the upload-collection block, so they weren't reached by the `_list-byline.scss` `--wrap` modifier which only targets `fd-list__title` and `fd-list__byline`

### `object-list.scss`
Added `fd-object-list--wrap` modifier.
- `__intro`, `__object-attribute`: removes `fd-ellipsis()`
- `__row-left`, `__row-right`: removes `fd-ellipsis()`; also overrides `fd-object-status__text` inside both `__row-left` and `__row-right` — each carries its own `fd-ellipsis()` that cannot be reached from the outer modifier
- `__object-identifier__title`: removes `fd-line-clamp(2)` (`display: block`, `-webkit-line-clamp: initial`, `overflow: visible`) — `fd-line-clamp()` is applied internally and cannot be reached from the outer modifier

### `tree.scss`
Added `fd-tree--wrap` list-level shorthand mirroring the per-node `__content--wrap` modifier.
- `__item-container`: adds `padding-block: 0.25rem` when wrapping is active, giving wrapped items vertical breathing room
- `__content`: applies `text-wrapping-behaviour()` (mirrors the per-node `__content--wrap` modifier)
- `__text`: applies `text-wrapping-behaviour()`; sets `line-height: var(--fdTree_Item_Wrap_Text_Line_Height)`

---

## Stories added

| Component | Story | Behaviour shown |
|-----------|-------|-----------------|
| Side Navigation | `Long text` | `fd-list--wrap` on vertical nav; collapsed state unaffected |
| Navigation List | `Long text` | `fd-navigation-list--wrap` on the nav list |
| Message Popover | `Long text` | `fd-list--wrap` on message-view list |
| Vertical Navigation | `Long text` | `fd-list--wrap` on vertical nav list |
| Object List | `Long text` | `fd-object-list--wrap` |
| Upload Collection | `Long text` | `fd-list--wrap` wraps long filenames (`__title`) and description text (`__description`) |
| Combobox Input | `Long text` | Dropdown wraps by default; `fd-list__title--no-wrap` to opt out |
| Multi ComboBox | `Long text` | Dropdown wraps by default; `fd-list__title--no-wrap` to opt out |
| Multi Input | `Long text` | `fd-list__label--wrap` on full-width checkbox option labels |
| Tree | `Long text` | `fd-tree--wrap` at multiple indent levels |
| Dialog | `Long text list` | `fd-list--wrap` + `fd-list__item--wrap` inside a dialog body |

---

## What is already wrapped by default (no modifier needed)

- `fd-list--dropdown` and `fd-list--multi-input` `__title` / `__secondary` — `white-space: normal` is the base style; use `fd-list__title--no-wrap` to opt out
- Upload collection title — inherits byline `fd-list__title` wrapping when `fd-list--wrap` is on the root

---


packages/styles/src/mixins/list/_list-message-view.scss
```diff
@@ -60,5 +60,18 @@
...
+    &.#{$block}--wrap {
+      .#{$block}__content {
+        overflow: visible;
+      }
+
+      .#{$block}__title,
+      .#{$block}__subtitle {
+        white-space: normal;
+        overflow: visible;
+        text-overflow: initial;
```
InnaAtanasova, line 73: This should be mixin

packages/styles/src/mixins/vertical-nav/_vertical-nav-navigation.scss
```diff
@@ -344,6 +344,35 @@ $fd-list-second-level-popover-position: 3.25rem !default;
...
+      .#{$blockList}__navigation-item {
+        height: auto;
+        min-height: $fd-list-navigation-item-height;
+        padding-block: 0.875rem;
+      }
+
+      .#{$blockList}__navigation-item-text {
+        white-space: normal;
+        overflow: visible;
+        text-overflow: initial;
```
InnaAtanasova, line 358: these should be a mixin

packages/styles/src/_message-view.scss
```diff
@@ -16,6 +16,11 @@ $block: #{$fd-namespace}-message-view;
     max-height: 50vh;
     max-width: 26rem;
 
+    .fd-list--wrap {
+      width: 26rem;
+      max-width: 100%;
+    }
```
InnaAtanasova, line 22: This should not be here. It violates the self-contained rule: components should not modify another component’s properties within their own styles.

Image

You have these classes: `fd-list fd-list--navigation-indication fd-list--navigation fd-list--message-view fd-list--wrap`. You can overwrite these properties inside list, where you have `fd-list--message-view`. There you can add specficity `fd-list--message-view fd-list--wrap`.

packages/styles/src/navigation-list.scss
```diff
@@ -384,4 +387,24 @@ $block: #{$fd-namespace}-navigation-list;
       display: none;
     }
   }
+
+  &--wrap {
+    .#{$block}__content {
+      height: auto;
+      min-height: var(--fdNavigationList_Content_Height);
+      overflow: visible;
+      padding-block: var(--fdNavigationList_Content_Wrap_Padding_Block);
```
InnaAtanasova, line 396: This should be:
`padding-block: var(--fdNavigationList_Content_Wrap_Padding_Block, 0.875rem);` 
⬆️ create a local CSS variable `--fdNavigationList_Content_Wrap_Padding_Block` and assign to it the value `0.875rem`.
Then:
```
@include fd-compact-or-condensed() {
   --fdNavigationList_Content_Wrap_Padding_Block: 0.5rem;
}
```
⬆️ Change the value of the CSS var in compact mode without using the property `padding-block`.

packages/styles/src/navigation-list.scss
```diff
@@ -384,4 +387,24 @@ $block: #{$fd-namespace}-navigation-list;
...
+
+      @include fd-compact-or-condensed() {
+        padding-block: var(--fdNavigationList_Content_Wrap_Compact_Padding_Block);
+      }
+    }
+
+    .#{$block}__text {
+      white-space: normal;
+      overflow: visible;
+      text-overflow: initial;
```
InnaAtanasova, line 406: should include the mixin for wrapping.

packages/styles/src/object-list.scss
```diff
@@ -147,4 +147,42 @@ $block: #{$fd-namespace}-object-list;
 
     color: var(--sapContent_LabelColor);
   }
+
+  &--wrap {
+    .#{$block}__intro {
+      white-space: normal;
+      overflow: visible;
+      text-overflow: initial;
```
InnaAtanasova, line 155: mixin

packages/styles/src/object-list.scss
```diff
@@ -147,4 +147,42 @@ $block: #{$fd-namespace}-object-list;
...
+    }
+
+    .#{$block}__object-identifier {
+      // fd-object-identifier__title uses fd-line-clamp() which cannot be reached from the outer modifier
+      .#{$fd-namespace}-object-identifier__title {
+        display: block;
+        -webkit-line-clamp: initial;
+        -webkit-box-orient: initial;
```
InnaAtanasova, line 183: I don't think this is needed.

packages/styles/src/tree.scss
```diff
@@ -201,6 +201,22 @@ $fd-tree-item-selected-border-bottom: var(--sapList_BorderWidth) solid var(--sap
...
+  &--wrap {
+    .#{$block}__item-container {
+      padding-block: 0.25rem;
+    }
+
+    .#{$block}__content {
+      @include text-wrapping-behaviour();
+
+      .#{$block}__text {
+        @include text-wrapping-behaviour();
```
InnaAtanasova, line 213: It's correct what you did, including the existing mixin. But since we are doing this for many components, maybe delete the local mixin on line 14 of this file and use the global one (less code to maintain).

packages/styles/src/upload-collection.scss
```diff
@@ -110,6 +110,20 @@ $fd-upload-collection-button-group-small-margin: 0.5rem !default;
...
+    .#{$block}__description {
+      white-space: normal;
+      overflow: visible;
+      text-overflow: initial;
+    }
+
+    .#{$block}__status-group-item-text {
+      white-space: normal;
+      overflow: visible;
+      text-overflow: initial;
```
InnaAtanasova, line 123: mixin

packages/styles/src/upload-collection.scss
```diff
@@ -110,6 +110,20 @@ $fd-upload-collection-button-group-small-margin: 0.5rem !default;
     }
   }
 
+  &.fd-list--wrap {
+    .#{$block}__description {
+      white-space: normal;
+      overflow: visible;
+      text-overflow: initial;
```
InnaAtanasova, line 117: mixin

packages/styles/src/upload-collection.scss
```diff
@@ -110,6 +110,20 @@ $fd-upload-collection-button-group-small-margin: 0.5rem !default;
     }
   }
 
+  &.fd-list--wrap {
```
InnaAtanasova, line 113: this one is tricky 😆 
1. You can use `fd-list--wrap` but you need to use it with the namespace: `.#{$fd-namespace}-list--wrap`.
2. You can create a modifier class for Upload Collection and use specificity, aka replace this with `&--wrap`.
Because we have `fd-list fd-list--byline fd-list--wrap fd-upload-collection`. Basically 2 block modifiers used together: `fd-list` and `fd-upload-collection`. You will always break the self-contained rule 😆
I suggest option 1 and just use the namespace `&.#{$fd-namespace}-list--wrap`.

MariaIDineva: Thanks, I did option 1.

InnaAtanasova (approved):

